Issue #10, opened 2020-02-04
Calendar Interface


jethrokuan:
A calendar view to see daily notes would be nice.

dit7ya:
Are you talking about Roam's endless daily notes page or something else?

jethrokuan:
Something like org's calendar view that shows days with daily notes and be able to quickly access them I guess

ryjm:
I wonder if you could just have a thin layer over the `diary` library for the daily notes. You would have to account for each daily note being a headline vs a separate file, but it would be nice to avoid the buildup of files for each date.

jethrokuan:
Do you foresee the file buildups to be a big issue? Typically they become a small proportion of the total note files. While the time complexity of building backlinks is linear in the number of files, it's also dependent on file size and complexity (using the org parser to get the AST, and filter out links). Keeping them as separate files also keeps the library simple.

ryjm:
I was mainly thinking that hooking into the diary would make building out the calendar interface easy. Probably not that hard to do it from scratch though.

I do think it might be valuable to treat headlines and files in the same way (i.e everything under the header *Roam* would be treated the same as if it was in the file ` roam.org ( http://roam.org/ ) `), but that's a different discussion.

jethrokuan:
Yeah, it seems like `calendar.el` doesn't have an easy way to plug in other sources of diary information. Pity. I'll think about it a bit more when this becomes higher priority.

> i.e everything under the header *Roam* would be treated the same as if it was in the file ` roam.org

This sounds like it would be confusing...

anthrolisp:
Maybe have interoperability with [org-journal](https://github.com/bastibe/org-journal), which allows for different options for how to keep the journal entries (one per year, month, day, etc)

jethrokuan:
Oh yeah, that should just work out of the box by default! Still no calendar view though, but I think I'm gonna do that.

jethrokuan:
Okay yeah I confirmed that org-journal works neatly out of the box within org-roam. Here's my config:

```
(use-package org-journal
  :bind
  ("C-c n j" . org-journal-new-entry)
  :custom
  (org-journal-date-prefix "#+TITLE: ")
  (org-journal-file-format "%Y-%m-%d.org")
  (org-journal-dir "/home/jethro/Dropbox/org/braindump/org/")
  (org-journal-date-format "%A, %d %B %Y"))
```

And calendar view marking dates with entries:

[...]

PR #25, opened 2020-02-07
Add documentation on synergistic packages

jethrokuan:
There are a number of packages that play nice with org-roam. Related: #10


jethrokuan:
Nice commit hash!

PR #28, opened 2020-02-07
Add org-roam-tomorrow and org-roam-date

steven-lang:
Similar to `org-roam-today`:

- `org-roam-tomorrow`: Add a file for tomorrow
- `org-roam-date`: Add a file for any date (choose via date picker)


jethrokuan:
LGTM!

PR #29, opened 2020-02-07
Add or fix symlink support

daniel-koudouna:
One of the conditions of `org-roam--maybe-update-buffer ` checks the true name of the current buffer. However, When populating `org-roam-files`, the true name is not used. This causes symlinks to not work, at least for me.

This PR adds the true file name to `org-roam-files` instead.


jethrokuan:
Thanks!

Issue #37, opened 2020-02-09
Cant get started: org-roam-directory doesn't work


rileyrg:
custom.el:-

'(org-roam-directory "~/.emacs.d/orgfiles/roam-files/")

![Screenshot from 2020-02-09 21-34-12](https://user-images.githubusercontent.com/362891/74109468-fd424880-4b83-11ea-88b5-032f9a20cf8d.png)

Have no idea what's happening here. 

`(expand-file-name org-roam-directory)
`
Shows the proper directory.

(.emacs.d is in fact a symlink to ~/Dropbox/homefiles/DotFiles/.emacs.d but that shouldnt matter)

jethrokuan:
Isn't this behaviour expected then? since your `org-roam-directory` resolves to that directory in the screenshot.

jackbaty:
Not sure if this is the same issue, but since an earlier update (today?) calling `org-roam-today` creates the new file relative to whichever the current directory is. For example, if I call `org-roam-today` while editing a file in ~/org, the new file is created as ~/org/2020-02-09.org. My `org-roam-directory` is configured as ~/notes/roam.

jethrokuan:
ah yes @jackbaty thanks, should be fixed!

rileyrg:
No it doesn’t. No such directory.

Sent from Mail for Windows 10

jethrokuan:
@rileyrg can you pull down the latest master, and see if things are fixed for you yet?

rileyrg:
Will do.

rileyrg:
Nope. Sorry.

  
![Screenshot from 2020-02-10 08-48-43](https://user-images.githubusercontent.com/362891/74130465-6c4e8a00-4be2-11ea-970c-a1097a85cc07.png)
![Screenshot from 2020-02-10 08-48-17](https://user-images.githubusercontent.com/362891/74130467-6ce72080-4be2-11ea-9224-51f4190694a1.png)
![Screenshot from 2020-02-10 08-46-13](https://user-images.githubusercontent.com/362891/74130468-6d7fb700-4be2-11ea-8d96-eba9f59cf068.png)

jethrokuan:
I don't see a problem with the value of `org-roam-directory`, it is exactly what you set it to be. You'll need to create the directory yourself.

rileyrg:
It is created. Look at the path in the third screenshot. The code is sonehow doubling up the symlink part of the path.

[...]

PR #44, opened 2020-02-10
Fix symlinks handling

jethrokuan:
Fixed some of the bad symlink handling. This is way more troublesome than I hoped it would be.


jethrokuan:
Should address #37

PR #55, opened 2020-02-11
Fix org-roam--update-buffer after org-capture

### LINKED ISSUE #54: org-roam--update-buffer breaks org-capture
If you org-capture, then `C-c C-c` this error occurs:

```
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  substring(nil 0 1)
  file-truename(nil)
  org-roam--update-buffer()
  org-roam--update-cache()
  run-hooks(after-save-hook)
  basic-save-buffer(nil)
  save-buffer()
  org-capture-finalize(nil)
  funcall-interactively(org-capture-finalize nil)
  call-interactively(org-capture-finalize nil nil)
  command-execute(org-capture-finalize)
```

aaronjensen:
Fixes #54

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -588,7 +588,8 @@ This needs to be quick/infrequent, because this is run at
 
 (defun org-roam--update-buffer ()
   "Update `orgo-roam-buffer'."
-  (org-roam-update (file-truename (buffer-file-name (window-buffer)))))
+  (when-let ((buffer-file-name (buffer-file-name (window-buffer))))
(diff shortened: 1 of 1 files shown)


jethrokuan:
Thanks for the attempt! Simply adding this here duplicates work for `org-roam--maybe-update-buffer`, which is called on `post-command-hook`, so I decided to remove the update-buffer function altogether.

aaronjensen:
Awesome, thanks for the quick fix.

Issue #53, opened 2020-02-11
Enter new Zettel name only once with org-roam-insert


chrispoole643:
A feature request really for anyone who can more quickly think how to do this...

When I invoke `org-roam-insert` and type a new name for a Zettel that doesn't yet exist, it drops me to a prompt to enter the File ID (`org-roam--insert-title`). Feature request for it to take whatever name I've just entered, a file that doesn't yet exist, and create that file with that name.

(Similar to how Deft functions with `deft-new-file`.)

jethrokuan:
Done!

chrispoole643:
Wow: quick! Don't think I can reopen it but tried it out and don't think it's doing quite what I intended.

If I run `org-roam-insert`, and enter a file title that isn't in the list (I'm using ivy here, for completion support...), the `[[title]]` in my Org file will be populated, but it still prompts for `Enter new file id: `. I want the file ID to be auto-populated too, as something like `{{title without special chars to work on all filesystems.org}}`. Preserve case, spaces, and perhaps remove the rest?

(I tried looking for a builtin elisp function to do this in a stadard Emacs way but can't find much...)

jethrokuan:
Oh I see, the above PR was to insert the `#+TITLE` attribute, but I can make a default.

jethrokuan:
@chrispoole643 done, now it autopopulates with something sensible. I think this whole ID thing is becoming a pain, I should just generate file names using unique timestamps.

Issue #23, opened 2020-02-06
Feature request: allow timestamps as file names


brabalan:
I have a simple implementation of a zettelkasten using deft and org files, which seems to be the basis behing org-roam. I let deft use timestamps as file names as it makes for stable links, and I can easily edit the title of a note without breaking links. It seems that you use the title itself as file name. Could it be adapted?

jethrokuan:
Yeah sure, I can add a function that creates time-stamped files. However, either you would need to know the relevant timestamp for `org-roam-insert`, or I would need to parse all the files to pull the title, and I think the latter is not likely to happen unless a cheap method of doing so presents itself. Would you be okay with that?

brabalan:
The workflow I'm thinking of, which I haven't coded yet, is something like this:
- use deft, or a local directory search (using `rg` and `ivy` or `helm`) to do a full-text search when searching for a place to link
- instrument the results to insert an org link with description taken from the title

At the moment I'm doing it manually: I switch to deft, find the destination file, and link it. This works but I dislike moving away from the file I'm linking from.

jethrokuan:
I wouldn't mind a PR the workflow you have in mind, i.e. use rg to get name of file, and then inserting the org link with title as the description.

brabalan:
I'll try to implement it and will let you know.

jethrokuan:
Here's something that works, if you have `counsel` and `rg`:

```
(defun org-roam-insert-action (x)
  (when (string-match "\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" x)
    (let* ((file-name (match-string-no-properties 1 x))
           (file (file-truename (expand-file-name
                                 file-name
                                 (ivy-state-directory ivy-last))))
           (id (org-roam--get-id file))
           (file-path (org-roam--get-file-path id)))
      (unless (file-exists-p file-path)
        (make-empty-file file-path))
      (insert (format "[[%s][%s]]"
                      (concat "file:" file-path)
                      (concat org-roam-zettel-indicator id)))
      (unless (eq ivy-exit 'done)
        (swiper--cleanup)
        (swiper--add-overlays (ivy--regex ivy-text))))))

(defun org-roam-rg ()
  (interactive)
  (let ((default-directory org-roam-directory)
        (counsel-ag-command "rg --with-filename --no-heading --line-number --color never %s"))
    (ivy-read (concat (car (split-string "rg")) ": ")
              #'counsel-ag-function
              :initial-input ""
[...]

PR #57, opened 2020-02-12
Add org-roam--populate-title

jethrokuan:
The title is now automatically populated by default. This behaviour is customizable via `org-roam-autopopulate-title`. The title is either provided during prompt, or inferred from the file name.


jethrokuan:
FIxes #53 , #23 .

PR #72, opened 2020-02-13
deduplicate async/non-async functions

jethrokuan:
There was a lot of duplication between the lambda for building the cache async, and the normal processing functions, because the processing functions needed to be available within the `async-start` function. I'm not sure of the proper way to inject defuns into `emacs-async`, but this seems to work. The downside is that help functions no longer work/point to the correct source.

Do you know of a better way? cc @l3kn

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -161,7 +161,6 @@ If called interactively, then PARENTS is non-nil."
         (unless org-roam-cache-initialized
           (user-error "The Org-Roam caches aren't built! Please run org-roam--build-cache-async"))
         nil)))
-
 (defun org-roam--find-files (dir)
   "Return all org-roam files in `DIR'."
   (if (file-exists-p dir)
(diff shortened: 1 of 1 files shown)


l3kn:
I ran into a similar issue a few days ago and solved it by putting all defun code inside the `async-start`, then adding a polyfill-marco so the code still works when async is not installed.

https://github.com/l3kn/org-fc/blob/org-el-cache/org-fc-review.el#L280

That only works because I have very little code that needs to run inside the `async-start`,
to add async loading to `org-el-cache`, I'll try a similar approach to yours.

I think if you use macros instead of defuns in the "wrapper functions", you can skip the `(eval)`.
I'll try that for the `org-el-cache` code later.

l3kn:
I've not found a lot of other emacs packages using async so far.

One of their examples uses a similar approach by inlining the function body:
https://github.com/jwiegley/emacs-async/blob/master/dired-async.el#L176

jethrokuan:
yeah, that was my inspiration for the previous version, but I need it to be a function instead. The macro version doesn't look much better.

l3kn:
If you move all code used inside the `async-start` to a separate file, you can also inject the `load-path` variable like this:

``` lisp
(async-start
 `(lambda ()
    (add-to-list 'load-path ,(file-name-directory (or load-file-name (buffer-file-name))))
    (require 'org-el-cache)
    ;; More code
    )
 (lambda (res) (message res)))
```

jethrokuan:
> If you move all code used inside the async-start to a separate file, you can also inject the load-path variable like this:

This was also suggested in https://github.com/jwiegley/emacs-async/issues/122, but I'll use the latter method mentioned there as well.

PR #87, opened 2020-02-15
Look for encrypted org-files (.org.gpg) in files and links

### LINKED ISSUE #85: Adding support for encryption
It would be useful to have support for encrypting individual notes using via GnuPG. These files have a `org.gpg` extension which are not picked up by org-roam currently.

Having just one encrypted note in your notes directory would require a password each time the cache is rebuilt (unless the password is cached). If this is done often it could be annoying, so perhaps the encrypted notes should only be cached once.

Would this feature be a useful addition to org-roam, or is it out of scope? Are there any potential problems adding a new file extension in addition to `.org` (this might be nice to have regardless of encryption)?

chip2n:
Seems to work both for building the initial cache as well as jumping to a note via a back link (password prompt is shown if password not cached).

Fixes #85

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -39,7 +39,7 @@
 (require 'cl-lib)
 
 (defun org-roam--org-file-p (path)
-  "Check if the path is pointing to an org file."
+  "Check if PATH is pointing to an org file."
   (let ((ext (file-name-extension path)))
     (or (string= ext "org")
         (and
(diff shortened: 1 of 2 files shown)


org-roam-utils.el
```diff
@@ -38,6 +38,14 @@
 (require 'subr-x)
 (require 'cl-lib)
 
+(defun org-roam--org-file-p (path)
+  "Check if the path is pointing to an org file."
```
jethrokuan:
```suggestion
  "Check if PATH is pointing to an org file."
```

chip2n:
There we go!

PR #88, opened 2020-02-16
Add a missed (file-truename org-roam-directory)

nivekuil:
Fixes org-roam--maybe-enable not working with a symlinked directory. By the way, should all the file-truename calls be optimized somehow?


jethrokuan:
Thanks!

> By the way, should all the file-truename calls be optimized somehow?

If you can think of a way, please do let me know.

PR #90, opened 2020-02-16
Add support for file encryption by default

chip2n:
If org-roam-encrypt-files is t, org-roam creates files with .org.gpg extension instead.

Might be a good idea to rename `org-roam--get-file-path` to `org-roam--make-new-file-path`, since given an existing ID it no longer reliably gets the path for it (could be either encrypted or unencrypted). I think it's only used for creating new files now anyways.

See #85

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.1.2 (TBD)
 
 ### New Features
-* [#87][gh-87] Support encrypted Org files (by [@chip2n](https://github.com/chip2n/))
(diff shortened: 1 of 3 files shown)


jethrokuan:
> I think it's only used for creating new files now anyways.

Yep you're right. Could you do that as well? Also, `org-roam-encrypt-files` could use a short para in the documentation. https://github.com/jethrokuan/org-roam/blob/develop/doc/configuration.md

chip2n:
Done!

rileyrg:
what/if any key variable does/can his default to? gnupg.conf default-key isnt used, epa-file-encrypt-to doesnt seem to work either. I'm constantly prompted to mark my only encrpytion key.

PR #99, opened 2020-02-17
(feature) Add link to backlink line to backlinks buffer

jdormit:
This adds a link that goes directly to the line containing the backlink for each entry in the backlinks buffer. It does this by changing the the data structure in the backlinks cache that stores content from being a list of content strings to a list of cons pairs, where the `car` of the pair is the content string and the `cdr` of the pair is the line number where the content was found. This structure allows the backlinks buffer to link directly to the content lines in the linking file rather than just linking to the file as a whole.

image

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -45,7 +45,7 @@ Like file-name-extension, but does not strip version number."
   (save-match-data
     (let ((file (file-name-nondirectory filename)))
       (if (and (string-match "\\.[^.]*\\'" file)
-	             (not (eq 0 (match-beginning 0))))
+                     (not (eq 0 (match-beginning 0))))
           (substring file (+ (match-beginning 0) 1))))))
 
 (defun org-roam--org-file-p (path)
@@ -86,10 +86,11 @@ Like file-name-extension, but does not strip version number."
                    (org-roam--org-file-p path))
           (goto-char start)
           (let* ((element (org-element-at-point))
+                 (begin (or (org-element-property :content-begin element)
+                            (org-element-property :begin element)))
                  (content (or (org-element-property :raw-value element)
                               (buffer-substring
-                               (or (org-element-property :content-begin element)
-                                   (org-element-property :begin element))
+                               begin
                                (or (org-element-property :content-end element)
                                    (org-element-property :end element)))))
                  (content (string-trim content))
@@ -97,14 +98,13 @@ Like file-name-extension, but does not strip version number."
                                 (file-truename (buffer-file-name (current-buffer))))))
             (list :from file-path
                   :to (file-truename (expand-file-name path (file-name-directory file-path)))
-                  :content content
-                  :line (line-number-at-pos start t))))))))
+                  :properties (list :content content :point begin))))))))
 
 (cl-defun org-roam--insert-item (item &key forward backward)
   "Insert ITEM into FORWARD and BACKWARD cache.
 
-ITEM is of the form: (:from from-path :to to-path :content preview-content :line line-number)."
-  (pcase-let ((`(:from ,p-from :to ,p-to :content ,content :line ,line) item))
+ITEM is of the form: (:from from-path :to to-path :properties (:content preview-content :point point))."
+  (pcase-let ((`(:from ,p-from :to ,p-to :properties ,props) item))
     ;; Build forward-links
     (let ((links (gethash p-from forward)))
       (if links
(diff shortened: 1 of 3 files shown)


jdormit:
Not sure if this is the best approach - it would be a lot more elegant if you could just press `enter` on each backlink entry to switch to that file and that line. But this is an easy-to-implement stopgap until I have time to flesh out a nicer solution.

org-roam-utils.el
```diff
@@ -87,13 +87,14 @@
                                 (file-truename (buffer-file-name (current-buffer))))))
             (list :from file-path
                   :to (file-truename (expand-file-name path (file-name-directory file-path)))
-                  :content content)))))))
+                  :content content
```
jethrokuan: Instead of this, change `:content` to something more generic like `:properties`, which contains a `plist`:

```
(:content content
 :point p)
```

`p` can be obtained from the org-element properties (`:content-begin` or `:begin`), but we need to be careful about this as this needs to match the actual preview content.

org-roam-utils.el
```diff
@@ -106,14 +107,14 @@ ITEM is of the form: (:from from-path :to to-path :content preview-content)."
     (let ((contents-hash (gethash p-to backward)))
       (if contents-hash
           (if-let ((contents-list (gethash p-from contents-hash)))
-              (let ((updated (cons content contents-list)))
+              (let ((updated (cons (cons content line) contents-list)))
```
jethrokuan: as with above, putting the entire `plist` in, rather than a cons cell.

org-roam.el
```diff
@@ -415,10 +415,13 @@ This is equivalent to removing the node from the graph."
                          (insert (format "** [[file:%s][%s]]\n"
                                          file-from
                                          (org-roam--get-title-or-slug file-from)))
-                         (dolist (content contents)
```
jethrokuan: we can:
1. propertize the content with the file and start-point.
2. create a sparse-keymap containing a binding for `RET` to `org-roam--jump-to-point`, using the text properties (file and point) to jump to the actual point.

```
     (let ((map (make-sparse-keymap)))
       (define-key map [mouse-2] 'operate-this-button)
       (put-text-property link-start link-end 'keymap map))
```

See https://www.gnu.org/software/emacs/manual/html_node/elisp/Clickable-Text.html

jethrokuan:
Sorry I made a pretty big change in #105 would mean you need to resolve some merge conflicts, hope you don't mind.

jdormit:
@jethrokuan thanks for the quick feedback! I'll work through this over the next couple of days.

org-roam-utils.el
jethrokuan: > needs to match the actual preview content.

You may change this if you wish if it's reasonable. The worry is pulling out headers (especially level 1 headers), which would mess with the formatting of the buffer.

org-roam-utils.el
```diff
@@ -45,7 +45,7 @@ Like file-name-extension, but does not strip version number."
   (save-match-data
     (let ((file (file-name-nondirectory filename)))
       (if (and (string-match "\\.[^.]*\\'" file)
-	             (not (eq 0 (match-beginning 0))))
+                     (not (eq 0 (match-beginning 0))))
```
jdormit: This was just changing a tab to a space

jethrokuan: Thanks! I copied this from Emacs, I guess they use tabs :-1:

org-roam.el
```diff
@@ -397,6 +397,28 @@ This is equivalent to removing the node from the graph."
   (let ((time (org-read-date nil 'to-time nil "Date:  ")))
     (org-roam--new-file-named (format-time-string "%Y-%m-%d" time))))
 
+(defun org-roam-jump-to-backlink ()
```
jethrokuan: I was thinking of having the jump be to the exact character at point, but I think this is good enough for me.

jdormit:
@jethrokuan sure thing! I think I want to make the backlink buffer mode a major mode rather than a minor mode to make sure the keybindings don't get overriden - what do you think?

jethrokuan:
@jdormit that's fine with me

jdormit:
@jethrokuan ok, I'm feeling good about this now. Want to take another look?

jdormit:
Oops! Added the changelog entry

jdormit:
Er, screwed up the formatting. Fixing now.

jdormit:
Hmm, maybe I did it right? At any rate the formatting matches that of the other entries.

Issue #79, opened 2020-02-14
Add test cases for all functions


jethrokuan:
# Functions
- [x] org-roam--build-cache-async
- [x] org-roam-insert
- cache updates on:
  - [ ] file creation
  - [ ] link addition
  - [ ] link removal
## Scenarios
- [x] flat structure
- [x] nested structure

jethrokuan:
Hmm the rest are hard to test, because they're interactive.

PR #105, opened 2020-02-17
(tests): add tests for org-roam-insert

jethrokuan:
Testing the basic combinations for `org-roam-insert`.


jethrokuan:
Part of #79

PR #107, opened 2020-02-17
(bugfix): force a cache update on making a new file

### LINKED ISSUE #102: Confusion around file names and creating existing links
Hi! 

Thank you so much for creating this. I'm in the process of emptying my org-brain repository and entering the information in here -- which is a much better fit for how I want to be able to think.

Something just happened, which I think is related to the default behavior of using timestamps for file names (or may be a bug in the background updater - I don't quite know how to troubleshoot the current state of things):

1. I created an org-roam file with org-roam-find-file for a link about how to handle projects in org-mode.
2. In that file, I pasted the link, I also used org-roam-insert-file to insert a link to a non-existing roam file called "org-mode". 
3. I then used org-roam-find-file to open the file `org-mode`, which _created a new file_ when my intent was to open the file I had just created with that name.

I'm not exactly sure what happened, but my intent was to have a single file with the name `org-mode`.

jethrokuan:
@BarryLiCMU can I have you confirm this fix works for you? I changed `find-file` to `find-file-noselect`. 

Fixes #102

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -239,7 +239,10 @@ If not provided, derive the title from the file name."
       (error (format "Aborting, file already exists at %s" file-path))
     (if org-roam-autopopulate-title
         (org-roam--populate-title file-path title)
-      (make-empty-file file-path))))
+      (make-empty-file file-path))
+    (save-excursion
+      (with-current-buffer (find-file-noselect file-path)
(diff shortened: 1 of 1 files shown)


BarryLiCMU:
Just tested this branch, everything works for me. Thanks for the quick response!

Issue #30, opened 2020-02-08
Predictable link-following behaviour


daniel-koudouna:
I'm wondering if anyone else is having inconsistent behaviour when following links.

The only way to follow links from org seems to be `org-open-at-point`. This doesn't mirror the behaviour of `find-file` for me. It sometimes opens a new window, and the org-roam buffer is updated with a considerable delay. For evil users, this also messes with evil jumping.

I've written a find function to use more predictable behaviour, Would this be useful to add as a PR?  As I don't use vanilla org-mode features, this might be a solved problem.

jethrokuan:
This is right, the org-roam buffer won't update on an `org-open-file-at-point`. That's either because the `post-command-hook` doesn't get run, or when it was run it the predicates don't pass. It gets updated on the next command.

There might be a more proper solution here, like advicing the function. The "opening in a new window" behaviour is also probably customizable somehow. I'll look at it later.

jethrokuan:
Regarding `org-open-at-point`: This helps alleviate the problem a little:

`(add-to-list 'org-link-frame-setup '(file . find-file))`

It helps when opening file links within the main org buffer, but it doesn't help for opening links in the org-roam buffer, because that's a side window. Seems like `org-open-at-point` eventually runs the `post-command-hook` and updates the org-roam-buffer though.

daniel-koudouna:
Would it be worthwhile to have a custom link following function for the org-roam buffer? The idea would be to open the link in the previously selected buffer, to keep the org-roam buffer open.

jethrokuan:
Yeah, but actually I'm thinking of eventually moving away from using org-mode as the major mode for the org-roam buffer, to magit-section where things can be customized more easily, so this problem should go away. See #5

l3kn:
A workaround that seems to work well is storing the window `org-roam` was called from in a variable,
then use a custom link type in the org-roam buffer

``` lisp
(defun org-roam-find-file (file)
  (with-selected-window org-roam-last-window
    (find-file file))
  (select-window org-roam-last-window))

(org-link-set-parameters "org_roam_file" :follow #'org-roam-find-file)
```

@jethrokuan Does that seem like a good solution to you? If so, I'll open a pull request for it.

jethrokuan:
@l3kn that does look like a good solution. Happy to have a PR. Thanks!

jethrokuan:
This is closed with #108 thanks @l3kn

PR #108, opened 2020-02-17
WIP: Add org_roam_file link type

l3kn:
Add a custom org-mode link type that reuses the last window `org-roam` was called from to open files,
for use in the org-roam buffer.

Closes issue #30

@jethrokuan Should this go under "New Features" or "Internal" in the changelog?

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -397,19 +397,22 @@ This is equivalent to removing the node from the graph."
 
 (defun org-roam--find-file (file)
   "Open FILE in the window `org-roam' was called from."
-  (if org-roam-last-window
+  (if (and org-roam-last-window (window-valid-p org-roam-last-window))
       (progn (with-selected-window org-roam-last-window
                (find-file file))
              (select-window org-roam-last-window))
     (find-file file)))
-
-(org-link-set-parameters "org_roam_file" :follow #'org-roam--find-file)
 
 (defun org-roam-update (file-path)
   "Show the backlinks for given org file for file at `FILE-PATH'."
   (org-roam--ensure-cache-built)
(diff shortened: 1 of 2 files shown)


jethrokuan:
Now that I think about it, this doesn't resolve links from the preview content actually. Let me think about it again, sorry for the trouble.

l3kn:
Oh, I didn't think about that.
The only way I can think of that would make that work would be changing all links to org-roam files to use the new link type. That doesn't seem like an elegant solution.

Maybe it would be possible to overwrite the link following behavior for `file:` links just for a single buffer, then use `org-roam--find-file` instead of `find-file`.

Then the question is whether to open non-org-roam files that way, too.

jethrokuan:
> Maybe it would be possible to overwrite the link following behavior for file: links just for a single buffer, then use org-roam--find-file instead of find-file.

This sounds good to me.

> Then the question is whether to open non-org-roam files that way, too.

In a way, this is being consistent. I'm fine with that.

jethrokuan:
@l3kn just tested it, looks like it's working for me.

> @jethrokuan Should this go under "New Features" or "Internal" in the changelog?

Let's put it in a new `Changes` section?

org-roam.el
```diff
@@ -458,6 +475,7 @@ Valid states are 'visible, 'exists and 'none."
 (defun org-roam ()
   "Pops up the window `org-roam-buffer' accordingly."
   (interactive)
+  (setq org-roam-last-window (get-buffer-window))
```
jethrokuan: I think this may not be intuitive, and maybe `org-roam-last-window` should be set to the last `org-roam` file being interacted with, but for now I think this is fine and should be equivalent 99% of the time, we'll see if people complain.

PR #121, opened 2020-02-18
Add autoload to org-roam-mode

CeleritasCelery:
I am super excited to start using this package. It looks really well done. 

With this change it is no longer required to load org-roam before use. You can just add the `org-roam-mode` to `org-mode-hook` and it will load the package when needed.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -543,6 +543,7 @@ This needs to be quick/infrequent, because this is run at
       (org-roam-update (expand-file-name
                         (buffer-local-value 'buffer-file-truename buffer))))))
 
+;;;###autoload
 (define-minor-mode org-roam-mode
   "Global minor mode to automatically update the org-roam buffer."
   :require 'org-roam


jethrokuan:
Do you think there are other places to add `autoload` too, or is this enough? I think I'd add it for the cache building too.

CeleritasCelery:
You could add it to `org-roam--build-cache-async` for the use case you outline in the README (calling after init). But in that case we should probably change the name to only have 1 `-` since that is the convention for functions that are meant to be called by the user (i.e. `org-roam-build-cache-async`).

jethrokuan:
I don't have a strong opinion on whether `org-roam--build-cache-async` should be autoloaded, what do you think?

jethrokuan:
@CeleritasCelery I think just this autoload is fine. This branch needs to be rebased though.

Perhaps the next release is bug-free enough to be published on MELPA

CeleritasCelery:
I moved `org-roam-build-cache-async` to the external namespace (`--` is conventional resevered for internal only functions) and autoloaded it. If you don't like that I can revert it. They will probably mention that if you submit it to melpa though.

jethrokuan:
I think we can revert it. This change breaks user config, and also requires a doc update.

CeleritasCelery:
done.

Issue #119, opened 2020-02-18
Cache updates on file deletion and file renaming


jethrokuan:
### Brief Abstract
Cache should remain consistent on deleting and renaming files.

### Long Description
When deleting and renaming files, the cache should be updated accordingly.

Also, when either of this operations happen, it would be good to warn when some links will break, or update the corresponding links during a rename.

### Proposed Implementation (if any)
Add advices on `delete-file` and `rename-file`.

### Please check the following:
- [x] No similar feature requests

Issue #36, opened 2020-02-09
Rename file - update corresponding links.


akoen:
Hi, thanks for a great project. Is it possible to have backlinks automatically update if you rename an org-roam file? I believe that this would lead to more long-term stability.

jethrokuan:
It is, but I think this will come later. I'd like for everything else to be stable first.

akoen:
Yes of course.

daniel-koudouna:
I can pick this up if you want. What would be the expected behaviour on the name of the link? An idea could be:

- unchanged link names would be changed to the new name, i.e
[[file:cool_stuff.org][$cool_stuff]] --> [[file:cool_things.org][cool_things]]

- custom/changed link names would be the same, i.e
[[file:cool_stuff.org][my_stuff]] --> [[file:cool_things.org][my_stuff]]

jethrokuan:
@daniel-koudouna are you on the develop branch? Ideally:

- unchanged link names should be the title of the Org file if it exists, or the file name if not
- custom/changed link names should be the same

I'd expect this to be an advice on `rename-file`.

jethrokuan:
(tbh I'd really like to push for not renaming files to not be part of a workflow, but this seems like a common request)

grimsleydl:
Is using `ord-id` to keep track of things out of the question this far along/to keep dependencies down?

jethrokuan:
> [[id:735f13ef-60e5-43fc3-92c9-6adfhwey][cool_stuff]]

The links are to files, so `org-id` wouldn't help.

jdiez17:
Is the goal of this issue/PR #124 to be able to change the title of an org-roam file, and update the link text of anything that references it, to be the new title? In my opinion this would be a more common use case than renaming the actual underlying file, but let me know what you think.

jethrokuan:
I'd say neither use-cases should be common. But #124 at least makes sure that the cache remains a consistent view of the current state of the files.

jethrokuan:
I believe this should be resolved by #124 .

PR #124, opened 2020-02-18
Update backlinks on file rename

daniel-koudouna:
This is a work in progress to update files with links when a file is renamed, as per #36 and #119 . However, there are a few things I'm not sure about:

- I'm advising around `rename file`, which triggers when emacs saves temporary files. I've found `rename file` passes a flag which seems to indicate if it's a temporary file or not, but I couldn't find any documentation regarding this.
- Currently, the org-roam cache is updated only from the current buffer, which means the files have to be visited. This seems like undesirable behaviour, and I'm wondering if there's an alternative.
- I'm doing standard regex replace to update the links, but someone versed in org-mode might know a better method.

In addition, this PR changes `org-roam--org-roam-file-p` to an actual predicate by requiring a parameter, since it is used to check files other than the current buffer.


CeleritasCelery:
The third argument to `rename-file` is `OK-IF-ALREADY-EXISTS` which means that it won't throw a `file-already-exists` error.

daniel-koudouna:
> 
> 
> The third argument to `rename-file` is `OK-IF-ALREADY-EXISTS` which means that it won't throw a `file-already-exists` error.

I see. In that case, I've changed it to use `auto-save-file-name-p `.

org-roam.el
```diff
@@ -162,11 +162,15 @@ If called interactively, then PARENTS is non-nil."
...
-(defun org-roam--org-roam-file-p ()
-  "Return t if file is part of org-roam system, false otherwise."
-  (and (buffer-file-name (current-buffer))
-       (f-descendant-of-p (file-truename (buffer-file-name (current-buffer)))
-                          (file-truename org-roam-directory))))
+(defun org-roam--org-roam-file-p (file)
+  "Return t if FILE is part of org-roam system, false otherwise."
+  (and file
+       (f-child-of-p (file-truename file)
+                     org-roam-directory)))
```
CeleritasCelery: We need to use the truename of `org-roam-directory` since we are using the truename of the file.

org-roam.el
```diff
@@ -162,11 +162,15 @@ If called interactively, then PARENTS is non-nil."
     (org-roam--build-cache-async)
     (user-error "Your Org-Roam cache isn't built yet! Please wait")))
 
-(defun org-roam--org-roam-file-p ()
-  "Return t if file is part of org-roam system, false otherwise."
-  (and (buffer-file-name (current-buffer))
-       (f-descendant-of-p (file-truename (buffer-file-name (current-buffer)))
-                          (file-truename org-roam-directory))))
+(defun org-roam--org-roam-file-p (file)
```
CeleritasCelery: My personal preference would be to make `file` an optional argument. If it is not present then use `buffer-file-name` instead. That will make it easier in debugging and make the function call cleaner.

org-roam.el
```diff
@@ -587,6 +591,42 @@ This needs to be quick/infrequent, because this is run at
...
+                         (insert-file-contents file-from)
+                         (replace-regexp
+                          (format "\\[\\[file:%s\\]\\[%s\\]\\]" path old-title)
+                          (format "[[file:%s][%s]]" relative-path new-title))
+                         (beginning-of-buffer)
+                         (replace-regexp
+                          (format "\\[\\[file:%s\\]\\[\\(.*\\)\\]\\]" path)
+                          (format "[[file:%s][\\1]]" relative-path)))
+                       (find-file file-from)
+                       (org-roam--update-cache)))
```
CeleritasCelery: it seems to make more sense to only update the cache after we are done with all files.

org-roam.el
```diff
@@ -587,6 +591,42 @@ This needs to be quick/infrequent, because this is run at
...
+                       (with-temp-file file-from
+                         (insert-file-contents file-from)
+                         (replace-regexp
+                          (format "\\[\\[file:%s\\]\\[%s\\]\\]" path old-title)
+                          (format "[[file:%s][%s]]" relative-path new-title))
+                         (beginning-of-buffer)
+                         (replace-regexp
+                          (format "\\[\\[file:%s\\]\\[\\(.*\\)\\]\\]" path)
+                          (format "[[file:%s][\\1]]" relative-path)))
+                       (find-file file-from)
```
CeleritasCelery: Why are we opening the file here?

org-roam.el
```diff
@@ -587,6 +591,42 @@ This needs to be quick/infrequent, because this is run at
...
+                         (org-roam--get-title-or-slug new-path)))
+           (new-title (format org-roam-link-title-format new-slug)))
+      (when files
+        (save-excursion
+          (maphash (lambda (file-from contents)
+                     (let* ((relative-path (file-relative-name new-path (file-name-directory file-from))))
+                       (with-temp-file file-from
+                         (insert-file-contents file-from)
+                         (replace-regexp
+                          (format "\\[\\[file:%s\\]\\[%s\\]\\]" path old-title)
```
CeleritasCelery: rather then adding all these backslashes we can use `regexp-quote` to do that work for us.

org-roam.el
```diff
@@ -587,6 +591,42 @@ This needs to be quick/infrequent, because this is run at
...
+
+           (new-slug (or (org-roam--get-title-from-cache (file-truename file))
+                         (org-roam--get-title-or-slug new-path)))
+           (new-title (format org-roam-link-title-format new-slug)))
+      (when files
+        (save-excursion
+          (maphash (lambda (file-from contents)
+                     (let* ((relative-path (file-relative-name new-path (file-name-directory file-from))))
+                       (with-temp-file file-from
+                         (insert-file-contents file-from)
```
CeleritasCelery: Rather then using a temp file, we can use `find-file-noselect` which will prevent us from needing to reopen buffers we already have open.

org-roam.el
```diff
@@ -587,6 +591,42 @@ This needs to be quick/infrequent, because this is run at
...
+           (new-path (org-roam--relative-path new-file))
+
+           (slug (org-roam--get-title-or-slug file))
+           (old-title (format org-roam-link-title-format slug))
+
+           (new-slug (or (org-roam--get-title-from-cache (file-truename file))
+                         (org-roam--get-title-or-slug new-path)))
+           (new-title (format org-roam-link-title-format new-slug)))
+      (when files
+        (save-excursion
```
CeleritasCelery: Move this `save-excursion` inside the function. Just outside of the replace-regexp. It is not doing anything out here.

org-roam.el
daniel-koudouna: As I said above, the cache is updated only for the current buffer. So if we want the backlinks to be updated we need to call update for each file. Are you suggesting another map over the files?

org-roam.el
daniel-koudouna: Do we not have more control over secondary effects using `with-temp-file`? One of which is actually displaying the buffer. As I see it, it would be ideal if we would never opened a file.

org-roam.el
CeleritasCelery: ah, I didn't understand that.

org-roam.el
CeleritasCelery: `find-file-noselect` will not display the buffer, it just returns the buffer. 

If we don't want to open extra files we could either check if it is open already with `get-file-buffer` or use `with-temp-buffer`. But we don't have a need for a temporary file. 

```elisp
(with-temp-buffer
  (insert-file-contents file-from)
  ;; do stuff
  (write-file file-from))
```

org-roam.el
jethrokuan: +1

org-roam.el
daniel-koudouna: That would be ideal. However, I think some functionality relies on at least the buffer name being associated with a file, so opening the file might be unavoidable if we want to update the cache.

jethrokuan:
Thanks for the work @daniel-koudouna , and the additional pair of eyes @CeleritasCelery .

Things are looking good so far, but there are some more glaring issues:

1. The links for the original file are not cleared.
2. Nested files are not updated (probably because of the `f-child-of-p` but changing it and rerunning the tests on my end doesn't fix it)

I think the best way to go about this is to write tests and make sure they all pass. I've done up some preliminary tests for you in this patch:

https://gist.github.com/jethrokuan/240296d726ecf50a9c80bb6692ded683

but they have to be more comprehensive. The instructions for running the tests are:

```
make sandbox=.sandbox install-deps=t install-linters=t
make v=vv sandbox=.sandbox test
```

org-roam.el
jethrokuan: The path is not always relative to the `org-roam-directory`. It needs to be the relative path between the current file, and the linked file.

org-roam.el
```diff
@@ -587,6 +590,47 @@ This needs to be quick/infrequent, because this is run at
...
+                         (while (re-search-forward slug-regex  nil t)
+                           (replace-match (format "[[file:%s][%s]]" relative-path new-title)))
+                         (goto-char (point-min))
+                         (while (re-search-forward named-regex nil t)
+                           (replace-match (format "[[file:%s][\\1]]" relative-path))))
+                       (find-file file-from)
+                       (org-roam--update-cache)))
+                   files))))))
+
+(advice-add 'rename-file :after 'org-roam--rename-file-links)
```
jethrokuan: It's a bit inappropriate for the advice to out here like this, I think eventually `org-roam-mode` needs to be made a global mode and the advice to be added and removed via toggling the global mode. But I'm good with merging this as is first, and then fixing it later.

tests/test-org-roam.el
```diff
@@ -148,3 +149,44 @@
...
+
+              (expect (->> org-roam-forward-links-cache
+                           (gethash (abs-path "f3.org"))
+                           (member (abs-path "meaningful-title.org"))) :not :to-be nil)
+
+              ;; Links are updated with the appropriate name
+              (expect (with-temp-buffer
+                        (insert-file-contents (abs-path "f3.org"))
+                        (buffer-string)) :to-match (regexp-quote "[[file:meaningful-title.org][meaningful-title]]"))
+              ))
```
jethrokuan: could you shift the parens to the above line?

org-roam.el
```diff
@@ -587,6 +590,47 @@ This needs to be quick/infrequent, because this is run at
...
+           (path (file-truename file))
+           (new-path (file-truename new-file))
+           (slug (org-roam--get-title-or-slug file))
+           (old-title (format org-roam-link-title-format slug))
+           (new-slug (or (org-roam--get-title-from-cache path)
+                         (org-roam--get-title-or-slug new-path)))
+           (new-title (format org-roam-link-title-format new-slug)))
+      (when files
+        (save-excursion
+          (maphash (lambda (file-from contents)
```
jethrokuan: could you change `contents` to `properties`? it contains more than just contents.

daniel-koudouna:
One more comment: the cache is not updated for the new file, i.e forward-links-cache is empty for the new file. Is this the desired behaviour? (since it will update on `after-command-hook`) or should I explicitly update the cache in the rename function? See test f1 -> new_f1

jethrokuan:
> should I explicitly update the cache in the rename function?

I think you should. The cache is updated on `after-save-hook`, which the user may not call.

daniel-koudouna:
Thanks for the feedback everyone. I've also switched to use `save-window-excursion` to avoid landing on a different buffer after the function is done.

PR #130, opened 2020-02-19
Add org-ref link support

ahendriksen:
Hi, 

Thank you for this very useful library! I use `org-ref` extensively and I have converted all my bibliographical notes to the roam format. I got carried away today and implemented support for finding `org-ref` style backlinks to a note. This might be a useful addition for others as well. 

This pull request adds support for org-ref links of the form
`cite:bibtex-key`. The code assumes that the notes for a paper are
stored in
```
/org/roam/directory/<bibtex-key>.org
```

Now, when you visit the note file for a paper `bibtex-key.org`, you will see all notes that cite this paper in your roam directory. 

I wanted this for myself  :) 

Due to `async', it is difficult to override the `org-roam--parse-content' function in init.el. Or at least, I did not get it working. Also, async code is very particular about which packages are loaded. 

###### Modifications

The function `org-roam--parse-buffer` had to be modified quite
extensively because an org-ref link can link to multiple papers at
once. In these multi-links, the bibtex keys are separated by commas,
e.g., `cite:zhu-2017-unpair-image,isola-2016-image-tokl`.

I have added the function `org-roam--link-get-content` that obtains
the content of a link. This function is used when handling both file
links and org-ref style links, thus minimizing code repetition.

###### Current code issues
I am not sure if the code is fit to merge in its current form. 

Previously, the property `:point` of an org-roam item would be
determined by the `:content-begin` or `:begin` property of the
`org-element-at-point'. Now, the `:point` property of an org-roam item
is populated by the `:begin' property of the link. I do not know and
have not tested if this leads to observably different behaviour.

I have added a dependency on dash.el because I wanted to use
`-flatten-n`.

The file path of a note on a paper is hard-coded to be `bibtex-key.org` in the `org-roam-directory`. 

Please let me know if this pull request looks workable and if and how it could be merged :)

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -37,6 +37,9 @@
 (require 'org-element)
 (require 'subr-x)
 (require 'cl-lib)
+(require 'dash)
+;; Optionally load org-ref so that we can also find "cite:" links.
+(require 'org-ref nil t)
 
 (defun org-roam--file-name-extension (filename)
   "Return file name extension for FILENAME.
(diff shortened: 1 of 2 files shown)


frigge:
hey great and very useful addition. I just converted all my notes to a representation compatible with org-roam, including my paper notes. Unfortunately for now I have  used to default org-roam naming convention which named the files "<timestamp>_<paper-title>.org". Do you happen to have a quick idea of how to easily rename the files to the cite keys or would you be interested to add such a function to your pull request? I'd also be willing to code such a function myself. Unfortunately my elisp knowledge is very limited, so if you could point me in the right direction, that would be awesome.

ahendriksen:
Thanks! I used a keyboard macro to copy my previous notes (one long bibliography.org file) over to individual notes. I have put a minimal description in a [gist](https://gist.github.com/ahendriksen/d7ee96de4009f8db82c8bab4c18f31b4). The macro for sure needs fine-tuning for your use case, but it might point you in the right direction.

jethrokuan:
I too use `org-ref` extensively, and like this idea a lot!

Couple of changes I would like to see:

- Separating the citations from the back-links cache.
- Adding some test cases (test file with citations), checking the citations cache if it's properly populated

> The file path of a note on a paper is hard-coded to be bibtex-key.org in the org-roam-directory.

This is fine, can't think of any other way to do it without introducing org-roam-specific org syntax.

org-roam-utils.el
```diff
@@ -77,28 +80,48 @@ Like file-name-extension, but does not strip version number."
...
+		   (list :from file-path
+			 :to (file-truename (expand-file-name path (file-name-directory file-path)))
+			 :properties (list :content content :point start))))
+		 ((string= type "cite")
+		  ;; Org-ref "cite:" links. These are only recognized if org-ref is loaded.
+		  (mapcar
+		   (lambda (key)
+		     (list :from file-path
+			   :to (expand-file-name (concat key ".org") org-roam-directory)
+			   :properties (list :content content :point start)))
```
jethrokuan: add a new property for the kind of link: `:type 'citation` and process it differently in org-roam--insert-item.

org-roam-utils.el
```diff
@@ -77,28 +80,48 @@ Like file-name-extension, but does not strip version number."
...
+                    (org-element-property :begin element)))
+         (content (or (org-element-property :raw-value element)
+		      (buffer-substring
+		       begin
+		       (or (org-element-property :content-end element)
+                           (org-element-property :end element)))))
+         (content (string-trim content)))
+    content))
 
 (cl-defun org-roam--insert-item (item &key forward backward)
```
jethrokuan: add a new key `citations`, and a new variable containing the citations cache

brabalan:
Instead of using the file name to get the key, could one have a line like
```
#+BIBKEY: foo2020
```
in the file?

jethrokuan:
How about something like `#+ROAM_BIBKEYS: key1,key2`

ahendriksen:
Great idea! I think the `#+ROAM_BIBKEYS: cite:key1,key2` syntax is pretty neat. 

To implement this feature, I would have to dive deeper into the org-roam internals. Perhaps it is a good strategy to hard-code the file path and then implement the `#+ROAM_BIBKEYS` idea as an additional feature?

jethrokuan:
> Perhaps it is a good strategy to hard-code the file path and then implement the #+ROAM_BIBKEYS idea as an additional feature?

I'd still prefer not to store the file path in the cache in this case. I'm thinking the cache would like like:

```
(bibkey1) -> {"note1.org" -> [(content1 point1), (content2, point2)...}
```

In `org-roam-update` we can check either:
1. file name matches bibkey, or
2. there are keys in `ROAM_BIBKEYS`

The latter isn't hard to add, it's pretty much the same as the `extract-title` function. We can refactor that out to a `extract-global-prop`, replacing the hardcoded `TITLE` string. Then it's a matter of simple string manip (trimming away the cite, split by comma).

bdarcus:
Two questions, which be dumb, since I don't have a ton of experience with org-ref, but:

1. Why would the variable be specific to roam?
2. Also, what would be the relation between this and the "CUSTOM_ID" property org-ref seems to use?

jethrokuan:
1. Because it's not a feature required anywhere else, as with functions, these special keys should be namespaced too.
2. i'm not aware of any custom-id thing by org-ref, at least not in my years of usage...

bdarcus:
There's more on custom_id, and I guess, label, here:

https://github.com/jkitchin/org-ref/issues/173

I just noticed that was the property I had in my org bib notes.

jethrokuan:
I don't think we'll be doing anything with `label`, and `custom_id` or `name` here.

ahendriksen:
I believe I changed this pull request in line with the changes you requested. 

I have not yet added support for `#+ROAM_BIBKEYS: cite:key1,key2`, but that should indeed be pretty doable. 

Consider this work in progress. I will squash/rebase/etc when all the work is done..

jethrokuan:
I'll work the org-ref functionality in after it hits MELPA.

ahendriksen:
> I'll work the org-ref functionality in after it hits MELPA.

Thank you! I was looking at rebasing this weekend, but with the amazing amount of progress you have made in the mean time, a rebase appeared quite daunting. 

I can tell from experience now that having backlinks for org-ref citations is really great! So I look forward to having the feature integrated into the mainline. If there is anything I can do to help, please let me know :)

jethrokuan:
yeah I think rebasing would've been harder than rewriting :P thanks for the attempt, I'll have something to reference!

PR #132, opened 2020-02-19
allow customization of graphviz node appearance

frigge:
This pull request allows the customization of the appearance of exported graphviz nodes.

Two new customizable variables were introduced:
- `org-roam-graph-max-title-length` (default value: `100`)
- `org-roam-graph-node-shape` (default value: `ellipse`)

If the title length is longer than `org-roam-graph-max-title-length`, the title is shortened and `...` is concatenated to the end of the title.

`org-roam-graph-node-shape` is simply a string that is passed to the shape option of the node definition.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -116,6 +116,16 @@ If nil, always ask for filename."
   :type 'string
   :group 'org-roam)
 
+(defcustom org-roam-graph-max-title-length 100
+  "Maximum length of titles in graphviz graph nodes"
+  :type 'number
+  :group 'org-roam)
+
+(defcustom org-roam-graph-node-shape "elipse"
+  "Maximum length of titles in graphviz graph nodes"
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -557,21 +567,33 @@ This needs to be quick/infrequent, because this is run at
...
+	(insert "digraph {\n")
+	(dolist (file (org-roam--find-all-files))
+	  (insert
+	   (let ((title (org-roam--get-title-or-slug file)))
+		 (let ((shortened-title (if (> (length title) org-roam-graph-max-title-length)
+									(concat (substring
+											 title
+											 0
+											 (min (length title) org-roam-graph-max-title-length))
+											"...")
```
jethrokuan: Just use `s-truncate`, since we're already importing `s.el`.

PR #136, opened 2020-02-20
Handle nil file names

acowley:
I ran into trouble evaluating source blocks in an org document having nothing to do with org-roam. This function was encountering an error whenever I evaluated the source block due to `nil` being passed to `file-truename`. We can detect that, and avoid the trouble.


jethrokuan:
Thank you!

PR #137, opened 2020-02-20
inject "load-path" variables before building cache asychronously

hackmalife:
The error I faced is reported in [here](https://github.com/jethrokuan/org-roam/issues/131)
you need to inject load-path otherwise it does not find `org-roam-utils`

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -323,6 +323,7 @@ If PREFIX, downcase the title before insertion."
   (interactive)
   (async-start
    `(lambda ()
+      ,(async-inject-variables "load-path")
       (setq load-path ',load-path)
       (package-initialize)
       (require 'org-roam-utils)


jethrokuan:
This is honestly very odd to me. `load-path` is a variable that Emacs sets, so your Emacs configuration must be untradititional enough to have the 'load-path` not set correctly before `org-roam` is required.

hackmalife:
I found this [link](https://stackoverflow.com/a/22843310) relevant to this topic,
"async Emacs process has no access to the functions and variables from your main process"
maybe it is because I run emacs daemon and use emacsclient

jethrokuan:
> "async Emacs process has no access to the functions and variables from your main process"

Yes, I think what it does is it spawns a child Emacs instance, and runs the lambda. But I'd imagine that your spawned Emacs instance should have set up the `load-path` correctly, without inheriting it from the parent Emacs process.

>  maybe it is because I run emacs daemon and use emacsclient
This is what I use too.

hackmalife:
Anyways..that added line did the trick for me.
By the way, thanks for this awesome package!

l3kn:
Weird, my understanding of [`async-inject-variables`](https://github.com/jwiegley/emacs-async/blob/86aef2c38e7d35e8509b7feeee3e989d825eba91/async.el#L89) is that it does the same as `(setq var ',var)`  with an additional `symbol-value` to get the global (not the lexically bound) value.

@hackmalife Can you try running the following code to see if the two differ on your system?

``` lisp
(async-start
   `(lambda ()
      ,(async-inject-variables "load-path")
      (setq load-path2 ',load-path)
      (cons load-path load-path2))
   (lambda (res)
     (pp res)))
```

CeleritasCelery:
Also note that `async-inject-variables` takes a regex, so we would need to add symbol boundaries to the code `(async-inject-variables "\\_<load-path\\_>")`. But I agree with @l3kn that this seems odd and I don't think this is the root of the issue.

juergenhoetzel:
> seems odd and I don't think this is the root of the issue

You might want to `(setq async-debug t)` so you can see what's going on.

Issue #151, opened 2020-02-21
org-roam--build-cache-async fails to decrypt .gpg files


juergenhoetzel:
### Description
`org-roam--build-cache-async` fails to decrypt .gpg files when the corresponding key isn't cached in `gpg-agent` 

#### Steps to Reproduce

1. Load Emacs
2. Run `gpgconf --kill gpg-agent` command
2. Run `org-roam--build-cache-async`

#### Expected Results
Org-roam cache built!
#### Actual Results
```
Debugger entered--Lisp error: (file-error "Opening input file" "Decryption failed" "")                                                                                                                                                                                                                                                                                                                                                                                                      
  signal(file-error ("Opening input file" "Decryption failed" ""))                                                                                                                                                                                                                                                                                                                                                                                                                          
  async-handle-result((closure (t) (pair) (let ((directory (car pair)) (cache (cdr pair))) (setq org-roam-directory directory org-roam-forward-links-cache (plist-get cache :forward) org-roam-backward-links-cache (plist-get cache :backward) org-roam-titles-cache (plist-get cache :titles) org-roam-cache-initialized t)) (message "Org-roam cache built!")) (async-signal (file-error "Opening input file" "Decryption failed" "")) #<buffer *emacs*<9>>)                             
  async-when-done(#<process emacs> "finished\n")  
```
### Versions
- Emacs (`C-h v emacs-version`):  26.3
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/d02d48e

jethrokuan:
Yeah, this is a problem that can't be helped, there's no way to pass authentication credentials to the async lambda.

juergenhoetzel:
> Yeah, this is a problem that can't be helped, there's no way to pass authentication credentials to the async lambda.

My first test was done on a remote (SSH) system.

Actually it works very well in most cases.

The unlocking happens outside of Emacs:

`gpg` uses `pinentry`. `pinentry` in turn uses the following programs successively on my system:

- `pinentry-gtk-2`
- `pinentry-curses`

`pinentry-curses` can't work without a terminal. But `pinentry-gtk2` will work even on remote systems if `X-Forwaring` is active (async process has also set the DISPLAY variable). 

It even works with GPG agent forwarding. The only cases where it does not work is when

- `X-Forwarding` is decativated
- `No X libraries are available on the remote system`

juergenhoetzel:
[...]

PR #152, opened 2020-02-21
(bugfix): Remove lock in case of async errors (WIP)

juergenhoetzel:
Refs #149, #151

When [org-roam--build-cache-async fails to decrypt .gpg files](https://github.com/jethrokuan/org-roam/issues/151) there is currently no possibility for a 2nd attempt, because the lock `org-roam--ongoing-async-build` is still set.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -378,24 +378,27 @@ If PREFIX, downcase the title before insertion."
   "Builds the caches asychronously."
   (interactive)
   (unless org-roam--ongoing-async-build
-    (setq org-roam--ongoing-async-build t)
-    (async-start
-     `(lambda ()
-        (setq load-path ',load-path)
-        (package-initialize)
-        (require 'org-roam-utils)
-        ,(async-inject-variables "org-roam-directory")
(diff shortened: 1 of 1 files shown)


herbertjones:
`unwind-protect` will run the unwind forms on success and failure.  That would cause the lock flag to be unset immediately after setting it, error or not.  In that case, opening several buffers at once would again spawn several async processes.  Look at `condition-case` to see if that will resolve the issue.  If it doesn't, and the async library isn't executing the lambda when an error occurs, then instead of a flag we should save the async promise and only spawn new processes if that promise is still valid(assuming there is an API for that).

juergenhoetzel:
> Look at `condition-case` to see if that will resolve the issue.

Ineed: The error must be handled in the async  `START-FUNC`.

juergenhoetzel:
> we should save the async promise and only spawn new processes if that promise is still valid(assuming there is an API for that).

Yes, thats the correct way to do it: [(bugfix): Check future status to prevent multiple async tasks](https://github.com/jethrokuan/org-roam/pull/160)

PR #159, opened 2020-02-22
(feature): add `org-roam-mute-cache-build` to mute cache build message

### LINKED ISSUE #153: Silence "Org-roam cache built!" message
### Brief Abstract

"Org-roam cache built!" message should be silence-able.

### Long Description

`org-roam--build-cache-async` eventually emits "Org-roam cache built!" as a message, but it does so via a lambda so silencing it with advice is not something that I know how to do, if it's even possible. 
If the message must remain because it's valuable to some, then a flag to disable it would be nice.

jethrokuan:

Resolves #153

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -65,6 +65,11 @@ If nil, default to the org-roam-directory (preferred)."
   :type 'directory
   :group 'org-roam)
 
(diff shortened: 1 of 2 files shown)


aaronjensen:
Looks like this went away and a new message was introduced which is not mutable.

PR #160, opened 2020-02-22
(bugfix): Check future status to prevent multiple async tasks

juergenhoetzel:
Prevents that the `org-roam--build-cache-async` process is blocked forever if an async
error occurs.

Refs #149, #151

When [org-roam--build-cache-async fails to decrypt .gpg files](https://github.com/jethrokuan/org-roam/issues/151) there is currently no possibility for a 2nd attempt, because the lock `org-roam--ongoing-async-build` is still set.


jethrokuan:
This is very neat, thanks!

Issue #114, opened 2020-02-17
Feature: Customize the org template used for `org-roam-insert`


leothelocust:
Maybe this is already built in somewhere but missing documentation, but I'm hoping to configure the initial org file that is created when triggering the `org-roam-insert` function.

In essence, I might want something more fleshed out like the following:
```txt
#+TITLE:   My New Note
#+STARTUP: showall
#+AUTHOR:  blah blah
... etc

---

* Metadata:
** Tags: 
** Links:
... etc
```
instead of the default:
```txt
#+TITLE: My New Note
```

Thanks for the great project!!! Loving this so far!
-Levi

jethrokuan:
I think this is a great idea. I'd deprecate `org-roam-autopopulate-title`, and have something like `org-roam-insert-template` instead.

I already do what you're suggesting, in a convoluted fashion. I have this in my `init.el`:

```
(defun jethro/deft-insert-boilerplate ()
    (interactive)
    (let ((setupfile (jethro/org-get-prop "SETUPFILE"))
          (title (jethro/org-get-prop "TITLE")))
      (when (and (not setupfile)
                 (not (s-match "hugo_setup"
                               (file-name-sans-extension
                                (buffer-file-name (current-buffer)))))
                 (not (s-match "private"
                               (file-name-sans-extension
                                (buffer-file-name (current-buffer))))))
        (goto-char (point-min))
        (insert "#+SETUPFILE:./hugo_setup.org\n")
        (insert "#+HUGO_SECTION: zettels\n")
        (insert (format "#+HUGO_SLUG: %s\n" (jethro/title-to-id (or title "EDITME"))))
        (goto-char (point-max)))))
```

which is evaluated in the directory containing my org files in via directory local variables.

leothelocust:
Thanks @jethrokuan 

Okay, so I understand the dir-locals part, but how do you call this function? Do you just call `jethro/deft-insert-boilerplate` instead of `org-roam-insert`?

Also, may I ask what you have in `#+SETUPFILE: ./hugo_setup.org`?  I'm not familiar with the `SETUPFILE` variable.  Also, how do you keep that `hugo_setup.org` out of the search cache?

Thanks,
-Levi

jethrokuan:
It's called as part of the dir-locals setup:

https://github.com/jethrokuan/braindump/blob/master/org/.dir-locals.el

My `hugo_setup.org` is here:

[...]

PR #165, opened 2020-02-22
(feature) add org-roam-template

jethrokuan:
Adds `org-roam-templates`. This allows users to customize the filename, and the content of the template. Here's my configuration:

```emacs-lisp
(defun jethro/org-roam-title-private (title)
    (let ((timestamp (format-time-string "%Y%m%d%H%M%S" (current-time)))
          (slug (org-roam--title-to-slug title)))
      (format "private-%s_%s" timestamp slug)))
  (setq org-roam-templates
        (list (list "default" (list :file #'org-roam--file-name-timestamp-title
                                    :content "#+SETUPFILE:./hugo_setup.org
#+HUGO_SECTION: zettels
#+HUGO_SLUG: ${slug}
#+TITLE: ${title}"))
              (list "private" (list :file #'jethro/org-roam-title-private
                                    :content "#+TITLE: ${title}")))))
```

With this configuration, when creating a non-existent file, I'm prompted for the template I want, and creates the file with the appropriate contents.

I'm going to leave this up here for awhile, because this is a big change.

output-2020-02-23-01:45:32

Addresses #114.


benide:
Nice, I like this a lot! What do you think of having the `:content` string accept something like `$?` to represent where the point should be, kind of like with an org-capture template? This way, a template could have `bibliography:references.bib` or something similar at the end with the point starting above it so typing can begin immediately.

jethrokuan:
> What do you think of having the :content string accept something like $? to represent where the point should be, kind of like with an org-capture template?

would be nice, but would also require significant amount of elisp (i.e. writing the templating engine).

Issue #80, opened 2020-02-14
Possible to have more than one org-roam-directory?


anthrolisp:
Is it possible to have more than one org-roam-directory for more than one separate org-roam zettelkastens?

jethrokuan:
I'm unlikely to merge such functionality in, but you can write some simple elisp for this. Here's some pseudo-code:

```
((completing read for new org-roam-directory)
 (setq org-roam-directory new-dir)
 (org-roam--build-cache-async))
```

should work for the most part.

jethrokuan:
I changed my mind about this, because I think it's not a particularly abusable feature and it might make sense for some to keep things separate (but not to me). I'll see what I can do.

anthrolisp:
Great!!!! Thank you!!!!!

herbertjones:
I developed a hacky workaround for the time being.  It relies on helm, though it would be easy to switch that to `completing-read` if needed.

https://gist.github.com/herbertjones/9718942aac44cd9f0ee5d353ad015fe3

juergenhoetzel:
> Is it possible to have more than one org-roam-directory for more than one separate org-roam zettelkastens?

I myself am not a big fan of implementing this in org-roam myself.

There are already several packages for Emacs to handle multiple workspaces or projects. I use projectile for all my Project
Switching workspaces is wired in my muscle memory.

 in this case the functionality can be easily implemented using a [projectile-after-switch-project-hook](https://projectile.readthedocs.io/en/latest/configuration/)

The main advantage of Emacs is its extensibility and adaptability to your own workflow instead of  imposing a predefined workflow.

juergenhoetzel:
How about adding another indirection to the cache: Using a new var `org-roam-current-dir`

Another new function `org-roam-set-current-dir`  could be called interactively or in a hook by `projectile`  to set this indirection.

jethrokuan:
@herbertjones can I have you document how to enable multiple org-roam-directories in the configuration section? After which I think we can close this issue.

anthrolisp:
Whenever I visit a file in the second org-roam instance, I keep getting the following prompt. It works for that session. But even after I answer with a "!", it re-prompts me the next time I open Emacs and visit a file in this directory. Any ideas?

The local variables list in /path/to/here/
contains values that may not be safe (*).Do you want to apply it?  You can type
y  -- to apply the local variables list.
n  -- to ignore the local variables list.
!  -- to apply the local variables list, and permanently mark these
[...]

PR #166, opened 2020-02-22
(feature): support multiple org-roam directories WIP

herbertjones:
ref #80

Adds functionality to org-roam-directory to permit multiple directories.  Which directory is used it based on context.  When there is insufficient context, the first directory is used. (Maybe it would be better to allow the user to pick when there isn't sufficient context, or make it customizable.)   There is a new command org-roam-find-file-pick-dir that uses completing read to choose the directory to use first.

I have multiple contexts that I use org-roam in.  I would like to use org-roam concurrently with my various repositories of org files.  At the same time, I want to keep them separate, as work files shouldn't link to my personal files and vice-versa.

###### Note
This is a big change, I've tested basic functionality and haven't found anything yet that isn't working, but I'd be wary to merge it in without multiple people testing out their own use cases.

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -137,12 +137,12 @@ ITEM is of the form: (:from from-path :to to-path :properties (:content preview-
         (org-element-property :value kw)))
     :first-match t))
 
-(defun org-roam--build-cache (dirs)
+(defun org-roam--build-cache (dir)
   "Build the org-roam caches in DIR."
   (let ((backward-links (make-hash-table :test #'equal))
         (forward-links (make-hash-table :test #'equal))
         (file-titles (make-hash-table :test #'equal)))
-    (let* ((org-roam-files (mapcan #'org-roam--find-files dirs))
+    (let* ((org-roam-files (org-roam--find-files dir))
            (file-items (mapcar (lambda (file)
                                  (with-temp-buffer
                                    (insert-file-contents file)
@@ -160,9 +160,13 @@ ITEM is of the form: (:from from-path :to to-path :properties (:content preview-
             (puthash file title file-titles)))
         org-roam-files))
     (list
+     :directory dir
      :forward forward-links
      :backward backward-links
      :titles file-titles)))
+
+(defun org-roam--build-cache-from-dirs (dirs)
+  (mapcar #' org-roam--build-cache dirs))
 
 (provide 'org-roam-utils)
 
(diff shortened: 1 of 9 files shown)


jethrokuan:
Could you add test cases for this? Have 2 org-roam directories, and replicate some of the basic tests and make sure they still work.

jethrokuan:
If it's not too much trouble, can this be shifted such that files from different org-roam-directories are stored in different caches? The reason is that when new features are added, I don't want to have to deal with the logic of handling multiple org-roam-directories. The changes here seem a bit invasive. 
All functions should only operate on the current active cache.

jethrokuan:
I really don't like the `&optional context` everywhere. What do you think about this approach:

1. Have a variable `org-roam-current-cache` indicating the currently active current cache
2. All functions, e.g. `org-roam-find-file`, operate only on the current cache.
3. To handle switching of directories, have a function that sets the variable of the current cache. to call `org-roam-find-file` with a different cache, have a macro:

```
(with-current-cache cache-id
  (org-roam-find-file))
```

juergenhoetzel:
> 3\. To handle switching of directories, have a function that sets the variable of the current cache. to call `org-roam-find-file` with a different cache, have a macro:

i would prefer to simply fall back on solid [solutions outside of org-roam](https://github.com/jethrokuan/org-roam/issues/80#issuecomment-590053303) and not reinvent the wheel and thus unnecessarily complicate the code base.

jethrokuan:
@juergenhoetzel I refer to [your comment](https://github.com/jethrokuan/org-roam/issues/80#issuecomment-590053303)

> in this case the functionality can be easily implemented using a projectile-after-switch-project-hook

The only problem with this is that then each time the project is switched, the cache would have to be rebuilt completely from scratch for that project, since org-roam only supports 1 cache directory at the moment.

I too don't really like the idea of this, and would like the changeset to be as minimal as possible, which is why I suggested the above approach.

herbertjones:
I've ported the code to have multiple caching directories to another branch(6957260f18ba2f94f999881d2d7d0dec446207ec) and may make another PR based off it.

juergenhoetzel:
> The only problem with this is that then each time the project is switched, the cache would have to be rebuilt completely from scratch for that project, since org-roam only supports 1 cache directory at the moment.

I see :+1:  i continue the discussion in [Possible to have more than one org-roam-directory?
](https://github.com/jethrokuan/org-roam/issues/80)

Issue #31, opened 2020-02-08
Feature: Finding unlinked references


jethrokuan:
For the currently open org-roam file, we can grep other files for the current file's title, to find any matches. Display any matches below the backlinks. One-click to replace the match with a file link to the current file.

dit7ya:
I have an ugly and basic but working implementation of this using `ripgrep`. If you think this implementation is going in the correct direction I will improve upon the code and open a PR. 

``` emacs-lisp

(defun org-roam-buffer--insert-unlinked ()
  "Insert all unlinked references in the org-roam-directory to the current buffer title."
  (insert "\n\n* Unlinked References \n\n")
  (setq-local rg-command (format "rg --pretty --pcre2 -i -t org '(?<!\\[)%s(?!\\.org)'" (org-roam--get-title-or-slug (buffer-file-name org-roam-buffer--current)))) ;; TODO exclude #+TITLE lines
  (setq-local srch-result (format "%s" (shell-command-to-string rg-command))) ;; TODO this needs to be an asynchronous process rather 
  (setq-local srch-result-list (s-split "\n\n" srch-result))
  (dolist (elem srch-result-list)
    (setq-local flname
                (car(s-split-up-to "\n" elem 1)))
    (insert (format "** [[file:%s][%s]] \n\n"
                    flname (org-roam--get-title-or-slug flname)))
    (setq-local content
                (car(cdr (s-split-up-to "\n" elem 1))))
    (dolist (ln (s-split "\n" content))
      ;; TODO truncate very long lines
      (setq-local ln-num
                  (car (s-split-up-to ":" ln 1))) ;; TODO use this ln-num to propertize point
      (setq-local ln-content
                  (car (cdr (s-split-up-to ":" ln 1))))
      (insert (format "%s\n\n" 
                      (s-trim (replace-regexp-in-string "^\\**" "" ln-content)))))))
```

d12frosted:
🤔 I think it should be open, no?

bbigras:
Yeah bad bot :robot: . Maybe closing stale issue should take longer than that.

bbigras:
I'm not mad bot. I'm disappointed.

PR #170, opened 2020-02-23
WIP (feature): Add support for unlinked references

daniel-koudouna:
This PR follows the motivation of #31 to add unlinked references to the org roam buffer. This also allows to covert the unlinked reference into a link by clicking on the entry. Some notes:

- The files are read on `org-roam-update`, which might be underisable. This can easily be converted to use a cache similar to backlinks.
- There is a list of `org-roam-unwanted-regexps` which can be modified for a user's workflow. An example included would be to ignore source blocks.
- A particular regexp is matched against the file, which includes the file name, the title and the lower case title (separated by word boundary). Suggestions or omissions very welcome.
- Format of the unlinked references (including count) mirror the backlink behaviour.
- Regex is used throughout. There could be better solutions. There are already some issues with getting the unformatted point vs the formatted point. There may be an easy solution to this to accomodate any possible formatting.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -494,18 +494,25 @@ This is equivalent to removing the node from the graph."
 Bindings:
 \\{org-roam-backlinks-mode-map}")
 
-(define-key org-roam-backlinks-mode-map [mouse-1] 'org-roam-jump-to-backlink)
-(define-key org-roam-backlinks-mode-map (kbd "RET") 'org-roam-jump-to-backlink)
+(define-key org-roam-backlinks-mode-map [mouse-1] 'org-roam-jump-to-or-create-backlink)
+(define-key org-roam-backlinks-mode-map (kbd "RET") 'org-roam-jump-to-or-create-backlink)
 
-(defun org-roam-jump-to-backlink ()
-  "Jumps to original file and location of the backlink content snippet at point"
+(defun org-roam-jump-to-or-create-backlink ()
+  "Jumps to original file and location of the backlink content snippet at point. Creates a backlink if the appropriate information is present."
   (interactive)
   (let ((file-from (get-text-property (point) 'file-from))
-        (p (get-text-property (point) 'file-from-point)))
+        (p (get-text-property (point) 'file-from-point))
+        (file-to (get-text-property (point) 'file-to))
+        (text-start (get-text-property (point) 'text-start))
+        (text-end (get-text-property (point) 'text-end)))
     (when (and file-from p)
       (find-file file-from)
       (goto-char p)
-      (org-show-context))))
+      (org-show-context)
+      (when (and file-to text-start text-end)
+        (let ((text (delete-and-extract-region text-start text-end))
+              (path (file-relative-name (file-truename file-to) (file-name-directory (file-truename file-to)))))
+          (insert (format "[[file:%s][%s]]" path text)))))))
 
 (defun org-roam--find-file (file)
   "Open FILE in the window `org-roam' was called from."
@@ -540,6 +547,7 @@ Bindings:
                             (org-toggle-link-display)
                             (while (re-search-forward match-regexp nil t)
                               (let* ((p (match-beginning 0))
+                                     (e (match-end 0))
                                      (context (buffer-substring (line-beginning-position) (line-end-position)))
                                      (rep (format "[[file:%s][\\&]]" other-file))
                                      (content (replace-regexp-in-string match-regexp rep context)))
@@ -551,6 +559,7 @@ Bindings:
                                                   unwanted-matches))
                                   (push (list
                                          :point p
+                                         :end e
                                          :content content)
                                         matches))
                                 (org-toggle-link-display)))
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -506,6 +522,53 @@ Bindings:
...
+                                         :end e
+                                         :content content)
+                                        matches))
+                                (org-toggle-link-display)))
+                            (when (> (length matches) 0)
+                              (list
+                               :file-from other-file
+                               :buffer (buffer-string)
+                               :matches matches))))))))
+    (delete nil (mapcar find-func (org-roam--find-all-files)))))
```
jethrokuan: This is slow. Can we fork this out to some shell program (rg, grep etc.)? Asynchronously perform the search, and update the buffer.

daniel-koudouna: That is one solution. However, if we did that, we would miss out on some benefits like knowing the precise point in the buffer for the link. Would it not be better to do it asynchronously in emacs itself (e.g using async)?

jethrokuan: > we would miss out on some benefits like knowing the precise point in the buffer for the link.

Most search programs return row and column number, I think that could be used to go to point as well.

> Would it not be better to do it asynchronously in emacs itself (e.g using async)?

Yeah, this is also a good solution.

daniel-koudouna:
I've updated the PR to use async. What would be a good way to ensure that, when the async function ends, the user hasn't navigated to another org-roam file (and hence upate the buffer with old information)?

org-roam.el
```diff
@@ -607,7 +617,16 @@ If item at point is not org-roam specific, default to Org behaviour."
         (org-roam--find-file file-from)
         (goto-char p)
         (org-show-context)
+        ;; Org-roam unlinked reference
+        (when-let ((file-to (get-text-property (point) 'file-to))
+                   (text-start (get-text-property (point) 'text-start))
+                   (text-end (get-text-property (point) 'text-end)))
+          (let ((text (delete-and-extract-region text-start text-end))
+                (path (file-relative-name (file-truename file-to) (file-name-directory (file-truename file-to)))))
+            (insert (format "[[file:%s][%s]]" path text))))
```
jethrokuan: Extract this logic into its own function

org-roam.el
```diff
@@ -619,6 +638,53 @@ If item at point is not org-roam specific, default to Org behaviour."
...
+                        (with-temp-buffer
+                          (insert-file-contents other-file)
+                          (let ((matches)
+                                (unwanted-matches))
+                            (dolist (unwanted-regexp org-roam-unwanted-regexps)
+                              (while (re-search-forward unwanted-regexp nil t)
+                                (push (list :start (match-beginning 0) :end (match-end 0)) unwanted-matches))
+                              (goto-char (point-min)))
+                            (org-mode)
+                            (org-toggle-link-display)
```
jethrokuan: Are these really needed?

org-roam.el
```diff
@@ -607,7 +617,16 @@ If item at point is not org-roam specific, default to Org behaviour."
...
+        ;; Org-roam unlinked reference
+        (when-let ((file-to (get-text-property (point) 'file-to))
+                   (text-start (get-text-property (point) 'text-start))
+                   (text-end (get-text-property (point) 'text-end)))
+          (let ((text (delete-and-extract-region text-start text-end))
+                (path (file-relative-name (file-truename file-to) (file-name-directory (file-truename file-to)))))
+            (insert (format "[[file:%s][%s]]" path text))))
         (throw 'ret t))
+
+
```
jethrokuan: Remove these whitespace

org-roam.el
```diff
@@ -659,7 +725,36 @@ If item at point is not org-roam specific, default to Org behaviour."
...
+                (setq load-path ',load-path)
+                (package-initialize)
+                (require 'org-roam)
+                (org-roam--find-unlinked-references ',file-path ',buffer-title ',(org-roam--find-all-files)))
+             (lambda (unlinked-references)
+               (let ((inhibit-read-only t))
+                 (with-current-buffer org-roam-buffer
+                   (if (> (length unlinked-references) 0)
+                       (progn
+                         (insert (format "\n\n* %d Unlinked references\n"
```
jethrokuan: Rather than inserting, maybe do a replace on everything under the "unlinked references" headline. Have a text property on the headline so you can check if the async return is for the correct file. 

```
(when (is-correct-file-p)
  (replace all content under "unlinked references" headline))
```

org-roam.el
daniel-koudouna: The issue is the location of the point. If you have any links in the file, `with-temp-buffer insert-file-contents` does not collapse them. This means the point is relative to the **raw** file. However, users see the formatted file, so the point will point in a different place, e.g:

```
[[linkhere][text]]

(point)unlinked here

some text after
```

```
_text_

unlinked here

som(point)e text after
```

 Is there a way to get the location of the text in another way?

jethrokuan:
Now that async is gone, my gut feel is that we should use not bring it back in, and use a tool like ripgrep instead. This kind of search can be easily parallelized by the system, whereas using `async` will result in a sequential runthrough.

daniel-koudouna:
I won't have time to work on this at the moment, but feel free to pick up where I left off or otherwise. As I said above, I don't know of an easy way to get the relative location of the point for the user without staying within emacs.

PR #173, opened 2020-02-24
Change "develop" to "master" in installation.md

visika:
As a newcomer, I spent hours trying to figure out why links didn't format properly, until I discovered that it was because I was on develop branch. Switching to master solved all my pains. I hope others don't have to figure it out in the future, maybe just put a disclaimer somewhere in the page :smiley:

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -8,7 +8,7 @@ The recommended method is using [use-package][use-package] and
       :after org
       :hook 
       (after-init . org-roam-mode)
-      :straight (:host github :repo "jethrokuan/org-roam" :branch "develop")
+      :straight (:host github :repo "jethrokuan/org-roam" :branch "master")
       :custom
       (org-roam-directory "/path/to/org-files/")
       :bind (:map org-roam-mode-map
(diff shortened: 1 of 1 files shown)


jethrokuan:
> spent hours trying to figure out why links didn't format properly

In what way?

> Switching to master solved all my pains.

Bugfixes are still streaming in: `master` isn't exactly stable. So I'd still recommend the `develop` branch, at least for now.

visika:
> In what way?

When I insert a link to a note it stays like this [[file:note.org][note]], black, not underlined. If I remove the column it formats properly and is clickable, but it's broken obviously. I thought initially it had to do with customize-face etc. (I'm pretty noob with emacs) but the behaviour doesn't change.

On the other side, on master the links get clickable.

jethrokuan:
Maybe your version of Org-mode is too old. This needs Org version at least `9`, which was released more than 3 years ago, but doesn't ship with some versions of Emacs.

visika:
Wow, I didn't expect that! I updated with straight and now it's all fine :rocket: I think I can close this, thanks a lot for the help!

alphapapa:
@jethrokuan You can declare Org 9 in the `Package-Requires` header.

PR #178, opened 2020-02-24
Allow dir locals override dir

herbertjones:
ref #80 

I have multiple contexts that I use org-roam in. I would like to use org-roam concurrently with my various repositories of org files. At the same time, I want to keep them separate, as work files shouldn't link to my personal files and vice-versa.

###### Summary of changes
This commit gives each org-roam-directory its own cache.  This was anyone that wants to have multiple org-roam directories can do so by configuring a `.dir-locals.el` for the extra repository.

Example .dir-locals.el file:
```lisp
((org-mode . ((eval . (setq-local org-roam-directory (expand-file-name "./"))))))
```

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -212,25 +212,25 @@ as a unique key."
   (setf (alist-get (org-roam-directory-normalized)
                    org-roam--cache nil nil #'equal) data))
 
-(defun org-roam-cache-initialized ()
+(defun org-roam--cache-initialized-p ()
   "Is cache valid?"
   (oref (org-roam--get-directory-cache) initialized))
 
-(defun org-roam-forward-links-cache ()
+(defun org-roam--forward-links-cache ()
   "Cache containing forward links."
   (oref (org-roam--get-directory-cache) forward-links))
 
-(defun org-roam-backward-links-cache ()
+(defun org-roam--backward-links-cache ()
   "Cache containing backward links."
   (oref (org-roam--get-directory-cache) backward-links))
 
-(defun org-roam-titles-cache ()
+(defun org-roam--titles-cache ()
   "Cache containing titles for org-roam files."
   (oref (org-roam--get-directory-cache) titles))
 
 (defun org-roam--ensure-cache-built ()
   "Ensures that org-roam cache is built."
-  (unless (org-roam-cache-initialized)
+  (unless (org-roam--cache-initialized-p)
     (org-roam--build-cache-async)
     (user-error "Your Org-Roam cache isn't built yet! Please wait")))
 
@@ -245,9 +245,9 @@ as a unique key."
 
 (defun org-roam--get-title-from-cache (file)
   "Return title of `FILE' from the cache."
-  (or (gethash file (org-roam-titles-cache))
+  (or (gethash file (org-roam--titles-cache))
       (progn
-        (unless (org-roam-cache-initialized)
+        (unless (org-roam--cache-initialized-p)
           (user-error "The Org-Roam caches aren't built! Please run org-roam--build-cache-async"))
         nil)))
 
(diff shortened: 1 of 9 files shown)


jethrokuan:
This is much nicer, thank you. @anthrolisp any thoughts?

bdarcus:
> This was anyone that wants to have multiple org-roam directories can do so by configuring a `.dir-locals.el` for the extra repository.

My understanding of lisp and emacs is really basic, but wouldn't it make more sense from a user perspective to simply redefine the org-roam directory variable to be a list?

Instead of having to create a special file?

jethrokuan:
>  but wouldn't it make more sense from a user perspective to simply redefine the org-roam directory variable to be a list?

But there is only one active org-roam directory at a time. Having the org-roam-directory point to the current active directory makes sense, and keeps the codebase simple.

herbertjones:
@jethrokuan Now when a file is opened and the cache is uninitialized, an async build will start and the buffer will have the org-roam hooks added once the build is complete.

org-roam.el
```diff
@@ -609,9 +669,21 @@ Applies `org-roam-link-face' if PATH correponds to a Roam file."
 (defun org-roam--find-file-hook-function ()
   "Called by `find-file-hook' when `org-roam-mode' is on."
   (when (org-roam--org-roam-file-p)
-    (org-link-set-parameters "file" :face 'org-roam--roam-link-face)
-    (add-hook 'post-command-hook #'org-roam--maybe-update-buffer nil t)
-    (add-hook 'after-save-hook #'org-roam--update-cache nil t)))
+    (setq org-roam-last-window (get-buffer-window))
+    (if (org-roam-cache-initialized)
+        (org-roam--setup-found-file)
+      (org-roam--build-cache-async
```
jethrokuan: what happens when you switch to multiple files in the same org-roam-directory with the cache still unbuilt? Will some of them be missing the hooks?

I think it's okay to setup the hooks first. `org-roam--update-cache` and `org-roam--maybe-update-buffer` can be a no-op when the caches aren't yet initialized.

herbertjones: Unfortunately in testing it where the hooks are added first, I get this message if I try to add the hooks before the cache is built:
```
Error in post-command-hook (org-roam--maybe-update-buffer): (user-error "Your Org-Roam cache isn’t built yet! Please wait")
```

However, if I add a check for initialization inside of `org-roam--maybe-update-buffer`, this fixes the problem and I can move the hooks.  This also likely fixes the issue I've experience in testing where a buffer that has the hooks and the cache doesn't get built causes emacs to not be able to change buffers or exit.

jethrokuan: Yep, I think `org-roam--maybe-update-buffer` can also be a no-op if the caches aren't initialized.

org-roam.el
```diff
@@ -648,7 +648,8 @@ This needs to be quick/infrequent, because this is run at
     (when (and (or redisplay
                    (not (eq org-roam--current-buffer buffer)))
                (eq 'visible (org-roam--current-visibility))
-               (buffer-local-value 'buffer-file-truename buffer))
+               (buffer-local-value 'buffer-file-truename buffer)
+               (org-roam-cache-initialized))
```
jethrokuan: move this check up? it might be cheaper.

herbertjones: I'm not sure if it is, but I'm made the switch.  I'll think about how to optimize the cache checks better.  There should be a better way than doing a string comparison.  Maybe the lookup is done once per buffer and then the cache object is made buffer local?  I'd have to not replace the cache object but rather update them.

jethrokuan: just a small nit, can be name helper predicate functions like `org-roam--cache-initialized-p`

org-roam.el
```diff
@@ -159,29 +160,77 @@ If called interactively, then PARENTS is non-nil."
...
   "Last window `org-roam' was called from.")
 
+(defvar org-roam--cache nil
+  "The list of cache separated by directory.")
+
 ;;; Utilities
+(defun org-roam-directory-normalized ()
+  "Get the org-roam-directory normalized so that it can be used
+as a unique key."
+  (s-chop-suffix "/" (file-truename org-roam-directory)))
```
jethrokuan: This might flake on windows, which uses forward slashes for directories. Use `directory-file-name`?

PR #196, opened 2020-02-27
(fix) Change locals to use hash tables for emacs 25

daniel-koudouna:
Pull request #178 regarding multiple org directories and following PRs use `alist-get` with the `equal` test function. Unless there is a good reason for association lists over hash tables, we could achieve the same effect with `gethash` and `puthash` to support emacs 25, since the `alist-get` with a test function argument was introduced in emacs 26.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -182,7 +182,7 @@ If called interactively, then PARENTS is non-nil."
 (defvar org-roam-last-window nil
   "Last window `org-roam' was called from.")
 
-(defvar org-roam--cache nil
+(defvar org-roam--cache (make-hash-table :test 'equal)
   "The list of cache separated by directory.")
 
 (defvar-local org-roam--local-cache-ref nil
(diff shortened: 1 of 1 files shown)


nivekuil:
This seems to have broken `org-roam--build-cache-async`:
```
Debugger entered--Lisp error: (wrong-type-argument hash-table-p #<process emacs>)
  gethash("/my/roam/directory" #<process emacs> nil)
  org-roam--build-cache-async()
  funcall-interactively(org-roam--build-cache-async)
  call-interactively(org-roam--build-cache-async record nil)
  command-execute(org-roam--build-cache-async record)
  helm-M-x-execute-command(org-roam--build-cache-async)
  helm-execute-selection-action-1()
  helm-execute-selection-action()
  helm-internal(...)
  helm(...)
  helm-M-x(nil)
  funcall-interactively(helm-M-x nil)
  call-interactively(helm-M-x nil nil)
  command-execute(helm-M-x)
```

jethrokuan:
you'll have to restart emacs after that change.

Issue #189, opened 2020-02-26
Powerup `roam://` links for note creation


jethrokuan:
### Brief Abstract
Create sub-protocols for `roam://` links, to facilitate note creation.

### Long Description

Now we're only using `roam://` links to open files. We can do much more.

For example, `roam://template://https://....` links create or open a note for the given template key. This can be used for quickly creating or opening notes for web-pages.

### Proposed Implementation (if any)

I haven't given this thorough thought yet, but the general idea is to follow `org-protocol`, in having sub-protocols.

### Please check the following:
- [x] No similar feature requests

jethrokuan:
Closed with #203

PR #203, opened 2020-02-28
(feature): deprecate roam-protocol, extend org-protocol instead

jethrokuan:
Rather than having our own protocol, we use org-protocol's underlying functionality and extend org-protocol instead.

We add 2 custom handlers:

1. `roam-file?file=path`: this simply opens the file at `path` in Emacs.
2. `roam-ref?ref=ref&template=roam-template&title=title&...`: attempts to open a roam note with a given `ROAM_KEY`. If the note doesn't exist, create one. Else, open it.

If you're curious, I uploaded a video of it on Twitter:
https://twitter.com/jethroksy/status/1233408971437821952?s=09

See #189

### DIFF
--- a/doc/roam_protocol.md
+++ b/doc/roam_protocol.md
@@ -1,5 +1,33 @@
-The setup is the same as org-protocol. Here `roam://` links are
-defined, and need to be associated with an application. 
+## What is Roam protocol?
+
+Org-roam defines several protocols that help boost productivity, by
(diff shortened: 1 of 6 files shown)


kljohann:
@jethrokuan I think passing the URL via `url=` instead of `ref=` is more common with `org-protocol` and also supported by some existing browser extensions. Is this something you might consider changing / introducing an option for?

PR #204, opened 2020-02-29
(bugfix): Only match protocol at beginning of URL

juergenhoetzel:
Prevents handling of non-roam links like org-roam:pros.txt

Executing

```
emacsclient org-roam:pros.txt
```
results in a text file `pros.txt` instead of `org-roam:pros.txt`


jethrokuan:
This is going away in #203

PR #212, opened 2020-02-29
Documentation: Installing with Doom Emacs

teesloane:

Partial contribution to https://github.com/jethrokuan/org-roam/issues/209: Add notes on installing `org-roam` with Doom Emacs.

---

Let me know if this is of use / needs work. Thank you for your hard work on this package!


doc/installation.md
```diff
@@ -88,3 +88,28 @@ If you are using Spacemacs, you can easily install org-roam by creating a simple
...
+
+```elisp
+;; ~/.doom.d/config.el
+(use-package! org-roam
+  :hook (after-init . org-roam-mode)
+  :config
+  (setq-default
+   org-roam-directory "~/<path>/<to>/<roam_directory>"
+   org-roam-link-title-format "%sº" ;; appends a  `º` to each Roam link.
+  (org-roam-build-cache))
```
jethrokuan: Not needed anymore

doc/installation.md
```diff
@@ -88,3 +88,28 @@ If you are using Spacemacs, you can easily install org-roam by creating a simple
...
+Subsequently, in your `~/.doom.d/config.el` file, configure org-roam:
+
+```elisp
+;; ~/.doom.d/config.el
+(use-package! org-roam
+  :hook (after-init . org-roam-mode)
+  :config
+  (setq-default
+   org-roam-directory "~/<path>/<to>/<roam_directory>"
+   org-roam-link-title-format "%sº" ;; appends a  `º` to each Roam link.
```
jethrokuan: Maybe leave this out, seems like personal preference. Also there's a less invasive option of setting the font face.

doc/installation.md
```diff
@@ -88,3 +88,28 @@ If you are using Spacemacs, you can easily install org-roam by creating a simple
...
+
+If you are using [Doom Emacs](https://github.com/hlissner/doom-emacs), configure packages as explained in the [getting started](https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#configuring-packages) guide. 
+
+Declare org-roam as a package in your `~/.doom.d/packages.el`:
+
+```elisp
+;; ~/.doom.d/packages.el
+
+(package! org-roam
+  :recipe (:host github :repo "jethrokuan/org-roam" :branch "develop"))
```
asymmetric: Is the `:branch` necessary, if `develop` is the default branch?

doc/installation.md
asymmetric: Do we need `org-roam--build-cache-async` instead?

doc/installation.md
```diff
@@ -88,3 +88,28 @@ If you are using Spacemacs, you can easily install org-roam by creating a simple
...
+```
+
+Subsequently, in your `~/.doom.d/config.el` file, configure org-roam:
+
+```elisp
+;; ~/.doom.d/config.el
+(use-package! org-roam
```
asymmetric: Do we need a `:after org` here? 

/cc @hlissner

doc/installation.md
jethrokuan: for now yes, I'll probably change it again when I push out 1.0.0.

doc/installation.md
jethrokuan: That function no longer exists in `develop`.

doc/installation.md
jethrokuan: No. If you wish to have the bindings available before Org is loaded, then you can leave out `:after org`.

teesloane:
> Are there no sane bindings? Don't use modal editing, so not sure how to do it.

It depends, I think. Doom manages a lot of it's bindings in a single place using the `(map! ...)` feature. I tend to mimic that, customizing my own block of bindings in a single place in my config. With that said, I've moved to add a "map!" usage into the use-package! declaration. It's much clearer anyway, so I'll prob start doing that in the future.

jethrokuan:
> I guess bindings can clash too, so maybe it's worth mentioning that?

Yeah probably is worth mentioning!

bdarcus:
I'm not sure if this is ideal, but this is what I have for the bindings.

```elisp
(map!
 :map (org-mode-map)
 :localleader
 (:prefix ("R" . "org-roam")
   "r" #'org-roam
   "f" #'org-roam-find-file
   "i" #'org-roam-insert
   "g" #'org-roam-show-graph))
```

teesloane:
@bdarcus That looks nice too -- I didn't know you could use the `:map` symbol like that. I grouped my bindings in with `notes` so that I could easily jump to any note from any mode ¯\_(ツ)_/¯

jethrokuan:
> I'm not sure if this is ideal, but this is what I have for the bindings.
> 
> ```emacs-lisp
> (map!
>  :map (org-mode-map)
>  :localleader
>  (:prefix ("R" . "org-roam")
>    "r" #'org-roam
>    "f" #'org-roam-find-file
>    "i" #'org-roam-insert
>    "g" #'org-roam-show-graph))
> ```

Doesn't this mean that the functions are only available in org-mode? The best is to bind them to the global `org-roam-mode`.

doc/installation.md
```diff
@@ -88,3 +88,34 @@ If you are using Spacemacs, you can easily install org-roam by creating a simple
...
+  :hook (after-init . org-roam-mode)
+  :config
+  (setq-default
+   org-roam-directory "/path/to/org-files/")
+  ;; Note bindings could clash / overwrite Doom Bindings. 
+  (map!
+      (:leader
+        (:prefix-map ("n" . "notes")
```
hlissner: I suggest `:prefix "n"` instead. `:prefix-map` shouldn't be used directly (it's purpose to define a prefix keymap, which isn't useful for the end user to do).

Also, no need to label a prefix that already has a label.

doc/installation.md
hlissner: If we're binding leader keys, we could bind them early and activate org-roam-mode when any of those commands autoload the package. i.e.

```elisp
(use-package! org-roam
  :commands (org-roam-insert org-roam-find-file org-roam)
  :init 
  (setq org-roam-directory "/path/to/org-files/")
  (map! :leader 
        :prefix "n"
        :desc "Org-Roam-Insert" "i" #'org-roam-insert
        :desc "Org-Roam-Find"   "/" #'org-roam-find-file
        :desc "Org-Roam-Buffer" "r" #'org-roam)
  :config
  (org-roam-mode +1))
```

What do you think?

PR #232, opened 2020-03-05
org-roam-show-graph: PREFIX to generate a graph but not display it

acowley:

This lets the viewer application watch the file, or have the user manually refresh the view rather than, say, opening a new browser tab.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1113,9 +1113,10 @@ into a digraph."
 	  (insert "}")
 	  (buffer-string)))
 
-(defun org-roam-show-graph ()
-  "Displays the generated Org-roam graph using `org-roam-graph-viewer'."
-  (interactive)
+(defun org-roam-show-graph (&optional prefix)
+  "Generates and displays the Org-roam graph using `org-roam-graph-viewer'.
+If PREFIX, then the graph is generated but the viewer is not invoked."
+  (interactive "P")
   (declare (indent 0))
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -1127,9 +1128,10 @@ into a digraph."
     (with-temp-file temp-dot
       (insert graph))
     (call-process org-roam-graphviz-executable nil 0 nil temp-dot "-Tsvg" "-o" temp-graph)
-    (if (and org-roam-graph-viewer (executable-find org-roam-graph-viewer))
-	      (call-process org-roam-graph-viewer nil 0 nil temp-graph)
-      (view-file temp-graph))))
+    (when (null prefix)
```
jethrokuan: `(unless prefix ...`

org-roam.el
```diff
@@ -1113,9 +1113,10 @@ into a digraph."
 	  (insert "}")
 	  (buffer-string)))
 
-(defun org-roam-show-graph ()
-  "Displays the generated Org-roam graph using `org-roam-graph-viewer'."
-  (interactive)
+(defun org-roam-show-graph (prefix)
```
jethrokuan: &optional prefix

acowley: I appreciate the improvements! FWIW, I used `org-roam-insert` as a style reference for the prefix argument, so perhaps we should change that, too.

Issue #223, opened 2020-03-04
Exclude certain note classes from graph


jdiez17:
### Brief Abstract

Currently, all of my journal entries appear on the org-roam graph. I would like to be able to exclude them, to only see the connections of the other notes.

### Long Description

With my current workflow (where I have an org-roam note for daily journal entries in addition to more "long-term" notes in the same org-roam directory), whenever I run `org-roam-show-graph`, the journal notes (prefixed with `private_`) also show as nodes on the graph. This adds noise to the graph, and I'm not particularly interested in how the journal notes are connected.

### Proposed Implementation (if any)

I'm an elisp noob so there may be better ways of doing it. But one possible implementation would be to add a user-defined-function/hook/whatever, which is called with the file name as an argument for each node that is being considered. If this function returns `t`, then the node is added to the graph.

### Please check the following:
- [X] No similar feature requests

jethrokuan:
You can now do this:
```
(setq org-roam-graph-exclude-matcher "private_")
```

should exclude all files with `private_` in their name.

jdiez17:
This works great. Thanks a lot!

Elvecent:
This feature appears not to be working. I'm using the MELPA release (1.0.0) and I see the matching code in org-roam.el, but setting `org-roam-graph-exclude-matcher` to a string or a list seems to have no effect.

jethrokuan:
I'm using a byte-compiled version of the package and its fine for me. What's the string you are setting it to?

Elvecent:
> I'm using a byte-compiled version of the package and its fine for me. What's the string you are setting it to?

The precise contents don't matter. I've just reproduced that by
1. Compile the package
2. Set `org-roam-graph-exclude-matcher` to a new value
3. Restart Emacs
4. Generate a graph
5. Observe that it uses the old exclude-matcher
6. Remove the compiled code
7. Restart Emacs again
8. Generate a graph
9. Observe that it works correcty

The exclude-matcher seems to be captured by the macro during load- or compile-time; so when the package is first installed without any settings, it's void and so nothing ever gets excluded unless the package is recompiled or reloaded from source.

BTW thank you so much for working on this package and responding so quickly!

philohistoria:
sorry, has this `org-roam-graph-exclude-matcher` variable been removed in org-ref v2? I cannot find it in the customization and any where in the org-mode el scripts. Thanks!

PR #233, opened 2020-03-06
(feature): org-roam-graph-exclude-matcher: exclude nodes from graph

jethrokuan:
Addresses #223 

Adds a string variable `org-roam-graph-exclude-matcher`, which is a regular string. If the file-path matches this string, it is excluded from the graph. defaults to nil, in which case no nodes are excluded from the graph.


alanz:
It might be nice to have a way to generate a graph anchored on the currently open buffer, showing `n` links from it only.  Something like the `org-brain` navigation view.

PR #247, opened 2020-03-07
(feature): introduce org-roam-backlink face

rasendubi:
I don't like org-block styling in backlinks buffer as it looks ugly
with my color scheme.

Add org-roam-backlink face for styling backlinks instead of using
org-block. It inherits from org-block, so that shouldn't affect the
default styling, but allows to override it without modifying
legitimate org-blocks.


jethrokuan:
LGTM!

PR #248, opened 2020-03-07
(chore): drop org-roam-filename-noconfirm

rasendubi:
With org-capture integration, it does not seem to be used anymore.


jethrokuan:
Thanks!

PR #253, opened 2020-03-08
(bugfix): Don't query about SQL proccess on Emacs exit

juergenhoetzel:
Graceful shutdown is already guaranteed via `kill-emacs-hook`.

Refs #200

Prevents needless interaction on Emacs exit:

```
Active processes exist; kill them and exit anyway? (yes or no)
```

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -140,6 +140,7 @@ Performs a database upgrade when required."
            (init-db (not (file-exists-p db-file))))
       (make-directory (file-name-directory db-file) t)
       (let ((conn (emacsql-sqlite db-file)))
+        (set-process-query-on-exit-flag (emacsql-process conn) nil)
         (puthash (file-truename org-roam-directory)
                  conn
                  org-roam--db-connection)


org-roam.el
```diff
@@ -140,6 +140,7 @@ Performs a database upgrade when required."
            (init-db (not (file-exists-p db-file))))
       (make-directory (file-name-directory db-file) t)
       (let ((conn (emacsql-sqlite db-file)))
+	(set-process-query-on-exit-flag (emacsql-process conn) nil)
```
jethrokuan: Indentation seems a bit off.

juergenhoetzel: > Indentation seems a bit off.

I used (default) <kbd>Tabs</kbd> and just amended the commit to  `untabify`.

A global `.editorconfig` would be nice to prevent this issues.

jethrokuan: Sorry it was right, it looked a bit off on my phone, which was why I resolved it after.

PR #257, opened 2020-03-09
Add a company-mode backend

jethrokuan:

Autocomplete links to existing notes while typing. This should increase the density of links, making sure we don't forget to link notes together.

output-2020-03-11-14:07:55


d12frosted:
I am sorry for the spam. I know that I can put a 👍 as reaction to the PR itself, but it doesn't express how cool this feature is. Big thanks to you for `org-roam` and your further work on it 💯

Issue #255, opened 2020-03-09
Feature: Tags


gambhiro:
### Brief Abstract

Recognizing links to tags files:

- adding them with a separate insert function
- filtering them from the graph
- variable to format them with a different link face and link prefix (such as `T:%s` where other links might be `R:%s`)

### Long Description

Links to tags (such as 'People', 'Quotes') are mostly a kind of taxonomy, to produce a list of category items, and don't usually imply connections in related knowledge.

They crowd out the graph as well, with lots of links to a 'Quotes' node for example, which is not too insightful.

Tags files could be created without a date, but marked with a prefix, such as `tag_people.org`.

The prefix would allow them to be offered as list when inserting with a specific function, and being filtered from the graph as well.

### Please check the following:
- [x ] No similar feature requests

Thank you for this package! It was such a great find. I started to write my own zettel functions, but when your article about smart notes came up, I switched over to org-roam. Fits like a glove.

jethrokuan:
> adding them with a separate insert function

You can do this by creating a template for tags:

```
(add-to-list 'org-roam-capture-templates
             '("t" "tags" plain (function org-roam--capture-get-point)
              ""
              :file-name "tag_${slug}"
              :head "#+TITLE: ${title}\n"
              :immediate-finish t))
```

> filtering them from the graph

You can then filter them by setting the matcher:

`(setq org-roam-graph-exclude-matcher "tag_")`

>  format them with a different link face and link prefix

You can override `org-roam--roam-link-face` to achieve your desired result:

```
(defun org-roam--roam-link-face (path)
  "Conditional face for org file links.
Applies `org-roam-link-face' if PATH correponds to a Roam file."
  (if (org-roam--org-roam-file-p path)
      'org-roam-link
    'org-link))
```

gambhiro:
Great, that's really flexible! And filtering when inserting tags could be then simply starting with a `Tag: ...` title.

Maybe it would have a space in the documentation, such as an 'Appendix' chapter.

gambhiro:
The above is working well, the only thing I found missing was a way to remove the `Tag:` prefix from titles when inserting tag links.

I'd like to ask if the `(format ...)` line in `org-roam-insert` could be extracted to a helper function which then could be easily customized.

```
(defun org-roam--title-format (title)
  (format org-roam-link-title-format title))
```

And in `org-roam-insert`:

```
;; From:
(description (format org-roam-link-title-format (if prefix
                                                    (downcase region-or-title)
[...]

PR #261, opened 2020-03-09
(feat): Allow function for org-roam-link-title-format

jethrokuan:
Addresses #255

One may do something fancy like:

```
(defun my-format-fn (title)
  "my format function"
  "hi")

(setq org-roam-link-title-format #'my-format-fn)
```


alanz:
btw, I noticed some other usages of `org-roam-link-title-format` in `org-roam.el`, that possibly also need updating.

PR #263, opened 2020-03-09
(feat) Added neato support

andresm:
neato is able to create more compact graphs than dot. It reduces the horizontal scrolling when navigating big graphs.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -99,12 +99,7 @@ Org-roam tries its best to locate the Graphviz executables from your
 
 ```
 (setq org-roam-graphviz-dot-executable "/path/to/dot")
-```
-
-or
-
-```
-(setq org-roam-graphviz-neato-executable "/path/to/dot")
+(setq org-roam-graphviz-neato-executable "/path/to/neato")
 ```
 
 So you can select between dot or neato:
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -20,7 +20,7 @@ considered part of the Org-roam ecosystem.
 
 Emacs supports directory-local variables, allowing the value of
 `org-roam-directory` to be different in different directories. It does
-this by checking for a file named `.dir-locals.el`. 
+this by checking for a file named `.dir-locals.el`.
 
 To add support for multiple directories, override the
 `org-roam-directory` variable using directory-local variables. This is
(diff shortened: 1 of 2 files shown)


alanz:
I have not run this, so have no idea what it looks like.  But a more vertical layout might make sense, as shown in https://dev2.wiki.innovateoregon.org/view/welcome-visitors/view/preview-next-diagram

The relevant bit is something like 
```
DOT strict digraph

rankdir=LR

  node [style=filled fillcolor=lightyellow penwidth=5 color=black fontname=\"Helvetica-bold\"]
  HERE NODE

    node [style=filled fillcolor=lightblue]
    WHERE /^Next/
      LINKS HERE -> NODE
          node [style=filled fillcolor=white]
          HERE NODE
            WHERE /^Next/
              LINKS HERE -> NODE

    node [style=filled fillcolor=white penwidth=3 color=black fontname=\"Helvetica\"]
    LINKS HERE -> NODE
       node [style=filled fillcolor=white penwidth=1 color=black fontname=\"Helvetica\"]
       HERE NODE
         LINKS HERE -> NODE
```
Note: this is a fedwiki template, from viewing the JSON of the diagram, in the footer.

bdarcus:
> I have not run this, so have no idea what it looks like. But a more vertical layout might make sense, as shown in https://dev2.wiki.innovateoregon.org/view/welcome-visitors/view/preview-next-diagram

That's a pretty awesome UI! Would be cool to see something like that in org-roam export.

jethrokuan:
Rather than add another customization option, couldn't this be achieved by setting the graphviz executable directly to neato?

```
(setq org-roam-graphviz-executable (executable-find "neato"))
```

I don't mind adding `overlap=false;` into the graph.

alanz:
Perhaps we should consider `rankdir=LR`?
http://martin-loetzsch.de/S-DOT/rankdir.html
This makes the long dimension vertical, like a normal web page, and does not penalize nodes with long names.

jethrokuan:
I'll add a variable that will allow adding extra options. E.g. 

```
(setq org-roam-graph-extra-options '(("rankdir" . "LR")))
```

I gave that option a try, and found that the average distance between nodes went up significantly.

Tbh graphviz at this point is a stopgap solution. I'm hoping to eventually move to something more sophisticated like d3.

- keyboard navigation of nodes
- hide/show arbitrarily deep layers
- proper force-based layout

Unfortunately my web skills are not that great 😬

andresm:
Here are examples of a graph generated by dot and neato:

[GraphViz dot file. Node as Chess order](https://stackoverflow.com/questions/20473268/graphviz-dot-file-node-as-chess-order)

Regarding the extra options: I kept both paths in the code to be able to switch between dot and neato easily, and allow the user to configure both paths in case the automatic identification fails.

jethrokuan:
> allow the user to configure both paths in case the automatic identification fails.

I think users can be expected to configure the path to the executable used appropriately themselves, the extra search for the dot executable can actually be removed

PR #273, opened 2020-03-10
(bugfix): use replace when rebuilding db

herbertjones:
Prevents SQL errors when rebuilding the database due to duplicate entries.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -360,22 +360,22 @@ This is equivalent to removing the node from the graph."
       (org-roam--db-clear-file file))
     (when all-files
       (org-roam-sql
-       [:insert :into files
+       [:replace :into files
         :values $v1]
        all-files))
     (when all-links
       (org-roam-sql
-       [:insert :into file-links
+       [:replace :into file-links
         :values $v1]
        all-links))
     (when all-titles
       (org-roam-sql
-       [:insert :into titles
+       [:replace :into titles
         :values $v1]
        all-titles))
     (when all-refs
       (org-roam-sql
-       [:insert :into refs
+       [:replace :into refs
         :values $v1]
        all-refs))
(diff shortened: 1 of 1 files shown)


herbertjones:
Since I use syncthing to share my files, and Org-roam doesn't recognize file changes, I occasionally need to rebuild the database.  This allows it to be rebuilt instead of failing with errors due to duplicate inserts.

jethrokuan:
Technically this shouldn't be necessary. Org-roam detects file changes by comparing the file hashes. What error are you seeing?

herbertjones:
> peculiar error: "UNIQUE constraint failed: files.file"

Without running `org-roam-build-cache` new files won't show up with `org-roam-find-file`.

jethrokuan:
I have no issue with this, just trying to understand what's going on.

https://github.com/jethrokuan/org-roam/blob/dee06452c7c0a3c2fe2fa09d0dba34b60c80dedf/org-roam.el#L346

In this line, Org-roam should have cleared any changed files from the cache. `all-files` will only be populated with items that have been cleared from the cache, which must mean that the cache clearing has some problem.

herbertjones:
Likely the new files aren't in the database.  I disabled syncing the db file directly since I had many syncthing "conflict" files.  I can re-enable it, but I'll likely still run into this error due to whichever version of the database syncthing ends up using might not have all the new files.

jethrokuan:
> Likely the new files aren't in the database.

If this were the case, then there wouldn't be a uniqueness conflict problem?

The files table should've been sufficient in doing a full rebuild of the database. My worry with the replace is that this just masks the problem, and the database isn't a proper reflection of the Org-roam directory files.

herbertjones:
Yeah, you're right.  It's due to having a local org-roam-directory that isn't propagated.

Issue #277, opened 2020-03-11
Descriptive Dates in Title like Org Journal


aswinmohanme:
### Brief Abstract
When creating new date entry with `org-roam-today` the title is populated as `#+TITLE: 2020-03-11
`. 

Creating a new entry with `org-journal` the title is `#+TITLE: Wednesday, 11 March 2020` . Is there any way to enable title to be in long form as it would make searching easier.

jethrokuan:
I personally don't use `org-roam-today`, and choose to use `org-journal`. Those functions are there mainly for historic reasons, and for a simple implementation of that org-journal functionality.

aswinmohanme:
I have tried with org-journal and it seems kind of glitchy. I can't use org-roam-date for a dated entry too. It would be really nice to have this feature integrated with org-roam.

I know how much you get peddled with feature requests and how much effort you are putting into this.

PR #280, opened 2020-03-12
Adds descriptive time-stamp like org-journal

aswinmohanme:
This commit adds a descriptive timestamp which gets added on file creation.

Fixes: #277


jethrokuan:
This is going to change the default behaviour for people who have been using the function for a while. I don't mind adding a customizable variable instead.

PR #282, opened 2020-03-12
Refactor org-roam-find-file and org-roam-insert

khinsen:
Replace completing-read by a plug-in in order to facilitate
the implementation of alternative completion methods such
as helm or ivy.

Avoid code duplication in add-ons. I'll propose Helm support based on this change as a separate pull request.


khinsen:
That's fine as well. I'll update accordingly.

khinsen:
Made obsolete by #284

PR #284, opened 2020-03-12
Customizable org-roam-completion-system

khinsen:
Currently implemented choices:
 - `'default` for Emacs' `completing-read`
 - `'helm` for Helm

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -101,6 +101,14 @@ Formatter may be a function that takes title as its only argument."
   :type 'boolean
   :group 'org-roam)
 
+(defcustom org-roam-completion-system 'default
+  "The completion system to be used by `org-roam'."
+  :type '(radio
+          (const :tag "Default" default)
+          (const :tag "Helm" helm)
+          (function :tag "Custom function"))
+  :group 'org-roam)
+
 ;;;; Dynamic variables
 (defvar org-roam--current-buffer nil
   "Currently displayed file in `org-roam' buffer.")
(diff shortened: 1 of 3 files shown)


jethrokuan:
Looks fantastic, thanks! Mind adding an entry into the changelog, and a short snippet in the configuration.md?

org-roam.el
```diff
@@ -594,6 +602,40 @@ specified via the #+ROAM_ALIAS property."
...
+Return user choice."
+  (cond
+   ((eq org-roam-completion-system 'default)
+    (completing-read prompt choices nil require-match initial-input))
+   ((eq org-roam-completion-system 'helm)
+    (unless (and (fboundp 'helm)
+                 (fboundp 'helm-build-sync-source))
+      (user-error "Please install helm from \
+https://github.com/emacs-helm/helm"))
+    (let* ((source (helm-build-sync-source "Title"
```
jethrokuan: should this be prompt instead of "Title"?

org-roam.el
```diff
@@ -594,6 +602,40 @@ specified via the #+ROAM_ALIAS property."
...
+                     :candidates (-map #'car choices)
+                     :filtered-candidate-transformer
+                     (and (not require-match)
+                          #'org-roam---helm-candidate-transformer)
+                     :fuzzy-match t))
+           (title (helm :sources source
+                        :buffer "*org-roam titles*"
```
jethrokuan: here too, should be made generic.

org-roam.el
khinsen: Yes, it should!

org-roam.el
khinsen: Yes, derived from the prompt.

PR #294, opened 2020-03-13
Less abusive use of org-capture - proof of concept

myshevchuk:
###### What's changed

- `org-roam--capture-new-file` does not write anything to a file, all the
   templating is delegated to vanilla `org-capture`.
- `org-roam--capture-get-point` injects the desired template adjustments and
   behavior using `org-capture-put`
- `org-roam-capture-templates` variables work as before
- `org-capture-templates` wildcards work as intended except for "%?"
- depending functions adjusted accordingly
- new variable to `org-roam--capture-immediate-finish` to control the capture
  process finishing unattended

###### Proof of concept

1. The original org-roam behaviour as presented to the end user did not change.
2. The changes, however, have not been thoroughly tested, so some things may not work.
3. Point is not set as indicated by "%?" in a template, this must be fixed.

1. The main idea is to use the `:target` property of `org-capture-templates` as designed by `org-capture`, that is 

> Most general way: write your own function which both visits the file and moves point to the right location.

See: https://orgmode.org/manual/Template-elements.html#Template-elements

2. This will ensure compatibility of org-roam and org-capture in the future.
3. The suggested changes also allow to use both `org-roam-capture-templates` and `org-capture-templates` goodies transparently.

###### Original discussion

https://github.com/jethrokuan/org-roam/issues/285

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -801,8 +801,6 @@ This function is used solely in Org-roam's capture templates: see
        (goto-char (point-max))))
     (_ (error "Invalid org-roam-capture-context"))))
 
-(defvar org-roam--capture-immediate-finish)
-
 (defun org-roam-capture (&optional goto keys)
   "Create a new file, and return the path to the edited file.
 The templates are defined at `org-roam-capture-templates'.  The
@@ -812,9 +810,7 @@ GOTO and KEYS argument have the same functionality as
         file-path)
     (when (= (length org-capture-templates) 1)
       (setq keys (caar org-capture-templates)))
-    (setq org-roam--capture-immediate-finish goto)
-    (org-capture nil keys)
-    ))
+    (org-capture nil keys)))
 
 ;;; Interactive Commands
 ;;;; org-roam-insert
@@ -925,7 +921,7 @@ INITIAL-PROMPT is the initial title prompt."
       (let* ((org-roam--capture-info (list (cons 'title title)
                                            (cons 'slug (org-roam--title-to-slug title))))
              (org-roam--capture-context 'title))
-        (org-roam-capture t)
+        (org-roam-capture)
         (add-hook 'org-capture-after-finalize-hook #'org-roam--capture-find-file)))))
 
 ;;;; org-roam-find-ref
--- a/org-roam.el
+++ b/org-roam.el
@@ -762,7 +762,6 @@ currently active Org-roam template."
          (file-path (org-roam--file-path-from-id new-id)))
     (when (file-exists-p file-path)
       (error (format "File exists at %s, aborting" file-path)))
-    (org-roam--touch-file file-path)
     file-path))
 
 (defun org-roam--capture-get-point ()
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -706,6 +700,15 @@ If there is no file with that ref, a file with that ref is created.
 
 This function is used solely in Org-roam's capture templates: see
 `org-roam-capture-templates'."
+  (org-capture-put :template
+                   (concat
+                    (org-roam--fill-template (or (org-capture-get :head)
+                                                 org-roam--capture-header-default)
+                                             org-roam--capture-info)
+                    "\n" (org-capture-get :template)))
```
jethrokuan: I like this change, this means that if org-capture is cancelled, the head isn't written.

org-roam.el
```diff
@@ -706,6 +700,15 @@ If there is no file with that ref, a file with that ref is created.
...
+
+  (org-capture-put :immediate-finish org-roam--capture-immediate-finish)
```
jethrokuan: not sure if this is necessary though.

jethrokuan:
Also if org-capture doesn't require the file to first exist, might want to remove the `touch-file` too, so if the capture process is cancelled then the file isn't created.

org-roam.el
myshevchuk: `org-roam--capture-immediate-finish` is used as follows
1. It's value is set to the first value passed to `org-roam-capture`, `t` or `nil` otherwise
2. We want it to be `t` when we create a new file using `org-roam-find-file` to finish the capture immediately (this simulates the original behaviour of `org-roam`
3. I forgot to mention another org-capture-finilize-hook added, if this variable is set to `t`, we will switch to the newly created file's buffer, again to simulate the original `org-roam-find-file` behaviour
4. It's value is `nil` in all other cases

myshevchuk:
@jethrokuan This is just a proof of concept. It's not ready for merging of course. Please, if you are interested in such changes to the capture mechanism, I will test and polish the things.

org-roam.el
myshevchuk: This and other smaller adjustments are needed to simulate the original org-roam behaviour,
in particular, immediately switching to the buffer newly created by `org-roam-find-file` but staying in the same buffer when creating a file with `org-roam-insert`.

myshevchuk:
> Also if org-capture doesn't require the file to first exist, might want to remove the `touch-file` too, so if the capture process is cancelled then the file isn't created.

As far as I saw, `org-capture` is capable of creating a file itself, so this may well be dropped.

org-roam.el
jethrokuan: I'd prefer changing the behaviour of the immediate finish in org-roam-capture for org-roam-find-file, because it's unintuitive. The org capture should go through, and the file should be visited at the end. That's the current behaviour in master now actually after the change earlier.

jethrokuan:
one additional thing that may need to be set when concatenating `:head` to the original template is enforcing the type to be capture template type to be `plain`.

org-roam.el
myshevchuk: Yes sure, fully agree. My patches are based on yesterday's develop branch, so I didn't have a chance to check up the new version. The things are changing blazingly fast :)

myshevchuk:
Rebased onto updated master and introduced the changes suggested here.

org-roam.el
```diff
@@ -810,10 +810,7 @@ GOTO and KEYS argument have the same functionality as
         file-path)
     (when (= (length org-capture-templates) 1)
       (setq keys (caar org-capture-templates)))
-    (org-capture goto keys)
-    (setq file-path org-roam--capture-file-path)
-    (setq org-roam--capture-file-path nil)
-    file-path))
+    (org-capture nil keys)))
```
jethrokuan:
```suggestion
    (org-capture goto keys)))
```

myshevchuk:
After multiple tests everything seems to work as expected.

jethrokuan:
Thanks for the work!

Issue #237, opened 2020-03-06
Choose the ordering of links in the backlinks window


brabalan:
If I read https://github.com/jethrokuan/org-roam/blob/b7afb02015a31a2cba02415ba863149ad7c0cee5/org-roam.el#L958 correctly, there is no particular ordering for the backlinks. It would be great to be able to configure it, for instance:
- most recently updated first (maybe this could be the default choice)
- oldest file first
- alphabetical order of title

jethrokuan:
All of these actually require some modification to the database.

1. most recently updated first: adding a time-added column
2. oldest file first: requires a join with the files table, not sure if it's worth it
3. alphabetical order of title: this can be done as a sort before output.

I thought about storing files as a foreign key in the `file-links` table previously, but eventually concluded that just storing the file path would be the most efficient. Will need to think about it more.

digitalsignalperson:
Without modifying database, can't you just get the list of files, then stat them to get creation/modification date and sort as needed?

MateuszSeczkowski007:
> If I read
> 
> https://github.com/jethrokuan/org-roam/blob/b7afb02015a31a2cba02415ba863149ad7c0cee5/org-roam.el#L958
> 
> correctly, there is no particular ordering for the backlinks. It would be great to be able to configure it, for instance:
> * most recently updated first (maybe this could be the default choice)
> * oldest file first
> * alphabetical order of title

MateuszSeczkowski007:
c539623

asymmetric:
> Without modifying database, can't you just get the list of files, then stat them to get creation/modification date and sort as needed?

Seems like this could scale poorly on large link lists?

jethrokuan:
you're only reordering backlinks, which shouldn't be large enough for it to matter.

cwittern:
It seems to me that adding a "created-date" and "modified-date" field to the database might come in handy later as well. I for one would also like to look at all notes and be able to sort them using these properties.

drobe504:
If this is still an intended feature, I'd like to suggest the following:
- `org-roam-backlink-compare-nodes` function that takes 2 nodes and returns 0, -1, 1 for comparison
- `org-roam-backlink-compare-property`
- default order behavior could use the aforementioned property variable in a default compare function to order the backlinks

That would allow users to override the aforementioned function with a custom comparator, or just specify a property to order by. Properties would be better than file created/modified, since they would work for nested org roam nodes also.

dustinfarris:
[...]

PR #300, opened 2020-03-15
natural ordering of backlinks

wannli:
Order backlinks by "file-from" file name, which seems preferable to the current non-sort. This would address (at least part of) #237.


jethrokuan:
This does seem better than a random order.

PR #303, opened 2020-03-15
Update roam_protocol.md

wannli:
With #299, EMP now supports org-protocol out of the box.


jethrokuan:
Does EMP run emacsclient, or does it launch a new Emacs instance?

wannli:
L223-226 of https://bitbucket.org/mituharu/emacs-mac/src/master/README-mac

> The Mac port doesn't support multi-tty with GUI.  The developer has no idea how to detach Emacs as a GUI application from Window Server or Dock without separating a GUI process (not thread) from the main Emacs (Lisp evaluator) process.  TTY-only multi-tty is supposed to work.

jethrokuan:
I'm not sure how well Org-roam works with multiple Emacs instances. I think the emacsql binary is compiled in a way that only allows a single process to access the database. The safe option is still to open the files via emacsclient.

PR #308, opened 2020-03-16
Fix Helm integration in org-roam--completing-read

### LINKED ISSUE #307: Helm integration broken again
### Description

When `org-roam--completing-read` is used with `org-roam-completion-system` set to `'helm`, entering a new note title (rather then selecting an existing one) will yield a title starting with two spaces.

I'd be happy to provide a patch, but if you again edit out the `s-trim-left` I put in to fix this, it will never work! I agree it doesn't look particularly elegant, but the only alternative I can see is creating another intermediate list of completion candidates, which is rather wasteful.

### Versions
- Emacs (`C-h v emacs-version`): vX.X.X

khinsen:
Fixes #307

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -659,15 +659,16 @@ https://github.com/emacs-helm/helm"))
                   (buf (concat "*org-roam "
                                (s-downcase (s-chop-suffix ":" (s-trim prompt)))
                                "*")))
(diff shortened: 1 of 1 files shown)


jethrokuan:
This just doesn't work in the general case. What if people want their "candidates" to start with spaces?

PR #324, opened 2020-03-18
For `ido-completing-read', `choices' must be a list of strings.

tsproisl:
When `org-roam-completion-system` is set to `ido`, the following errors occur:
* After `M-x org-roam-switch-to-buffer`:
  > Error in post-command-hook (ido-exhibit): (wrong-type-argument sequencep #\<buffer filename.org>)
* When hitting tab after `M-x org-roam-find-file`:
  > ido-file-lessp: Wrong type argument: listp, "/path/to/filename.org"

These errors occur because `ido-completing-read` can only deal with a list of strings. Suggested fix: Only pass the names.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -629,7 +629,8 @@ Return user choice."
     (setq res
           (cond
            ((eq org-roam-completion-system 'ido)
-            (ido-completing-read prompt choices nil require-match initial-input))
+	    (let ((candidates (mapcar #'car choices)))
+	      (ido-completing-read prompt candidates nil require-match initial-input)))
            ((eq org-roam-completion-system 'default)
             (completing-read prompt choices nil require-match initial-input))
            ((eq org-roam-completion-system 'ivy)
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -629,7 +629,8 @@ Return user choice."
     (setq res
           (cond
            ((eq org-roam-completion-system 'ido)
-            (ido-completing-read prompt choices nil require-match initial-input))
+	    (let ((candidates (mapcar #'car choices)))
```
jethrokuan: Indentation seems a bit off.

This is a good temp fix, but we have to be sure that for all calls to `org-roam--completing-read`,  `choices` is an alist, which happens to be the case right now.

tsproisl:
I corrected the indendation.

> we have to be sure that for all calls to `org-roam--completing-read`, `choices` is an alist
Should I add an assertion? Something like this, perhaps? (No idea if that is the best way to assert an alist or if an assertion makes sense at all…)
```elisp
           ((eq org-roam-completion-system 'ido)
            (cl-assert (seq-every-p #'consp choices) t)
            (let ((candidates (mapcar #'car choices)))
              (ido-completing-read prompt candidates nil require-match initial-input)))
```

jethrokuan:
> Should I add an assertion?

It's fine. Thanks for the patch!

Issue #298, opened 2020-03-14
Suggestions to `org-roam-capture-templates`


myshevchuk:
#### Brief Abstract
Below are several suggestions related to `org-roam-capture-templates`.

#### Long Description
1. Hide `:target` in `org-roam-capture-templates` from the user 
Since `(function org-roam--capture-get-point)` is the only valid target for org-capture, it may be desirable to hard-code it and hide from the user.
2. Hide `:type` from the user
The above applies here.
3.  Ensure `"%?"` is present in the template string
There is a bug in the original org-capture, more specifically in`org-capture-place-plain-text`:
When `:unnarrowed` is set to `t` AND `"%?"` is missing from the template string, the capture process is interrupted and the following error is thrown:
    > condition-case: Capture template ‘%s’: Invalid search bound (wrong side of point)

    Since in org-roam both `:unnarrowed t` and the plain type are present by default, it may be         desirable to ensure the user doesn't catch the error if they accidentally or intentionally remove `"%?"` from the template string.

### Proposed Implementation (if any)
1. & 2. in `org-roam-capture`
```
  (let* ((type 'plain)
        (target '(function org-roam--capture-get-point))
        (org-capture-templates (--map (--> it
                                           (-insert-at 2 type it)
                                           (-insert-at 3 target it))
                                      org-roam-capture-templates))
        file-path)
```
3. in `org-roam--capture-get-point`
```
      (let* ((head (org-roam--fill-template (or (org-capture-get :head)
                                            org-roam--capture-header-default)
                                        org-roam--capture-info))
         (template (org-capture-get :template))
         (full-template (concat head template)))
    (org-capture-put :template
                     (if (s-contains-p "%?" full-template)
                         full-template
                       (concat "%?" full-template))))
```
### Please check the following:
- [ x] No similar feature requests

jethrokuan:
1: I think it may be desirable to keep the structure of the original org capture template definitions. I know it's not great, but at least it's familiar. Things like DOCT tries to fix it.  

2: the type `plain` is only compulsory when the header template is injected. Other types like `entry` should be allowed.

3 is fine with me.

myshevchuk:
Ok. 

Regarding the `plain` type, the header template is inserted anyway, either a user one or the default one, so the `plain` type is actually enforced currently, shall this be changed?
```
  (org-capture-put :template
                   (concat
                    (org-roam--fill-template (or (org-capture-get :head)
                                                 org-roam--capture-header-default)
[...]

PR #330, opened 2020-03-20
ensure "%?" is present in the template string

myshevchuk:
- addresses #298
- fixes `org-capture-place-plain-text` throwing 'invalid search bound'
  when both :unnarowed t and "%?" is missing from the template string

See #298

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -823,7 +823,7 @@ the file if the original value of :no-save is not t and
                      ;; may become unnecessary when the upstream bug is fixed
                      (if (s-contains-p "%?" roam-template)
                          roam-template
-                       (concat "%?" roam-template))
+                       (concat roam-template "%?"))
                      :roam-new-file t
                      :type 'plain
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -809,15 +809,21 @@ the file if the original value of :no-save is not t and
...
-                     (concat
-                      (or (org-capture-get :head)
-                          org-roam--capture-header-default)
-                      (org-capture-get :template))
+                     ;; Fixes org-capture-place-plain-text throwing 'invalid search bound'
+                     ;; when both :unnarowed t and "%?" is missing from the template string;
+                     ;; may become unnecessary when the upstream bug is fixed
+                     (if (s-contains-p "%?" roam-template)
+                         roam-template
+                       (concat "%?" roam-template))
```
jethrokuan: Put "%?" at the back rather than at the front? Usually it makes sense for the cursor to be at the end.

PR #337, opened 2020-03-21
Use org-link-make-string

kljohann:
This takes care of properly escaping backslashes in file names.


jethrokuan:
Nice, didn't know that existed. Could you also change the occurrences in `org-roam--rename-file-advice`?

PR #350, opened 2020-03-24
(feature): add `org-roam-db-location`

### LINKED ISSUE #342: Ability to set database location for working with file sync
### Brief Abstract
I use syncthing to keep my org files in sync and as the db is in the same dir as the org files it gets locked as read only during that sync. It would be good to be able to set the DB location to somewhere custom.
### Long Description
If the DB is just a fast cache of the data in the org files then could this be moved to a local cache dir rather than the org dir as it's can be re-created.

- [ x] No similar feature requests

jethrokuan:

Fixes #342.

`org-roam-db-location` sets the location of the Org-roam sqlite database.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Changelog
 
+## 1.0.1 (TBD)
+
+### New Features
+* [#350][gh-350] Add `org-roam-db-location`
+
 ## 1.0.0 (23-03-2020)
 
 Org-roam is now on MELPA! We have squashed most of the bugs, and Org-roam has
 been stable for the most part.
 
-## New Features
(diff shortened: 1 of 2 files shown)


rasendubi:
FYI: The old `org-roam-db-filename` worked in the same way.

I had the following piece in my configuration and it placed db file in `~/.emacs.d/cache/org-roam.db`. (Because `expand-file-name` in `org-roam-db--get` does nothing if given an absolute path.)
```elisp
(setq org-roam-db-location (expand-file-name "cache/org-roam.db" user-emacs-directory))
```

I just noticed that db is created in `org-roam-directory` again :slightly_smiling_face:

Issue #312, opened 2020-03-16
Turn org-roam-backlinks-mode into a minor-mode?


hubisan:
Does `org-roam-backlinks-mode` have to be a major-mode? Or could it be a minor-mode? 

It might/does cripple some packages that work well with org as major-mode. 
Example: The very useful [link-hint](https://github.com/noctuid/link-hint.el).

jethrokuan:
It became a major mode primarily to avoid overriding of keybindings, either way the backlinks buffer is due for a revamp.

kljohann:
> It became a major mode primarily to avoid overriding of keybindings, either way the backlinks buffer is due for a revamp.

I think this could be achieved by removing `org-roam-open-at-point` in favor of a hook function that is put into `org-open-at-point-functions` when the (now minor) `org-roam-backlinks-mode` is active. Then the native `org-open-at-point` would "do the right thing" and as a plus, the user would not need to change their custom key bindings for this function. (Of course, if it was intentional to keep the original behavior of `org-open-at-point` in the backlinks buffer this would not work.)
This is possible because the first thing `org-open-at-point` does is to call
```elisp
(unless (run-hook-with-args-until-success 'org-open-at-point-functions)
  ...)
```
I think this change should not be too much effort and maybe it's worth it as an incremental step before revamping the backlinks buffer? If you are interested I can work on a pull request.

jethrokuan:
@kljohann that sounds good. That may also make the minor mode obsolete, but I think some people depend on it being active to run some hooks.

hubisan:
Key-bindings from a minor-mode should override major-mode ones.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Active-Keymaps.html#Active-Keymaps

> Usually, the active keymaps are: (i) the keymap specified by the keymap property, (ii) the keymaps of enabled minor modes, (iii) the current buffer's local keymap, and (iv) the global keymap, in that order. Emacs searches for each input key sequence in all these keymaps.

hungpham2511:
I also find that having org-mode as the major mode useful. For instance navigation using ace-link is not possible in the backlink buffer.

PR #355, opened 2020-03-26
Turn org-roam-backlinks-mode into a minor mode

kljohann:

This change makes `org-roam-open-at-point` non-interactive and adds it to
`org-open-at-point-functions`.  As a consequence, `org-open-at-point` will try
to visit an Org-roam link or preview text, before falling back to the default
behavior of Org mode.

In addition, `org-roam-backlinks-mode` is turned into a minor mode as discussed in #312.

There are some questions/remarks that came up during this change, I'll add them inline.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1158,12 +1158,18 @@ Applies `org-roam-link-face' if PATH corresponds to a Roam file."
     map)
   "Keymap for `org-roam-backlinks-mode'.")
 
-(define-derived-mode org-roam-backlinks-mode org-mode "Backlinks"
-  "Major mode for the `org-roam-buffer'.
+(define-minor-mode org-roam-backlinks-mode
+  "Minor mode for the `org-roam-buffer'.
 
 \\{org-roam-backlinks-mode-map}"
-  (add-hook 'org-open-at-point-functions
-            'org-roam-open-at-point nil 'local))
+  :lighter " Backlinks"
+  :keymap org-roam-backlinks-mode-map
+
+  (if org-roam-backlinks-mode
+      (add-hook 'org-open-at-point-functions
+                'org-roam-open-at-point nil 'local)
+    (remove-hook 'org-open-at-point-functions
+                 'org-roam-open-at-point 'local)))
 
 (defun org-roam-open-at-point ()
   "Open an Org-roam link or visit the text previewed at point.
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -1152,45 +1152,54 @@ Applies `org-roam-link-face' if PATH corresponds to a Roam file."
 
 ;;;; org-roam-backlinks-mode
 (defvar org-roam-backlinks-mode-map
-  (let ((km (make-sparse-keymap)))
-    (define-key km [mouse-1] 'org-roam-open-at-point)
-    (define-key km (kbd "RET") 'org-roam-open-at-point)
-    km)
+  (let ((map (make-sparse-keymap)))
+    (define-key map [mouse-1] 'org-open-at-point)
+    (define-key map (kbd "RET") 'org-open-at-point)
```
kljohann: I'm still keeping the bindings, as by default in Org mode `RET` is bound to `org-return`, which would work correctly for the "visit links" feature, but would not work for the "visit preview text" feature, as `org-return` only calls `org-open-at-point` if both  `org-return-follows-link` is set _and_ point is on a link.

org-roam.el
```diff
@@ -1227,7 +1236,9 @@ If item at point is not Org-roam specific, default to Org behaviour."
          (cons '(file . org-roam--find-file) org-link-frame-setup))
         (let ((inhibit-read-only t))
           (erase-buffer)
-          (unless (eq major-mode 'org-roam-backlinks-mode)
+          (unless (eq major-mode 'org-mode)
+            (org-mode))
+          (unless org-roam-backlinks-mode
             (org-roam-backlinks-mode))
           (make-local-variable 'org-return-follows-link)
```
kljohann: This could in principle also be moved into the mode definition. But see also the comment regarding `org-return` above.

org-roam.el
```diff
@@ -1227,7 +1236,9 @@ If item at point is not Org-roam specific, default to Org behaviour."
          (cons '(file . org-roam--find-file) org-link-frame-setup))
         (let ((inhibit-read-only t))
           (erase-buffer)
-          (unless (eq major-mode 'org-roam-backlinks-mode)
+          (unless (eq major-mode 'org-mode)
+            (org-mode))
+          (unless org-roam-backlinks-mode
```
kljohann: This gated activation of the major and minor mode has the effect that the hooks are not run each time, which I think is intentional? In any case, I tried to keep the current behavior there.

org-roam.el
```diff
@@ -1152,45 +1152,54 @@ Applies `org-roam-link-face' if PATH corresponds to a Roam file."
...
+  :lighter " Backlinks"
+  :keymap org-roam-backlinks-mode-map
+
+  (if org-roam-backlinks-mode
+      (add-hook 'org-open-at-point-functions
+                'org-roam-open-at-point nil 'local)
+    (remove-hook 'org-open-at-point-functions
+                 'org-roam-open-at-point 'local)))
 
 (defun org-roam-open-at-point ()
```
kljohann: Note that this function now only does the Org-roam part and as a consequence is no longer interactive (in order not to confuse people). Would you prefer to keep an interactive `org-roam-open-at-point` with the previous behavior (to fall back to `org-open-at-point`)? Or is it okay if we direct users to change their bindings/habits to directly make use of `org-open-at-point`?

PR #358, opened 2020-03-26
Modularize org-roam-graph

progfolio:
Move org-roam-graph to separate feature.
Ensure functions/variables share have consistent org-roam-graph prefix.
Remove direct references to Graphviz.

See: https://github.com/jethrokuan/org-roam/issues/357


progfolio:
I suspect this will fail until the rest of the code is moved into features and the require's are worked out.

progfolio:
I'm going to modularize things on my fork and link it in the issue I made earlier. Makes more sense than leaving a broken pull request here.

Issue #357, opened 2020-03-26
Modularize org-roam features


progfolio:
org-roam has a lot of potential for growth. It would make sense to break some functionality out into its own feature/file. For example, the org-roam-graph functionality seems like a good starting point. It is pretty well contained and could be made more consistent/generic so that its possible to support graphing software other than Graphviz in the future. Users who do not need the graphing functionality could skip the load time for that feature as well. Org does something similar with `org-modules'.

progfolio:
Looking over the source I feel like the database, graphing and capture logic could all be made into their own features.

progfolio:
I realize this is a very large commit, but the first step is getting each feature into its own file and normalizing the function/variable names:

https://github.com/progfolio/org-roam/commit/5bbcf5e0326ec7eabab676416731ccc3ea737cdd

Happy to work on this more if you think it's a viable direction.

jethrokuan:
Most of the features that can be moved into their own files are already separated by in sections using `;;;`.

I think moving things into their own files is good for maintainability. But I'd like a defalias to  the original function names, at least for the interactives, so people's keybindings don't break.

progfolio:
> Most of the features that can be moved into their own files are already separated by in sections using ;;;.

I followed roughly that pattern.

> I think moving things into their own files is good for maintainability. But I'd like a defalias to the original function names, at least for the interactives, so people's keybindings don't break.

I've added the command aliases and rebased on top of https://github.com/jethrokuan/org-roam/commit/df29da1b6db56d44f0f1232815612e8084dc177d

Here: https://github.com/progfolio/org-roam/tree/refactor-modularize-features

I'm running makem.sh locally and I've addressed all relevant errors:

```
Starting new Ispell process /usr/bin/aspell with default dictionary... \ 
Starting new Ispell process /usr/bin/aspell with default dictionary...done
company-org-roam.el:60: Error checking word UNIX using aspell with default dictionary
company-org-roam.el:118: Error checking word BACKEND using aspell with default dictionary
company-org-roam.el:132: Error checking word BACKEND using aspell with default dictionary
company-org-roam.el:139: Error checking word BACKEND using aspell with default dictionary
org-roam-capture.el:190: Error checking word PREPENDS using aspell with default dictionary
org-roam-capture.el:246: Error checking word ORG-ROAM using aspell with default dictionary
org-roam-completion.el:51: Error checking word ORG-ROAM using aspell with default dictionary
[...]

PR #363, opened 2020-03-27
Modularize features

progfolio:
Make distinction between org-roam features. Migrating each feature to its own file will separate development concerns and allow easier maintenance.
Addresses #357

### DIFF
--- a/company-org-roam.el
+++ b/company-org-roam.el
@@ -34,7 +34,6 @@
 ;;     (company-org-roam-init)
 
 ;;; Code:
-
 (require 'cl-lib)
 (require 'company)
 (require 'org-roam)
@@ -89,7 +88,7 @@ The string match is case-insensitive."
 Entries with no title do not appear in the completions."
   (let ((dir (file-truename org-roam-directory))
         (ht (make-hash-table :test #'equal)))
-    (dolist (row (org-roam-sql [:select [titles file] :from titles]))
+    (dolist (row (org-roam-db-query [:select [titles file] :from titles]))
       (let ((titles (car row))
             (file (cadr row)))
         (dolist (title titles)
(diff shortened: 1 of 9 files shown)


org-roam-protocol.el
```diff
@@ -39,19 +39,12 @@
 (require 'org-protocol)
 (require 'org-roam)
 
+;;;; Declarations
+(defvar org-roam-capture-ref-templates)
```
jethrokuan: The default template seems to be gone here.

progfolio:
> Can we also alias variables:
> 
> `org-roam-capture-ref-templates` <-> `org-roam-ref-capture-templates`

Taking care of this now. Do you want these and the aliased commands marked obsolete? This will generate warnings for users so they know how to upgrade appropriately.

jethrokuan:
> Do you want these and the aliased commands marked obsolete? 

yeah, am fine with this.

org-roam-protocol.el
progfolio: That variable was moved to org-roam-capture.el: 
https://github.com/progfolio/org-roam/blob/70bee2679080e6752207988a2084f0671c4bfc0c/org-roam-capture.el#L106-L114

We can actually remove that entire set of declarations since we're requiring org-roam which requires org-roam-capture itself.

progfolio:
Would you like the variables that are marked as internal aliased as well? Some of these seem like they may be better off as user facing. e.g.:

`org-roam--capture-file-name-default` -> `org-roam-capture-file-name-default`

jethrokuan:
> org-roam--capture-file-name-default -> org-roam-capture-file-name-default 

I don't think there's a need to expose this to users, there are other ways for users to customize the template (i.e. through the org-roam capture template syntax).

progfolio:
Looks like were good to go unless you see anything else that needs to be addressed.

company-org-roam.el
```diff
@@ -89,7 +88,7 @@ The string match is case-insensitive."
 Entries with no title do not appear in the completions."
   (let ((dir (file-truename org-roam-directory))
         (ht (make-hash-table :test #'equal)))
-    (dolist (row (org-roam-sql [:select [titles file] :from titles]))
+    (dolist (row (org-roam-db-query [:select [titles file] :from titles]))
```
jethrokuan: One more thing, can you alias this as well? I'm using it in `company-org-roam`.

progfolio: Taken care of

PR #374, opened 2020-03-29
(feat): Add org-ref citation support

jethrokuan:
Add support for viewing backlinks on citations. This is a follow up on #130.

image


bdarcus:
Very cool.

Can I just ask a high-level question, which might inform how you document this?

What's the difference between a cite and backlink, conceptually? They're both references, but how do they differ?

jethrokuan:
One are file links, and the other are `cite:` links generated by org-ref.

bdarcus:
So the "file links" are really (at a higher level) links of two notes.

The cite: links are links between notes and (I guess more abstract)
bibliographic sources, which may themselves have notes?

The practical import of the question is how I can best structure all this
best.

I tentatively setup things so I have a `biblio` subdir, where I put
individual bib notes.

So I'm just trying to understanding how best to do linking within those
bib notes.

Is it likely I'd want to liberally include cite: links AND link to their
related notes?

jethrokuan:
I'd cite where the source is indeed that particular reference, and use file links everywhere else. `cite:` links is not a thing invented by org-roam, it's an org-ref thing. There's some discussion on Slack about getting org-ref's note-taking capabilities and org-roam to play nicely, but I haven't looked at it myself.

bdarcus:
> I'd cite where the source is indeed that particular reference, and use file links everywhere else.

OK, that's helpful.

So in that case, using the `cite:` link is effectively a way to designate an org-roam note as a bibliographic note.

Issue #378, opened 2020-03-30
Support for navigating daily notes.


tmalsburg:
### Brief Abstract
When I'm looking at the notes for a day, e.g. `org-roam-today`, I'd like to have function (bound to some key) that allows me to go back and forward a day at a time.

### Long Description
One question: If I'm at today and I go back one, but there is no note for yesterday, do we want to create a note for yesterday, or simply go to the note before today that is the closest in time? Personally, I'd prefer if the missing note was created on the fly.  It may stay empty but that's not a big deal.

### Proposed Implementation (if any)
If the title of the current note looks like a date, take that date, subtract (or add) one day, open note for that new date or create on the fly if non-existent.  If the title does not look like a date, show some message.

I can make a PR if you think this would be a useful addition.  If there is already a way to navigate daily notes, please let me know.

### Please check the following:
- [X] No similar feature requests

jethrokuan:
Do you think it's a good idea to create an empty note with no content for the day while searching forward and backwards? Wouldn't this create noise in your directory?

fwiw if you're using the daily notes stuff heavily, please consider using org-journal instead. There's a lot of functionality there that works well with org-roam (like navigating between daily notes), and I don't think it's a good idea to reinvent the wheel here. The daily notes functionality is sort-of legacy, I'm using org-journal in place of it.

tmalsburg:
> Do you think it's a good idea to create an empty note with no content for the day while searching forward and backwards? Wouldn't this create noise in your directory?

Yes, reasons: 1. I will have an entry for most days anyways. 2. Even if the note is empty, I may link to it from elsewhere, and it may be useful to open the empty note to see the backlinks. 3. I'm simply not worried about having many notes and think notes should be created liberally.  They are not getting in my way anyway (and if they do, that's a UI problem).

> fwiw if you're using the daily notes stuff heavily, please consider using org-journal instead. There's a lot of functionality there that works well with org-roam (like navigating between daily notes), and I don't think it's a good idea to reinvent the wheel here. The daily notes functionality is sort-of legacy, I'm using org-journal in place of it.

Didn't know about org-journal. In this case, I suggest removing the diary functions from org-roam.  IMO org-roam is too young to maintain legacy code, and I think other new users may walk down the same garden path as I did.

jethrokuan:
Org-journal has a nice calendar interface that shows you which days have entries, and will get polluted when there are empty journal dates. 

[...]

PR #379, opened 2020-03-30
Functions for navigating to next/previous day

tmalsburg:
Addresses #378.

Note the comment in the code.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -503,6 +503,32 @@ This uses the templates defined at `org-roam-capture-templates'."
   (let ((time (org-read-date nil 'to-time nil "Date:  ")))
     (org-roam--file-for-time time)))
 
+(defun org-roam--move-days (days)
+  "If the current note has a title in date format yyyy-mm-dd, add
+DAYS days (can be negative) and open the note for the resulting
+new date.  If that note does not exist, it will be created."
+  (let* ((title (car (plist-get (org-export-get-environment) :title))))
+    (unless (string-match-p "^[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]$" title)
+      (user-error "The title is not a date with format yyyy-mm-dd."))
+    (let* ((date (org-parse-time-string title))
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -503,6 +503,32 @@ This uses the templates defined at `org-roam-capture-templates'."
   (let ((time (org-read-date nil 'to-time nil "Date:  ")))
     (org-roam--file-for-time time)))
 
+(defun org-roam--move-days (days)
+  "If the current note has a title in date format yyyy-mm-dd, add
+DAYS days (can be negative) and open the note for the resulting
+new date.  If that note does not exist, it will be created."
+  (let* ((title (car (plist-get (org-export-get-environment) :title))))
+    (unless (string-match-p "^[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]$" title)
```
tmalsburg: Here it would be good to use `org-roam-date-title-format` to parse the string. But it appears there is no standard elisp function for parsing string following a specific format?  I must be overlooking something.

jethrokuan:
After some thought, I have decided to modularize the daily notes functionality, and update the functionality there. I intend for it to initially be a subset of what org-journal does, with some upgrades (e.g. a fuller starting template) it should also work out of the box without tweaking.

org-roam.el
tmalsburg: [Someone wrote](https://emacs.stackexchange.com/questions/57460/check-whether-a-string-is-a-date-in-a-specific-format-e-g-y-m-d) an Elisp wrapper for strptime(3) that could be used here.

tmalsburg:
Great!  I actually stuck with org-roam's daily notes so far.  Just works for my purposes.

Is the new code pushed already?  Couldn't find it.

jethrokuan:
Nope nothing is pushed yet!

jethrokuan:
@zaeph is working on upgrading the dailies functionality. Should have some news soon.

zaeph:
PR is #978.

PR #380, opened 2020-03-30
Allow top and bottom placement of the sidebar in addition to left and right

philip-bl:
I like to have two files opened in my emacs side by side. Having a sidebar on the left or on the right messes that up, because I don't have enough horizontal space for 3 windows. So I wanted to be able to open it on the bottom instead.

##### Description of changed code
Since there already was code that opened the sidebar on the side, it was trivial to modify to do it on top or on the bottom. I have manually tested that it works.

##### New and modified `customize-group org-roam` variables
* Added `org-roam-buffer height`
* `org-roam-buffer-position` can now be either `top`, or `bottom`, or `left`, or `right`

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -702,14 +702,12 @@ Valid states are 'visible, 'exists and 'none."
         (display-buffer-in-side-window
          `((side . ,org-roam-buffer-position)))
         (select-window))
-    (progn
-      (if (or (eq org-roam-buffer-position 'left)
-              (eq org-roam-buffer-position 'right))
-          (org-roam--set-width
-           (round (* (frame-width) org-roam-buffer-width)))
-          (org-roam--set-height
-           (round (* (frame-height) org-roam-buffer-height))))
-      (select-window window))))
+    (pcase org-roam-buffer-position
+      ('right  (org-roam--set-width  (round (* (frame-width)  org-roam-buffer-width))))
+      ('left   (org-roam--set-width  (round (* (frame-width)  org-roam-buffer-width))))
+      ('top    (org-roam--set-height (round (* (frame-height) org-roam-buffer-height))))
+      ('bottom (org-roam--set-height (round (* (frame-height) org-roam-buffer-height)))))
+    (select-window window)))
 
 (defun org-roam ()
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -672,17 +684,32 @@ Valid states are 'visible, 'exists and 'none."
...
   (let ((window (get-buffer-window)))
     (-> (get-buffer-create org-roam-buffer)
         (display-buffer-in-side-window
          `((side . ,org-roam-buffer-position)))
         (select-window))
-    (org-roam--set-width
-     (round (* (frame-width)
-               org-roam-buffer-width)))
-    (select-window window)))
+    (progn
```
jethrokuan: this `progn` looks unnecessary

org-roam.el
```diff
@@ -672,17 +684,32 @@ Valid states are 'visible, 'exists and 'none."
...
+      (if (or (eq org-roam-buffer-position 'left)
+              (eq org-roam-buffer-position 'right))
+          (org-roam--set-width
+           (round (* (frame-width) org-roam-buffer-width)))
+          (org-roam--set-height
+           (round (* (frame-height) org-roam-buffer-height))))
```
jethrokuan: indentation looks off, at least. Maybe it's better to change it to a pcase:

```
(pcase org-roam-buffer-position
  ('right ...)
  ('left ...)
  ('top ...)
  ('bottom ...)
  (pos (error "invalid org-roam-buffer-position %s" pos)))
```

org-roam.el
philip-bl: Fixed in https://github.com/jethrokuan/org-roam/pull/380/commits/006679bee58bdc22d1ad592e7f180a9de6409807

org-roam.el
philip-bl: Fixed in https://github.com/jethrokuan/org-roam/pull/380/commits/006679bee58bdc22d1ad592e7f180a9de6409807

philip-bl:
@jethrokuan I've made both of your suggested fixes.

philip-bl:
And I've just rebased my branch on your current master branch.

PR #383, opened 2020-03-31
Update org-roam-graph variable names

progfolio:
See: https://github.com/jethrokuan/org-roam/issues/382

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -98,14 +98,14 @@ Org-roam tries its best to locate the Graphviz executable from your
 `PATH`, but if it fails to do so, you may set it manually:
 
 ```
-(setq org-roam-graphviz-executable "/path/to/dot")
+(setq org-roam-graph-executable "/path/to/dot")
 ```
 
 You may also choose to use `neato` in place of `dot`, which generates a more compact graph layout.
 
 ```
-(setq org-roam-graphviz-executable "/path/to/neato")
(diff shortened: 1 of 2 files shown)


jethrokuan:
is there a reason why `org-roam-grapher-extra-options` is does not have the `org-roam-graph` prefix?

progfolio:
> is there a reason why org-roam-grapher-extra-options does not have the org-roam-graph prefix?

It probably slipped through the linting because it matches the regexp `^org-roam-graph.*`
looks like `makem.sh` is using `package-lint` which checks with the following:

```emacs-lisp
(defun package-lint--get-package-prefix ()
  "Return package prefix string (i.e. the symbol the package `provide's).
Prefix is returned without any `-mode' suffix."
  (let ((feature (package-lint--provided-feature)))
    (when feature
      (replace-regexp-in-string "-mode\\'" "" feature))))
```

I agree it's not a very good name. How about changing it to something like `org-roam-graph-executable-options`? Or any other suggestions?

jethrokuan:
These options are not for the graphviz executable, but injected into the graphviz output to configure the graph output. Maybe `org-roam-graph-extra-config` or something? And alias the past 2 names.

progfolio:
Ah, understood. Declaring the aliased versions obsolete as well?

jethrokuan:
> Ah, understood. Declaring the aliased versions obsolete as well?

Sure!

PR #388, opened 2020-04-01
Refactor org-roam--setup-buffer

progfolio:
Change function name org-roam--setup-buffer to org-roam--set-up-buffer.
Notify user and default to 'right when org-roam-buffer-position is
an invalid position.
Make better use of pcase to avoid repeating code.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -694,18 +694,26 @@ Valid states are 'visible, 'exists and 'none."
        ((< (window-height) h)
         (enlarge-window (- h (window-height))))))))
 
-(defun org-roam--setup-buffer ()
-  "Setup the `org-roam' buffer at the `org-roam-buffer-position'."
-  (let ((window (get-buffer-window)))
+(defun org-roam--set-up-buffer ()
+  "Set up the `org-roam' buffer at the `org-roam-buffer-position'."
+  (let ((window (get-buffer-window))
+        (position
+         (if (member org-roam-buffer-position '(right left top bottom))
+             org-roam-buffer-position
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -694,18 +694,25 @@ Valid states are 'visible, 'exists and 'none."
...
-  "Setup the `org-roam' buffer at the `org-roam-buffer-position'."
-  (let ((window (get-buffer-window)))
+(defun org-roam--set-up-buffer ()
+  "Set up the `org-roam' buffer at the `org-roam-buffer-position'."
+  (let ((window (get-buffer-window))
+        (position
+         (if (member org-roam-buffer-position '(right left top bottom))
+             org-roam-buffer-position
+           (let ((text-quoting-style 'grave))
+             (message "Invalid org-roam-buffer-position: %s. Defaulting to 'right"
```
jethrokuan, line 704: Maybe use `display-warning` here.

jethrokuan:
Many thanks for all the refactorings so far, my elisp-fu still needs some work

org-roam.el
progfolio: Would you prefer `display-warning` or `lwarn`? With display warning we would have to pass the string into `format` to embed the invalid value, `lwarn` allows format specifiers in the warning message string.

progfolio:
Thank you! This has been a very fun project to work on and I'm excited for its potential.

org-roam.el
jethrokuan: `lwarn` works too!

PR #392, opened 2020-04-01
(feature): Add numeric prefix argument to `org-roam-tomorrow`

juergenhoetzel:
Also add prefix argument to `org-roam-yesterday`.

This makes it possible to jump back or forward several days.

I often plan several days in the future

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -499,15 +499,17 @@ This uses the templates defined at `org-roam-capture-templates'."
   (interactive)
   (org-roam--file-for-time (current-time)))
 
-(defun org-roam-tomorrow ()
-  "Create and find the file for tomorrow."
-  (interactive)
-  (org-roam--file-for-time (time-add 86400 (current-time))))
+(defun org-roam-tomorrow (n)
+  "Create and find the file for tomorrow.
+With numeric argument N, use N days in the future."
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -499,15 +499,17 @@ This uses the templates defined at `org-roam-capture-templates'."
...
-
-(defun org-roam-yesterday ()
-  "Create and find the file for yesterday."
-  (interactive)
-  (org-roam--file-for-time (time-add -86400 (current-time))))
+(defun org-roam-tomorrow (n)
+  "Create and find the file for tomorrow.
+With numeric argument N, use N days in the future."
+  (interactive "p")
+  (org-roam--file-for-time (time-add (* n 86400) (current-time))))
```
jethrokuan, line 506: is the default arg 1?

juergenhoetzel: Yes:
```
— Function: prefix-numeric-value arg
This function returns the numeric meaning of a valid raw prefix argument value, arg. The argument may be a symbol, a number, or a list. If it is nil, the value 1 is returned; if it is -, the value −1 is returned; if it is a number, that number is returned; if it is a list, the car of that list (which should be a number) is returned.
```

PR #398, opened 2020-04-03
Allow to create graph only of nodes connected to current entry

kljohann:

As I found the graph increasingly unwieldy with a growing number of nodes, I wanted to only display the connected component containing the current node, which is implemented in the interactive function `org-roam-graph-show-connected-component`. It's based on `org-roam-db--connected-component`, which uses recursive common table expressions to retrieve all related file names.

To this end, I also adapted `org-roam-graph--build` and `org-roam-graph-show` to accept a node query, which was nice for playing around with different ideas. I'm not sure whether it should stay part of the interface of `org-roam-graph-show`, though. It could be moved to a separate function containing the current implementation.

In `org-roam-graph--build` the node query will be used to retrieve only those edges from the database, whose nodes should be displayed.

As there are no additional indexes, I would be interested in feedback w.r.t. performance/responsiveness from someone with a considerable number of notes.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -257,6 +257,21 @@ This is equivalent to removing the node from the graph."
                            file
                            :limit 1)))
 
+(defun org-roam-db--connected-component (file)
+  "Return all files reachable from/connected to FILE, including the file itself.
+If the file does not have any connections, nil is returned."
+  (let* ((query "WITH RECURSIVE
+                   links_of(file, link) AS
+                     (SELECT \"from\", \"to\" FROM links UNION
+                      SELECT \"to\", \"from\" FROM links),
+                   connected_component(file) AS
+                     (SELECT link FROM links_of WHERE file = $s1
+                      UNION
+                      SELECT link FROM links_of JOIN connected_component USING(file))
+                   SELECT * FROM connected_component;")
+         (files (mapcar 'car-safe (emacsql (org-roam-db) query file))))
+    files))
+
 ;;;;; Updating
 (defun org-roam-db--update-titles ()
   "Update the title of the current buffer into the cache."
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -38,6 +38,8 @@
 (defvar org-roam-directory)
 (declare-function org-roam-db--ensure-built  "org-roam-db")
 (declare-function org-roam-db-query          "org-roam-db")
+(declare-function org-roam-db--connected-component "org-roam-db")
+(declare-function org-roam--org-roam-file-p  "org-roam")
 (declare-function org-roam--path-to-slug     "org-roam")
 
 ;;;; Options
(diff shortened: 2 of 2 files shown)


kljohann:
(One additional minor improvement would be to color the currently selected node, which should be relatively easy. Maybe the node properties could be computed by calling a hook or function, which could be let-bound in `org-roam-graph-show-connected-component` or even user-provided.)

org-roam-graph.el
```diff
@@ -122,21 +123,18 @@ set WHERE to true if WHERE query already exists."
...
   (org-roam--with-temp-buffer
-    (let* ((node-query `[:select [file titles]
-                                 :from titles
-                                 ,@(org-roam-graph--expand-matcher 'file t)])
-           (nodes (org-roam-db-query node-query))
-           (edges-query `[:select :distinct [file-to file-from]
-                                  :from file-links
-                                  ,@(org-roam-graph--expand-matcher 'file-to t)
-                                  ,@(org-roam-graph--expand-matcher 'file-from t t)])
+    (let* ((nodes (org-roam-db-query node-query))
```
kljohann, line 133: Alternatively we might accept just the `:where` expression, instead of passing in the full query. (Since the rest of this function depends on the "file" and "titles" columns being returned in the correct order.

org-roam-db.el
```diff
@@ -254,6 +254,20 @@ This is equivalent to removing the node from the graph."
...
+  "Return all files reachable from/connected to FILE."
+  (let* ((query "WITH RECURSIVE
+                   links_of(file, link) AS
+                     (SELECT file_from, file_to FROM file_links UNION
+                      SELECT file_to, file_from FROM file_links),
+                   connected_component(file) AS
+                     (SELECT link FROM links_of WHERE file = $s1
+                      UNION
+                      SELECT link FROM links_of JOIN connected_component USING(file))
+                   SELECT * FROM connected_component;")
```
kljohann, line 267: Unfortunately I was not able to express two CTEs in a single query using the emacsql s-expressions.

org-roam-graph.el
kljohann: The benefit of the current approach is that we could pass in e.g.
```elisp
[:select [file titles]
 :from file-links
 :join titles :on (or (= file file-from)
                      (= file file-to))
 ,@(org-roam-graph--expand-matcher 'file t)]
```
to display only notes that have any connections at all.

org-roam-db.el
jethrokuan: this is fine.

org-roam-graph.el
```diff
@@ -190,6 +191,16 @@ If PREFIX, then the graph is generated but the viewer is not invoked."
 	        (call-process org-roam-graph-viewer nil 0 nil temp-graph)
         (view-file temp-graph)))))
 
+(defun org-roam-graph-show-connected-component (&optional prefix)
```
jethrokuan, line 194: This function probably needs a check for whether it's an org-roam file via `org-roam--org-roam-file-p`.

kljohann: In principle this should not be necessary, since non-roam files are not present in the database, consequently the query does not select any nodes. I tested it and the current behavior is an empty graph for non-roam files.
Should we still introduce an explicit check? And if yes: Keep the empty graph behavior or emit a warning?

jethrokuan: Yes, it doesn't error out. The point is to not generate the graph at all for a non-roam file, since it doesn't make sense to.

jethrokuan:
This also needs a rebase on top of #374 : mostly renaming file_links -> links, file_to -> to, file_from -> from

org-roam-graph.el
kljohann: Sounds good, I'll also take care of the rebase.

d12frosted:
This is really great. Thank you very much! 💯 

Would be also nice if this new function would respect `org-roam-graph-exclude-matcher`. As far as I understand it doesn't use it now.

kljohann:
Thanks! I started to work on respecting the exclude matcher in #452, but I did not yet get around to finish it.

d12frosted:
BTW, it would also be cool to be able to exclude by tags 😸 For example, it would allow me to exclude all my daily files (cause I mention different projects in daily files and it becomes messy haha).

Issue #346, opened 2020-03-23
Multiple levels in the backlinks buffer


bbardsley:
In the [blog post](https://www.nateliason.com/blog/roaml) by Nate Liason, I saw that he creates entries with certain dates and below it puts a to-do list, ordered by lower-level headings. Then, when referring to the file with the date, one can view the to-do list in the side bar. In other words, one can see all the levels below the level where the link to the current file is placed. However, in the current build of org-roam, one can only view one level, is it possible to see any lower level headings?

jingnanshi:
This feature will be very helpful. The link provided is broken. [Here](https://www.nateliason.com/blog/roam) is the actual link to the blog post.

rigond:
This feature will allow you to build content feeds indirectly from multiple files based on a link or tag that captures a lower (or same) level content block. No additional effort will be required. Such feeds can, for example, act as an inboxes for your ideas, content, or tasks. 

Here is an example of what I mean: https://youtu.be/A6PIrVZoZAk?t=1250.

However, you would have to be able to filter and order the contents of the side-bar to search or to reduce clutter.
The side-bar can also act as a preview window to another note as well.

remedialchaos:
I've been thinking about this lately. When I add notes in daily, I usually just create a heading with the file link, and add dump everything I have below that level. Then when I'm consolidating all the dumps, I see which dates I have entries for, and only the heading. 

I understand it will clutter the backlinks buffer if everything under the heading is shown all the time, but perhaps we can have an option to expand the buffer in same org-mode view styles (top only, headings only, everything) in rotation on tab press. That will combine best of both worlds of being spartan by default and showing extra information by default when wanted. this way, user doesn't have to open the entire file and lose the context of open file in front just to refer and copy something from under the heading.

hpfr:
Fewer upvotes but more discussion at https://github.com/org-roam/org-roam/issues/245

Issue #323, opened 2020-03-18
Show notes that have links in common with the current one


khinsen:
### Brief Abstract

Backlinks are useful for finding related notes. Another feature of related notes is having links in common with the current one. 

### Long Description

This feature adds a second section to the `org-roam` buffer that quotes notes with common links, in the same way it displays backlinks. I have a working implementation, which I have been using for a few days and found useful so far. I opened this issue in order to get feedback. Does this look useful to others? Could it be made even more useful?

### Proposed Implementation (if any)

https://github.com/khinsen/org-roam/tree/show-common-links-in-org-roam-buffer

### Please check the following:
- [X] No similar feature requests

bdarcus:
Could you post a screenshot of it here?

khinsen:
<img width="881" alt="Screenshot 2020-03-18 at 20 45 32" src="https://user-images.githubusercontent.com/94934/77001007-8b66d700-6959-11ea-81db-50866fc8f2a5.png">

bdarcus:
Is this use case related to #31?

khinsen:
It's a similar idea indeed. What #31 proposes is a full-text search for the current note's title. That would probably yield more hits than my approach, which analyzes only the link structure. For me that makes more sense because I carefully think about what I want to link to in order to exploit the link structure later on. I create linked notes for anything that looks like a keyword I might be interested in later on.

My idea for implementing #31 would be a kind of database maintenance operation, launched manually from time to time, that checks all notes for the titles of all other notes that are not already links, and proposes to convert them into links.

jethrokuan:
I think something like `org-agenda-view` is due for the backlink buffer. We need some way to customize what is shown on the backlink buffer, similar to how org-agenda does it:

```
(setq org-roam-backlinks-view '((org-roam-backlinks
                                 :depth 2)
                                (org-roam-common-links)))
```

I'm thinking of having the backlinks buffer just be a special case of the advanced search buffer (with search term "has tag:current-file". These are huge projects that I don't have the bandwidth right now to tackle, but can expect in a 2 months after I'm done with my studies.

khinsen:
My current code has minimal customization (yes/no for each section), but that could indeed be made more flexible, at the cost of leaving the comfort zone of Emacs' customization system (which is fine with me).

[...]

Issue #245, opened 2020-03-07
Provide an option to show the entire contents of the linked documents in Backlinks


ecvgit:
### Brief Abstract
I create tags in my org mode documents. However, when I navigate to the note for a tag, I only see a small portion of the original document in the back links. It would be good to provide an option to show the entire content of the document in the backlinks. So that one can see the entire document that is being linked to without going back and forth.
### Long Description
Let us say I have a document like 
#+Title: Title 1
tags :: tag1, tag2
** Heading 2
  Some content 3
** Heading 3
 ** Some content 3

Now, when I navigate to tag1, the backlinks section in org-roam buffer shows only the title and tags line. I want it to show the entire contents of Title 1 document when necessary.

### Proposed Implementation (if any)
Not sure how to implement this. May be something like org-show-context-detail can fix this?

### Please check the following:
- [x] No similar feature requests

jethrokuan:
This will likely destroy the formatting of the org-roam buffer. Rather than view the full contents within the narrow sidebar, wouldn't just opening the file be better?

ecvgit:
My use case is this:
Let us say I have a tag called 'China' - I want to see all the notes I have written on china without having to navigate back and forth between individual notes. Just having the title does not help. 
I think this will be a powerful feature.

Or one can provide a separate command, that will collect all the backlinks in a new buffer with the entire contents, which the user can export, save, etc. 

I understand your point about it destroying the formatting. But may be there is a way to work around this by providing making it a sub node of a new node (or something like this)?

jethrokuan:
Yeah okay, got it. It'll probably be a separate interface (search interface, that will allow complex queries like has tag "china" and tag "korea").

ecvgit:
Thank you. That was precisely what I was thinking about. 

Also, I think there is value to having an option similar to the current setup as well (where only related context is shown). So it would be nice to make it an option.

herbertjones:
Here's a proof of concept that provides `org-roam-search`.  It allows searching for all backlinks to a file in a new buffer, and has a toggle to insert just the reference line, or the entire article.

https://gist.github.com/herbertjones/a55131b7bb9e74d61159f5ba14613ef8

jethrokuan:
@herbertjones nice! This was how I imagined it to be as well (inspired from https://github.com/Wilfred/deadgrep)

Wetlize:
[...]

Issue #219, opened 2020-03-03
"Sticky'ing" the org-roam window 


yangnom:
### Brief Abstract
It would be good if I could have the org-roam window be the backlinks for a certain org-roam file when I'm looking at another org-roam file

### Long Description
I've been using org-roam for about a week, and amassing notes, but I've run into two problems and the larger one is that I'll have the file that I want to amass all of my notes into from my daily notes, let's say it's my `emacs` file.
I'll open up the org-roam window on the right to see all the backlinks, and then I'll open a backlink (into a new buffer window below my main `emacs` file, let's say it's `2020-02-29`), but as I'm going through those notes, the org-roam window switches to show me the backlinks for the `2020-02-29`, which is generally pretty useless as I don't think I'll ever have backlinks for the `2020-02-29` type of daily org-roam file.
It would be great if I could "sticky" the `emacs` org-roam backlinks so I can go through the backlinks and open them up in the lower window buffer without losing where I am in the `emacs` backlinks windows.

I hope I've described myself well enough and apologize if I haven't.

The second thing would be if the backlinks could be in reverse order from _newest_ to _oldest_ but that wouldn't be so important if I could "sticky" the org-roam sidebar to a certain files backlinks.

### Please check the following:
- [x] No similar feature requests

jethrokuan:
Maybe this is related to tracking file trails: #118 .

I can see the losing of context happening, and do face it a little myself. I'd need to think about the interface a bit harder.

yangnom:
My best workflow to get around this problem is to make a new workspace, open all the back links in seperate buffers in that workspace, split the screen between the cleanup notes, `emacs` and the daily notes like `2020-02-29`, and then go through all the backlinked buffers closing them off one by one as I copy and redo the relevant information in them to the main notes.

PR #406, opened 2020-04-07
Modularize org-roam-buffer

progfolio:
Split functionality related to the org-roam-buffer into its own
feature/file.

Introduce `org-roam-buffer-prepare-hook' to allow customizations of the
org-roam-buffer content.

Related issues: #346 #323 #245 #219 #198 #95 #14

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -1,0 +1,266 @@
+;;; org-roam-buffer.el --- Roam Research replica with Org-mode -*- coding: utf-8; lexical-binding: t -*-
+
+;; Copyright © 2020 Jethro Kuan <jethrokuan95@gmail.com>
+
+;; Author: Jethro Kuan <jethrokuan95@gmail.com>
+;; URL: https://github.com/jethrokuan/org-roam
+;; Keywords: org-mode, roam, convenience
+;; Version: 1.0.0-rc1
+;; Package-Requires: ((emacs "26.1") (dash "2.13") (f "0.17.2") (s "1.12.0") (org "9.3") (emacsql "3.0.0") (emacsql-sqlite "1.0.0"))
+
+;; This file is NOT part of GNU Emacs.
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to the
+;; Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+;; Boston, MA 02110-1301, USA.
+
+;;; Commentary:
+;;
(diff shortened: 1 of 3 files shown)


org-roam-buffer.el
```diff
@@ -0,0 +1,266 @@
...
+         `((side . ,position)))
+        (select-window))
+    (pcase position
+      ((or 'right 'left)
+       (org-roam-buffer--set-width  (round (* (frame-width)  org-roam-buffer-width))))
+      ((or 'top  'bottom)
+       (org-roam-buffer--set-height (round (* (frame-height) org-roam-buffer-height)))))
+    (select-window window)))
+
+(defun org-roam-buffer-show ()
```
jethrokuan, line 256: This function also hides the buffer if it is shown, so maybe both the docstring and function name needs to change.

org-roam.el
```diff
@@ -798,17 +608,17 @@ Otherwise, behave as if called interactively."
...
 (defun org-roam--find-file-hook-function ()
   "Called by `find-file-hook' when mode `org-roam-mode' is on."
   (when (org-roam--org-roam-file-p)
     (setq org-roam-last-window (get-buffer-window))
-    (add-hook 'post-command-hook #'org-roam--maybe-update-buffer nil t)
+    (add-hook 'post-command-hook #'org-roam-buffer--update-maybe nil t)
     (add-hook 'after-save-hook #'org-roam-db--update-file nil t)
     (org-link-set-parameters "file" :face 'org-roam--roam-link-face)
-    (org-roam--maybe-update-buffer :redisplay nil)))
+    (org-roam-buffer--update-maybe :redisplay t)))
```
jethrokuan, line 621: the value here changed, but afaict it shouldn't affect behaviour, since find-file should cause an update anyway.

progfolio: I believe I changed this due to another, unrelated issue I was investigating. Happy to change it back if you like.

org-roam-buffer.el
```diff
@@ -0,0 +1,266 @@
...
+                        (org-roam--extract-ref)))
+            (key-backlinks (org-roam--get-backlinks (s-chop-prefix "cite:" roam-key)))
+            (grouped-backlinks (--group-by (nth 0 it) key-backlinks)))
+      (progn
+        (insert (format "\n\n* %d Cite backlinks\n"
+                        (length key-backlinks)))
+        (dolist (group grouped-backlinks)
+          (let ((file-from (car group))
+                (bls (cdr group)))
+            (insert (format "** [[file:%s][%s]]\n"
```
kljohann, line 111: Maybe also use `(org-link-make-string link description)` here (and in backlinks function), as it will take care of proper escaping?

org-roam-buffer.el
```diff
@@ -0,0 +1,266 @@
...
+                            file-from
+                            (org-roam--get-title-or-slug file-from)))
+            (dolist (backlink bls)
+              (pcase-let ((`(,file-from _ ,props) backlink))
```
kljohann, line 115: Maybe instead:
```suggestion
              (cl-destructuring-bind (file-from _ props) backlink
```
or even replace `dolist` by an appropriate `cl-loop` form.

org-roam-buffer.el
progfolio: I agree this is the right thing to do. Probably want to wrap `org-link-make-string' in a conditional alias so we can cater to pre 9.3 versions of Org as well. I'll look into this today.

progfolio: See: #412 This would allow us to take care of backward compatibility in a clean way.

jethrokuan:
The above are non-blockers, so I'll merge this in first.

PR #412, opened 2020-04-08
(internal): Move aliases to org-roam-compat.el

progfolio:
Same strategy Org uses. Consolidates backward compatibility into a
single location.


jethrokuan:
Great idea!

HaoZeke:
Unfortunately this seems to break with `Cannot open load file: No such file or directory, org-roam-compat` using the latest `org-roam` and `company-org-roam`

```
  ✓ (36/345) org-roam updated (6e97003 -> dfb8449)
  ✓ (37/345) company-org-roam updated (0913d86 -> 063581d)

```

Issue #399, opened 2020-04-05
`org-roam-capture--in-process` remains `t` even when the capture buffer is closed


hungpham2511:
#### Steps to Reproduce
1. Insert a new link;
2. Close the capture buffer manually;

#### Expected Results
1. org-roam-capture--in-process is nil, thus allows subsequent captures.

#### Actual Results
1. org-roam-capture--in-process is t, prevent subsequent captures.

### Versions
- Emacs (`C-h v emacs-version`): 28.0.50
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/4ca2606a28f8898299f2e0cc01c01b57cf848006

hungpham2511:
A related issue, which actually is the reason why I happened to "close the capture buffer manually" is that org-roam-insert basically hang when inserting a new note. This behavior is quite random and I can't reproduce it.

isolet:
I got `user-error: Org-roam capture in process` every time I'm trying to run `org-roam-find-file`.

juergenhoetzel:
Fixed by #413

PR #413, opened 2020-04-08
(bugfix): Use predicate to check for a `org-roam-capture` process

juergenhoetzel:
Instead of using a global variable to check for a existing capture,
leverage the :org-roam plist property of the buffer-local variable
`org-roam-capture-current-plist`.

This prevents a lockout if the capture buffer is killed.

When killing the capture buffer instead of using <kbd>C-c C-k</kbd> I can't start new capture processes:
```
user-error: Nested Org-roam capture processes not supported
```

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -364,16 +364,16 @@ If PREFIX, downcase the title before insertion."
           (when region ;; Remove previously selected text.
             (delete-region (car region) (cdr region)))
           (insert (org-roam--format-link target-file-path link-description)))
-      (if org-roam-capture--in-process
-          (user-error "Nested Org-roam capture processes not supported")
-        (let ((org-roam-capture--info (list (cons 'title title)
-                                            (cons 'slug (org-roam--title-to-slug title))))
-              (org-roam-capture--context 'title))
-          (add-hook 'org-capture-after-finalize-hook #'org-roam-capture--insert-link-h)
-          (setq org-roam-capture-additional-template-props (list :region region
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -364,16 +364,16 @@ If PREFIX, downcase the title before insertion."
...
-              (org-roam-capture--context 'title))
-          (add-hook 'org-capture-after-finalize-hook #'org-roam-capture--insert-link-h)
-          (setq org-roam-capture-additional-template-props (list :region region
-                                                                 :link-description link-description
-                                                                 :capture-fn 'org-roam-insert))
-          (org-roam--capture))))))
+      (when (org-roam-capture--in-process-p)
+	(user-error "Nested Org-roam capture processes not supported"))
+      (let ((org-roam-capture--info (list (cons 'title title)
+					  (cons 'slug (org-roam--title-to-slug title))))
```
progfolio, line 370:
```suggestion
      (let ((org-roam-capture--info `((title . ,title) (slug . ,(org-roam--title-to-slug title))))
```

jethrokuan:
Nice, this is smart, and should also fix #399

org-roam.el
juergenhoetzel: Thanks for your Feedback. I prefer to follow git best practices and use single-purpose commits. 
I myself often got sidetracked to code cleanups an introduced non-related bugs :roll_eyes:

PR #415, opened 2020-04-09
Org roam org ref extra

myshevchuk:
This small library offers a tighter integration between org-roam and org-ref by providing `org-roam-extra-org-ref-notes-fn` function, which upon installing into `org-ref-notes-function` will enable org-ref to use org-roam as a backend for managing bibliography notes.

Optionally, `org-roam-capture-templates` can be dynamically preformatted with bibtex field values.

Optionally, automatic switching to the perspective (persp-mode) with the notes project (projectile) is possible.


org-roam-extra.el
```diff
@@ -0,0 +1,268 @@
...
+;; with the notes project (projectile) is possible.
+
+;;; Code:
+;;;; Library Requires
+(require 'org-roam)
+
+(autoload 'org-ref-find-bibliography "Find the bibliography in the buffer.")
+(autoload 'projectile-relevant-open-projects "projectile" "Return a list of open projects.")
+(autoload 'persp-switch "persp-mode" "Switch to the perspective with name NAME.")
+(autoload 'persp-names "persp-mode" "")
```
jethrokuan, line 51: Do we need to declare autoloads for externally provided functions? seems odd to me

org-roam-extra.el
```diff
@@ -0,0 +1,268 @@
...
+      (setf (nth 4 template) tp))
+    template))
+
+(defun org-roam-extra--switch-perspective ()
+  "Helper function for `org-roam-extra-org-ref-note-fn'."
+  (when
+      (and (featurep 'projectile)
+           (require 'projectile nil t)
+           (featurep 'persp-mode)
+           (require 'persp-mode nil t))
```
jethrokuan, line 182: not sure if the `featurep` checks are needed, I think just the `require` clauses are needed.

org-roam-extra.el
```diff
@@ -0,0 +1,268 @@
...
+
+4. Optionally, if you are using projectile and persp-mode and
+have a dedicated workspace to work with your org-roam collection,
+you may want to set the perspecive name and project path in
+`org-roam-extra-persp-project' and `org-roam-extra-switch-persp' to
+t. In this case, the perspective will be switched to the org-roam
+notes project before calling any org-roam functions."
+  (when
+      (and (featurep 'org-ref)
+           (require 'org-ref))
```
jethrokuan, line 241: Same here.

org-roam-extra.el
```diff
@@ -0,0 +1,268 @@
...
+      (org-roam-mode +1))
+    (let ((note-info (list (cons 'ref citekey))))
+      ;; Optionally switch to the notes perspective
+      (when org-roam-extra-switch-persp
+        (org-roam-extra--switch-perspective))
+      ;; Find org-roam reference with the CITEKEY
+      (unless (ignore-errors (org-roam-find-ref note-info))
+        ;; Call org-roam-find-file
+        (let* ((bibtex-completion-bibliography (org-ref-find-bibliography))
+               (entry (ignore-errors (bibtex-completion-get-entry citekey)))
```
jethrokuan, line 252: This seems to use things from `helm-bibtex.el`, so a require might be needed. Maybe this dependency gets pulled in transitively, but it's better to do an explicit require in case the upstream repository changes.

org-roam-extra.el
myshevchuk: This is to silence the compiler about not defined functions. I don't like it either, although the idea is not mine. I'll try to figure a better way to silence the compiler and not to require these optional features unconditionally in the top level.

org-roam-extra.el
myshevchuk: As far as I remember, org-ref pulls bibtex-completion, which in turn pulls helm-bibtex. I look closer at it.

myshevchuk:
Thank you!

I absolutely agree with all the points. 

I initially thought about the`org-roam-org-ref-`  namespace but then it looked like kind of too long. Maybe `org-roam-or-`?

org-roam-extra.el
jethrokuan: Silencing the compiler is typically done with `declare-function`, rather than `autoload`.

myshevchuk: Thank you, I knew "Emacs had a keybinding for that"

org-roam-extra.el
myshevchuk: I had to move `(require 'org-ref)` into the function body because the project would not pass build check otherwise, see above commit d79b245.

This, to my opinion, doesn't make much sense because it is actually a plugin function for org-ref and it should of course be loaded by then. So `(require 'org-ref)` must be either a top-level for or omitted altogether.

bdarcus:
> ###### Motivation for this change
> This small library ... will enable org-ref to use org-roam as a backend for managing bibliography notes.

I'm confused. How, specifically, does this work?

org-roam-extra.el
myshevchuk: Ok, `bibtex-completion` is the backend which is supported by two frontends `helm-bibtex` and `ivy-bibtex`. Although they each belong to the same package, `bibtex-completion` does not depend on them. And since `bibtex-completion` is a dependency of `org-ref`, on which `org-roam-org-ref` depends too, it seems to me that `(require 'org-ref)` should be enough.

myshevchuk:
@bdarcus I'll make a gif.

Basically, put `org-roam-extra-org-ref-notes-fn` in `org-ref-notes-function`. Then press `org-open-at-point` while in an org-mode buffer on at a `cite:` link , select `3. Open notes` and voilà.

To see how the template pre-formatting works, use one of the examples provided in doc-string to `org-roam-extra-preformat-keywords` and set `org-roam-extra-preformat-templates` to t.

bdarcus:
A gif would be perfect.

FWIW, a couple of us are discussing/working on a little biblio module for doom emacs. Perhaps this could be incorporated, based on whether a user has the "+roam" flag selected.

myshevchuk:
Cool!

bdarcus:
> Basically, put `org-roam-extra-org-ref-notes-fn` in `org-ref-notes-function`. 

So this is the suggested setup for multiple files in the org-ref documentation, which is what I use.

``` elisp
(setq org-ref-notes-function
       (lambda (thekey)
         (let ((bibtex-completion-bibliography (org-ref-find-bibliography)))
           (bibtex-completion-edit-notes
            (list (car (org-ref-get-bibtex-key-and-file thekey)))))))
```

You're saying that with this function, we can simplify to something like this?

``` elisp
(setq org-ref-notes-function org-roam-extra-org-ref-notes-fn)
```

Either way, could include the answer in the docs please?

myshevchuk:
Right it is a multi-file setup for org-ref. Just don't forget to quote the function.
```(setq org-ref-notes-function #'org-roam-extra-org-ref-notes-fn
                org-ref-notes-directory "path/to/org-roam-directory")
```

myshevchuk:
Sorry, I could not create a gif, it records my desktop background instead of the Emacs window. Maybe later.
Meanwhile here is a video cast on youtube, in low resolution though, sorry. As far as I understood, the hi-res will appear later after processing.

https://youtu.be/75dbqV3TH3I

org-roam-extra.el
```diff
@@ -0,0 +1,268 @@
...
+Set `org-ref-notes-function' to this function if your
+bibliorgaphy notes are managed by org-roam and you want some extra
+integration between the two packages.
+
+This function allows to use org-roam as a backend for org-ref's
+handling of bibliography notes.
+
+Implementation details and features:
+
+1. This funciton first calls `org-roam-find-ref' trying to find the
```
bdarcus, line 206: typo

jethrokuan:
@myshevchuk can you move this file into a new folder under `contrib/lisp`?

org-roam-org-ref.el
```diff
@@ -0,0 +1,281 @@
...
+(declare-function projectile-relevant-open-projects "projectile")
+(declare-function persp-switch "persp-mode" (name &optional frame (window (selected-window)) (called-interactively-p (called-interactively-p 'any))))
+(declare-function persp-names "persp-mode" (&optional (phash *persp-hash*) (reverse t)))
+
+(defvar org-roam-org-ref-preformat-templates nil
+  "Non-nil to enable template preformatting.
+See `org-roam-org-ref-edit-notes' for details.")
+
+(defvar org-roam-org-ref-preformat-keywords '("citekey" . "=key=")
+  "The template prompt wildcards for preformatting.
```
TimQuelch, line 66: Should these be `defcustom`s rather than `defvar`? If I understand correctly it is intended for these to be user customisable.

myshevchuk: Thank you. Yes definitely, all the user options must be `defcustom`.

jethrokuan:
@myshevchuk it looks like the recommendation from the MELPA side is to package these as a separate repo, and separate package.

myshevchuk:
@jethrokuan ok. I don't think I'll bother with MELPA though.

bdarcus:
> I don't think I'll bother with MELPA though.

Maybe you can work with @Zaeph on a common package, and he can submit to MELPA?

Zaeph:
> > I don't think I'll bother with MELPA though.
> 
> Maybe you can work with @Zaeph on a common package, and he can submit to MELPA?

@myshevchuk: I've shot you a quick email this morning because I didn't know where to include it, but I'll jut paste it here for legacy:

> After checking your code, I’ve come to the conclusion that my code was barely adding anything to yours, and If I wanted to do more, I’d have to write similar functions to yours.
> 
> How about we join forces and co-maintain a new repo together that we’d have on MELPA?  Since you’re not basing your project off org-ref anymore, maybe we could call it org-roam-bib?
> 
> If you’re tight on time, I can take care of the repo creation and add you as maintainer.

jethrokuan:
Even if it doesn't go up on MELPA, exposing it as a GitHub repo still allows people to easily install it using other package managers. I have other packages that aren't on MELPA as well.

In addition, I've found the MELPA reviewers to be very helpful throughout the process (e.g. with code review) so it might be a good experience.

myshevchuk:
@Zaeph Yes, sure. I just haven't had a chance to answer your email yet.

Zaeph:
> @Zaeph Yes, sure. I just haven't had a chance to answer your email yet.

Splendid, then.  I'll take care of it later today.  I'll try to some `git` alchemy to to a subrepo out of the work you've done in the PR to preserve the ownership of the commits, I'll add you as a maintainer, and I'll make a PR on MELPA.

Good work everyone!

myshevchuk:
@jethrokuan  I just don't think it's worth a package on its own to be honest. Just a few helpful snippets. 

@zaeph I don't think preserving the commit history that important. It would be pretty straightforward with online receipts such as this one https://help.github.com/en/github/using-git/splitting-a-subfolder-out-into-a-new-repository if the file were from the very beginning in its own subdirectory. I'm not sure how easy it is in this particular case.

Zaeph:
I've created the new repo:
https://github.com/Zaeph/org-roam-bibtex

PR #418, opened 2020-04-09
Allow to limit the number of jumps when creating graph

kljohann:

As a follow-up to #398 I wanted to investigate whether it's feasible to only traverse the link graph up to a given distance. This is helpful to get a quick (local) overview once the number of connections grow. (Because if all your notes are connected in one way or another, #398 is not going to help…)

To give you an idea, here is the entry on the "Kalman Filter" from https://github.com/jethrokuan/braindump with a maximum distance of one in the first image and two in the second image.

**M-1 M-x org-roam-graph-show-connected-component RET**
graph-1-jump-kalman

**M-2 M-x org-roam-graph-show-connected-component RET**
graph-2-jumps-kalman

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -278,6 +278,7 @@ including the file itself.  If the file does not have any connections, nil is re
                       SELECT lo.link, json_insert(cc.trace, '$[' || json_array_length(cc.trace) || ']', lo.link) FROM
                       connected_component AS cc JOIN links_of AS lo USING(file)
                       WHERE (
+                        -- Avoid cycles by only visiting each file once.
                         (SELECT count(*) FROM json_each(cc.trace) WHERE json_each.value == lo.link) == 0
                         -- Note: BFS is cut off early here.
                         AND json_array_length(cc.trace) < $s2))
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -262,6 +262,31 @@ If the file does not have any connections, nil is returned."
          (files (mapcar 'car-safe (emacsql (org-roam-db) query file))))
     files))
 
+(defun org-roam-db--links-with-max-distance (file max-distance)
+  "Return all files reachable from/connected to FILE in at most MAX-DISTANCE steps,
+including the file itself.  If the file does not have any connections, nil is returned."
+  (let* ((query "WITH RECURSIVE
+                   links_of(file, link) AS
+                     (SELECT \"from\", \"to\" FROM links UNION
+                      SELECT \"to\", \"from\" FROM links),
+                   -- Links are traversed in a breadth-first search.  In order to calculate the
+                   -- distance of nodes and to avoid following cyclic links, the visited nodes
+                   -- are tracked in 'trace'.
+                   connected_component(file, trace) AS
+                     (VALUES($s1, json('[]'))
+                      UNION
+                      SELECT lo.link, json_insert(cc.trace, '$[' || json_array_length(cc.trace) || ']', lo.link) FROM
+                      connected_component AS cc JOIN links_of AS lo USING(file)
+                      WHERE (
+                        (SELECT count(*) FROM json_each(cc.trace) WHERE json_each.value == lo.link) == 0
+                        -- Note: BFS is cut off early here.
+                        AND json_array_length(cc.trace) < $s2))
+                   SELECT DISTINCT file, min(json_array_length(trace)) AS distance
+                   FROM connected_component GROUP BY file ORDER BY distance;")
+         ;; In principle the distance would be available in the second column.
+         (files (mapcar 'car-safe (emacsql (org-roam-db) query file max-distance))))
+    files))
+
 ;;;;; Updating
 (defun org-roam-db--update-titles ()
   "Update the title of the current buffer into the cache."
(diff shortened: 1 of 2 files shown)


org-roam-graph.el
```diff
@@ -186,18 +186,22 @@ If PREFIX, then the graph is generated but the viewer is not invoked."
 	        (call-process org-roam-graph-viewer nil 0 nil temp-graph)
         (view-file temp-graph)))))
 
-(defun org-roam-graph-show-connected-component (&optional prefix)
+(defun org-roam-graph-show-connected-component (&optional max-distance no-display)
```
kljohann, line 189: I'm not too happy about breaking the existing interface here, so maybe this should be moved to a separate function now. (I just started this experiment by extending the existing "connected-component" function.)
Previously the prefix was used to suppress showing of the graph. But I found that a numeric prefix was a very convenient way to set the maximum distance, so I changed that behavior…

org-roam-db.el
```diff
@@ -262,6 +262,31 @@ If the file does not have any connections, nil is returned."
...
+                   -- Links are traversed in a breadth-first search.  In order to calculate the
+                   -- distance of nodes and to avoid following cyclic links, the visited nodes
+                   -- are tracked in 'trace'.
+                   connected_component(file, trace) AS
+                     (VALUES($s1, json('[]'))
+                      UNION
+                      SELECT lo.link, json_insert(cc.trace, '$[' || json_array_length(cc.trace) || ']', lo.link) FROM
+                      connected_component AS cc JOIN links_of AS lo USING(file)
+                      WHERE (
+                        (SELECT count(*) FROM json_each(cc.trace) WHERE json_each.value == lo.link) == 0
```
kljohann, line 281: Add comment: "Avoid cycles by only visiting each file once."

org-roam-db.el
```diff
@@ -262,6 +262,31 @@ If the file does not have any connections, nil is returned."
...
+including the file itself.  If the file does not have any connections, nil is returned."
+  (let* ((query "WITH RECURSIVE
+                   links_of(file, link) AS
+                     (SELECT \"from\", \"to\" FROM links UNION
+                      SELECT \"to\", \"from\" FROM links),
+                   -- Links are traversed in a breadth-first search.  In order to calculate the
+                   -- distance of nodes and to avoid following cyclic links, the visited nodes
+                   -- are tracked in 'trace'.
+                   connected_component(file, trace) AS
+                     (VALUES($s1, json('[]'))
```
kljohann, line 276: Currently the root file will be visited a second time, if there is a back-link. This could be avoided by including it in the trace here. However this second occurence is filtered out by the `GROUP BY` in any case.

org-roam-db.el
```diff
@@ -262,6 +262,31 @@ If the file does not have any connections, nil is returned."
...
+(defun org-roam-db--links-with-max-distance (file max-distance)
+  "Return all files reachable from/connected to FILE in at most MAX-DISTANCE steps,
+including the file itself.  If the file does not have any connections, nil is returned."
+  (let* ((query "WITH RECURSIVE
+                   links_of(file, link) AS
+                     (SELECT \"from\", \"to\" FROM links UNION
+                      SELECT \"to\", \"from\" FROM links),
+                   -- Links are traversed in a breadth-first search.  In order to calculate the
+                   -- distance of nodes and to avoid following cyclic links, the visited nodes
+                   -- are tracked in 'trace'.
```
kljohann, line 274: This "trace" is not the most beautiful solution, but what I came up with given little time and the constraints of SQLite. :man_shrugging: I think it should not be too bad, as this function should in any case only be called when the distance (and thus the number of entries in this array) has an upper bound.

kljohann:
Or using the `concentrate=true` graph attribute:

graph-1-jumps-kalman-concentrate

graph-2-jumps-kalman-concentrate

org-roam-graph.el
jethrokuan: I think breakage now is still fine, since this functionality was only recently merged, shouldn't affect many.

I do think it's a bit odd for a function called `org-roam-graph-show`, to not actually show anything, the building logic can probably be extracted away from the showing logic.

kljohann: I like the idea of having two functions instead of using the prefix to control whether the graph is shown or not. :+1: The numerical prefix could then be used to indicate the maximum distance for both the interactive "build" function and the interactive "build&show" function.

kljohann: Would it be fine for you if I then also adapted `org-roam-graph-show` to ignore the prefix and instead added a `org-roam-graph-build` interactive function?

jethrokuan: > Would it be fine for you if I then also adapted `org-roam-graph-show` to ignore the prefix and instead added a `org-roam-graph-build` interactive function? (Maybe also in a follow-up PR, with a separate commit on the master branch if you prefer.)

Yes, this is fine.

kljohann: Okay, then I'll create a second PR as soon as this one is merged. :slightly_smiling_face:

kljohann: → #426, which I'll update after you hit merge on the current PR. :+1:

PR #422, opened 2020-04-09
Add customization to set no-delete-other-windows

jsravn:
While navigating links and back-links I'll end up with several buffers
open due to auto-splitting. When I `delete-other-windows` the org-roam
window will also close. With this option set to `t` it will keep the
org-roam window around even when closing other windows.

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -83,6 +83,12 @@ Has an effect if and only if `org-roam-buffer-position' is `top' or `bottom'."
   :type 'hook
   :group 'org-roam)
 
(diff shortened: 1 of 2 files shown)


jsravn:
I find this really convenient. It might even be worth defaulting to `t but I left it to nil to maintain existing behavior.

jethrokuan:
Could you rebase this on top of master, and rename `org-roam-delete-other-windows` to `org-roam-buffer-delete-other-windows`?

jsravn:
@jethrokuan okay, done.

Issue #423, opened 2020-04-10
Accessing and creating org-roam notes via helm-bibtex 


Zaeph:
Hello,

### Brief Abstract
Just as `org-ref`'s `cite:` links are integrated into `org-roam`, partially integrating `helm-bibtex` seems like a good idea.

### Long Description

I'm reading Ahrens (2017), and a contrast is made between *literature* notes and regular *slip-box* notes.  The former are the notes you take as you read, which are then refined and integrated into the slip-box as a series of the latter.

I was pleased to see that `org-roam` already has some partial integration with `org-ref`, and I felt I could write something to do the same with `helm-bibtex`.  On the user-level, it does two things:
1. Add an action to `helm-bibtex` to `Edit slip-box notes`, which creates the slip-box note if it doesn't exist.
2. Add a tag for entries which already have a slip-box note.

Behind the curtain, as you can imagine, it checks for the existence of said slip-box notes before accessing or creating them.

### Proposed Implementation (if any)

More than a *proposed* implementation, I have an *actual* implementation.

I've tried to integrate my functions with what was already written.  For instance, the slip-box note created by `bibtex-completion-edit-sb-notes` (custom-made) reuses the `ref` template as defined in `org-roam-capture-ref-templates`.

Here's a screencast:
![org-roam-bibtex](https://user-images.githubusercontent.com/12202828/78980778-e934ac00-7b1e-11ea-9dab-9305b9ed602e.gif)

I've submitted this as a feature *request* rather than  PR because I'm not yet sure how I want to implement it into `org-roam`, since I have to `(require 'helm-bibtex nil t)`.  If you have any suggestion on that front, I'm all ears.

Small caveat: I've only written the commands for the `helm` portion of `helm-bibtex`, which also implements an `ivy` equivalent.  It should be trivial to add, but I haven't investigated it yet.

HTH.

### Please check the following:
- [x] No similar feature requests

### References
- Ahrens, Sönke.  *How to Take Smart Notes: One Simple Technique to Boost Writing, Learning and Thinking for Students, Academics and Nonfiction Book Writers*, 2017. ([WorldCat)](https://www.worldcat.org/title/how-to-take-smart-notes-one-simple-technique-to-boost-writing-learning-and-thinking-for-students-academics-and-nonfiction-book-writers/oclc/1003209782&referer=brief_results)

jethrokuan:
[...]

PR #427, opened 2020-04-10
(feat) cond display relative subdir in title for selection

Zaeph:
A title can now be constructed as `subdir/file-title` where `subdir` is the sub-directory relative from `org-roam-directory`.

When `org-roam-title-include-subdirs` is non-nil, `org-roam--extract-titles` now outputs titles in this format.

Hello,

This PR is a result of the discussions in #423.  In it, we discussed the possibility of having files with the same `#:TITLE:` (or alias) across different sub-directories.  The problem is that, currently, having both `lit/ahrens2017.org` with `#:TITLE: ahrens2017` and `refs/ahrens2017.org` with `#:TITLE: ahrens2017` would cause problems with `org-roam-find-file`, since the parser would generate a list with `ahrens2017` appearing two times.

I've created a switch to enable this option, which is disabled by default.

Please let me know if there's anything I've missed.

HTH.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -324,7 +324,7 @@ If the file does not have any connections, nil is returned."
                   (cons (vector file contents-hash time) all-files))
             (when-let (links (org-roam--extract-links file))
               (setq all-links (append links all-links)))
-            (let ((titles (org-roam--extract-titles)))
+            (let ((titles (org-roam--extract-titles file)))
               (setq all-titles (cons (vector file titles) all-titles)))
             (when-let ((ref (org-roam--extract-ref)))
               (setq all-refs (cons (vector ref file) all-refs))))
--- a/org-roam.el
+++ b/org-roam.el
@@ -246,7 +246,27 @@ it as FILE-PATH."
                       link-type
                       (list :content content :point begin)))))))))
 
-(defun org-roam--extract-titles ()
+(defcustom org-roam-title-include-subdirs nil
+  "Include sub-directories relative to `org-roam-directory' in titles."
+  :type 'boolean
+  :group 'org-roam)
+
+(defun org-roam--format-title (title file-path)
+  "Format TITLE with relative sub-directory from `org-roam-directory'."
+  (if org-roam-title-include-subdirs
+      (let* ((root (expand-file-name org-roam-directory))
+             (path (or file-path
+                       (-> (or (buffer-base-buffer)
+                               (current-buffer))
+                           (buffer-file-name)
+                           (file-truename))))
+             (dir (file-name-directory path)))
+        (concat (unless (equal root dir)
+                  (file-relative-name dir root))
+                title))
+    title))
+
+(defun org-roam--extract-titles (&optional file-path)
   "Extract the titles from current buffer.
 Titles are obtained via the #+TITLE property, or aliases
 specified via the #+ROAM_ALIAS property."
@@ -254,9 +274,11 @@ specified via the #+ROAM_ALIAS property."
          (aliases (cdr (assoc "ROAM_ALIAS" props)))
          (title (cdr (assoc "TITLE" props)))
          (alias-list (org-roam--aliases-str-to-list aliases)))
-    (if title
-        (cons title alias-list)
-      alias-list)))
+    (mapcar (lambda (title)
+              (org-roam--format-title title file-path))
+            (if title
+                (cons title alias-list)
+              alias-list))))
 
 (defun org-roam--extract-ref ()
   "Extract the ref from current buffer."


Zaeph:
I've been tracking a nasty bug for the last hour.  Aside from the `title` and the `path`, we're also caching the `ref`, which means that if we have `#+ROAM_KEY: cite:ahrens2017`, the reference will be cached.

For #423, I've created the function `helm-bibtex-edit-sb-notes` which opens a slip-box note if it exists, and create it if it doesn't.  The `org-roam` `ref`-template that I use is configured to place the notes thusly created in `refs/`, but the problem is that I already had a file in `lit/` with `#+ROAM_KEY: cite:ahrens2017`.  There must be a low-level function checking for the existence of this `#+ROAM_KEY:` because I can't for the life of me get it to create `refs/ahrens2017` whilst `#+ROAM_KEY:` is in `lit/ahrens2017`.  As soon as I remove it, the file can be created.

Also, I forgot to mention that such a feature would also require us to modify the parsing done by `org-roam-find-file`.  It'd be nice to have it place files in sub-directories directly.  For instance, typing in `foo/Bar` in the `org-roam-find-file` prompt would:
1. Ensure that `foo/` exists, and create it if it doesn't
2. Create the note `bar` inside `foo/`
3. Fill the title of the note as `Bar` rather than `foo/Bar`

Zaeph:
> I've been tracking a nasty bug for the last hour. Aside from the `title` and the `path`, we're also caching the `ref`, which means that if we have `#+ROAM_KEY: cite:ahrens2017`, the reference will be cached.

I think the problem could easily be solved by catching `ref` duplicates and warning the user about it.  Within the frame of this PR, I originally thought that it could be a good idea to prioritise the `#+ROAM_KEY:` present `refs/`, but I think it think it'd be better to prevent the behaviour entirely.

org-roam.el
```diff
@@ -246,17 +246,39 @@ it as FILE-PATH."
                       link-type
                       (list :content content :point begin)))))))))
 
-(defun org-roam--extract-titles ()
+(defcustom org-roam-title-include-subdirs nil
+  "Include sub-directories relative to `org-roam-directory' in titles."
```
jethrokuan, line 250:
```suggestion
  "If t, include sub-directories relative to `org-roam-directory' in title completions."
```

org-roam.el
```diff
@@ -246,17 +246,39 @@ it as FILE-PATH."
...
+  :type 'boolean
+  :group 'org-roam)
+
+(defun org-roam--format-title (title file-path)
+  "Format TITLE with relative sub-directory from `org-roam-directory'."
```
jethrokuan, line 255: Please ensure `FILE-PATH` appears in the docstring.

jethrokuan:
> There must be a low-level function checking for the existence of this `#+ROAM_KEY:` because I can't for the life of me get it to create `refs/ahrens2017` whilst `#+ROAM_KEY:` is in `lit/ahrens2017`. As soon as I remove it, the file can be created.

The low level thing is the database itself, there's a unique constraint on the ref.

> Also, I forgot to mention that such a feature would also require us to modify the parsing done by org-roam-find-file. It'd be nice to have it place files in sub-directories directly.

Not sure if I'm supportive of this since the `find-file` command shouldn't be the one responsible for placing files in correct directories, but the templates. Even this change feels slightly weird, but there does need to be a way to differentiate the files from different directories having same titles.

Re refs, the key is by design supposed to be unique.

Zaeph:
> The low level thing is the database itself, there's a unique constraint on the ref.

Could we run a check that produces a warning for the user?  I don't think we're doing that quite yet.

> Not sure if I'm supportive of this since the `find-file` command shouldn't be the one responsible for placing files in correct directories, but the templates. Even this change feels slightly weird, but there does need to be a way to differentiate the files from different directories having same titles.

I've slept on it, and yes, I agree with you: templates should be used to notes in sub-directories.  We'd need to make this explicit in the doc.

Also, I'm not wondering whether using `/` as the separator was a good idea.  The other one I had in mind was `:`.  Maybe turn the separator into a variable?

Zaeph:
> Also, I'm not wondering whether using `/` as the separator was a good idea. The other one I had in mind was `:`. Maybe turn the separator into a variable?

I've implemented it in c23880b.

Zaeph:
> Not sure if I'm supportive of this since the `find-file` command shouldn't be the one responsible for placing files in correct directories, but the templates. Even this change feels slightly weird, but there does need to be a way to differentiate the files from different directories having same titles.

I've racked my head on how we could articulate `org-roam-find-file` and `org-roam-templates`, and no option is fully satisfying.  The one that comes the closest is having `org-roam-find-file` check if there is a template whose `file-name:` contains the sub-directory inputted, and only ask the user which template to use in case of ambiguity (say, two templates with the sub-directory in `file-name:`.  But it's a lot of hassle to avoid having to choose a template.  I also don't like the fact that we'd be mimicking a file-hierarchy instead of using *actually* using the browsing possibilities provided by `ivy` or other completion engines.

org-roam.el
```diff
@@ -253,17 +253,50 @@ it as FILE-PATH."
...
+  "When non-nil, include sub-directories relative to
+`org-roam-directory' in title completions."
+  :type 'boolean
+  :group 'org-roam)
+
+(defcustom org-roam-title-subdir-separator "/"
+  "String to use to separate subdirs when
+`org-roam-title-include-subdirs' is non-nil."
+  :type 'string
+  :group 'org-roam)
```
Wetlize, line 266: Can we use some sort of format-string instead, like we already do so with `org-roam-link-title-format`? So given `org-roam-directory/biblio/title.org` and `"(%s) %s"` (or `"(=subdir=) =title="`) set as a format-string it would yield to `(biblio) Title`.

It can be more complicated, but would give users more flexibility (e.g. placing subdirectory at the end of the string).

zaeph: That's a good idea.  I'll see what I can do.

Wetlize: @zaeph maybe instead of the formatting string it would be simpler to preset this to a function that would take subdirectory and title as a 2 separate arguments. The return value would yield the final title. This approach is even more flexible (user can decide whether format something and how) and might be more simpler.

zaeph: Sounds good.  I like the fact that we're allowing most of our switches to accept functions for extra customisation.

zaeph: Let me finish wrapping my head around `org-roam-backlink` and I'll have a look.

zaeph: Implemented in cb1b2c01cbfbc20308da00e5316aa01c3bb8dfc6.

org-roam.el
```diff
@@ -292,6 +355,15 @@ Titles are obtained via:
         (cons title alias-list)
       alias-list)))
 
+(defun org-roam--extract-and-format-titles (&optional file-path)
+  "Extract the titles from the current buffer and format them.
+If FILE-PATH is not provided, the file associated with the
+current buffer is used."
+  (let ((titles (org-roam--extract-titles)))
```
jethrokuan, line 362: I think we can use dash to clean this up.

```
(-> (org-roam--extract-titles) (org-roam--format-title file-path))
```

Apologies for the poor formatting, typing elisp on github.

org-roam-db.el
```diff
@@ -304,11 +304,13 @@ including the file itself.  If the file does not have any connections, nil is re
 ;;;;; Updating
 (defun org-roam-db--update-titles ()
   "Update the title of the current buffer into the cache."
-  (let ((file (file-truename (buffer-file-name))))
+  (let ((file (file-truename (buffer-file-name)))
+        title)
```
jethrokuan, line 308: can use a `let*` bind here for title.

org-roam.el
```diff
@@ -359,10 +359,9 @@ Titles are obtained via:
   "Extract the titles from the current buffer and format them.
 If FILE-PATH is not provided, the file associated with the
 current buffer is used."
-  (let ((titles (org-roam--extract-titles)))
-    (mapcar (lambda (title)
-              (org-roam--format-title title file-path))
-            titles)))
+  (->> (org-roam--extract-titles)
```
jethrokuan, line 362: oh right nvm, forgot the threading macro doesn't do a map.

```suggestion
  (mapcar (lambda (title) (org-roam--format-title title file-path)) (org-roam--extract-titles))
```

PR #439, opened 2020-04-11
Add edges in graph relating to org-ref cite links

TimQuelch:
Now that citations appear as backlinks in the backlinks buffer, it makes sense for them to also show
up in the graph.

I'm quite a novice as SQL, so this query probably could have been merged into the previous, but this
works for now.

Citation links appear in the backlinks buffer, so they should probably appear in the graph view

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * [#380][gh-380] Allow `org-roam-buffer-position` to also be `top` or `bottom`
 * [#385][gh-385] Add `org-roam-graph-node-extra-config` to configure Graphviz nodes
 * [#435][gh-435] Add `org-roam-graph-edge-extra-config` to configure Graphviz edges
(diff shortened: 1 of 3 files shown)


jethrokuan:
I agree! Maybe cite links should also get a different color to distinguish them from normal links.

TimQuelch:
> I agree! Maybe cite links should also get a different color to distinguish them from normal links.

Agreed. I've added in a configuration variable to control the style. I've set it to default to red for citation links. Could be worth just having a default of black though?

TimQuelch:
I've extended the queries in `org-roam-db--conected-component` and `org-roam-db--links-with-max-distance` to ensure that they count citation links as connections. 

I'm quite new to SQL, so I'm not sure how efficient/correct these queries are, but they are working on my graphs. Review is welcome on https://github.com/jethrokuan/org-roam/pull/439/commits/be1329a3ad6d7b806cd78cec31e29f68db98ab61

PR #443, opened 2020-04-12
(fix) use dash.el’s -if-let forms rather than subr-x.el’s

Zaeph:


jethrokuan:
Closing in favour of #429

jethrokuan:
Closing in favour of #429

PR #444, opened 2020-04-12
(doc) fix broken format+typo in win-install (#438)

nobiot:
Thank you for accepting my earlier PR: #438. 
The format didn't turn out to be as intended. Here is my attempt to fix that, along with minor grammar and typo issues.

I don't have access MKDocs, so I was not able to ensure the format in the official documentation would be correct. This time, I was more careful and tried previewing the change in GitHub itself -- hopefully this is good enough to check the format in the final MKDocs format... I am noting GFM might be different from Python-Markdown that MKDocs uses.


jethrokuan:
Thanks!

PR #452, opened 2020-04-13
(bugfix) Respect exclude matcher in connected components graph

kljohann:

I think for consistency the excluded files filter should also be applied to the "connected components" graph. What do you think?

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -250,7 +250,9 @@ If MAX-DISTANCE is non-nil, only nodes within the given number of steps are show
                     (list file)))
          (query `[:select [file titles]
                   :from titles
-                  :where (in file [,@files])]))
+                  :where (in file [,@files])
+                  ,@(org-roam-graph--expand-matcher
+                     'file 'negate 'and)]))
     (org-roam-graph-build query)))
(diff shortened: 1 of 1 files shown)


org-roam-graph.el
```diff
@@ -250,7 +250,9 @@ If MAX-DISTANCE is non-nil, only nodes within the given number of steps are show
                     (list file)))
```
kljohann, line 250: Note that the filter is only taken into account /after/ computing the connected component. In the worst case this might lead to unexpected behavior by including too many nodes, if two clusters are only connected via a single node. E.g.
```
cluster A -> node B -> cluster C
```
Would lead to nodes from both clusters being displayed, while only excluding node B.

jethrokuan:
There's a different problem with this exclude-matcher function, that I haven't been able to figure out how to fix. The macro i sexpanded, and the function is defined once, with the exclude sql clause once, which means that changing the value at run time via `setq` won't change the filters.

Maybe the way it's implemented can be changed altogether

kljohann:
Ah, I see what you mean. Probably makes sense to look at this at the same time. :+1:

PR #456, opened 2020-04-14
(bugfix): fix org-roam-graph--open

### LINKED ISSUE #455: `temp-graph` should be `file`
https://github.com/jethrokuan/org-roam/blob/b33300a23de2b7b8478b21dce15655c86cde1463/org-roam-graph.el#L230

I get a `org-roam-graph-show: Symbol’s value as variable is void: temp-graph` error that I can fix by changing `temp-graph` to `file`

jethrokuan:
Fixes #455


kljohann:
Oh, seems like I forgot to stage that change. Thanks for the fix!

Issue #222, opened 2020-03-03
Discovering / repairing broken references after moving files


chrischambers:
### Brief Abstract
When moving a file into a different subdirectory under `org-roam-directory` (via, say, a file manager like OSX Finder or `git mv`) or when renaming a file using the terminal, perhaps, it would be excellent if org-roam could detect the changes and prompt me to update the broken references.

### Proposed Implementation 
This would require a slightly more extensive change than https://github.com/jethrokuan/org-roam/pull/124 - you'd need some kind of heuristic. I see you're already storing the file hash under roam-db:files - perhaps a job which traverses `org-roam-directory` for matching hashes but different filenames would suffice?

### Please check the following:
- [x] No similar feature requests

chrischambers:
Awesome project, by the way. :)

asymmetric:
I personally don't think org-roam should become a kitchen sink of all possible useful features.

Couldn't this functionality be implemented more generically for any org-mode file?

chrischambers:
Here's a usecase to explain why this isn't just a "kitchen sink" feature:

I start off creating some programming notes - I make a note on the solid principles, another on tdd, another on bdd, one as a reference for elisp, another as a reference for scheme, and so on. In the meantime I'm creating backlinks between them - all is good. 

But now I have enough programming notes to warrant creating a new subdirectory so they aren't mixed up with my other notes - so I move them via the terminal or vifm or some such. Now all the backlinks are broken. I have to do a lot of careful grepping to find out which file links are broken, and then maybe a brittle `sed -i` command to update them. Assuming I remember to do that.

Time goes on and I want to create new subdirectories - one on processes (where solid_principles.org might go), another on testing (where the tdd/bdd notes go), another for languages (scheme, elisp, etc.). If every time I do this all my backlinks get broken, the system won't be sustainable. I think it's a pretty common usecase.

As to whether this could be a generic org-mode feature - I'm not sure about that. 
It only works presuming there's a finite set of locations/location (e.g. `org-roam-directory`) under which all the interlinked files sit. Some people choose not to collect all their org notes under `org-directory`. It also seems easier to implement for org-roam than org in general because of the sqlite db storing file hashes.

asymmetric:
Thanks for the usecase :)

I’ll only say that to me, this idea of having directories representing categories seems antithetical to the Roam approach of having all notes live in a flat hierarchy, and have the relationships between the notes kind of emerge and still be fluid.

[...]

PR #464, opened 2020-04-15
Add prefill-current argument to org-roam-insert

zaeph:
This PR modifies `org-roam-insert` to accept a new optional, boolean argument, `prefill-current`.  When it is non-nil, `org-roam-insert` prefills the completion template with the title of the current entry.  It's useful for quickly creating notes from a topic card.

Here's a screencast:
output-2020-04-15-08:39:39

Note that this is something that can be easily expanded.  For instance, I'd like to create a command to create a new version of the current note.  For example, if I'm on the note 'Metaphor', and I don't agree with what I've written anymore, I could run a command and have it automatically create a `Metaphor v2` note.  The command would also check the current version-number and increment it for further versions.

How do you like this?

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -308,9 +308,17 @@ specified via the #+ROAM_ALIAS property."
      (concat "file:" (file-relative-name target here))
      description)))
 
-(defun org-roam-insert (prefix)
+(defcustom org-roam-title-separator ": "
+  "Separator to use when breaking off titles.
+See `org-roam-insert' for details."
+  :type 'string
+  :group 'org-roam)
+
+(defun org-roam-insert (prefix &optional prefill-current)
   "Find an Org-roam file, and insert a relative org link to it at point.
-If PREFIX, downcase the title before insertion."
+If PREFIX, downcase the title before insertion.
+If PREFILL-CURRENT is non-nil, insert the title of the current note in the
+completion form."
   (interactive "P")
   (unless (org-roam--org-roam-file-p
            (buffer-file-name (buffer-base-buffer)))
@@ -328,7 +336,7 @@ If PREFIX, downcase the title before insertion."
                  "File: " completions
                  :initial-input (if prefill-current
                                     (concat title-current
-                                            ": "
+                                            org-roam-title-separator
                                             region-text)
                                   region-text)))
          (region-or-title (or region-text title))
--- a/org-roam.el
+++ b/org-roam.el
@@ -314,11 +314,21 @@ See `org-roam-insert' for details."
   :type 'string
   :group 'org-roam)
 
-(defun org-roam-insert (prefix &optional prefill-current)
+(defun org-roam--format-title-as-topic (title &optional string)
+  "Format TITLE as a topic.
+Concatenate TITLE and `org-roam-title-separator', and return the
+resulting string.
+If STRING is non-nil, also concatenates it before returning."
+  (concat title
+          org-roam-title-separator
+          string))
+
+(defun org-roam-insert (prefix &optional prefill-fn)
   "Find an Org-roam file, and insert a relative org link to it at point.
 If PREFIX, downcase the title before insertion.
-If PREFILL-CURRENT is non-nil, insert the title of the current note in the
-completion form."
+PREFILL-FN is the name of function to prefill the completion
+form.  The function should accept two arguments: the title of the
+current buffer, and the content of the region if it is active."
   (interactive "P")
   (unless (org-roam--org-roam-file-p
            (buffer-file-name (buffer-base-buffer)))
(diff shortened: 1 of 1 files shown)


Zaeph:
> Note that this something that can be easily expanded. For instance, I'd like to create a command to create a new version of the current note. For example, if I'm on the note 'Metaphor', and I don't agree with what I've written anymore, I could run a command and have it automatically create a `Metaphor v2` note. The command would also check the current version-number and increment it for further versions.

I think this would be best done by changing `PREFILL-CURRENT` to `PREFILL-FN`, which would be the name of a function who’d take one argument: the title of the current entry.  That way, we could do plenty of neat stuff.

Zaeph:
> I think this would be best done by changing `PREFILL-CURRENT` to `PREFILL-FN`, which would be the name of a function who’d take one argument: the title of the current entry. That way, we could do plenty of neat stuff.

I went ahead and did it in 7997e3d.  Instead of taking one argument, though, the functions take two: the title of the current note, and the content of the region if it is active.

asymmetric:
When is this useful in practice? i.e. what could come after `Metaphor: `?

Zaeph:
> When is this useful in practice? i.e. what could come after `Metaphor: `?

You could have:
- `Metaphor: Comparison with simile`
- `Metaphor: Cognitive dimension`
- `Metaphor: Uses in rhetorics`

Those aren't *atomic* notes, but notes which aim to relate `Metaphor` to other notes/topic whilst keeping `Metaphor` as a soft-parent.

asymmetric:
As I mentioned in #222 (not that my opinion matters that much tho! :smile: ), I think notions of parenting should not be part of org-roam, due to the inherent fragility of the concept of a "hierarchy of ideas".

In your examples, I guess I would simply keep the notes in one file, but you could obviously also split them up and link them back to a `Metaphor` note.

asymmetric:
I wonder if [this](https://github.com/jethrokuan/org-roam/issues/317) issue (not implemented yet) could also work in your direction.

Zaeph:
> As I mentioned in #222 (not that my opinion matters that much tho! smile ), I think notions of parenting should not be part of org-roam, due to the inherent fragility of the concept of a "hierarchy of ideas".

I agree with you, and this is echoed in Ahrens (2017):

> The quicker we get from the index to the concrete notes, the quicker we move our attention from mentally preconceived ideas towards the fact-rich level of interconnected content, where we can conduct a fact-based dialogue with the slip-box.

So, an index, and that’s it.  However, a point is also made about the importance of *rewriting* notes:

> What we regard as being relevant for a topic and how we structure it **will change over time**. This change might lead to another note with a different, more adequate topic structure, which then can be seen as a comment on the previous note. Thankfully, it won’t make all the other notes redundant. As mentioned before: **All we have to do is to change the entry in the index to this new note and/or indicate on the old note that we now consider a new structure more fitting**. [emphasis added]

So, in that regard, the PR would provides the facility to support those types of notes which have a *meta*-link to the previous entry.

To come back to the idea of prepending the title of new entries with something like `Metaphor: `, I’ve mentioned the idea of *soft-parent*.  When you think about the way Luhmann incorporated the method, he only branched off onto another slash (i.e. from `12/a` to `12/a/1`), it was because he wanted to continue in a similar direction, but there was already a note after it.  That’s why I say that the note thus created is a *soft-child* of the previous one because its existence was warranted *as a result* of its *soft-parent*.  Since we’re not bound to paper limitations with `org-roam`, you can also think of creation-times and modification-times as *softly* linking notes.  They shouldn’t take precedence on the manual associations we made between notes, however far-reaching they may be, but I think there is value in having them.

> In your examples, I guess I would simply keep the notes in one file, but you could obviously also split them up and link them back to a `Metaphor` note.

See, I’d tried that in my first notes, and I ended up with gargantuan notes which weren’t all that easy to browse.  But I think you’ve brought up a valuable issue, and I’d love to hear the opinions of others on this.

> I wonder if [this](https://github.com/jethrokuan/org-roam/issues/317) issue (not implemented yet) could also work in your direction.

Thanks for pointing me to it.  Indeed, it seems related, but it also looks like it hasn’t been robustly implemented yet.  I’ll have a deeper look when I get the chance.

### References
<div class="csl-bib-body" style="line-height: 1.35; margin-left: 2em; text-indent:-2em;">
  <div class="csl-entry">Ahrens, Sönke. <i>How to Take Smart Notes: One Simple Technique to Boost Writing, Learning and Thinking for Students, Academics and Nonfiction Book Writers</i>, 2017.</div>
  <span class="Z3988" title="url_ver=Z39.88-2004&amp;ctx_ver=Z39.88-2004&amp;rfr_id=info%3Asid%2Fzotero.org%3A2&amp;rft_id=urn%3Aisbn%3A978-1-5428-6650-7&amp;rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Abook&amp;rft.genre=book&amp;rft.btitle=How%20to%20Take%20Smart%20Notes%3A%20One%20Simple%20Technique%20to%20Boost%20Writing%2C%20Learning%20and%20Thinking%20for%20Students%2C%20Academics%20and%20Nonfiction%20Book%20Writers&amp;rft.aufirst=S%C3%B6nke&amp;rft.aulast=Ahrens&amp;rft.au=S%C3%B6nke%20Ahrens&amp;rft.date=2017&amp;rft.isbn=978-1-5428-6650-7&amp;rft.language=English"></span>
</div>

PR #476, opened 2020-04-16
(fix) ensure that directory exists before capturing to dir

Zaeph:
Trying to create a note with `org-roam-find-file` inside a sub-directory that does not exist results in a silent failure.

This PR fixes it by ensuring the existence of the sub-directory (and its parents).


Zaeph:
> Trying to create a note with `org-roam-find-file` inside a sub-directory that does not exist results in a *silent failure*. [emphasis added]

Actually, it's not silent at all: it just complains that the file/directory doesn't exist.

PR #478, opened 2020-04-17
(docs) Provide examples of using template expansion and function overrides.

grzm:

Those excited about org-roam but new to Emacs and org-mode may not be familiar with the flexibility they can take advantage of. Providing some examples can help with this.

### DIFF
--- a/doc/templating.md
+++ b/doc/templating.md
@@ -50,7 +50,7 @@ org-capture expands `%<%Y%m%d%H%M%S>` with timestamp: e.g.
 All of the flexibility afforded by emacs and org-mode are
 available. For example, if you want to encode a UTC timestamp in the
 filename, you can take advantage of org-mode's `%(EXP)` template
-expansion to call `format-time-string` directy to provide its third
+expansion to call `format-time-string` directly to provide its third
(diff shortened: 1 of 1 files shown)


doc/templating.md
```diff
@@ -47,6 +47,43 @@ For example, take the template: `"%<%Y%m%d%H%M%S>-${title}"`, with the title
 org-capture expands `%<%Y%m%d%H%M%S>` with timestamp: e.g.
 `20200213032037-Foo`.
 
+All of the flexibility afforded by emacs and org-mode are
+available. For example, if you want to encode a UTC timestamp in the
+filename, you can take advantage of org-mode's `%(EXP)` template
+expansion to call `format-time-string` directy to provide its third
```
jethrokuan, line 53:
```suggestion
expansion to call `format-time-string` directly to provide its third
```

Zaeph:
> Those excited about org-roam but new to Emacs and org-mode may not be familiar with the flexibility they can take advantage of. Providing some examples can help with this.

Thanks for taking the time to write beginners-oriented documentation.  I really like the two examples you've written.

PR #481, opened 2020-04-17
Optional `FILTER` argument for `org-roam-find-file` and `org-roam-insert`

### LINKED ISSUE #474: Don't propose "bibliography ressources" when using org-roam-insert
Let me first describe my workflow using `org-roam` which is I guess, quite usual:
- each paper/book I read has its own org file with `#+ROAM_KEY: cite:bibtex_entry` at the top

Then: 
- When I want to reference something in my bibliography from the note file, I generally use `helm-bibtex` that will insert `cite:bibtex_entry` to the file and everything work (back-links, ...)
- When I want to link to another "topic/tag" (and not a bibliography element), I use `org-roam-insert`.
  In such case, all the papers/books appear in the helm buffer and it is not very convenient to find the "topic/ag" I want to link.

Is there a way to do **not** show the files with a Roam key that look like `cite:something`?

That will be very useful.

Thanks a lot

Zaeph:
Motivation is described in #474.

Implements:
- Optional `FILTER` argument for `org-roam-find-file` and `org-roam-insert`
- `org-roam--get-non-ref-path-completions` to only get non-ref notes
- `org-roam-find-non-ref-file` & `org-roam-insert-non-ref` as convenience commands

Closes #474.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -357,6 +357,13 @@ If PREFIX, downcase the title before insertion."
 							       :capture-fn 'org-roam-insert))
 	(org-roam--capture)))))
 
+(defun org-roam-insert-non-ref (prefix)
+  "Find an Org-roam, non-ref file, and insert a relative org link to it at point.
+See `org-roam-insert' and
+`org-roam--get-non-ref-path-completions' for details."
+  (interactive "P")
+  (org-roam-insert prefix #'org-roam--get-non-ref-path-completions))
+
 ;;;; org-roam-find-file
 (defun org-roam--get-title-path-completions ()
   "Return a list of cons pairs for titles to absolute path of Org-roam files."
@@ -410,6 +417,13 @@ CANDIDATES is a an alist of candidates to consider.  Defaults to
             (path (cdr candidate)))
         (unless (member path refs-path)
           (push (cons title path) completions))))))
+
+(defun org-roam-find-non-ref-file (&optional initial-prompt)
+  "Find and open an Org-roam, non-ref file.
+See `org-roam-find-files' and
+`org-roam--get-non-ref-path-completions' for details."
+  (interactive)
+  (org-roam-find initial prompt #'org-roam--get-non-ref-path-completions))
 
 ;;;; org-roam-find-directory
 (defun org-roam-find-directory ()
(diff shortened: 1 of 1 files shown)


Zaeph:
This PR follows the same direction as #464 in adding arguments to `org-roam-find-file` and `org-roam-insert`.

jethrokuan:
While I can see adding the optional filter argument to the function, I don't think we should ship these custom interactives within org-roam. The use-cases are niche, and should go into a user's config, what org-roam should do is to allow easy creation of these custom interactives.

Zaeph:
> While I can see adding the optional filter argument to the function, I don't think we should ship these custom interactives within org-roam. The use-cases are niche, and should go into a user's config, what org-roam should do is to allow easy creation of these custom interactives.

I could ship those functions (renamed) in `org-roam-bibtex`.  Should I remove them from the PR?

jethrokuan:
> Should I remove them from the PR?

Yes, I'm okay with adding a filter function as argument.

Zaeph:
> Yes, I'm okay with adding a filter function as argument.

The PR is ready for that, then.

org-roam.el
```diff
@@ -316,9 +316,12 @@ specified via the #+ROAM_ALIAS property."
      (concat "file:" (file-relative-name target here))
      description)))
 
-(defun org-roam-insert (prefix)
+(defun org-roam-insert (prefix &optional filter)
```
jethrokuan, line 319:
```suggestion
(defun org-roam-insert (prefix &optional filter-fn)
```

org-roam.el
```diff
@@ -316,9 +316,12 @@ specified via the #+ROAM_ALIAS property."
...
   "Find an Org-roam file, and insert a relative org link to it at point.
-If PREFIX, downcase the title before insertion."
+If PREFIX, downcase the title before insertion.
+FILTER is the name of a function to apply on the candidates which
```
jethrokuan, line 322:
```suggestion
FILTER-FN takes as argument an alist of path-completions, and returns a filtered alist.
```

org-roam.el
Zaeph: Done in 3280e1c.

Zaeph:
Thanks for the merge! 🎉

PR #482, opened 2020-04-17
(feat) implement after-find-file-hook for org-roam-capture

Zaeph:
Allows user to run functions when an `org-roam-capture` is successful.  This might come in handy for moving point after finding the new file, since right now, it is at `(point-min)`.

For `org-roam-bibtex`, someone asked me if we could directly open `org-noter` on the main heading after finishing the capture process.  I took inspiration from this:
https://github.com/jethrokuan/org-roam/blob/6175739b33b0abaa862c0137e216e6029f24d5cd/org-roam.el#L387

And this:
https://github.com/jethrokuan/org-roam/blob/6175739b33b0abaa862c0137e216e6029f24d5cd/org-roam-capture.el#L152-L158

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -296,8 +296,9 @@ This function is used solely in Org-roam's capture templates: see
     (append (nreverse converted) `(:org-roam ,org-roam-plist))))
 
 (defun org-roam-capture--find-file-h ()
-  "Run the hooks defined in `org-roam-capture-after-finalize-hook'.
-This is added as a hook to `org-capture-finalize-hook'."
+  "Opens the newly created template file.
+This is added as a hook to `org-capture-after-finalize-hook'.
+Run the hooks defined in `org-roam-capture-after-find-file-hook'."
   (unless org-note-abort
     (when-let ((file-path (org-roam-capture--get :file-path)))
       (find-file file-path))
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -295,14 +295,14 @@ This function is used solely in Org-roam's capture templates: see
         (push val converted)))
     (append (nreverse converted) `(:org-roam ,org-roam-plist))))
 
-(defun org-roam-capture--run-after-find-file-hook ()
+(defun org-roam-capture--find-file-h ()
   "Run the hooks defined in `org-roam-capture-after-finalize-hook'.
 This is added as a hook to `org-capture-finalize-hook'."
   (unless org-note-abort
     (when-let ((file-path (org-roam-capture--get :file-path)))
       (find-file file-path))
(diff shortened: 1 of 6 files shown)


org-roam-capture.el
```diff
@@ -303,6 +301,20 @@ This function is used solely in Org-roam's capture templates: see
         (push val converted)))
     (append (nreverse converted) `(:org-roam ,org-roam-plist))))
 
+(defun org-roam-capture--run-after-find-file-hook ()
+  "Run the hooks defined in `org-roam-capture-after-finalize-hook'.
+This is added as a hook to `org-capture-finalize-hook'."
+  (unless org-note-abort
+    (org-roam-capture--find-file-h)
+    (run-hooks 'org-roam-capture-after-find-file-hook))
+  (remove-hook 'org-capture-after-finalize-hook #'org-roam-capture--run-after-find-file-hook))
```
jethrokuan, line 310: No need to introduce a new hook here. Substitute line 308 with the `when-let` clause in the previous hook.

Zaeph: Let me wrap my head around your suggestion, and I'll make the modification.

Zaeph: Did you mean something like in 22d96fe?
https://github.com/jethrokuan/org-roam/blob/22d96fe338bf6607b80d875eed0db48b00d935ac/org-roam-capture.el#L298-L305

Zaeph: I've removed `org-roam-capture--run-after-find-file-hook` from the code, but there was another mention of it here (L34):
https://github.com/jethrokuan/org-roam/blob/ba80a31fe0ddb48d41c446560c35c5677214a6ce/org-roam-dailies.el#L49-L55

Not sure if this has an impact, since I don't use that functionality.

jethrokuan:
@Zaeph could you keep the original hook name?

Zaeph:
> @Zaeph could you keep the original hook name?

Done in ffb1599778119172f80b8407b83d7aa8b044570b.

org-roam-capture.el
```diff
@@ -303,6 +295,21 @@ This function is used solely in Org-roam's capture templates: see
         (push val converted)))
     (append (nreverse converted) `(:org-roam ,org-roam-plist))))
 
+(defun org-roam-capture--find-file-h ()
+  "Run the hooks defined in `org-roam-capture-after-finalize-hook'.
```
jethrokuan, line 299: the docstring is incorrect now.

Zaeph: On it.

Zaeph: Fixed in a6ac2b58961efbfdee7479f16bbbd1201ca4f9a8.

PR #486, opened 2020-04-17
(feat) remove user-error for org-roam-insert

### LINKED ISSUE #477: Allow org-roam-insert from anywhere
### Brief Abstract

`org-roam-insert` currently only works inside org-roam files, but it would be useful to be able to create links to notes in other org files. 

### Long Description

I would like to be able to easily refer to notes from other org files that are not in org roam (for example in project todo items or outlines). 

As far as I can see this makes sense as an operation, but `org-roam-insert` gives an error if you are not in an org file. 

Remove the check and error call: https://github.com/jethrokuan/org-roam/blob/master/org-roam.el#L325

The error you get was mentioned in one issue #377 but it turned out to be a problem with the installation, not a request for changed behaviour.

Zaeph:
Closes #477.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -320,9 +320,6 @@ specified via the #+ROAM_ALIAS property."
   "Find an Org-roam file, and insert a relative org link to it at point.
 If PREFIX, downcase the title before insertion."
   (interactive "P")
-  (unless (org-roam--org-roam-file-p
-           (buffer-file-name (buffer-base-buffer)))
-    (user-error "Not in an Org-roam file"))
   (let* ((region (and (region-active-p)
                       ;; following may lose active region, so save it
                       (cons (region-beginning) (region-end))))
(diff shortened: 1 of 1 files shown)


Zaeph:
> Might have to check if this works, in say, a scratch buffer, I think it might error out.
> 
> i.e. a buffer with no base file.

You're right: when a buffer has no base-file, there's trouble.  I'm investigating.

ricklupton:
Following my comment in #477, I guess using absolute paths would be a solution to that, if the problem is that the buffer might not have a path?

Either in this situation only, or whenever org-roam-insert is called from a non org-roam file. The latter would be more consistent.

Zaeph:
> Following my comment in #477, I guess using absolute paths would be a solution to that, if the problem is that the buffer might not have a path?

It's part of the problem, yes.
https://github.com/jethrokuan/org-roam/blob/a617da891179d8d36575e8c3df5c005a12de1653/org-roam.el#L310-L314

jethrokuan:
`org-roam` actually works regardless of whether the links are absolute or relative, it's a trade-off to make with regards to portability (e.g. syncing via dropbox over different machines), so it's possible to also make this a choice globally.

Zaeph:
Should we just run a check in that function, and use absolute-path if there isn't any file associated with the current buffer?

Issue #440, opened 2020-04-11
Allow providing a function for `org-roam-graph-viewer`


grzm:
### Brief Abstract

On macOS, the common way of opening a file in Safari from the command line is `open -a Safari $file`. `Currently, org-roam-graph-viewer is expected to be the path to an executable.` It would be convenient to be able to provide a function for `org-roam-graph-viewer` so that one can customize within Emacs how the file is opened.

### Long Description

The current implementation of `org-roam-graph--open` assumes `org-roam-graph-viewer` is the path to an executable (checking using `executable-find`), and calls `call-process` to open the generated svg file.

https://github.com/jethrokuan/org-roam/blob/4af4d2e4d5f18422fe0f90d5d83fe4c10ef1a20f/org-roam-graph.el#L203-L208

This implementation does not permit passing any additional arguments to the executable. On macOS, the executable is `open`, and to specify the application (Safari), one passes `-a Safari` as an option. This can be worked around by creating a simple script that wraps `open -a Safari $1`, and provide the path to the script as `org-roam-graph-viewer` (which is what I've currently done). The downside of this approach is that it requires the creation of this additional script. While not unduly burdensome, it's less than ideal.

Another option as mentioned in https://github.com/jethrokuan/org-roam/issues/115#issuecomment-587248405 is to assign Safari as default application for svg files. This seems too be too big of a hammer, however. Choosing an svg viewer for org-roam specifically shouldn't require designating the default application for all svg files.

One could also override `org-roam-graph--open` entirely to provide the desired functionality, but the issue seems general enough that an in-the-box solution 

### Proposed Implementation (if any)

In `org-roam-graph--open`, check if org-roam-graph-viewer is a string or a function. If it's a function, call the function passing the file as an argument. If it's a string, assume its the path to an executable and use the current `call-process` path.

### Please check the following:
- [X] No similar feature requests

The topic has been mentioned tangentially in https://github.com/jethrokuan/org-roam/issues/115#issuecomment-587248405 (elaborated on above).

progfolio:
I implemented this a little bit ago, forgot to make a pull request for it:
https://github.com/progfolio/org-roam/commit/7a0191f761a7f0e2e267e4fd8341c0e34b3feaa6
I'll take a look at it later tonight and submit pull.

progfolio:
@grzm: sorry for the delay. Didn't forget, just had to get some other work out of the way first.

grzm:
[...]

PR #488, opened 2020-04-17
Allow function for org-roam-graph-viewer

progfolio:
User may specify a function to open the org-roam graph.
Adds more error handling to org-roam-graph--open.

See #440


RCoeurjoly:
How do you use this?
I tried 
(setq org-roam-graph-viewer eww-open-file)
but it tells me Symbol’s value as variable is void: eww-open-file

jethrokuan:
you need to quote it: `(setq org-roam-graph-viewer #'eww-open-file)`

Issue #450, opened 2020-04-13
Consolidate org-roam-graph commands


progfolio:
Currently we have four commands that do similar things from a user perspective:

- `org-roam-graph-build`
- `org-roam-graph-show`
- `org-roam-graph-build-connected-component`
- `org-roam-graph-show-connected-component`
  
Instead of exposing four similar commands, we could make three of them private functions
and dispatch based on the prefix argument in `org-roam-show`.
The following prefix arguments could dispatch like so:

    | Prefix | Effect                           |
    |--------+----------------------------------|
    | nil    | show  graph                      |
    | N      | show connected with max N steps  |
    | C-u    | build graph                      |
    | C-u -  | build connected                  |
    | C-u -N | build connected with max N steps |

# implementation sketch

After making the other three commands private functions,
we dispatch based on the prefix-arg:

```emacs-lisp
(defun org-roam-show (&optional arg node-query)
  (interactive "P")
  (let ((dispatched
         (pcase arg
           ('nil "showing")
           ('(4) "only building")
           ('-   "only building (connected)")
           ((pred (> 0))  "only building (connected with steps)")
           ((pred (< 0))  "showing (connected with steps)"))))
    ;;etc
    ))
```

Thoughts?

jethrokuan:
Maybe we can expose instead expose these functionality using `transient`, or `hydra`, which allows for a bit more discoverability.

progfolio:
That's not a bad idea for users that prefer using something like hydra or transient. For those who don't, this would be a way to allow more flexible functionality from a single key-binding/command. Hydra/transient users could still bind to multiple keys by wrapping the command and passing the appropriate arg. e.g.

```emacs-lisp
(defun my/org-roam-build-graph ()
  "Wrapper for `org-roam-show-graph' which only builds the graph."
  (org-roam-show-graph '(4)))
```

kljohann:
I would also humbly ask to not introduce too many _required_ dependencies (e.g. while I use transient in my configuration I'm trying to avoid pulling in hydra). :upside_down_face:

On the other hand having a single entry-point into the graph generation (like you suggest) sounds convenient. Though should the user maybe still have the possibility to conveniently/directly call into the underlying functions, if they want a more specific key binding? Also, separate commands might still be more discoverable. Maybe the dispatcher could be built on top of them?

progfolio:
[...]

PR #490, opened 2020-04-18
(refactor): org-roam-graph

progfolio:
- Consolidate graph build/display commands into org-roam-graph command
  See #450
- Require org-roam-db
  Rather than declaring its functions.
- Move obsolete variable org-roam-graph-node-shape to org-roam-compat
- org-roam-graph--build-connected-component accepts a file argument
  Allows building a graph without having the target file as the current
  buffer
- Eliminate repeating code
- Fix checkdoc warnings

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -33,12 +33,10 @@
 (require 'xml) ;xml-escape-string
 (require 's)   ;s-truncate, s-replace
 (require 'org-roam-macs)
+(require 'org-roam-db)
 
 ;;;; Declarations
 (defvar org-roam-directory)
-(declare-function org-roam-db--ensure-built  "org-roam-db")
-(declare-function org-roam-db-query          "org-roam-db")
-(declare-function org-roam-db--connected-component "org-roam-db")
 (declare-function org-roam--org-roam-file-p  "org-roam")
 (declare-function org-roam--path-to-slug     "org-roam")
 
@@ -100,27 +98,18 @@ are excluded."
           (list :tag "Matchers"))
   :group 'org-roam)
 
-(make-obsolete-variable 'org-roam-graph-node-shape  'org-roam-graph-node-extra-config "2020/04/01")
-(defcustom org-roam-graph-node-shape "ellipse"
-  "Shape of graph nodes."
-  :type 'string
-  :group 'org-roam)
-
 ;;;; Functions
 (defun org-roam-graph--expand-matcher (col &optional negate where)
   "Return the exclusion regexp from `org-roam-graph-exclude-matcher'.
 COL is the symbol to be matched against.  if NEGATE, add :not to sql query.
 set WHERE to true if WHERE query already exists."
-  (let ((matchers (cond ((null org-roam-graph-exclude-matcher)
-                         nil)
-                        ((stringp org-roam-graph-exclude-matcher)
-                         (cons (concat "%" org-roam-graph-exclude-matcher "%") nil))
-                        ((listp org-roam-graph-exclude-matcher)
-                         (mapcar (lambda (m)
-                                   (concat "%" m "%"))
-                                 org-roam-graph-exclude-matcher))
-                        (t
-                         (error "Invalid org-roam-graph-exclude-matcher"))))
+  (let ((matchers (pcase org-roam-graph-exclude-matcher
+                    ('nil nil)
+                    ((pred stringp) `(,(concat "%" org-roam-graph-exclude-matcher "%")))
+                    ((pred listp) (mapcar (lambda (m)
+                                            (concat "%" m "%"))
+                                          org-roam-graph-exclude-matcher))
+                    (_ (error "Invalid org-roam-graph-exclude-matcher"))))
         res)
     (dolist (match matchers)
       (if where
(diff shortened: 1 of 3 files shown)


org-roam-graph.el
```diff
@@ -144,120 +141,109 @@ into a digraph."
     (let* ((nodes (org-roam-db-query node-query))
            (edges-query
             `[:with selected :as [:select [file] :from ,node-query]
-              :select :distinct [to from] :from links
-              :where (and (in to selected) (in from selected))])
+                    :select :distinct [to from] :from links
+                    :where (and (in to selected) (in from selected))])
```
kljohann, line 145: This was due to `emacsql-fix-vector-indentation`:
```
When called, advise ‘calculate-lisp-indent’ to stop indenting vectors.
Once activate, vector contents no longer indent like lists.
```
I'm not sure which indentation style we want to keep. The one with `emacsql-fix-vector-indentation` seems a little cleaner, but then all contributors would be required to call that function before making changes to these queries. :man_shrugging:

org-roam-graph.el
```diff
@@ -144,120 +141,109 @@ into a digraph."
...
-                                     (concat (car n) "=" (cdr n))))
-                           (s-join ","))))
-      (insert (format "  edge [%s];\n"
-                      (->> org-roam-graph-edge-extra-config
-                           (mapcar (lambda (n)
-                                     (concat (car n) "=" (cdr n))))
-                           (s-join ","))))
-
+        (insert (org-roam-graph--dot-option option) ";\n"))
+      (dolist (node-type '("node" "edge"))
```
kljohann, line 157: maybe `attr-type` or `attr-kind`? Since "edge" isn't really a "node type".

org-roam-graph.el
```diff
@@ -144,120 +141,109 @@ into a digraph."
...
           (insert
-           (format "  \"%s\" [%s];\n"
-                   file
-                   (->> node-properties
-                        (mapcar (lambda (n)
-                                  (concat (car n) "=" "\"" (cdr n) "\"")))
-                        (s-join ","))))))
+           (format "  \"%s\" [%s];\n" file
+                   (mapconcat (lambda (n)
+                                (org-roam-graph--dot-option n nil "\""))
```
kljohann, line 175: I wonder whether `...-dot-option` should rather always wrap string-type arguments, keep numbers alone and convert symbols into unwrapped strings? This would be a breaking change towards existing configurations, though.

org-roam-graph.el
```diff
@@ -144,120 +141,109 @@ into a digraph."
...
                           (org-roam--path-to-slug file)))
                (shortened-title (s-truncate org-roam-graph-max-title-length title))
-               (node-properties (list (cons "label" (s-replace "\"" "\\\"" shortened-title))
-                                      (cons "URL" (concat "org-protocol://roam-file?file="
-                                                          (url-hexify-string file)))
-                                      (cons "tooltip" (xml-escape-string title)))))
+               (node-properties
+                `(("label"   . ,(s-replace "\"" "\\\"" shortened-title))
+                  ("URL"     . ,(concat "org-protocol://roam-file?file=" (url-hexify-string file)))
+                  ("tooltip" . ,(xml-escape-string title)))))
```
kljohann, line 171: It feels like `label` should be treated the same as tooltip? (I haven't tried it out, though.)

org-roam-graph.el
```diff
@@ -144,120 +141,109 @@ into a digraph."
...
          (temp-graph (make-temp-file "graph." nil ".svg")))
     (call-process org-roam-graph-executable nil 0 nil
                   temp-dot "-Tsvg" "-o" temp-graph)
     temp-graph))
 
+(defun org-roam-graph--build-connected-component (file &optional max-distance)
+  "Build a graph of nodes connected to FILE.
+If MAX-DISTANCE is non-nil, limit nodes to MAX-DISTANCE steps."
+  (unless (org-roam--org-roam-file-p file)
+    (user-error "%s is not an org-roam file" file))
```
kljohann, line 209: Maybe put quotes around the file name in the message? Also, in the case of a non-file-visiting buffer the message will read `nil is not an...`, which might be confusing?

org-roam-graph.el
progfolio: Good to know. I wasn't aware that was being used. I agree with both points (that's a bit cleaner looking, but also a hurdle for potential contributors).  One possibility is  to add per-file before-save-hook functions to ensure it is run. Though, I'm not too fond of evaling code when opening a file.

org-roam-graph.el
progfolio: good catch. Will change.

org-roam-graph.el
progfolio: I think I'll leave this for the moment, because I'm not all that familiar with the DOT language yet. Definitely could be more sophisticated if needed later on.

org-roam-graph.el
progfolio: https://www.graphviz.org/doc/info/shapes.html#html
>  As HTML strings are processed like HTML input, any use of the ", &, <, and > characters in literal text or in attribute values need to be replaced by the corresponding escape sequence. For example, if you want to use & in an href value, this should be represented as &amp;. 

So we probably should be escaping entities.

org-roam-graph.el
progfolio: Good point. I thought on this and think it would actually make sense to move the error checking to the `org-roam-graph` command, since that is the user entry-point. I'll add a case for each condition there so we can give more helpful user-error messages.

org-roam-graph.el
progfolio: https://renenyffenegger.ch/notes/tools/Graphviz/attributes/label/index
>  HTML like lables
Typically, the label is just a string that is enclosed in quotes (label = "some text").
However, if angle brackes are used (label = < <table> … >), Graphviz renders labels with HTML tags.

scratch that, plain text is valid for the label and escaping would revert #314.
The HTML labels look like they could enable some interesting features by allowing us to show more information in each node, but that's for another issue.

PR #491, opened 2020-04-18
(feat): Use TITLE as description when linking before first heading

### LINKED ISSUE #487: Capture note title as link description when using org-store-link
### Brief Abstract

`org-store-link` can be used to store a link to an org-roam file, but it does not store the title of the note. This means when you insert the stored link into an org file you get the path only with no description. It would be more useful if it stored the title of an org roam file as the link description, as well as the file name. 

### Long Description

As discussed in #477. 

@Zaeph wrote:

> At first, I was on the fence because I think that `org-roam-insert` should specifically work with `org-roam` files (i.e. inside `org-roam-directory`. After all, we can use `org-store-link` to get a link to the file. Problem is, if `org-before-first-heading-p` is `t`, we would only get the file-name, which might not be very informative.
>
> We could create a new link-handler for `org-roam` which would fill the visible part of a link with the `#+TITLE` of a note. We might also want to _avoid_ the usual behaviour which, when `org-before-first-heading-p` is `nil`, would link you to the nearest parent heading.

Create a new store link handler as described above. 

juergenhoetzel:
Fixes #487

#487 
IMO, this is especially useful because the file name is automatically generated in org roam.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -540,6 +540,15 @@ for Org-ref cite links."
                       :order-by (asc from)]
                      target))
 
+(defun org-roam-store-link ()
+  "Store a link to an `org-roam' file."
+  (when (org-before-first-heading-p)
+    (when-let ((title (cdr (assoc "TITLE" (org-roam--extract-global-props '("TITLE"))))))
+      (org-link-store-props
+       :type        "file"
+       :link        (format "file:%s" (abbreviate-file-name buffer-file-name))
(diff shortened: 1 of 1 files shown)


Zaeph:
Looks good, I'll try it out in a bit.

Zaeph:
Everything works as it should.

output-2020-04-18-11:47:11

Issue #104, opened 2020-02-17
Speed up `org-roam-find-files` using external tools


jethrokuan:
Shelling out to `find` should lead to significant speed improvements over the current e-lisp implementation. The elisp implementation will be a fallback for environments without these unix tools (i.e. windows).

asymmetric:
Please also consider ripgrep.

jethrokuan:
Closed with #664

PR #497, opened 2020-04-19
(internal): Simplify `org-roam--list-files` implementation.

juergenhoetzel:
Use destructive elisp functions to optimize for speed and memory
usage.

Refs #104

slight speedup:
Original:
```
ELISP> (benchmark 1000 '(org-roam--list-files org-roam-directory))
"Elapsed time: 29.403994s (6.193575s in 122 GCs)"
```
Commit:
```
ELISP> (benchmark 1000 '(org-roam--list-files org-roam-directory))
"Elapsed time: 27.186161s (5.973992s in 117 GCs)"
```

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -174,22 +174,11 @@ If FILE is not specified, use the current buffer's file-path."
 (defun org-roam--list-files (dir)
   "Return all Org-roam files located within DIR, at any nesting level.
 Ignores hidden files and directories."
-  (if (file-exists-p dir)
-      (let ((files (directory-files dir t "." t))
-            (dir-ignore-regexp (concat "\\(?:"
-                                       "\\."
-                                       "\\|\\.\\."
-                                       "\\)$"))
-            result)
-        (dolist (file files)
-          (cond
-           ((file-directory-p file)
-            (unless (string-match dir-ignore-regexp file)
-              (setq result (append (org-roam--list-files file) result))))
-           ((and (file-readable-p file)
-                 (org-roam--org-file-p file))
-            (setq result (cons (file-truename file) result)))))
-        result)))
+  (let ((regex (concat "\\.\\(?:"(mapconcat #'regexp-quote org-roam-file-extensions "\\|" )"\\)\\(?:\\.gpg\\)?\\'"))
+	result)
+    (dolist (file (directory-files-recursively dir regex) result)
+      (when (and (file-readable-p file) (org-roam--org-file-p file))
+	(push file result)))))
 
 (defun org-roam--list-all-files ()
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
BTW removing `file-truename` for each file would result in a huge speedup:

```
ELISP> (benchmark 1000 '(org-roam--list-files org-roam-directory))
"Elapsed time: 3.529275s (0.933762s in 18 GCs)"
```
Only downside: Symbolic linked org files don't get resolved.

org-roam.el
```diff
@@ -174,22 +174,10 @@ If FILE is not specified, use the current buffer's file-path."
...
-          (cond
-           ((file-directory-p file)
-            (unless (string-match dir-ignore-regexp file)
-              (setq result (append (org-roam--list-files file) result))))
-           ((and (file-readable-p file)
-                 (org-roam--org-file-p file))
-            (setq result (cons (file-truename file) result)))))
-        result)))
+  (let (result)
+    (dolist (file (directory-files-recursively dir "\\.\\(org\\|org.gpg\\)\\'" nil) result)
```
jethrokuan, line 178: The regexp for this needs to be built from `org-roam-file-extensions`.

jethrokuan:
The elisp implementation of finding these files should be a backup, this is tracked in an old issue #104

juergenhoetzel:
> The regexp for this needs to be built from org-roam-file-extensions.

Thanks: I amended the commit.

> I'm overall in favour of this, but the old code had a dir-ignore-regexp that tries to be smart by ignoring certain hidden directories (like those that start with ., e.g. .git), not sure if the performance is slow when there are such directories in org-roam-directory with a lot of objects.

I was able to remove the `file-truename` bottleneck (from `directory-files-recursively` docstring):


> Each file name appears in the returned list in its absolute form.


This results in a huge performance speedup:

Original Code:

```
ELISP> (benchmark 1000 '(org-roam--list-files org-roam-directory))
"Elapsed time: 21.594839s (7.690849s in 156 GCs)"
```
Commit:

```
ELISP> (benchmark 1000 '(org-roam--list-files org-roam-directory))
"Elapsed time: 6.259151s (1.435158s in 29 GCs)"
```
I would guess the ".git" special handling would not improve performance, because git uses pack-files with corresponding indices for large repos and the elisp code would be more complex.

timorunge:
@juergenhoetzel, @jethrokuan: If you're using tools like e.g. [Syncthing](https://syncthing.net) which has local versioning (the data is stored in a hidden folder) you will be able to see all versioned files. 

The correct RegEx to exclude hidden folders would be something like `^[^.](.*).(?:(org))(?:.gpg)?` but due to my limited lisp knowledge I was not able to get it working...

PR #508, opened 2020-04-21
(docs): Fix org-roam-protocol JS Bookmarklet

jmenglis:
JS Bookmarklet for org-roam-protocol was missing extra `/` in documentation.

### DIFF
--- a/doc/roam_protocol.md
+++ b/doc/roam_protocol.md
@@ -20,7 +20,7 @@ To use this, create a Firefox bookmarklet as follows:
 
 ```javascript
 javascript:location.href =
-'org-protocol:/roam-ref?template=r&ref='
+'org-protocol://roam-ref?template=r&ref='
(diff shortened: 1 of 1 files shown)


zaeph:
Just to be clear: it did not prevent the bookmarklet from forwarding the information, right?

jmenglis:
> Just to be clear: it did not prevent the bookmarklet from forwarding the information, right?

You are correct still works as intended.  Just for matching to [org-protocol](https://orgmode.org/worg/org-contrib/org-protocol.html#org120a44e).

zaeph:
>  Just for matching to [org-protocol](https://orgmode.org/worg/org-contrib/org-protocol.html#org120a44e).

I agree that we should keep to what's recommended in their manual for consistency.

Thanks for spotting the typo and making a PR!

PR #511, opened 2020-04-22
(bugfix): Fix invalid usage of error handler

juergenhoetzel:
The error handler was never invoked when executing
`org-roam-graph-viewer` failed

No signal when executing `org-roam-graph-viewer` failed.

BTW: It would be nice to make byte compile errors fatal in the CI jobs. The above error was also revealed when byte-compiling:

```
In toplevel form:
org-roam-graph.el:212:1:Warning: Unused lexical variable ‘err’


```

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -216,7 +216,7 @@ into a digraph."
      (if (executable-find org-roam-graph-viewer)
          (condition-case err
              (call-process org-roam-graph-viewer nil 0 nil file)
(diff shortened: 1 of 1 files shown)


progfolio:
Apologies! Thank you for catching that.

progfolio:
> It would be nice to make byte compile errors fatal in the CI jobs. The above error was also revealed when byte-compiling

Agreed. We should be aiming for more test coverage as the project matures, too. I'm guilty of making quite a few changes without adding to the tests.

PR #512, opened 2020-04-22
add keybinding to qutebrowser

Accacio:

### DIFF
--- a/doc/roam_protocol.md
+++ b/doc/roam_protocol.md
@@ -25,6 +25,14 @@ javascript:location.href =
 + '&title='
 + encodeURIComponent(document.title)
 ```
+or as a keybinding in qutebrowser, adding the following to the autoconfig.yml file:
+```yml
+settings:
+  bindings.commands:
+    global:
+      normal:
+        gc: open javascript:void(location.href='org-protocol://roam-ref?template=r&ref='+encodeURIComponent(location.href)+'&title='+encodeURIComponent(document.title))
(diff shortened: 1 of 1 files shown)


The-Compiler:
qutebrowser maintainer here, just discovered this - yay! :smile: 

However, let me note the top of qutebrowser's `autoconfig.yml`:

```yaml
# DO NOT edit this file by hand, qutebrowser will overwrite it.
# Instead, create a config.py - see :help for details.
```

So instead of recommending to edit `autoconfig.yml` by hand, this should probably either recommend the `:bind` command, or a snippet for `config.py` (see qutebrowser's [configuration documentation](https://github.com/qutebrowser/qutebrowser/blob/master/doc/help/configuring.asciidoc)).

Also I think you could use `:jseval` instead of `:open` here, but if that works as well, I guess it doesn't make much of a difference.

jethrokuan:
There's also #753 recently, but I couldn't find this PR that added the docs. I'll have them changed.

The-Compiler:
> but I couldn't find this PR that added the docs

FWIW I can recommend the `-G` option to `git log`, which lets you search for changes inside git commits. In this case, `git log -G qutebrowser` find this commit/PR (and then some newer doc restructurings).

Accacio:
> However, let me note the top of qutebrowser's `autoconfig.yml`:
@The-Compiler, I'm sorry about the bad practice 😢

Issue #494, opened 2020-04-19
Generalize org-roam-graph to allow multiple graph building backends


progfolio:
Currently dot/graphviz are the only option. We could generalize similarly to #488 and allow either a function or a string for `org-roam-graph-executable`.  Then when we build the graph, if a function is supplied we can simply defer to that. This would allow alternative graph visualization back ends to be easily developed for org-roam as separate packages. This would also allow the  graphviz/dot functionality to be moved to its own module and offer more/cleaner customization options as well.

progfolio:
Not packaged yet, but implementing this would allow users to plug in something like: https://github.com/alphapapa/org-graph-view

Accacio:
it could also be interesting to generate a html page instead of a svg, so we could use visjs to plot the graph as in [https://visjs.github.io/vis-network/examples/network/data/dotLanguage/dotEdgeStyles.html](https://github.com/jethrokuan/org-roam/issues/url)

zaeph:
> it could also be interesting to generate a html page instead of a svg, so we could use visjs to plot the graph as in https://visjs.github.io/vis-network/examples/network/data/dotLanguage/dotEdgeStyles.html

I believe there's already been a project exploring a JS implementation of the graph on Slack.  The author said he might open a PR in the future, so we'll have to keep an eye out for that.

PR #513, opened 2020-04-23
(feat) implement org-roam-find-index

zaeph:
Implement a new command `org-roam-find-index` and helper function `org-roam--find-index` (note the `--` for internal) for finding an index file in `org-roam-directory`.

The resources I've read on the slip-box indicated that one should maintain an index file to lead you to your topic-notes.  From Ahrens (2017):
> After adding a note to the slip-box, **we need to make sure it can be found again**. This is what the index is for.
[…]
The organisation of the notes is in the network of references in the slip-box, so all we need from the index are entry points. A few wisely chosen notes are sufficient for each entry point. **The quicker we get from the index to the concrete notes, the quicker we move our attention from mentally preconceived ideas towards the fact-rich level of interconnected content, where we can conduct a fact-based dialogue with the slip-box**. [emphasis added]

So, I think it might be useful to *recommend* users to create an index file, and to provide a command to access it.  I've proposed an extensible way to match the index.  For now:
- The path to the index file, relative or absolute, can be defined in`org-roam-index-file`.
- Otherwise, the default is to select an Org-roam note whose `#+TITLE:` is `Index`.
- If neither of those options is satisfied, the command offer to create it for you via `org-roam-find-file`.

References:
- <div class="csl-bib-body" style="line-height: 1.35; margin-left: 2em; text-indent:-2em;">
  <div class="csl-entry">Ahrens, Sönke. <i>How to Take Smart Notes: One Simple Technique to Boost Writing, Learning and Thinking for Students, Academics and Nonfiction Book Writers</i>, 2017.</div>
  <span class="Z3988" title="url_ver=Z39.88-2004&amp;ctx_ver=Z39.88-2004&amp;rfr_id=info%3Asid%2Fzotero.org%3A2&amp;rft_id=urn%3Aisbn%3A978-1-5428-6650-7&amp;rft_val_fmt=info%3Aofi%2Ffmt%3Akev%3Amtx%3Abook&amp;rft.genre=book&amp;rft.btitle=How%20to%20Take%20Smart%20Notes%3A%20One%20Simple%20Technique%20to%20Boost%20Writing%2C%20Learning%20and%20Thinking%20for%20Students%2C%20Academics%20and%20Nonfiction%20Book%20Writers&amp;rft.aufirst=S%C3%B6nke&amp;rft.aulast=Ahrens&amp;rft.au=S%C3%B6nke%20Ahrens&amp;rft.date=2017&amp;rft.isbn=978-1-5428-6650-7&amp;rft.language=English"></span>
</div>

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -422,9 +422,9 @@ whose title is 'Index'."
                  (wrong-type (signal 'wrong-type-argument
                                      `((functionp stringp)
                                        ,wrong-type))))))
-    (pcase index
-      ((pred (f-relative-p)) (concat (file-truename org-roam-directory) path))
-      (_ index))))
+    (if (f-relative-p index)
+        (concat (file-truename org-roam-directory) path)
+      index)))
 
 (defun org-roam-find-index ()
   "Find the index file in `org-roam-directory'.
--- a/org-roam.el
+++ b/org-roam.el
@@ -417,7 +417,7 @@ whose title is 'Index'."
   (let* ((index org-roam-index-file)
          (path (pcase index
                  ((pred functionp) (funcall index))
-                 ((pred stringp) index)
+                 ((or (pred stringp) 'nil) index)
                  (wrong-type (signal 'wrong-type-argument
                                      `((functionp stringp)
                                        ,wrong-type))))))
--- a/org-roam.el
+++ b/org-roam.el
@@ -403,7 +403,7 @@ absolute.
 If it is not defined, the index is assumed to be a note in
 `org-roam-directory' whose title is 'Index'.")
 
-(defun org-roam--find-index ()
+(defun org-roam--get-index-path ()
   "Return the path to the index in `org-roam-directory'.
 The path to the index can be defined in `org-roam-index-file'.
 Otherwise, it is assumed to be a note in `org-roam-directory'
@@ -423,7 +423,7 @@ Otherwise, the function will look in your `org-roam-directory'
 for a note whose title is 'Index'.  If it does not exist, the
 command will offer you to create one."
   (interactive)
-  (let ((index (org-roam--find-index)))
+  (let ((index (org-roam--get-index-path)))
     (if (and index
              (file-exists-p index))
         (find-file index)
--- a/org-roam.el
+++ b/org-roam.el
@@ -436,7 +436,7 @@ command will offer you to create one."
     (if (and index
              (file-exists-p index))
         (find-file index)
-      (when (y-or-n-p "Index file does not exist.  Would you like to create it?")
+      (when (y-or-n-p? "Index file does not exist.  Would you like to create it?")
         (org-roam-find-file "Index")))))
 
 ;;;; org-roam-find-ref
--- a/org-roam.el
+++ b/org-roam.el
@@ -437,7 +437,7 @@ command will offer you to create one."
     (if (and index
              (file-exists-p index))
         (find-file index)
-      (when (y-or-n-p? "Index file does not exist.  Would you like to create it?")
+      (when (y-or-n-p "Index file does not exist.  Would you like to create it? ")
         (org-roam-find-file "Index")))))
 
 ;;;; org-roam-find-ref
--- a/org-roam.el
+++ b/org-roam.el
@@ -421,12 +421,10 @@ whose title is 'Index'."
                  (wrong-type (signal 'wrong-type-argument
                                      `((functionp stringp)
                                        ,wrong-type))))))
-    (or (and org-roam-index-file
-             (if (f-relative-p path)
-                 (concat (file-truename org-roam-directory) path)
-               path))
-        (cdr (assoc "Index"
-                    (org-roam--get-title-path-completions))))))
+    (pcase index
+      ('nil (cdr (assoc "Index" (org-roam--get-title-path-completions))))
+      ((pred (f-relative-p)) (concat (file-truename org-roam-directory) path))
+      (_ index))))
 
 (defun org-roam-find-index ()
   "Find the index file in `org-roam-directory'.
--- a/org-roam.el
+++ b/org-roam.el
@@ -414,10 +414,13 @@ assumed to be a note in `org-roam-directory' whose title is
 The path to the index can be defined in `org-roam-index-file'.
 Otherwise, it is assumed to be a note in `org-roam-directory'
 whose title is 'Index'."
-  (let ((path (--> org-roam-index-file
-                   (if (functionp it)
-                       (funcall it)
-                     it))))
+  (let* ((index org-roam-index-file)
+         (path (pcase index
+                 ((pred functionp) (funcall index))
+                 ((pred stringp) index)
+                 (wrong-type (signal 'wrong-type-argument
+                                     `((functionp stringp)
+                                       ,wrong-type))))))
     (or (and org-roam-index-file
              (if (f-relative-p path)
                  (concat (file-truename org-roam-directory) path)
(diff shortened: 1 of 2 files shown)


ocharles:
Does the graph not act as that index? It certainly allows all notes to be found

zaeph:
> Does the graph not act as that index? It certainly allows all notes to be found

It does, and it's a great way to find orphan notes which isn't possible with the paper slip-box.  However, I think there is value in allowing the users to maintain an `Index` file themselves.  There's a degree of intention that comes with it which should be encouraged in my eyes, but I'm open to feedback.

progfolio:
> Does the graph not act as that index? It certainly allows all notes to be found

We could dynamically build a textual index. This would be very flexible and reuse existing code once #494 is implemented.

zaeph:
> We could dynamically build a textual index. This would be very flexible and reuse existing code once #494 is implemented.

It could indeed be a good way to create a textual index for those who cannot access the visual, node-based representation.  I think this is a little different from the `Index` that I'm suggesting we implement, but we could differentiate the two with `Automatic Index` vs `Index`.

progfolio:
> but we could differentiate the two with `Automatic Index` vs `Index`
Perhaps `org-roam-index-file` should allow a function as well.
If it's value is a string, we use the logic you've already implemented to find that file.
If it is a function, `org-roam-find-index` defers to that function.

zaeph:
> Perhaps `org-roam-index-file` should allow a function as well.

Sounds good, I'll implement this.

progfolio:
Just to be clear, I think the custom function should return the path to the index file. It can do more (e.g. dynamically create the file), but as far as `org-roam-find-index` is concerned, it just provides the path.

zaeph:
> Just to be clear, I think the custom function should return the path to the index file. It can do more (e.g. dynamically create the file), but as far as `org-roam-find-index` is concerned, it just provides the path to the index.

This is what I had in mind when I create this internal function:
https://github.com/jethrokuan/org-roam/blob/a31e4f862beddfb934d93195452fd00539809efc/org-roam.el#L406-L417

It's then used in `org-roam-find-index`:
https://github.com/jethrokuan/org-roam/blob/a31e4f862beddfb934d93195452fd00539809efc/org-roam.el#L419-L431

progfolio:
fantastic. Looks like a simple check in `org-roam--find-index` on `org-roam-index-file`'s type.
e.g.

```emacs-lisp
(let ((path (if (functionp org-roam-index-file) (funcall org-roam-index-file) ;;etc
```
Another suggestion: maybe rename `org-roam--find-index` to `org-roam--index-path`?

zaeph:
> fantastic. Looks like a simple check in `org-roam--find-index` on `org-roam-index-file`'s type.
> e.g.
> 
> ```emacs-lisp
> (let ((path (if (functionp org-roam-index-file) (funcall org-roam-index-file) ;;etc
> ```

Done.

I've kept the code to create the index if it doesn't exist, which we might remove as soon as we implement the default function to generate the index.

> Another suggestion: maybe rename `org-roam--find-index` to `org-roam--index-path`?

I've looked at the rest of the codebase, and I've found that `org-roam--get-index-path` might be a better fit.  [What do you think?](https://xkcd.com/910/)

org-roam.el
```diff
@@ -398,17 +398,26 @@ which takes as its argument an alist of path-completions.  See
...
+(defun org-roam--get-index-path ()
   "Return the path to the index in `org-roam-directory'.
 The path to the index can be defined in `org-roam-index-file'.
 Otherwise, it is assumed to be a note in `org-roam-directory'
 whose title is 'Index'."
-  (let ((path org-roam-index-file))
+  (let ((path (--> org-roam-index-file
+                   (if (functionp it)
+                       (funcall it)
+                     it))))
```
progfolio, line 420: Should probably verify that `org-roam-index-file` is a function or a string explicitly. Throw an error in cases where it is not.

progfolio:
```emacs-lisp

  (pcase org-roam-index-file
    ((pred functionp) (funcall org-roam-index-file))
    ((pred stringp)   org-roam-index-file)
    (wrong-type (signal 'wrong-type-argument `((functionp stringp) ,wrong-type))))
```

progfolio:
> I've looked at the rest of the codebase, and I've found that org-roam--get-index-path might be a better fit. What do you think?

lol @ the comic. That's a fine name, too.

org-roam.el
zaeph: …Wow.  I didn't know you could do [all this with `pcase`](http://newartisans.com/2016/01/pattern-matching-with-pcase/)!  Colour me impressed, and thanks for the discovery!

zaeph: Not only that, but you've taught me how to use `signal`.  I'm completely mindblown by the amount of stuff I've discovered in just 4 lines.  Really, thanks!

progfolio: Glad to help! Pattern matching is a powerful feature.
At some point we may want to issue our own types of errors, but for now using signal with the standard errors is better than nothing at all.

org-roam.el
```diff
@@ -395,6 +395,51 @@ which takes as its argument an alist of path-completions.  See
...
+Otherwise, the function will look in your `org-roam-directory'
+for a note whose title is 'Index'.  If it does not exist, the
+command will offer you to create one."
+  (interactive)
+  (let ((index (org-roam--get-index-path)))
+    (if (and index
+             (file-exists-p index))
+        (find-file index)
+      (when (y-or-n-p "Index file does not exist.  Would you like to create it? ")
+        (org-roam-find-file "Index")))))
```
jethrokuan, line 441: it seems presumptuous to hardcode the file title here. Would this not ignore the `org-roam-index-file` variable?

zaeph: > it seems presumptuous to hardcode the file title here. Would this not ignore the `org-roam-index-file` variable?

The first arg of `org-roam-find-file` is for pre-filling the completion, so the user can still change it.  i can remove the `y-or-n-p` altogether if you want.

jethrokuan: Won't it then create an index file, at a path that's not the same as `org-roam-index-file` (if that's set)?

zaeph: It's because @progfolio suggested that we automatically-generate an index based on what's available in the db.  I've left the legacy `y-or-no-p` creation of file as a temporary measure.  I can look into better options if you want, or we can just `error` on `file-exists-p` in the meantime.

zaeph: Does 6d3cf8110d73e8f2a68591f8112598749e220532 look better to you?

org-roam.el
```diff
@@ -395,6 +395,51 @@ which takes as its argument an alist of path-completions.  See
...
+         (path (pcase index
+                 ((pred functionp) (funcall index))
+                 ((pred stringp) index)
+                 ('nil (user-error "You need to set `org-roam-index-file' before you can jump to it"))
+                 (wrong-type (signal 'wrong-type-argument
+                                     `((functionp stringp)
+                                       ,wrong-type))))))
+    (pcase index
+      ((pred (f-relative-p)) (concat (file-truename org-roam-directory) path))
+      (_ index))))
```
progfolio, line 427:
```suggestion
    (if (f-relative-p index)
    (concat (file-truename org-roam-directory) path)
  index)
```
pcase is usually overkill when there's a single if/else branch.

zaeph: Thanks, I forgot to refactor it after removing the conditions that warranted it.

Fixed in f4399737271ffde04cb17179e2e3dda83be39514.

jethrokuan:
LGTM, but this needs some documentation, otherwise it'll go pretty much undiscovered/unused.

zaeph:
> LGTM, but this needs some documentation, otherwise it'll go pretty much undiscovered/unused.

I'll add something to the readme in a bit.

PR #516, opened 2020-04-24
Restore the org-roam-backlink face

### LINKED ISSUE #515: org-roam-backlink face never applies
### Description

When opening the backlink buffer `org-roam-backlink` will never apply its face to links that are backlinks. Grepping the codebase didn't show me anywhere where `org-roam-backlink` should actually apply, so I guess the code is missing somewhere?

#### Expected Results
`org-roam-backlink` should apply its face on backlinks in the backlink buffer.

#### Actual Results
The backlink buffer renders all backlinks with `org-roam-link` face.

### Versions
- Emacs (`C-h v emacs-version`): 26.1
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/d01f7b2dafcbcb4f7ca3e5f9ab4bec125070e1d4

zaeph:
Fixes #515.

However, there are edge cases that are proving surprisingly difficult to address.  Consider the following structure in your `org-roam-directory`:
```
sub1
├── foo.org
└── sub3
    └── foo.org
sub2
└── foo.org
```
And the following links:
- `sub2/foo.org` → `sub1/foo.org`
- `sub1/sub3/foo.org` → `sub1/foo.org`

With the basic implementation (i.e., not the one tagged `(experimental)`), the `org-roam-backlink` over-applies to the file on which the link was found, sine they share their `file-name-non-directory`.  Backlinks are in maroon:
screenshot-0jgT0Z0Xxn

The main problem is that `org-roam--roam-link-face' only receives the PATH of the link from `org-link-set-parameters', and it can be relative.  Without having access to the file itself, we cannot reliably compute the absolute path of the file to which the backlink is pointing to ascertain that it is the same as `org-roam-buffer--current'.

I've tried to solve it on my own in 734618a2fd70ee433466a404a67e5dd00fce4835, but this is a surprisingly tough nut to crack.

We could merge the quick-fix, and leave this branch running until we can figure out a better way to do it.


zaeph:
I'm well on my way for a more elegant solution, so please don't start any work on this.

zaeph:
Phew.  Blood, sweat, and tears on that one.

34cb891f7a25ac3121613f30bd2951ad8d5d90ed implements a robust check for backlink target.  it'll work as long as we keep the same format for the side-window.  If we change it, we'd have to find a new way to retrieve `origin-file` (which shouldn't be too hard to do, mind you).

org-roam.el
```diff
@@ -459,12 +459,46 @@ INFO is an alist containing additional information."
   "Face for org-roam backlinks in backlinks buffer"
   :group 'org-roam-faces)
 
+(defun org-roam--in-buffer-p ()
+  "Return t if in the Org-roam buffer."
+  (string= (buffer-name)
+           org-roam-buffer))
```
zaeph, line 465: This isn't a particularly robust way to check if we're in the side-window, but I don't think there was anything allowing this check in the codebase.  It'd probably be a good idea to create a minor-mode for the side-window, and we should keep this in mind when we get to the revamp.

jethrokuan:
Maybe `org-roam-backlink` isn't the right name for the face: it doesn't suggest to me that it's the currently viewed file.

zaeph:
> Maybe `org-roam-backlink` isn't the right name for the face: it doesn't suggest to me that it's the currently viewed file.

- `org-roam-current-link`?
- `org-roam-link-to-current`?  (Looks like a function.)

jethrokuan:
maybe the former? I think it's better than `org-roam-backlink`.

zaeph:
> maybe the former? I think it's better than `org-roam-backlink`.

I agree.  What about `org-roam-backlinks-mode`, though?

zaeph:
> maybe the former? I think it's better than `org-roam-backlink`.

Addressed in c8b9d702ee7617913fc60d53cca42813400e4af9.

> What about `org-roam-backlinks-mode`, though?

I've left it as is for now.

org-roam.el
```diff
@@ -451,20 +451,54 @@ INFO is an alist containing additional information."
...
+  "Return t if backlink is to the current Org-roam file."
+  (let* ((current (buffer-file-name org-roam-buffer--current))
+         (backlink-dest (org-roam--retrieve-link-path))
+         ;; Retrieve path from file-heading
+         (origin-file (save-excursion
+                        (org-back-to-heading)
+                        (re-search-forward "^\\*+ " nil t)
+                        (org-roam--retrieve-link-path)))
+         ;; Compute absolute path to `backlink-dest'
+         (backlink-dest-abs (relative-expand-file-name
```
Wetlize, line 484: Where's `relative-expand-file-name` comes from? This causes `Error during redisplay: (jit-lock-function 1) signaled (void-function relative-expand-file-name)` in my Emacs.

zaeph: Damn, you're right.  I think it was native, but it's from `load-relative`.  I took me quite a while to find a function that did what I wanted, i.e. to compute an absolute path from 1) another absolute path, and 2) a relative path from that 2nd absolute path.

I'll see what I can do.

zaeph: It's easy to refactor.  I'll take care of it now.

zaeph: Addressed in 85975de650d8dd13b781d3b8553efd654fcac560.

zaeph:
We might want to hold off on this: this is merely addressing an effect rather than the source.

A better solution would be to replace all the links in the `org-roam-buffer` with their absolute counterparts.

zaeph:
> We might want to hold off on this: this is merely addressing an effect rather than the source.

#520 is an attempt to address the source.  If it were to be merged (refactored or not), this PR could be pruned and applied on top to fix #515.

zaeph:
> If it were to be merged (refactored or not), this PR could be pruned and applied on top to fix #515.

Did it in https://github.com/zaeph/org-roam/commit/1524e06e44c0c3a2a7a99a0533aa29890715c135 .

zaeph:
Now that #520 has been merged, I've cherry picked the relevant commits in 4163667a0311365e2aa2c426a57de86b52fa9cae.

org-roam.el
```diff
@@ -603,20 +603,45 @@ INFO is an alist containing additional information."
...
   :group 'org-roam-faces)
 
+(defun org-roam--in-buffer-p ()
+  "Return t if in the Org-roam buffer."
+  (string= (buffer-name)
+           org-roam-buffer))
+
+(defun org-roam--retrieve-link-path (&optional pom)
+  "Retrieve the path of the link at point or POM.
+POM can be a position in the current buffer or a marker."
```
jethrokuan, line 621:
```suggestion
(defun org-roam--retrieve-link-path (&optional point-or-marker)
  "Retrieve the path of the link at POINT-OR-MARKER.
  If nil, POINT-OR-MARKER is set to the current point."
```

No need to abbrieviate here.

zaeph: It's actually what Org-mode uses in some of their commands:
```el
(defmacro org-with-point-at (pom &rest body)
  "Move to buffer and point of point-or-marker POM for the duration of BODY."
  (declare (debug (form body)) (indent 1))
  (org-with-gensyms (mpom)
    `(let ((,mpom ,pom))
       (save-excursion
	 (when (markerp ,mpom) (set-buffer (marker-buffer ,mpom)))
	 (org-with-wide-buffer
	  (goto-char (or ,mpom (point)))
	  ,@body)))))
```

I don't mind changing it, I'll leave it up to you.

jethrokuan: hmm maybe follow their docstring. I was initially confused by the "point or POM"

org-roam.el
zaeph: You're right.

Addressed in 0d210932fe63c00945f888b3961f6b909c3c3f96.

org-roam.el
zaeph: I think it's not clear.  Do you prefer 36199f15bd8dda34961404e8a7e2562a88a1b6bb?

PR #520, opened 2020-04-25
(feat) expand rel-links upon extraction

### LINKED ISSUE #517: The backlink buffer doesn't correctly open links from the preview area using sub-directories
### Description
When you have your files spread across multiple sub-directories the links of which point to files across all other notes in sub-directories of `org-roam-directory`, opening them from the preview area of the backlink buffer will not properly resolve them (i.e. it will not open the correct file). Read further for more details.

This only happens when you use `org-roam-open-at-point` at the preview area with the point put on a link itself. If you use `org-roam-open-at-point` at a pure (non-linking) text, then it will properly resolve to the destination of the text, as expected.

#### Steps to Reproduce
In `org-roam-directory` create files copying the next structure:
- `org-roam-directory`
  - foobar1.org
  - foobar2.org
  - foo/
    - foo1.org
    - foo2.org
  - bar/
    - bar1.org
    - bar2.org

Put this in `foo1.org`:
```
[[file:../foobar1_org.org][Foobar]] from the main root, [[file:../foobar2.org][Foobar2]] from the main root, [[file:foo1.org][(foo) Foo1]] (self-loop) and [[file:foo2.org][(foo) Foo2]] from =foo= subdir, [[file:../bar/bar1.org][(bar) Bar1]] and [[file:../bar/bar2.org][(bar) Bar2]] from =bar= subdir.
```

Then try open all these links from the preview area of the backlink buffer. It doesn't matter which file you use select as your `org-roam-buffer--current` candidate, since all of them will behave the same.

#### Expected Results
Links that are part of the preview content should properly resolve their paths to correct destinations.

#### Actual Results
The links resolve instead the next way:
- links to `foobar1.org`, `foobar2.org`, `bar1.org`, `bar2.org` will always resolve to `foo1.org` file.
- link to `foo1.org` and `foo2.org` will always create new buffers in the main root of `org-roam-directory`.

### Versions
- Emacs (`C-h v emacs-version`): 26.1
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/963692f353090359b0513cc75abe92e7e7546bfc, #516 and #427 (tried all merged and separated).

zaeph:
Fixes #517.

-----

Here's a prototype to address #517 (and help with #516).

The links created by `org-roam-insert` are relative.  This is well an good until we need to handle those relative links in `org-roam-buffer`.  The problem is `org-roam-buffer` is buffer-agnostic, and therefore file-path-agnostic as well.  Since we do not have the root from which to compute the relativity, we cannot reliably get the absolute path of a backlink.

#516 attempted to resolve the issue by exploiting the fact that `org-roam-buffer--insert-backlinks` inserted those backlinks under a heading which contained the absolute path to the file in which those backlinks were found.  It worked, yes, but it was hackish.  Also, it only covered an effect of the problem rather than its cause.

#517 then came along with its targeting problem, and it led me to think about the way we store those backlinks in the database.  This in turn led me to `org-roam--extract-links`.

There's a lot going on in that function, but I'm only modifying the way `content` is stored in the database.  Think of `content` as the neighbouring content, the line you see below a backlink (highlighted below):
screenshot-43ITJMHZDj

I'm running the content through the `org-roam--expand-links` which expands all the relative links, and returns the resulting string.

-----

This implementation is working, but there are a couple of things bugging me:
1. I've had to replace `buffer-substring` with `buffer-substring-no-properties` because I wanted to avoid modifying them.  As far as I can see, this is not a problem for now since `org-roam-buffer` still depends on `org-mode`, which means that the proper fontification is applied.  Since the formatting of the links also depends on `org-mode`, I think this is not a big deal.
2. The whole thing has this hack-ish quality to it that I don't like.  We might be able to find a better to store the content inside the database.  I'm not good with text-properties or with `org-element-property`, so I could only do a rudimentary job.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -200,18 +200,24 @@ The search terminates when the first property is encountered."
         (push (cons prop p) res)))
     res))
 
+(defvar org-roam--org-link-file-bracket-re
+  "\\[\\[file:\\(\\(?:[^][\\]\\|\\\\\\(?:\\\\\\\\\\)*[][]\\|\\\\+[^][]\\)+\\)]\\(?:\\[\\(\\(?:.\\|
+\\)+?\\)]\\)?]"
+  "Matches a 'file:' link in double brackets.")
+
 (defun org-roam--expand-links (content path)
   "Crawl CONTENT for relative links and expand them.
 PATH should be the root from which to compute the relativity."
-  (let ((dir (file-name-directory path)))
+  (let ((dir (file-name-directory path))
+        (re org-roam--org-link-file-bracket-re))
     (with-temp-buffer
       (insert content)
       (goto-char (point-min))
       ;; Loop over links
-      (while (re-search-forward org-link-bracket-re (point-max) t)
+      (while (re-search-forward re (point-max) t)
         (goto-char (match-beginning 1))
         ;; Strip 'file:'
-        (setq link (substring (match-string 1) 5))
+        (setq link (match-string 1))
         ;; Delete relative link
         (when (f-relative-p link)
           (delete-region (match-beginning 1)
--- a/org-roam.el
+++ b/org-roam.el
@@ -200,6 +200,26 @@ The search terminates when the first property is encountered."
         (push (cons prop p) res)))
     res))
 
+(defun org-roam--expand-links (content path)
+  "Crawl CONTENT for relative links and expand them.
+PATH should be the root from which to compute the relativity."
+  (let ((dir (file-name-directory path)))
+    (with-temp-buffer
+      (insert content)
+      (goto-char (point-min))
+      ;; Loop over links
+      (while (re-search-forward org-link-bracket-re (point-max) t)
+        (goto-char (match-beginning 1))
+        ;; Strip 'file:'
+        (setq link (substring (match-string 1) 5))
+        ;; Delete relative link
+        (when (f-relative-p link)
+          (delete-region (match-beginning 1)
+                         (match-end 1))
+          (insert (expand-file-name
+                   (concat dir link)))))
+      (buffer-string))))
+
 (defun org-roam--extract-links (&optional file-path)
   "Extracts all link items within the current buffer.
 Link items are of the form:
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -200,6 +200,26 @@ The search terminates when the first property is encountered."
...
+        (goto-char (match-beginning 1))
+        ;; Strip 'file:'
+        (setq link (substring (match-string 1) 5))
+        ;; Delete relative link
+        (when (f-relative-p link)
+          (delete-region (match-beginning 1)
+                         (match-end 1))
+          (insert (expand-file-name
+                   (concat dir link)))))
+      (buffer-string))))
```
Wetlize, line 221: This needs to be a bit more robust, since it will break all non `file:` based links.

I've also tried to include an absolute `file:` link to somewhere outside of `org-roam-directory` and it breaks it too. Not sure though if it someone would do so in a real case?

zaeph: > This needs to be a bit more robust, since it will break all non `file:` based links.

Thanks for spotting this.  I'd shielded the code at the start, but through the iterations, I've forgot to reimplement the check.

> I've also tried to include an absolute `file:` link to somewhere outside of `org-roam-directory` and it breaks it too. Not sure though if someone would do so in a real case?

What do you mean by breaking?  It's working for me.
screenshot-E0cumC3rgA

Wetlize: > What do you mean by breaking?

Huh, It didn't follow it from the org-roam-buffer and instead it would open for me `foo1.org` file. Cannot reproduce this after restarting Emacs.

zaeph: I think this is expected behaviour.  Clicking anywhere on the text which isn't an `org-roam-link` or `org-roam-link-current` brings you to the file where the link was found.

Wetlize: Now I can reproduce and yes, you're correct. This logic is written in `org-roam-open-at-point`.

zaeph: > This needs to be a bit more robust, since it will break all non `file:` based links.

Fixed in 80f6a8affc2788d744534b11e4cc79d45e330b11.

zaeph:
I've tried all my current PRs ( #427, #516, #513, #521, #520) in https://github.com/zaeph/org-roam/tree/next, and everything works as it should without conflict.

Wetlize:
Can confirm, https://github.com/zaeph/org-roam/tree/next works as expected and fixes all these issues. I'm switching to it until merge on the upstream. Thanks for your work.

zaeph:
> Can confirm, https://github.com/zaeph/org-roam/tree/next works as expected, fixes all my issues and generally improves the workflow with sub-directories. I'm switching to it until all this gets merged on the upstream. Thank you for your work!

Glad to hear!
Just be careful because my `next` branch is unstable, and I often hard-reset it to test combinations of PR.  I've created `maint` for you which is pinned to where `next` was when you checked it out, so you should probably get on that branch instead.

PR #521, opened 2020-04-25
(feat) conditionally pluralize heading in org-roam-buffer

zaeph:
Prompted by #520.

I've noticed that we didn't have a switch between singular and plural for the titles in `org-roam-buffer`:
screenshot-43ITJMHZDj

So, I went ahead and implemented it.

screenshot-W1OafD3HcY

Note that I've also done it for cite backlinks.

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -108,8 +108,9 @@ When non-nil, the window will not be closed when deleting other windows."
             (key-backlinks (org-roam--get-backlinks (s-chop-prefix "cite:" roam-key)))
             (grouped-backlinks (--group-by (nth 0 it) key-backlinks)))
       (progn
-        (insert (format "\n\n* %d Cite backlinks\n"
-                        (length key-backlinks)))
+        (insert (let ((l (length key-backlinks)))
+                  (format "\n\n* %d Cite backlink%s\n"
+                          l (if (> l 1) "s" ""))))
         (dolist (group grouped-backlinks)
           (let ((file-from (car group))
                 (bls (cdr group)))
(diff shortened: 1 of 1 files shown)


org-roam-buffer.el
```diff
@@ -108,8 +108,9 @@ When non-nil, the window will not be closed when deleting other windows."
             (key-backlinks (org-roam--get-backlinks (s-chop-prefix "cite:" roam-key)))
             (grouped-backlinks (--group-by (nth 0 it) key-backlinks)))
       (progn
-        (insert (format "\n\n* %d Cite backlinks\n"
-                        (length key-backlinks)))
+        (insert (let ((l (length key-backlinks)))
+                  (format "\n\n* %d Cite backlink%s\n"
+                          l (if (> l 1) "s" ""))))
```
jethrokuan, line 113: Better implemented as a helper function (org-roam-pluralize) or something.

zaeph: Sure, on it.

zaeph: Uhm, I'm actually not sure how to best implement this.  db6a477360e9bef930a92e48f434f9f7bd557a5f is hardly any better.  I'm thinking about a double `format` expansion with `%%s` for the plurals.

asymmetric:
```suggestion
                          l (when (> l 1) "s"))))
```

Is this idiomatic Lisp? (and does it work?)

zaeph: `format` would replace `%s` with `nil` (i.e. literal `nil`) if we used `when`.  Empty-string is not the same as `nil` in that regard.

asymmetric: True, `nil` is formatted as `"nil"`.

PR #532, opened 2020-04-27
Reenable nested captures

### LINKED ISSUE #527: Support nested org-roam-capture
### Brief Abstract
We've disabled nested `org-roam-capture` because it'd be complicated to handle well.  However, from the user's perspective, running `org-roam-insert` inside an `org-roam-capture` buffer prompted by another `org-roam-insert` should be possible.

### Long Description
#327 brought up some problems with nested `org-roam-capture`, which were addressed in #329 by disabling nesting.  @jethrokuan mentioned [here](https://github.com/jethrokuan/org-roam/issues/327#issuecomment-601665965) that solving this issue would be complex, but I think this is necessary.

Reenable nested `org-roam-capture`, and create the infrastructure to have *parents* be aware of their *children*.
- When `org-roam-insert` calls `org-roam-capture` inside an `org-roam-capture` buffer (the parent), we make the *child* buffer aware of its *parent*, and the *parent* aware of its child.  This can be achieved with local variables in the two buffers.
- In the *parent* buffer, shadow all the commands that would result in the killing of the `org-roam-capture` buffer (success, abort, `kill-buffer`) as long as the *child* buffer exists.  The commands should now warn that a child-`org-roam-capture` is underway, and that it should be completed first.
- If a parent has multiple *children*, simply store them in a list, and check for the existence of all of them before killing the *parent* buffer.
- In the *child* buffer, on a successful capture, ensure that we insert the link inside the *parent* `org-roam-capture` buffer.  Point might be an issue, but we could create a marker in the *parent*, and have the *child* be aware of it.

We now have a hierarchy between the buffers that ensures that the links are inserted where they should, and that the captures aren't aborted or completed before we need them to be.

I'm going to start working on an implementation next week.  In the meantime, if you have any suggestion, I'm all ears!

zaeph:
Fixes #527.

This is a going to be a long-running PR to address the nesting of `org-roam-capture` processes.

*Some* work has been done on it already, but it's very alpha.  Check the commit-messages for clarifications.

I'm busy this week, but I'll resume work on this next weekend.

Small demo of the current state (click it because it's quite long, height-wise):
output-2020-04-27-00:48:47

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -152,6 +152,19 @@ buffer."
 		  (plist-get (buffer-local-value 'org-capture-current-plist (current-buffer)) :org-roam)))
 	   (buffer-list)))
 
+(define-minor-mode org-roam-capture-mode
+  "Minor mode for the `org-roam-captue'."
+  :lighter " orc"
+  ;; :keymap org-roam-capture-mode-map
+
+  (when (and org-roam-capture-mode
+             (eq (org-roam-capture--get :capture-fn)
+                 'org-roam-insert))
+    (add-hook 'org-capture-before-finalize-hook
+                #'org-roam-capture--insert-link-h nil 'local)))
+
+(add-hook 'org-capture-mode-hook #'org-roam-capture-mode)
+
 (defun org-roam-capture--fill-template (str &optional info)
   "Expand the template STR, returning the string.
 This is an extension of org-capture's template expansion.
@@ -180,7 +193,8 @@ This is added as a hook to `org-capture-after-finalize-hook'."
           (desc (org-roam-capture--get-local :link-description)))
       (org-with-point-at (org-roam-capture--get-local :insert-at)
         (insert (org-roam--format-link path desc)))))
-  (remove-hook 'org-capture-before-finalize-hook #'org-roam-capture--insert-link-h))
+  ;; (remove-hook 'org-capture-before-finalize-hook #'org-roam-capture--insert-link-h)
+  )
 
 (defun org-roam-capture--save-file-maybe-h ()
   "Save the file conditionally.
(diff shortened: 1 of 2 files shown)


org-roam-capture.el
```diff
@@ -113,28 +113,60 @@ applies.
...
+  (let ((p (org-roam-capture--prepare-plist stuff)))
     (setq org-capture-plist
           (plist-put org-capture-plist :org-roam p))))
 
+(defun org-roam-capture--put-local (&rest stuff)
+  "Put properties from STUFF into the `org-roam-capture-template'."
+  (let ((p (org-roam-capture--prepare-plist stuff)))
+    (setq org-capture-current-plist
+          (plist-put org-capture-current-plist :org-roam p))))
+
```
progfolio, line 147: Perhaps just combine these into a singular function that takes a `local` argument similar to `org-roam-capture--get`.

zaeph: I wanted to, but I think there was something that stopped me at the time.  I'll get back to that PR later next week, but if you want to have a go at it, feel free!

org-roam-capture.el
```diff
@@ -113,28 +113,60 @@ applies.
...
+When LOCAL is non-nil, get the local value for the current
+buffer."
+  (let ((p (if local
+               org-capture-current-plist
+             org-capture-plist)))
+    (plist-get (plist-get p :org-roam) keyword)))
+
+(defun org-roam-capture--get-local (keyword)
+  "Get the local value for KEYWORD from the `org-roam-capture-template'."
+  (org-roam-capture--get keyword t))
```
progfolio, line 127: Instead of a wrapper function, we could pass a descriptive symbol in calls to `org-roam-capture--get`.
e.g.
```emacs-lisp
(org-roam-capture--get :capture-fn 'local)
```

zaeph: Sounds good to me.

zaeph:
Since we're considering dropping `org-capture` in favour of something simpler that wouldn't require us to jump through as many hoops, this PR might become obsolete in the future.  I'm changing it to _draft_ because it's what it should have been from the start.

jethrokuan:
Superceded by #966

PR #542, opened 2020-04-30
graph: add an option of wrapping the title

myshevchuk:
###### Description

Add an option to wrap the title in the graphviz graph instead of truncating it.
The value of `org-roam-graph-max-title-length` will be used for wrapping

Things will look nicer and no information is lost!

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -99,7 +99,7 @@ Example:
 
 (defcustom org-roam-graph-short-titles t
   "When this is non-nil, truncate titles in graph nodes.
-The title will be trancated according to `org-roam-graph-max-title-length'.
+The title will be truncated according to `org-roam-graph-max-title-length'.
 
 This can also be set to symbol 'wrap, in which case long titles
 will be wrapped with `org-roam-graph-max-title-length' as the
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -97,17 +97,17 @@ Example:
   :type 'number
   :group 'org-roam)
 
-(defcustom org-roam-graph-short-titles t
-  "When this is non-nil, truncate titles in graph nodes.
-The title will be truncated according to `org-roam-graph-max-title-length'.
+(defcustom org-roam-graph-shorten-titles 'truncate
+  "Determines how long titles appear in graph nodes.
+Recognized values are the symbols `truncate' and `wrap', in which
+cases the title will be truncated or wrapped, respectively, if it
+is longer than `org-roam-graph-max-title-length'.
 
-This can also be set to symbol 'wrap, in which case long titles
-will be wrapped with `org-roam-graph-max-title-length' as the
-maximum line width."
+All other values including nil will have no effect."
   :type '(choice
-          (const :tag "Yes" t)
-          (const :tag "Wrap" wrap)
-          (const :tag "No" nil))
+          (const :tag "truncate" truncate)
+          (const :tag "wrap" wrap)
+          (const :tag "no" nil))
   :group 'org-roam)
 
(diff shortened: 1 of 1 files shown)


myshevchuk:
``` elisp
(setq org-roam-graph-max-title-length 50)
```

Screenshot 2020-04-30 at 11 55 22

org-roam-graph.el
```diff
@@ -97,6 +97,12 @@ Example:
   :type 'number
   :group 'org-roam)
 
+(defcustom org-roam-graph-wrap-title t
```
jethrokuan, line 100: We should make the variable something like `org-roam-graph-title-shortener`, and the values be `'truncate`, `'wrap`, nil, and later below do a pcase on this to apply the appropriate function.

myshevchuk: what should be the default?

jethrokuan: the original (truncate), to minimize surprise.

myshevchuk:
`org-roam-graph-short-titles` can be 
1. `nil`, in which case titles are exported as-is
2.  `'wrap`, in which case they will be wrapped
3. Anything else, in which case they will be truncated (default behaviour)

org-roam-graph.el
```diff
@@ -97,10 +97,17 @@ Example:
...
-  :type 'boolean
+(defcustom org-roam-graph-short-titles t
+  "When this is non-nil, truncate titles in graph nodes.
+The title will be trancated according to `org-roam-graph-max-title-length'.
+
+This can also be set to symbol 'wrap, in which case long titles
+will be wrapped with `org-roam-graph-max-title-length' as the
+maximum line width."
+  :type '(choice
+          (const :tag "Yes" t)
```
jethrokuan, line 108: let's deprecate `org-roam-graph-max-title-length`, and move that configuration into this `defcustom`. so the choices are:

1. A number. If number, then truncate to this length.
2. `'wrap`. If `wrap`, wrap it.
3. `nil`. If nil, return the title as is.

org-roam-graph.el
```diff
@@ -97,10 +97,17 @@ Example:
   :type 'number
   :group 'org-roam)
 
-(defcustom org-roam-graph-wrap-title t
-  "When this is non-nil, wrap the title instead of truncating it.
-The value of `org-roam-graph-max-title-length` is used for wrapping."
-  :type 'boolean
+(defcustom org-roam-graph-short-titles t
```
jethrokuan, line 100:
```suggestion
(defcustom org-roam-graph-shorten-titles 100
```

org-roam-graph.el
```diff
@@ -97,10 +97,17 @@ Example:
...
+  "When this is non-nil, truncate titles in graph nodes.
+The title will be trancated according to `org-roam-graph-max-title-length'.
```
jethrokuan: trancated -> truncated, docstring needs to be updated accordingly.

org-roam-graph.el
myshevchuk: The problem here is that `s-word-wrap` also requires a number. It of course can be hard-coded, which is I guess less desirable. Or should it be passed together with `'wrap` in a cons cell? Like hardcoded default if just `'wrap` is provided and customizable `('wrap . 70)` to set wrapping after a particular number of symbols?

jethrokuan: Ah I see, I stand corrected. Could you just change `'t` to `'truncate` then? It's a bit more explicit than a truth value.

myshevchuk:
`org-roam-graph-shorten-titles` can be

`'truncate`: truncate the title (default)
`'wrap`: wrap the title
Anything else: no effect

PR #543, opened 2020-04-30
(doc): Add key-binding for index in default configs

zaeph:
I've forgotten to add the new command for finding the index in the default key-bindings.  This will further help the discoverability of the command.

I'm not sure about the key used for the key-binding, though.  I opted for `j` because it's easy to access, and you can kind of remember it like `jump to index`.  If we bring this idea to its logical conclusion, we might want to rename the command to `org-roam-jump-to-index`.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -88,7 +88,7 @@ In Org-roam, you can define the path to your index file by setting `org-roam-ind
 (setq org-roam-index-file "index.org")
 ```
 
-You can then bind `org-roam-find-index` in your configuration to access it (see [Basic Install and
+You can then bind `org-roam-jump-to-index` in your configuration to access it (see [Basic Install and
 Configuration](installation.md/#basic-install-and-configuration) to review how
 to set key-bindings).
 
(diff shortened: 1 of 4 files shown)


zaeph:
> If we bring this idea to its logical conclusion, we might want to rename the command to `org-roam-jump-to-index`.

Went ahead and did it in e45539c00121167e79b0d12f307c6625b6f5b9cf.

jethrokuan:
> Went ahead and did it in [651445b](https://github.com/jethrokuan/org-roam/commit/651445b438b4a202b1559c8dc89929e91162879f).

You should then add a corresponding obsolete function declaration to `org-roam-compat.el`.

zaeph:
> > Went ahead and did it in [651445b](https://github.com/jethrokuan/org-roam/commit/651445b438b4a202b1559c8dc89929e91162879f).
> 
> You should then add a corresponding obsolete function declaration to `org-roam-compat.el`.

Done in 1d8ed0a7fc5759e13c318d5e3d81edb23855054b (rebased).

PR #545, opened 2020-05-01
Ensure that multicite links are correctly split

### LINKED ISSUE #544: Multiple org-ref citations are not recognised as multiple links
### Description
org-ref multi-citations are handled by `cite:key1,key2`. In the org-roam databases this is stored as a single "cite" link to the (non-existing) key "key1,key2". This means that cite backlinks and cite graph behaviour does not work correctly for these links.

This looks like it is caused in
https://github.com/jethrokuan/org-roam/blob/08aa5c630de354440e8eed21b28507f4b9e6ac6a/org-roam.el#L229 ... https://github.com/jethrokuan/org-roam/blob/08aa5c630de354440e8eed21b28507f4b9e6ac6a/org-roam.el#L267-L272 
Here we're extracting the path of the link element. For the example below this is "key1,key2". 

Instead we should be splitting this somehow and generating two separate links. This could either be done as we generate it or we could do a pass through the mapping afterwards to split up keys containing commas.

#### Steps to Reproduce
Source note 1:
```
#+TITLE: Source note 1
#+ROAM_KEY: key1
```
Source note 2:
```
#+TITLE: Source note 2
#+ROAM_KEY: key2
```
Referencing note:
```
#+TITLE: Referencing note

cite:key1,key2
```

#### Expected Results
Two separate "cite" links in the links table of the database: One pointing to "key1" and one pointing to "key2"

#### Actual Results
A single "cite" link pointing to "key1,key2"

### Versions
- Emacs (`C-h v emacs-version`): v28.0.50
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/08aa5c630de354440e8eed21b28507f4b9e6ac6a

TimQuelch:
org-ref multi-citations are handled by "cite:key1,key2". In the org-roam databases this was
previously stored as a single "cite" link to the (non-existing) key "key1,key2". This means that
cite backlinks and cite graph behaviour does not work correctly for these links.

This fix will split any cite links with a comma into separate links to each individual ref.

To be honest, I don't really  like how this patch changes the readability of `org-roam--extract-links`. Because I've basically wrapped the existing code in a `mapccan`, the code to split the multicite links now appears before the code to actually extract the links; in my mind these should be the other way around. In addition, properly indenting would cause a gross diff (I've currently left the original indentation). Not sure how to fix this with my limited emacs-lisp knowledge so suggestions are welcome.

Fixes #544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
 * [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
+* [#545](https://github.com/jethrokuan/org-roam/pull/545) fix `org-roam--extract-links` to ensure that multiple citations (`cite:key1,key2`) are split correctly
 
 ### Features
 * [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title 
--- a/org-roam.el
+++ b/org-roam.el
@@ -238,6 +238,14 @@ in temp buffers.  In cases where this occurs, we do know the file path, and pass
 it as FILE-PATH."
   (let ((file-path (or file-path
                        (file-truename (buffer-file-name)))))
+    (mapcan (lambda (linkvec)           ; Split any multi-cite citations
+              (if (string= (aref linkvec 2) "cite")
+                  (mapcar (lambda (ref) (vector (aref linkvec 0)
+                                                ref
+                                                (aref linkvec 2)
+                                                (aref linkvec 3)))
+                          (split-string (aref linkvec 1) ","))
+                (list linkvec)))
     (org-element-map (org-element-parse-buffer) 'link
       (lambda (link)
         (let* ((type (org-element-property :type link))
@@ -270,7 +278,7 @@ it as FILE-PATH."
                             ((string= link-type "cite")
                              path))
                       link-type
-                      (list :content content :point begin)))))))))
+                      (list :content content :point begin))))))))))
 
 (defcustom org-roam-title-include-subdirs nil
   "When non-nil, include subdirs in title completions.
--- a/org-roam.el
+++ b/org-roam.el
@@ -273,12 +273,12 @@ it as FILE-PATH."
                        link-type
                        (list :content content :point begin)))))))
      (mapcan (lambda (linkvec)          ; Split any multi-cite citations
-               (if (string= (aref linkvec 2) "cite")
+               (if (and (string= (aref linkvec 2) "cite") (require 'org-ref nil t))
                    (mapcar (lambda (ref) (vector (aref linkvec 0)
                                                  ref
                                                  (aref linkvec 2)
                                                  (aref linkvec 3)))
-                           (split-string (aref linkvec 1) ","))
+                           (org-ref-split-and-strip-string (aref linkvec 1)))
                  (list linkvec)))))))
 
 (defcustom org-roam-title-include-subdirs nil
--- a/org-roam.el
+++ b/org-roam.el
@@ -265,24 +265,19 @@ it as FILE-PATH."
                    (content (string-trim content))
                    ;; Expand all relative links to absolute links
                    (content (org-roam--expand-links content file-path)))
-              (let ((context (list :content content :point begin)))
-                (pcase link-type
-                  ("roam"
-                   (push (vector file-path
-                                 (file-truename (expand-file-name path (file-name-directory file-path)))
-                                 link-type
-                                 context)
-                         links))
-                  ("cite"
-                   (require 'org-ref nil t)
-                   ;; Separate citekeys and process them individually
-                   (seq-do (lambda (citekey)
-                             (push (vector file-path
-                                           citekey
-                                           link-type
-                                           context)
-                                   links))
-                           (org-ref-split-and-strip-string path))))))))))
+              (let ((context (list :content content :point begin))
+                    (names (pcase link-type
+                             ("roam"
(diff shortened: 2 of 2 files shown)


org-roam.el
```diff
@@ -238,6 +238,14 @@ in temp buffers.  In cases where this occurs, we do know the file path, and pass
 it as FILE-PATH."
   (let ((file-path (or file-path
                        (file-truename (buffer-file-name)))))
+    (mapcan (lambda (linkvec)           ; Split any multi-cite citations
+              (if (string= (aref linkvec 2) "cite")
+                  (mapcar (lambda (ref) (vector (aref linkvec 0)
+                                                ref
+                                                (aref linkvec 2)
+                                                (aref linkvec 3)))
+                          (split-string (aref linkvec 1) ","))
```
jethrokuan, line 247: `org-ref` provides a utility: `org-ref-split-and-strip-string`

```
(defun org-ref-split-and-strip-string (string)
  "Split key-string and strip keys in STRING.
Assumes the key-string is comma delimited."
  (mapcar 'org-ref-strip-string (split-string string ",")))
```

which seems to do more robust replacements. Might be good to use this function instead.

zaeph:
To avoid the long argument in `mapcan`, I've pushed a slight refactoring with `->>`.  Make sure to pull!

org-roam.el
zaeph: Uhm, I'm not sure if it's behaving as it should, though.
```el
(org-ref-split-and-strip-string "cite:ahrens2017,landy2012")
;; Result: ("cite:ahrens2017" "landy2012")
```

zaeph: Ah wait, maybe the `cite:` part is stripped before.

TimQuelch: Sounds good, I'll swap it to use that function

> Ah wait, maybe the `cite:` part is stripped before.

Yep this is stripped out earlier

TimQuelch:
> To avoid the long argument in `mapcan`, I've pushed a slight refactoring with `->>`. Make sure to pull!

Oh nice, that's exactly what I wanted. I didn't know about the dash library, it looks very useful

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
 * [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
+* [#545](https://github.com/jethrokuan/org-roam/pull/545) fix `org-roam--extract-links` to ensure that multiple citations (`cite:key1,key2`) are split correctly
```
zaeph, line 12: Way more diligence than I showed in of my PR right there. 🤡

Thanks!

org-roam.el
```diff
@@ -238,39 +238,48 @@ in temp buffers.  In cases where this occurs, we do know the file path, and pass
...
+                                  begin
+                                  (or (org-element-property :content-end element)
+                                      (org-element-property :end element)))))
+                    (content (string-trim content))
+                    ;; Expand all relative links to absolute links
+                    (content (org-roam--expand-links content file-path)))
+               (vector file-path
+                       (cond ((string= link-type "roam")
+                              (file-truename (expand-file-name path (file-name-directory file-path))))
+                             ((string= link-type "cite")
```
zaeph, line 271: I'm wondering if we couldn't distribute the `cite` at this point in the code.  That would prevent us from running an extra cond-check on every entries to check if they're a `cite` link.

TimQuelch: I did think about that, but I'm not sure how that would work. `org-element-map` acts like a `mapcar`, so one element can't be transformed into multiple elements (as far as I'm aware)

TimQuelch: I do agree though it would be ideal to do it there

TimQuelch: I guess we could have `org-element-map` map to a list of lists and then flatten that?

zaeph: > I did think about that, but I'm not sure how that would work. `org-element-map` acts like a `mapcar`, so one element can't be transformed into multiple elements (as far as I'm aware)

I think we might be able to get away with it with some Elisp Magic™✨, i.e., probably `pcase` and `setf`.  I'm working on it.

zaeph: Fixed in 734da3d250a965b5e0f6f33a0c6699c6aa67c19b.  Found a better way to do it.

org-roam.el
```diff
@@ -264,13 +265,25 @@ it as FILE-PATH."
...
+                   (push (vector file-path
+                                 (file-truename (expand-file-name path (file-name-directory file-path)))
+                                 link-type
+                                 context)
+                         links))
+                  ("cite"
+                   (require 'org-ref nil t)
+                   ;; Separate citekeys and process them individually
+                   (seq-do (lambda (citekey)
+                             (push (vector file-path
```
zaeph: Some repetition can be avoided here.  Working on it. 🔧

zaeph: Fixed in 10467d728d6f4489cba0f91375acbdf5f9a36fc5.

org-roam.el
TimQuelch: Nice. Looks good

zaeph:
Welp, the code might be more readable, and I'd certainly deserve some golf-points for it, it isn't that much optimised:
```el
;; Original
(with-selected-window (next-window)
  (benchmark 100 '(org-roam--extract-links)))
;; "Elapsed time: 5.322224s (1.574320s in 17 GCs)"

;; Optimised...?
(with-selected-window (next-window)
  (benchmark 100 '(org-roam--extract-links)))
;; "Elapsed time: 5.352579s (1.565723s in 17 GCs)"
```

I guess it'd be better if we could get `org-element-map` to not accumulate the result, but unless we send a patch upstream (i.e. Org-mode) to get it to accept a new argument, it's not going to happen.

Oh well. ¯\\\_(ツ)\_/¯

zaeph:
Thanks for your work on this!

Issue #526, opened 2020-04-26
Add an interactive Org-roam tutorial


jethrokuan:
### Brief Abstract

Add an interactive org-roam tutorial that's accessible using a key-binding.

### Long Description

Org-roam is growing increasingly complex, and I think we really need to work on the onboarding experience. I propose the creation of a tutorial (similar to the tour) that's accessible within Emacs.

The tutorial should link to relevant Info nodes: we also need to make the documentation available within Emacs.

Checklist:
- [ ] Cover the basic interactive commands
- [ ] Check that it works in a vanilla Emacs installation with only org-roam

### Please check the following:
- [x] No similar feature requests

PR #546, opened 2020-05-01
(internal): use Make and Org to build manual

jethrokuan:

This is partly motivated by #526 . We use Makefiles inspired by Magit and use-package, writing our manuals in Org, to be converted into pdf, html, texi, info and epub.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,15 @@
 /.sandbox/
 **/*.elc
+/doc/dir
+/doc/*.info
+/doc/*.html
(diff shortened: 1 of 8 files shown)


jethrokuan:
`AUTHORS.md` is auto-generated, and uses the email you use to commit with Git.

jethrokuan:
I think the manual is in alright shape now, but still not in ideal condition. It's still useful for those who want to refer to the manual within Emacs, although incomplete. Shall merge this in first, and then think about next steps.

PR #547, opened 2020-05-01
Adding key column to db

goktug97:
Refer https://github.com/jethrokuan/org-roam/issues/540
The key column is added to the refs table to be able to properly link references in the graph view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.1.2 
+
+### Bugfixes
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Fix unlinked references by changing db refs table structure.
+
 ## 1.1.1 (TBD)
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Features
 * [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title
 * [#553](https://github.com/jethrokuan/org-roam/pull/553) Add prefix argument to `org-roam-db-build-cache` for forcing rebuilds
+* [#560](https://github.com/jethrokuan/org-roam/pull/560) Apply 'error face to distinguish broken links
 
 ## 1.1.0 (21-04-2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,18 @@
 # Changelog
 
-## 1.1.2 
-
-### Bugfixes
-* [#547](https://github.com/jethrokuan/org-roam/pull/547) Fix unlinked references by changing db refs table structure.
-
 ## 1.1.1 (TBD)
 
 ### Breaking Changes
 * [#523](https://github.com/jethrokuan/org-roam/pull/523) remove `org-roam-completion-fuzzy-match` in favor of using completion mechanism's configuration options directly 
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Deprecate `org-roam-db--maybe-update`, in favour of `org-roam-db--update-maybe`.
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Upgrade the database version from 2 to 3.
 
 ### Bugfixes
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
 * [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
 * [#545](https://github.com/jethrokuan/org-roam/pull/545) fix `org-roam--extract-links` to ensure that multiple citations (`cite:key1,key2`) are split correctly
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Fix unlinked citations.
 
 ### Features
 * [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Bugfixes
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
+* [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
 
 ## 1.1.0 (21-04-2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
 * [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
 
+### Features
+* [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title 
+
 ## 1.1.0 (21-04-2020)
 
 To the average user, this release is mainly a bugfix release with additional options to customize. However, the changes made to the source is significant. Most notably, in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix external org files being tracked in database
 * [#537](https://github.com/jethrokuan/org-roam/pull/537) quote graphviz node and edge configuration options to allow multi-word configurations
+* [#545](https://github.com/jethrokuan/org-roam/pull/545) fix `org-roam--extract-links` to ensure that multiple citations (`cite:key1,key2`) are split correctly
 
 ### Features
 * [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Breaking Changes
 * [#523](https://github.com/jethrokuan/org-roam/pull/523) remove `org-roam-completion-fuzzy-match` in favor of using completion mechanism's configuration options directly 
 * [#547](https://github.com/jethrokuan/org-roam/pull/547) Deprecate `org-roam-db--maybe-update`, in favour of `org-roam-db--update-maybe`.
-* [#547](https://github.com/jethrokuan/org-roam/pull/547) Upgrade the database version from 2 to 3.
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Added `type` column to the `refs` table.
 
 ### Bugfixes
 * [#509](https://github.com/jethrokuan/org-roam/pull/509) fix backup files being tracked in database
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@
 * [#545](https://github.com/jethrokuan/org-roam/pull/545) fix `org-roam--extract-links` to ensure that multiple citations (`cite:key1,key2`) are split correctly
 
 ### Features
-* [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title 
+* [#538](https://github.com/jethrokuan/org-roam/pull/538) Optionally use text in first headline as title
+* [#553](https://github.com/jethrokuan/org-roam/pull/553) Add prefix argument to `org-roam-db-build-cache` for forcing rebuilds
 
 ## 1.1.0 (21-04-2020)
 
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -257,7 +257,7 @@ If the file does not have any connections, nil is returned."
                    links_of(file, link) AS
                      (WITH roamlinks AS (SELECT * FROM links WHERE \"type\" = '\"roam\"'),
                            citelinks AS (SELECT * FROM links
-                                                  JOIN refs ON links.\"to\" = refs.\"ref\"
+                                                  JOIN refs ON links.\"to\" = refs.\"key\"
                                                             AND links.\"type\" = '\"cite\"')
                       SELECT \"from\", \"to\" FROM roamlinks UNION
                       SELECT \"to\", \"from\" FROM roamlinks UNION
@@ -278,7 +278,7 @@ including the file itself.  If the file does not have any connections, nil is re
                    links_of(file, link) AS
                      (WITH roamlinks AS (SELECT * FROM links WHERE \"type\" = '\"roam\"'),
                            citelinks AS (SELECT * FROM links
-                                                  JOIN refs ON links.\"to\" = refs.\"ref\"
+                                                  JOIN refs ON links.\"to\" = refs.\"key\"
                                                             AND links.\"type\" = '\"cite\"')
                       SELECT \"from\", \"to\" FROM roamlinks UNION
                       SELECT \"to\", \"from\" FROM roamlinks UNION
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -43,6 +43,8 @@
 (declare-function org-roam--extract-ref         "org-roam")
 (declare-function org-roam--extract-links       "org-roam")
 (declare-function org-roam--list-files          "org-roam")
+(declare-function org-roam--cite-prefix         "org-roam")
+(declare-function org-roam--ref-type            "org-roam")
 (declare-function org-roam-buffer--update-maybe "org-roam-buffer")
 
 ;;;; Options
@@ -226,23 +228,10 @@ This is equivalent to removing the node from the graph."
     :values $v1]
    (list (vector file titles))))
 
-(defun org-roam-db--ref-type (ref)
-  (let* ((cite-prefix (seq-find
-                       (lambda (prefix) (s-prefix? prefix ref))
-                       (-map (lambda (type) (concat type ":"))
-                             org-ref-cite-types)))
-         (is-website (seq-some
-                      (lambda (prefix) (s-prefix? prefix ref))
-                      '("http" "https")))
-         (type (cond (cite-prefix "cite")
-                     (is-website "website")
-                     (t "roam"))))
-    type))
-
 (defun org-roam-db--insert-ref (file ref)
   "Insert REF for FILE into the Org-roam cache."
-  (let* ((type org-roam-db--ref-type ref)
-         (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
+  (let* ((type org-roam--ref-type ref)
+         (key (cond ((string= "cite" type) (s-chop-prefix (org-roam--cite-prefix ref) ref))
                     (t ref))))
     (org-roam-db-query
      [:insert :into refs :values $v1]
@@ -381,8 +370,8 @@ including the file itself.  If the file does not have any connections, nil is re
             (let ((titles (org-roam--extract-and-format-titles file)))
               (setq all-titles (cons (vector file titles) all-titles)))
             (when-let* ((ref (org-roam--extract-ref))
-                        (type (org-roam-db--ref-type ref))
-                        (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
+                        (type (org-roam--ref-type ref))
+                        (key (cond ((string= "cite" type) (s-chop-prefix (org-roam--cite-prefix ref) ref))
                                    (t ref))))
               (setq all-refs (cons (vector key file type) all-refs))))
           (remhash file current-files))))
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -304,11 +304,12 @@ including the file itself.  If the file does not have any connections, nil is re
 ;;;;; Updating
 (defun org-roam-db--update-titles ()
   "Update the title of the current buffer into the cache."
-  (let ((file (file-truename (buffer-file-name))))
+  (let* ((file (file-truename (buffer-file-name)))
+         (title (org-roam--extract-and-format-titles file)))
     (org-roam-db-query [:delete :from titles
                         :where (= file $s1)]
                        file)
-    (org-roam-db--insert-titles file (org-roam--extract-titles))))
+    (org-roam-db--insert-titles file title)))
 
 (defun org-roam-db--update-refs ()
   "Update the ref of the current buffer into the cache."
@@ -361,7 +362,7 @@ including the file itself.  If the file does not have any connections, nil is re
                   (cons (vector file contents-hash time) all-files))
             (when-let (links (org-roam--extract-links file))
               (setq all-links (append links all-links)))
-            (let ((titles (org-roam--extract-titles)))
+            (let ((titles (org-roam--extract-and-format-titles file)))
               (setq all-titles (cons (vector file titles) all-titles)))
             (when-let ((ref (org-roam--extract-ref)))
               (setq all-refs (cons (vector ref file) all-refs))))
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -55,7 +55,7 @@ when used with multiple Org-roam instances."
   :type 'string
   :group 'org-roam)
 
-(defconst org-roam-db--version 2)
+(defconst org-roam-db--version 3)
 (defconst org-roam-db--sqlite-available-p
   (with-demoted-errors "Org-roam initialization: %S"
     (emacsql-sqlite-ensure-binary)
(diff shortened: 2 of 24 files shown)


org-roam-db.el
```diff
@@ -227,10 +228,11 @@ This is equivalent to removing the node from the graph."
 
 (defun org-roam-db--insert-ref (file ref)
   "Insert REF for FILE into the Org-roam cache."
-  (org-roam-db-query
-   [:insert :into refs
-    :values $v1]
-   (list (vector ref file))))
+  (let ((key (elt (split-string ref ":") 1)))
```
TimQuelch, line 231: Won't this run into trouble with other types of links in the ROAM_KEY? e.g. `https://example.com` will get stripped to `//example.com`

Maybe we should check if the prefix is in `org-ref-cite-types` before splitting, and then strip off that prefix. Otherwise leave the key untouched.

Another issue this might run into is if a citekey has a ':' in it, this will strip off the stuff afterwards (after a quick search it appears that ':' is allowed in a bibtex citekey). e.g. in a pathological case: `cite:key:withcolon` would get  stripped to `key` instead of `key:withcolon`, or `key:withcolon` would get stripped to just `withcolon`.

org-roam-db.el
```diff
@@ -364,8 +366,9 @@ including the file itself.  If the file does not have any connections, nil is re
               (setq all-links (append links all-links)))
             (let ((titles (org-roam--extract-and-format-titles file)))
               (setq all-titles (cons (vector file titles) all-titles)))
-            (when-let ((ref (org-roam--extract-ref)))
-              (setq all-refs (cons (vector ref file) all-refs))))
+            (when-let* ((ref (org-roam--extract-ref))
+                        (key (elt (split-string ref ":") 1)))
```
TimQuelch, line 370: Same issue as above

org-roam-db.el
TimQuelch:
```suggestion
  (let ((key ref))
      (dolist (prefix org-ref-cite-types)
        (setq key (string-remove-prefix (concat prefix ":") key)))
```

org-roam-db.el
```diff
@@ -133,7 +133,8 @@ SQL can be either the emacsql vector representation, or a string."
 
     (refs
      [(ref :unique :not-null)
-      (file :not-null)])))
+      (file :not-null)
+      (key :not-null)])))
```
jethrokuan, line 137: Let's call this `type` instead, same as the `links` table. Key will then take on values:

1. `cite` if the prefix is an org-ref cite link (as per `org-ref-cite-types`)
2. `website` if it's a URL (optional, nothing is being done with this)
3. `roam' otherwise.

TimQuelch: Wouldn't this break setups where the `ROAM_KEY` is just the citekey?

e.g. `ROAM_KEY: ahrens2017` would be detected as a roam type rather than a cite type.

having the `key` (or whatever column name is chosen) just be the original `ROAM_KEY` stripped of any `org-ref-cite-types` prefixes should allow both `cite:ahrens2017` and `ahrens2017` to work

zaeph: I think it's fine to keep the `cite:` in `ROAM_KEY` as a way to identify bibliographic notes.  It comes at very little cost to us, and I agree with @jethrokuan that it helps with entry if you're not using `org-roam-bibtex` to generate those files.

This stems from the desire to have different types of keys co-exists with just one file-metadata tag.  Otherwise, another option which wouldn't be as elegant would be to have different file-metadata tags like `ROAM_CITEKEY`, `ROAM_URLKEY`, etc.

TimQuelch: Fair enough. I can see the logic for future extendability. `ROAM_CITEKEY` and `ROAM_URLKEY` does not sound like a nice solution IMO.

I might submit a quick PR to fix the backlinks buffer not catering for all cite types

goktug97:
> Let's call this `type` instead, same as the `links` table. Key will then take on values:
> 
>     1. `cite` if the prefix is an org-ref cite link (as per `org-ref-cite-types`)
> 
>     2. `website` if it's a URL (optional, nothing is being done with this)
> 
>     3. `roam' otherwise.

Maybe something like this?

(defun org-roam-db--insert-ref (file ref)
  "Insert REF for FILE into the Org-roam cache."
  (let* ((url (url-generic-parse-url ref))
         (type (url-type url)))
    (cond ((cl-find type org-ref-cite-types :test #'string=)
           (setq ref (string-remove-prefix (concat type ":")))
           (setq type "cite"))
          ((cl-member type '("http" "https") :test #'string=)
           (setq type "website"))
          (t
           (setq type "roam")))
    (org-roam-db-query
     [:insert :into refs :values $v1]
     (list (vector ref file type)))))

bdarcus:
I'm not following the details here, but did just want to mention somewhere that native org citation support is finally moving forward, and I expect to be merged to master "soon."

TimQuelch:
> Maybe something like this?

I don't think that `cl-find` will work correctly.. `parencite:key` will match to `cite`. You'll then try to remove the prefix `cite:` which does not exist. Therefore `ref` will remain `parencite:key`

TimQuelch:
@goktug97 how about something like this?
```
(defun org-roam-db--insert-ref (file ref)
  "Insert REF for FILE into the Org-roam cache."
  (let* ((cite-prefix (seq-find
                       (lambda (prefix) (s-prefix? ref))
                       (-map (lambda (type) (concat type ":"))
                             org-ref-cite-types)))
         (is-website (seq-find
                      (lambda (prefix) (s-prefix? ref))
                      '("http" "https")))
         (type (cond (cite-prefix "cite")
                     (is-website "website")
                     (t "roam")))
         (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
                    (t ref))))
    (org-roam-db-query
     [:insert :into refs :values $v1]
     (list (vector key file type)))))
```

goktug97:
> @goktug97 how about something like this? (untested)
> 
> ```emacs-lisp
> (defun org-roam-db--insert-ref (file ref)
>   "Insert REF for FILE into the Org-roam cache."
>   (let* ((cite-prefix (seq-find
>                        (lambda (prefix) (s-prefix? ref))
>                        (-map (lambda (type) (concat type ":"))
>                              org-ref-cite-types)))
>          (is-website (seq-some
>                       (lambda (prefix) (s-prefix? ref))
>                       '("http" "https")))
>          (type (cond (cite-prefix "cite")
>                      (is-website "website")
>                      (t "roam")))
>          (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
>                     (t ref))))
>     (org-roam-db-query
>      [:insert :into refs :values $v1]
>      (list (vector key file type)))))
> ```

It works perfectly but prefix was missing in `s-prefix?` so I added that.
```elisp
(defun org-roam-db--insert-ref (file ref)
  "Insert REF for FILE into the Org-roam cache."
  (let* ((cite-prefix (seq-find
                       (lambda (prefix) (s-prefix? prefix ref))
                       (-map (lambda (type) (concat type ":"))
                             org-ref-cite-types)))
         (is-website (seq-some
                      (lambda (prefix) (s-prefix? prefix ref))
                      '("http" "https")))
         (type (cond (cite-prefix "cite")
                     (is-website "website")
                     (t "roam")))
         (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
                    (t ref))))
    (org-roam-db-query
     [:insert :into refs :values $v1]
     (list (vector key file type)))))
```

To test it;
```elisp
(dolist (ref '("cite:test" "https://www.github.com" "file:///home/user"))
 (let* ((cite-prefix (seq-find
                      (lambda (prefix) (s-prefix? prefix ref))
                      (-map (lambda (type) (concat type ":"))
                            org-ref-cite-types)))
        (is-website (seq-some
                     (lambda (prefix) (s-prefix? prefix ref))
                     '("http" "https")))
        (type (cond (cite-prefix "cite")
                    (is-website "website")
                    (t "roam")))
        (key (cond ((string= "cite" type) (s-chop-prefix cite-prefix ref))
                   (t ref))))
   (print type)))
```
and it prints
```
"cite"

"website"

"roam"
nil
```

TimQuelch:
> It works perfectly but prefix was missing in `s-prefix?` so I added that.

oops my bad

goktug97:
Okay if we are choosing this route what are we going to do here.
https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam-graph.el#L175-L176

goktug97:
I added preview functionality to the `org-roam-graph-server`,
as soon as we are done with this PR, I will push the changes to the repo and open a pull request for the MELPA
image

goktug97:
The latest commits are not working. I thought they are working but after resetting the emacs I realized the new functions weren't properly loaded in the previous session.

goktug97:
I don't know how MELPA works but should we also update [this](https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam.el#L8) to 1.1.2

zaeph:
> I don't know how MELPA works but should we also update [this](https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam.el#L8) to 1.1.2

I don't think this plays a rôle in MELPA, but yes, this is something that we should update.

goktug97:
I should have appended the changelog to 1.1.1 instead of creating a new version? If that's true, do you recommend `ammend` or a new commit?

zaeph:
> I should have appended the changelog to 1.1.1 instead of creating a new version? If that's true, do you recommend `ammend` or a new commit?

Don't worry about it: we'll figure out when we merge.

zaeph:
But thanks for all the work you've put into this, it's looking great, especially the screenshot you've linked.  I've been busy with `org-roam-bibtex`, so I haven't had a chance to pore over your code, but this looks solid.

goktug97:
> But thanks for all the work you've put into this, it's looking great, especially the screenshot you've linked. I've been busy with org-roam-bibtex, so I haven't had a chance to pore over your code quite yet, but this looks solid.

Thank you too!

TimQuelch:
> Okay if we are choosing this route what are we going to do here.
> https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam-graph.el#L175-L176

I think this now needs to have an added clause in that `and` for `(= refs:type "cite")`

org-roam.el
```diff
@@ -374,6 +374,26 @@ current buffer is used."
...
+         (is-website (seq-some
+                      (lambda (prefix) (s-prefix? prefix ref))
+                      '("http" "https")))
+         (type (cond (cite-prefix "cite")
+                     (is-website "website")
+                     (t "roam"))))
+    type))
+
+(defun org-roam--cite-prefix (ref)
+  "Return citation prefix of REF if belongs to `org-ref-cite-types`."
```
TimQuelch, line 390:
```suggestion
  "Return the citation prefix of REF (e.g. \"cite:\", \"parencite:\" etc.) if it has a prefix of
  one of the `org-ref-cite-types`, return `nil` otherwise. This can be used to determine if a ref
  should be a \"cite\" type"
```
A little bit more verbose for the docstring

org-roam.el
```diff
@@ -374,6 +374,26 @@ current buffer is used."
...
+(defun org-roam--ref-type (ref)
+  "Determine the type of the REF from the prefix."
+  (let* ((cite-prefix (org-roam--cite-prefix ref))
+         (is-website (seq-some
+                      (lambda (prefix) (s-prefix? prefix ref))
+                      '("http" "https")))
+         (type (cond (cite-prefix "cite")
+                     (is-website "website")
+                     (t "roam"))))
+    type))
```
TimQuelch, line 387: I wonder if now `org-roam--extract-ref` should return the processed "ref" as well as the "type". I can't think of too many reasons we will want the raw `ROAM_KEY` anymore. 

Merging it might reduce the number of `let*` and `when-let*` required when using the ref

jethrokuan: Yeah, it might be more efficient to instead write a function that takes in the roam key, and returns a cons cell containing `(type processed-ref)`

zaeph: I agree, that'd be useful for checking the type.  That would allow for a convenient predicate:
```el
(defun org-roam--ref-type-p (type)
  "Return t if the ref from current buffer is TYPE."
  (let ((current (car (org-roam--extract-ref))))
    (eq current type )))
```

goktug97:
> > Okay if we are choosing this route what are we going to do here.
> > https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam-graph.el#L175-L176
> 
> I think this now needs to have an added clause in that `and` for `(= refs:type "cite")`

I tried it and unfortunately, it makes edges-cites `nil`

> I agree, that'd be useful for checking the type. That would allow for a convenient predicate:
> 
> ```emacs-lisp
> (defun org-roam--ref-type-p (type)
>   "Return t if the ref from current buffer is TYPE."
>   (let ((current (car (org-roam--extract-ref))))
>     (eq current type)))
> ```

Added.

Made all the suggested changes and tested, it works.

zaeph:
> Made all the suggested changes and tested, it works.

Thanks for that.

I've checked out the PR, cleared the db, tried to rebuilt it, and here's what I got:
> EmacSQL had an unhandled condition: "table refs has 2 columns but 3 values were supplied"

Can you identify the problem?

org-roam.el
goktug97: > > Made all the suggested changes and tested, it works.
> 
> Thanks for that.
> 
> I've checked out the PR, cleared the db, tried to rebuilt it, and here's what I got:
> 
> > EmacSQL had an unhandled condition: "table refs has 2 columns but 3 values were supplied"
> 
> Can you identify the problem?

I fixed by removing the .db file

TimQuelch:
> I've checked out the PR, cleared the db, tried to rebuilt it, and here's what I got:
> 
> > EmacSQL had an unhandled condition: "table refs has 2 columns but 3 values were supplied"
> 
> Can you identify the problem?

@zaeph I got that too, I manually deleted org-roam.db and rebuilt and that fixed it



> > I think this now needs to have an added clause in that `and` for `(= refs:type "cite")`
> 
> I tried it and unfortunately, it makes edges-cites `nil`
> 
> >

@goktug97 I added it in mine and it seems to be working?
```emacs-lisp
           (edges-cites-query
            `[:with selected :as [:select [file] :from ,node-query]
                    :select :distinct [file from]
                    :from links :inner :join refs :on (and (= links:to refs:ref)
                                                           (= links:type "cite")
                                                           (= refs:type "cite"))
                    :where (and (in file selected) (in from selected))])
```

goktug97:
> > Made all the suggested changes and tested, it works.
> 
> Thanks for that.
> 
> I've checked out the PR, cleared the db, tried to rebuilt it, and here's what I got:
> 
> > EmacSQL had an unhandled condition: "table refs has 2 columns but 3 values were supplied"
> 
> Can you identify the problem?

I fixed it by removing the .db file.
 Maybe instead of clearing the database, it can be deleted directly?
https://github.com/jethrokuan/org-roam/blob/e698ed7f5378106da8a8fec4537658392157657c/org-roam-db.el#L184-L191

zaeph:
> @zaeph I got that too, I manually deleted org-roam.db and rebuilt and that fixed it

That did it, thanks.  It's weird that resetting wasn't enough, though.  It might cause problems for the merge for people with existing configs.

zaeph:
> Maybe instead of clearing the database, it can be deleted directly?

We could add a `force` argument that would delete the file rather than clearing it, although I'm not sure if there's a particular reason why we're keeping the tables whilst clearing.  Any idea @jethrokuan?

goktug97:
> @goktug97 I added it in mine and it seems to be working?
> 
> ```emacs-lisp
>            (edges-cites-query
>             `[:with selected :as [:select [file] :from ,node-query]
>                     :select :distinct [file from]
>                     :from links :inner :join refs :on (and (= links:to refs:ref)
>                                                            (= links:type "cite")
>                                                            (= refs:type "cite"))
>                     :where (and (in file selected) (in from selected))])
> ```

I've tried it again and it works for me too now. I don't know why it didn't at the beginning

org-roam.el
```diff
@@ -371,9 +371,19 @@ current buffer is used."
...
 (defun org-roam--extract-ref ()
-  "Extract the ref from current buffer."
-  (cdr (assoc "ROAM_KEY" (org-roam--extract-global-props '("ROAM_KEY")))))
+  "Extract the ref from current buffer and return the type and the key of the ref."
+  (let* ((ref (cdr (assoc "ROAM_KEY" (org-roam--extract-global-props '("ROAM_KEY")))))
+         (type (org-roam--ref-type ref))
+         (key (cond ((string= "cite" type)
+                     (s-chop-prefix (org-roam--cite-prefix ref) ref))
+                    (t ref))))
+    (cons type key)))
```
TimQuelch, line 386:
```suggestion
  "Extract the ref from current buffer and return the type and the key of the ref."
  (if-let ((ref (cdr (assoc "ROAM_KEY" (org-roam--extract-global-props '("ROAM_KEY"))))))
      (let ((type (org-roam--ref-type ref))
            (key (cond ((string= "cite" type)
                        (s-chop-prefix (org-roam--cite-prefix ref) ref))
                       (t ref))))
        (cons type key))
    nil))
```

goktug97:
I think we should enforce it somehow. So that it will automatically rebuild the database if there is a breaking change in it.

org-roam.el
goktug97: https://github.com/jethrokuan/org-roam/pull/547/commits/ade062d5de05f642d1fa712cf6d51960d9809bac

jethrokuan:
> I think we should enforce it somehow. So that it will automatically rebuild the database if there is a breaking change in it.

See `org-roam-db`. You'll need to bump the version at `(org-roam-db--version)` to `3`, and add a handler at  `org-roam-db--maybe-update`, something like:

```
(when (= version 2)
      (progn
      ;; Add column to refs table  
      ))
```

goktug97:
>  (emacsql db [:alter-table refs   :add-column type :not-null])

Can't we just do the same thing ?
https://github.com/jethrokuan/org-roam/blob/4cad2cf6e6ec0617001f216ca79ce9cf649717c2/org-roam-db.el#L152-L153

zaeph:
> Can't we just do the same thing?

We have to make sure that people who used an older version of the database rebuild their files from scratch, which is what the the version of the db is about.

goktug97:
```elisp
(defun org-roam-db--maybe-update (db version)
  "Upgrades the database schema for DB, if VERSION is old."
  (emacsql-with-transaction db
    'ignore
    (if (< version org-roam-db--version)
        (progn
          (message (format "Upgrading the Org-roam database from version %d to version %d"
                        version org-roam-db--version))
          (delete-file (org-roam-db--get))
          (org-roam-db-build-cache))))
  version)
```

jethrokuan:
> ```elisp
> (defun org-roam-db--maybe-update (db version)
>   "Upgrades the database schema for DB, if VERSION is old."
>   (emacsql-with-transaction db
>     'ignore
>     (if (< version org-roam-db--version)
>         (progn
>           (message (format "Upgrading the Org-roam database from version %d to version %d"
>                         version org-roam-db--version))
>           (delete-file (org-roam-db--get))
>           (org-roam-db-build-cache))))
>   version)
> ```

That works, but is the lazy way I guess. Now you can replace that with `(org-roam-db-build-cache t)`

zaeph:
Should that go in `db` or in `compat`, actually?

goktug97:
When the database is upgraded from an old version org-roam-buffer doesn't work.
Either org-roam-mode must be disabled and reenabled again or emacs must be restarted for it to work again.

goktug97:
Also can we https://github.com/jethrokuan/org-roam/blob/69d4f05a131921d7ebb1e79cd7cf0eac61f030df/org-roam-db.el#L145 make this `org-roam-db--update-maybe` like https://github.com/jethrokuan/org-roam/blob/69d4f05a131921d7ebb1e79cd7cf0eac61f030df/org-roam-buffer.el#L197

zaeph:
Yes, let's standardise this.

CHANGELOG.md
```diff
@@ -4,12 +4,15 @@
 
 ### Breaking Changes
 * [#523](https://github.com/jethrokuan/org-roam/pull/523) remove `org-roam-completion-fuzzy-match` in favor of using completion mechanism's configuration options directly 
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Deprecate `org-roam-db--maybe-update`, in favour of `org-roam-db--update-maybe`.
+* [#547](https://github.com/jethrokuan/org-roam/pull/547) Upgrade the database version from 2 to 3.
```
jethrokuan, line 8: Can we reword this to specify exactly what changed:

"Added `type` column to the `refs` table"

goktug97:
> When the database is upgraded from an old version org-roam-buffer doesn't work.
> Either org-roam-mode must be disabled and reenabled again or emacs must be restarted for it to work again.

https://github.com/jethrokuan/org-roam/pull/547/commits/4fc87ae075a306342d52e895e7f72d9091aff811

Updating the database after the hook initialization fixed the problem.

zaeph:
LGTM.

I'm wondering if we should add a `defcustom` to `org-roam--get-ref-path-completions` which, when `t`, would replicate the old behaviour, i.e. formatting the `car` as `type:key`.

zaeph:
> I'm wondering if we should add a `defcustom` acting on `org-roam--get-ref-path-completions` which, when `t`, would replicate the old behaviour, i.e. formatting the `car` as `type:key`.

Obviously, we'd only do that for interactive calls.  As of now, we only have `org-roam-find-ref` making use of `org-roam--get-ref-path-completions`, but I can see other function/commands making use of it.

I'm working on something.

zaeph:
> I'm working on something.

I can't push to your branch.  Could you allow maintainers to do so?  It should be a setting for you in the PR.

zaeph:
Alternatively, I've opened a PR on your fork: goktug97/org-roam/pull/1.

[Here's the diff.](https://github.com/goktug97/org-roam/pull/1/files)

goktug97:
`Allow edits by maintainers` was on but I merged the PR

zaeph:
Thanks.  I think we're good for the merge.

org-roam.el
```diff
@@ -375,9 +375,41 @@ current buffer is used."
...
+         (is-website (seq-some
+                      (lambda (prefix) (s-prefix? prefix ref))
+                      '("http" "https")))
+         (type (cond (cite-prefix "cite")
+                     (is-website "website")
+                     (t "roam"))))
+    type))
+
+(defun org-roam--cite-prefix (ref)
+  "Return the citation prefix of REF (e.g. \"cite:\",\"parencite:\" etc.) if it has a prefix of
```
jethrokuan, line 405: The docstring needs to be tweaked to suit checkdoc requirements.

1. There's a faux indentation somewhere on this line.
2. The docstring needs start with a full sentence and REF must appear somwhere in it. Descriptions can come in later sentences.

org-roam.el
```diff
@@ -402,13 +402,12 @@ current buffer is used."
...
 
 (defun org-roam--cite-prefix (ref)
-  "Return the citation prefix of REF (e.g. \"cite:\",\"parencite:\" etc.) if it has a prefix of
-  one of the `org-ref-cite-types`, return `nil` otherwise. This can be used to determine if a ref
-  should be a \"cite\" type."
-  (seq-find
-   (lambda (prefix) (s-prefix? prefix ref))
-   (-map (lambda (type) (concat type ":"))
-         org-ref-cite-types)))
+  "Return the citation prefix of REF (e.g. \"cite:\",\"parencite:\" etc.) if it has a prefix of one of the `org-ref-cite-types`, otherwise, return `nil`.  This can be used to determine if a ref should be a \"cite\" type."
```
jethrokuan, line 405:
```suggestion
  "Return the citation prefix of REF, or nil otherwise.
The prefixes are defined in `org-ref-cite-types`.
Examples:   
   (org-roam--cite-prefix \"cite:foo\") -> \"cite:\"
   (org-roam--cite-prefix \"https://google.com\") -> nil"
```

goktug97:
```elisp
(defun org-roam--get-ref-path-completions (&optional interactive)
  "Return a list of cons pairs for refs to absolute path of Org-roam files."
```
org-roam.el:598: Argument ‘interactive’ should appear (as INTERACTIVE) in the doc string

what do you suggest for this one?

zaeph: Let me fix this.

zaeph: 4def9c6cf0a97c76d838e102ddc452ac9858d8e0 addresses it.

zaeph:
> `Allow edits by maintainers` was on but I merged the PR

Now I know.  It's because you've been making your changes on `master` instead of a feature branch.

jethrokuan:
CI fails without #562 merged, but #562 causes CI to fail because of compilation errors in org-ref, which I'm tracking here: https://github.com/jkitchin/org-ref/pull/735

zaeph:
> CI fails without #562 merged, but #562 causes CI to fail because of compilation errors in org-ref, which I'm tracking here: [jkitchin/org-ref#735](https://github.com/jkitchin/org-ref/pull/735)

Thanks for the work.  I'm not sure if it's needed for `bibtex-completion` to require Emacs 26.1.

goktug97:
https://github.com/jethrokuan/org-roam/pull/547/commits/289a792f3b8ffc798765d1b0e314ab32db36d6c9
https://github.com/jethrokuan/org-roam/pull/547/commits/7d1fc3ec9d9fafa517bd85a0a15086b232317e42
Sorry it seems you've already made the changes

TimQuelch:
Woooo tests passing. That red x has been haunting this PR for a long time

PR #553, opened 2020-05-03
(feat): add prefix argument to force database rebuild

jethrokuan:

Addresses https://github.com/jethrokuan/org-roam/pull/547#issuecomment-623080000

Running `C-u M-x org-roam-db-build-cache` will rebuild the database from scratch.


zaeph:
LGTM!

PR #559, opened 2020-05-03
(ehancement): Improve org-roam-graph

progfolio:
Allow showing/building the whole graph from any buffer.
Allow programatically specifying connected component path.

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -271,15 +271,16 @@ ARG may be any of the following values:
   - `\\[universal-argument]' -   build the graph for FILE.
   - `\\[universal-argument]' -N  build the graph for FILE limiting nodes to N steps."
   (interactive "P")
-  (let ((file (or file (buffer-file-name))))
-    (unless file
-      (user-error "Cannot build graph for nil file. Is current buffer visiting a file?"))
-    (unless (org-roam--org-roam-file-p file)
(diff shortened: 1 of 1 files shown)


progfolio:
Simple enough change, and everything seems to work fine on my end. Appreciate it if anyone can give it a try before we push.

zaeph:
It works on my end.

zaeph:
I'm wondering if we could change the `buffer-file-name` to also include indirect buffers.

zaeph:
> I'm wondering if we could change the `buffer-file-name` to also include indirect buffers.

Done in e6ee56549ca71d62b2bf8a676f6dfcedb6b09f99.  What do you think?

zaeph:
Whoops, good catch on f8dbdc97c1b4d7b943a539e1bd5cbb589458998b.

Ready to merge?

progfolio:
Sorry! Was working on it at the same time. I think we can just pass the result of `(base-buffer-name)` as the arg to `buffer-file-name`. If it is nil, the current buffer will still be used. Sound good?

zaeph:
> Sorry! Was working on it at the same time. I think we can just pass the result of `(base-buffer-name)` as the arg to `buffer-file-name`. If it is nil, the current buffer will still be used. Sound good?

Yup, that sounds good to me.  Merge?

PR #561, opened 2020-05-04
(internal): make lint errors fail CI, and disable spellcheck during linting

jethrokuan:

Lint errors such as compile errors now often slip past, since they don't cause the CI to fail.


zaeph:
Do you think this is the reason why CI was failing randomly at times?

jethrokuan:
No, that's unfortunately not the case.

zaeph:
> No, that's unfortunately not the case.

Oh well.  We'll track this down eventually.

PR #563, opened 2020-05-04
(chore): inherit org-roam-link-current from org-link

jethrokuan:

A better default for org-roam-link-current is the original org-link
face, rather than org-block.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -667,7 +667,7 @@ ARG is used to forward interactive calls to
   :group 'org-roam-faces)
 
 (defface org-roam-link-current
-  '((t :inherit org-block))
+  '((t :inherit org-link))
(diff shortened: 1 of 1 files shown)


zaeph:
Could we actually do the same for links without valid destination?

jethrokuan:
> Could we actually do the same for links without a valid destination?

Yes, but we'd need to define a new face for this. Right now it's just using `'error`, which is also what Doom does.

zaeph:
> > Could we actually do the same for links without a valid destination?
> 
> Yes, but we'd need to define a new face for this. Right now it's just using `'error`, which is also what Doom does.

Ah, I thought we could inherit both faces.  Let me play with it.

PR #564, opened 2020-05-04
(feat): add custom face for links without destinations

zaeph:
Follow-up to #563.

There's a small oddity going on with the syntax-highlighting of `error` in `emacs-lisp-mode` which formats it as the `symbol-function`, but I think this is just cosmetics:
https://github.com/jethrokuan/org-roam/blob/9ae2984bc7ebaf4c19db78618558c7e28c609ac8/org-roam.el#L675

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -674,7 +674,7 @@ ARG is used to forward interactive calls to
 (defface org-roam-link-invalid
   '((t :inherit (error org-link)))
   "Face for Org-roam links that are not valid.
-Typically, this face is used for links without a destination."
+This face is used for links without a destination."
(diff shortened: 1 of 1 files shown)


jethrokuan:
Maybe `org-roam-link-invalid` would be more generic and applicable elsewhere.

zaeph:
> Maybe `org-roam-link-invalid` would be more generic and applicable elsewhere.

Done in e5d18f9fc82dfc18a75b705792416bd98ad80410.

zaeph:
Not sure if I should modify the docstring, too.

jethrokuan:
I'd drop the "Typically", but not a blocker.

PR #567, opened 2020-05-04
(feat): simplify format of `INFO` for find-ref and add `FILTER`

### LINKED ISSUE #565: Format of the `INFO` argument for `org-roam-find-ref`
Do we have a specific reason for keeping the format of the `INFO` argument for `org-roam-find-ref` as `(ref . "key")`?
https://github.com/jethrokuan/org-roam/blob/11e15594ccd2066b93de2cdf33bbad53f16d7ca8/org-roam.el#L617-L628

Now that we've separated `type` and `key` in the db, it feels like we could either have `INFO` be the `key`, or have it be a cons like `(type . "key")`.

zaeph:
The format of the `INFO` argument of `org-roam-find` has been simplified, thanks
to the split of `type` and `ref` in the db permitted by #547.

The non-interactive behaviour of `org-roam-find-ref` has also been isolated
into `org-roam--find-ref`.  Even if it makes us repeat ourselves a bit, it
avoids having to nil the interactive argument just to get the non-interactive
us, which I think is better syntax.

Fixes #565.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -598,21 +598,36 @@ See `org-roam--get-ref-path-completions' for details."
   :type 'boolean
   :group 'org-roam)
 
-(defun org-roam--get-ref-path-completions (&optional interactive)
+(defun org-roam--get-ref-path-completions (&optional interactive filter)
   "Return a list of cons pairs for refs to absolute path of Org-roam files.
-When INTERACTIVE `org-roam-include-type-in-ref-path-completions'
-are non-nil, format the car of the completion-candidates as
-'type:ref'."
+When `org-roam-include-type-in-ref-path-completions' and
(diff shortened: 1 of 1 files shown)


zaeph:
0d1502ac656fb8e3b9feed6a53711f1fbee3a970 implements a `FILTER` arg for which allows you to filter out the candidates according to `type`:
https://github.com/jethrokuan/org-roam/blob/0d1502ac656fb8e3b9feed6a53711f1fbee3a970/org-roam.el#L645-L651

I didn't go for `seq-filter` because I didn't want to go over all the candidates twice.  I think the code has lost a little bit of readability because of it, but sometimes, that's the price to pay for optimisation.

PR #570, opened 2020-05-05
(feat): org-roam-doctor: fix broken links

### LINKED ISSUE #222: Discovering / repairing broken references after moving files

jethrokuan:

This library provides `org-roam-doctor`, a utility for diagnosing and fixing
Org-roam files. Running `org-roam-doctor` launches a list of checks defined
by `org-roam-doctor--checkers`. Every checker is an instance of
`org-roam-doctor-checker`.

Each checker is given the Org parse tree (AST), and is expected to return a
list of errors. The checker can also provide "actions" for auto-fixing errors
(see `org-roam-doctor--remove-link' for an example).

The UX experience is inspired by both org-lint and checkdoc, and their code
is heavily referenced.

doctor

Closes #222

### DIFF
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -1,0 +1,225 @@
+;;; org-roam-doctor.el --- Rudimentary Roam replica with Org-mode -*- coding: utf-8; lexical-binding: t; -*-
+;;
+;; Copyright © 2020 Jethro Kuan <jethrokuan95@gmail.com>
+
+;; Author: Jethro Kuan <jethrokuan95@gmail.com>
+;; URL: https://github.com/jethrokuan/org-roam
+;; Keywords: org-mode, roam, convenience
+;; Version: 1.1.0
+;; Package-Requires: ((emacs "26.1") (dash "2.13") (f "0.17.2") (s "1.12.0") (org "9.3") (emacsql "3.0.0") (emacsql-sqlite "1.0.0"))
+
+;; This file is NOT part of GNU Emacs.
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to the
+;; Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+;; Boston, MA 02110-1301, USA.
+
+
+;;; Commentary:
+;;
+;; This library provides `org-roam-doctor', a utility for diagnosing and fixing
+;; Org-roam files. Running `org-roam-doctor' launches a list of checks defined
+;; by `org-roam-doctor--checkers'. Every checker is an instance of
+;; `org-roam-doctor-checker'.
+;;
+;; Each checker is given the Org parse tree (AST), and is expected to return a
+;; list of errors. The checker can also provide "actions" for auto-fixing errors
+;; (see `org-roam-doctor--remove-link' for an example).
+;;
+;; The UX experience is inspired by both org-lint and checkdoc, and their code
+;; is heavily referenced.
+;;
+;;; Code:
+;; Library Requires
+(require 'cl-lib)
+(require 'org)
+(require 'org-element)
+
+(declare-function org-roam-insert "org-roam")
+(declare-function org-roam--get-roam-buffers "org-roam")
+(declare-function org-roam--list-all-files "org-roam")
+(declare-function org-roam--org-roam-file-p "org-roam")
+
+(cl-defstruct (org-roam-doctor-checker (:copier nil))
+  (name 'missing-checker-name)
+  (description "")
+  (actions nil))
+
+(defconst org-roam-doctor--checkers
+  (list
+   (make-org-roam-doctor-checker
+    :name 'org-roam-doctor-broken-links
+    :description "Fix broken links."
+    :actions '(("d" . ("Unlink" . org-roam-doctor--remove-link))
+               ("r" . ("Replace link" . org-roam-doctor--replace-link))
+               ("R" . ("Replace link (keep label)" . org-roam-doctor--replace-link-keep-label))))))
+
+(defun org-roam-doctor-broken-links (ast)
+  "Checker for detecting broken links.
+AST is the org-element parse tree."
+  (org-element-map ast 'link
+    (lambda (l)
+      (when (equal "file" (org-element-property :type l))
+        (let ((file (org-element-property :path l)))
+          (or (file-exists-p file)
+              (file-remote-p file)
+               `(,(org-element-property :begin l)
+                 ,(format (if (org-element-lineage l '(link))
+                              "Link to non-existent image file \"%s\"\
+ in link description"
+                            "Link to non-existent local file \"%s\"")
+                          file))))))))
+
+(defun org-roam-doctor--check (buffer checkers)
+  "Check BUFFER for errors.
+CHECKERS is the list of checkers used."
+  (with-current-buffer buffer
+    (save-excursion
+      (goto-char (point-min))
+      (let* ((ast (org-element-parse-buffer))
+             (errors (sort (cl-mapcan
+                            (lambda (c)
+                              (mapcar
+                               (lambda (report)
+                                 (list (set-marker (make-marker) (car report))
+                                       (nth 1 report) c))
+                               (save-excursion
+                                 (funcall
+                                  (org-roam-doctor-checker-name c)
+                                  ast))))
+                            checkers)
+                           #'car-less-than-car)))
+        (dolist (e errors)
+          (pcase-let ((`(,m ,msg ,checker) e))
+            (switch-to-buffer buffer)
+            (goto-char m)
+            (org-reveal)
+            (undo-boundary)
+            (org-roam-doctor--resolve msg checker)
+            (set-marker m nil)))
+        errors))))
+
+;;; Actions
+(defun org-roam-doctor--recursive-edit ()
+  "Launch into a recursive edit."
+  (message "When you're done editing press C-M-c to continue.")
+  (recursive-edit))
+
+(defun org-roam-doctor--skip ()
+  "Skip the current error."
+  (message "Skipping..."))
+
+(defun org-roam-doctor--replace-link ()
+  "Replace the current link with a new link."
+  (save-match-data
+    (unless (org-in-regexp org-link-bracket-re 1)
+      (user-error "No link at point"))
+    (let ((orig (buffer-string))
+          (p (point)))
+      (condition-case nil
+          (save-excursion
+            (replace-match "")
+            (org-roam-insert))
+        (quit (progn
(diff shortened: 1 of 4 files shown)


zaeph:
I like the idea, and I like making this into its own library.

An earlier idea I had for implementing this would have been to run `org-lint` on every `org-roam--list-all-files`, since I believe it is checking for broken links.

jethrokuan:
oh... that saved me a lot of trouble

org-roam-doctor.el
```diff
@@ -0,0 +1,199 @@
...
+                                  ast))))
+                            checkers)
+                           #'car-less-than-car)))
+        (dolist (e errors)
+          (pcase-let ((`(,m ,msg ,checker) e))
+            (switch-to-buffer buffer)
+            (goto-char m)
+            (org-reveal)
+            (org-roam-doctor--resolve msg checker)
+            (set-marker m nil)))
```
zaeph, line 109: I'm not sure if we should have an `undo-boundary` here to have a more granular history.  Right now, it's one undo per buffer visited.

org-roam-doctor.el
```diff
@@ -0,0 +1,199 @@
...
+  "Launch into a recursive edit."
+  (message "When you're done editing press C-M-c to continue.")
+  (recursive-edit))
+
+(defun org-roam-doctor--replace-link ()
+  "Replace the current link with a new link."
+  (unless (org-in-regexp org-link-bracket-re 1)
+    (user-error "No link at point"))
+  (save-excursion
+    (delete-region (match-beginning 0) (match-end 0))
```
zaeph, line 123: I think this should be done in a temp-buffer.  Otherwise, if you `C-g` in a middle of the `org-roam-insert` meant to fix it, the link is removed from the current buffer.

A better option might be to catch errors/interrupts with a `condition-case` which would revert the buffer to its pre-doctor state.

org-roam-doctor.el
```diff
@@ -0,0 +1,199 @@
...
+  (actions nil))
+
+(defconst org-roam-doctor--checkers
+  (list
+   (make-org-roam-doctor-checker
+    :name 'org-roam-doctor-broken-links
+    :description "Fix broken links."
+    :actions '(("d" . ("Unlink" . org-roam-doctor--remove-link))
+               ("r" . ("Replace link" . org-roam-doctor--replace-link))
+               ("R" . ("Replace link (keep label)" . org-roam-doctor--replace-link-keep-label))))))
```
zaeph, line 66: An option to quit would be appreciated, and so would graceful exits with `C-g`.

Maybe a skip as well?

org-roam-doctor.el
```diff
@@ -0,0 +1,199 @@
...
+(defun org-roam-doctor--replace-link-keep-label ()
+  "Replace the current link with a new link, keeping the current link's label."
+  (unless (org-in-regexp org-link-bracket-re 1)
+    (user-error "No link at point"))
+  (save-excursion
+    (let ((label (if (match-end 2)
+                     (match-string-no-properties 2)
+                   (org-link-unescape (match-string-no-properties 1)))))
+      (delete-region (match-beginning 0) (match-end 0))
+      (org-roam-insert nil nil label))))
```
zaeph, line 135: If we want to keep the label, chances are we might be interested in pre-selecting any candidate that starts with that label.  For instance, if I have a broken link with the label `Metaphor`, fixing it should pre-select `Metaphor` in my list.  This could be done by adding a `PRE-SELECT` argument to `org-roam-insert` which, when set, would attempt to place the matched candidate at the top of the completions.  Sorting also has the benefit of being completion-engine-agnostic.

org-roam-doctor.el
```diff
@@ -0,0 +1,199 @@
...
+        (user-error "Not in an org-roam file"))
+      (setq files (list (buffer-file-name))))
+    (dolist (f files)
+      (let ((buf (find-file-noselect f)))
+        (with-current-buffer buf
+          (org-roam-doctor--check buf org-roam-doctor--checkers))
+        (unless (memq buf existing-buffers)
+          (save-buffer buf)
+          (kill-buffer buf)))))
+  (message "Linting completed."))
```
zaeph, line 195: Maybe we could print stats here by incrementing variables for fixed and skipped links.

org-roam-doctor.el
jethrokuan: 1db8a08

org-roam-doctor.el
jethrokuan: `org-roam-insert` doesn't really work in a temp-buffer though.

org-roam-doctor.el
jethrokuan: Skip was added in 54e1178.

org-roam-doctor.el
zaeph: Then the second option I've mentioned might be best.

jethrokuan: done... seems a bit inefficient though.

org-roam-doctor.el
jethrokuan: graceful quitting added: help window destroyed, original window configuration restored regardless.

org-roam-doctor.el
jethrokuan: can this be done with regular `completing-read`? Not familiar with how to do this.

org-roam-doctor.el
```diff
@@ -0,0 +1,220 @@
...
+      (when (equal "file" (org-element-property :type l))
+        (let ((file (org-element-property :path l)))
+          (and (not (file-remote-p file))
+               (not (file-exists-p file))
+               (list (org-element-property :begin l)
+                     (format (if (org-element-lineage l '(link))
+                                 "Link to non-existent image file \"%s\"\
+ in link description"
+                               "Link to non-existent local file \"%s\"")
+                             file))))))))
```
progfolio, line 82:
```suggestion
(defun org-roam-doctor-broken-links (ast)
  "Checker for detecting broken links.
AST is the org-element parse tree."
  (org-element-map ast 'link
    (lambda (l)
      (when (equal "file" (org-element-property :type l))
        (let ((file (org-element-property :path l)))
          (and (not (or (file-exists-p file)
                        (file-remote-p file)))
               `(,(org-element-property :begin l)
                 ,(format (if (org-element-lineage l '(link))
                              "Link to non-existent image file \"%s\"\
 in link description"
                            "Link to non-existent local file \"%s\"")
                          file))))))))
```
I grouped the file predicates in an `or` and reversed the order. This should be more efficient, as the `or` will short-circuit on the first nil case. I believe that will be `file-exists-p` more often than `file-remote-p`. It's probably a micro-optimization, but it should be logically equivalent.

org-roam-doctor.el
```diff
@@ -0,0 +1,220 @@
...
+      (setq files (list (buffer-file-name))))
+    (save-window-excursion
+      (dolist (f files)
+        (let ((buf (find-file-noselect f)))
+          (with-current-buffer buf
+            (org-roam-doctor--check buf org-roam-doctor--checkers))
+          (unless (memq buf existing-buffers)
+            (save-buffer buf)
+            (kill-buffer buf))))))
+  (message "Linting completed."))
```
progfolio, line 216: Perhaps we could introduce a customizable variable, `org-roam-doctor-files` which defaults to `(org-roam-llist-all-files)`. `org-roam-doctor` would run its checks against the files in this list. Users/commands could then customize that list to check a subset.
e.g.
```emacs-lisp
(let ((org-roam-doctor-files `("example.org" "etc..."))) 
  (org-roam-doctor))
```
 similar to `org-agenda-files`.

org-roam.el
```diff
@@ -462,11 +463,12 @@ Examples:
...
      description)))
 
-(defun org-roam-insert (prefix &optional filter-fn)
+(defun org-roam-insert (&optional prefix filter-fn description)
   "Find an Org-roam file, and insert a relative org link to it at point.
 If PREFIX, downcase the title before insertion.
 FILTER-FN is the name of a function to apply on the candidates
-which takes as its argument an alist of path-completions.  See
+which takes as its argument an alist of path-completions.
+If DESCRIPTION is provided, use this as the link label. See
```
progfolio, line 471: If `prefix` is meant to downcase the title before insertion, can we give it a descriptive name? e.g. `downcase` or `lowercase`?

org-roam-doctor.el
```diff
@@ -0,0 +1,226 @@
...
+  "Perform a check on Org-roam files to ensure cleanliness.
+If THIS-BUFFER, run the check only for the current buffer."
+  (interactive "P")
+  (let (files)
+    (if (not this-buffer)
+        (setq files (org-roam--list-all-files))
+      (unless (org-roam--org-roam-file-p)
+        (user-error "Not in an org-roam file"))
+      (setq files (list (buffer-file-name))))
+    (org-roam-doctor files org-roam-doctor--checkers)))
```
progfolio, line 209:
```suggestion
;;;###autoload
(defun org-roam-doctor (&optional all)
  "Perform a check on the current Org-roam file to ensure cleanliness.
If ALL is non-nil, run the check only for all org-roam files."
  (interactive "P")
  (let (files (if all
                  (org-roam--list-all-files)
                (unless (org-roam--org-roam-file-p)
                  (user-error "Not in an org-roam file"))
                `(,(buffer-file-name))))
    (org-roam-doctor-start files org-roam-doctor--checkers)))
```
I think the last line is intended to call `org-roam-doctor-start`, correct?

Do you think reversing the semantics of `org-roam-doctor`'s argument would be better UX?
The larger one's set of files gets, the longer this command will block Emacs.
Do we want to present that as the typical use-case, or as an optional use-case?
Ideally we could provide the best of both worlds and implement the whole thing asynchronously, but that comes with challenges as well (e.g. what happens if you're editing the files as they're being checked...).

jethrokuan: > Do you think reversing the semantics of org-roam-doctor's argument would be better UX?

Hmm perhaps so. 

>  The larger one's set of files gets, the longer this command will block Emacs.

I want to add a `linted` column to the files table, so org-roam-doctor can mark a file as linted once it does a pass-through. Modifying a file would then reset the value of that column to `nil`. That should make linting fast enough to run, to `org-roam-db-build-cache`.

jcguu95:
Is it possible to `org-roam-doctor-this-file`? So I can fix the file name of the current file, and update all links to this file all at once?

zaeph:
Renaming the file in Emacs will already update all the links.  Try to do it in Dired.

jcguu95:
@zaeph that's definitely new to me! Would you mind giving a bit more pointers?

zaeph:
Well, there isn't any.  When `org-roam-mode` is active, renaming a file in your `org-roam-directory` will run a number of checks to make sure that  all the files that were pointing to the file being renamed will be updated.

jcguu95:
That's surprising.. say I renamed `1.org` to `1_.org`, how would `org-roam` know that `1_.org` used to be `1.org`?

zaeph:
Because the hook we use, `rename-file`, is aware of the previous name and the new name.
https://github.com/org-roam/org-roam/blob/7a4b15fd368bc8ac55bfe96392f5a300f71f009c/org-roam.el#L1310
https://github.com/org-roam/org-roam/blob/7a4b15fd368bc8ac55bfe96392f5a300f71f009c/org-roam.el#L1226-L1274

PR #574, opened 2020-05-07
(internal): add longer sleep time to tests

jethrokuan:

Tests are failing more often now, spuriously. Let's see if adding some sleep time helps.


zaeph:
It definitely seems to be more cooperative now.  Thanks for that!

jethrokuan:
Yeah it does seem so!

Issue #573, opened 2020-05-06
Add org-roam-capture-templates to documentation


edumerco:
### Description
The template system is well documented, but where/how to define them is not. 

Maybe the variable org-roam-capture-templates could be included (surrounding?) the examples so as to try to make it easier for the new users to customize their templates.

edumerco:
Thanks. :+1: :smiley:

zoechi:
@edumerco 

> The template system is well documented

Do you mean org-mode capture templates?
I haven't used them yet (still an Emacs rookie).
I couldn't find much info about org-roam capture templates.

edumerco:
It's in the documentation included. Online, it's  in https://org-roam.readthedocs.io/en/master/templating/

christabella:
I believe the readthedocs is deprecated 😰  moving so fast! In the updated docs, it is at https://org-roam.github.io/org-roam/manual/Template-Walkthrough.html#Template-Walkthrough

RaphaelPour:
https://www.orgroam.com/manual.html#Template-Walkthrough

@christabella Your link got deprecated, too :see_no_evil:

PR #575, opened 2020-05-07
(docs): mention where to modify org-roam templates

jethrokuan:

Addresses #573


alanz:
Should it reference `org-roam-capture-ref-templates` as well? or somewhere?

jethrokuan:
It's mentioned in the roam protocol page.

PR #577, opened 2020-05-07
(docs): link to discourse in README

zoechi:

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Important links:
 
 - **[Documentation][docs]**
 - **[Org-roam Discourse][discourse]**
+- **[Org-roam Slack][slack]**
 
 ## A Preview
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,7 +27,7 @@ As of February 2020, it is in a very early stage of development.
 Important links:
 
 - **[Documentation][docs]**
-- **[Org-roam Slack][slack]**
+- **[Org-roam Discourse][discourse]**
```
zaeph, line 30: I don't think we should be removing the Slack.  It's still useful for live support.

README.md
```diff
@@ -64,6 +64,12 @@ Here's a sample configuration with using `use-package`:
               (("C-c n i" . org-roam-insert))))
 ```
 
+`org-roam-graph` by default expects to find the `dot` executable 
+from the `graphviz` package in the `$PATH`.
+Ensure `graphviz` is installed if you want to use this feature
+or customize your configuration for `org-roam-graph` to use a 
+different tool.
```
jethrokuan, line 71: This section is not entirely correct. It doesn't have to be in env variable `$PATH`, but it needs to be in Emacs-lisp list `exec-path`. Ensuring graphviz is installed doesn't make the path to the executable available to Emacs either, that requires a bit more careful customization, but is typically resolved with package `exec-path-with-shell`.

README.md
zoechi: Ok, I thought new users shouldn't be directed there. I'll fix.

README.md
zoechi: Thanks for the suggestion. I changed it a bit.

README.md
```diff
@@ -28,6 +28,7 @@ Important links:
 
 - **[Documentation][docs]**
 - **[Org-roam Discourse][discourse]**
```
jethrokuan, line 30: Could you also change this to just "Discourse"? seems a bit more consistent.

PR #579, opened 2020-05-07
(fix): improve error message on missing executable

### LINKED ISSUE #576: Weird behavior when `dot` executable is nout found.
#### Steps to Reproduce
- ensure `dot` executable is not found
- `M-x` `org-roam-graph`

> user-error: Can’t find nil executable.  Please check if it is in your path

- look up in source code what executable is required (`nil`) is not helpful
```sh
sudo apt install graphviz
```
- `M-x` `org-roam-graph`

> user-error: Can’t find nil executable.  Please check if it is in your path

- execute
```lisp
(setq org-roam-graph-executable "dot")
```
(or restart Emacs)

- `M-x` `org-roam-graph`

- Graph is finally shown

#### Expected Results
It should say what executable it is looking for (`dot`) and not drop this default value if `dot` is not found.
Then it would produce a proper error message 
and would try to find `dot` again after it was installed and all would be fine.

#### Actual Results
Confusing behavior

### Versions
- `GNU Emacs 28.0.50 (build 6, x86_64-pc-linux-gnu, GTK+ Version 3.24.5, cairo version 1.16.0) of 2020-04-27`
- Org-roam commit: latest from elpa

zaeph:
Fixes #576.

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -53,7 +53,7 @@ It may be one of the following:
           (const    :tag "view-file"))
   :group 'org-roam)
 
-(defcustom org-roam-graph-executable "dot"
+(defcustom org-roam-graph-executable (executable-find "dot")
   "Name of graphing executable."
   :type 'string
   :group 'org-roam)
@@ -219,10 +219,9 @@ into a digraph."
 
 (defun org-roam-graph--build (&optional node-query)
   "Generate a graph showing the relations between nodes in NODE-QUERY."
-  (unless (executable-find org-roam-graph-executable)
-    (user-error (concat "Cannot find executable \"%s\" to generate the graph.  "
-                        "Please adjust `org-roam-graph-executable'")
-                org-roam-graph-executable))
+  (unless org-roam-graph-executable
+    (user-error (concat "Cannot find executable to generate the graph.  "
+                        "Please adjust `org-roam-graph-executable'")))
   (let* ((node-query (or node-query
                          `[:select [file titles]
                            :from titles
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -221,8 +221,7 @@ into a digraph."
   "Generate a graph showing the relations between nodes in NODE-QUERY."
   (unless org-roam-graph-executable
     (user-error (concat "Couldn’t find the executable to generate the graph.  "
-                        "Please adjust `org-roam-graph-executable'")
-                org-roam-graph-executable))
+                        "Please adjust `org-roam-graph-executable'")))
   (let* ((node-query (or node-query
                          `[:select [file titles]
                            :from titles
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -219,8 +219,8 @@ into a digraph."
 
 (defun org-roam-graph--build (&optional node-query)
   "Generate a graph showing the relations between nodes in NODE-QUERY."
-  (unless org-roam-graph-executable
     (user-error (concat "Couldn’t find the executable %s to generate the graph.  "
+  (unless (executable-find org-roam-graph-executable)
                         "Please adjust `org-roam-graph-executable'")
                 org-roam-graph-executable))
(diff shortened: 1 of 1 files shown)


zaeph:
I think we could run `executable-find` inside `graph--build` so that we can output which executable we were looking for.  Sure, we'd be retrieving the variable from path on every call, but that feels like a minor hurdle.

zaeph:
> I think we could run `executable-find` inside `graph--build` so that we can output which executable we were looking for. Sure, we'd be retrieving the variable from path on every call, but that feels like a minor hurdle.

Done in 7b876e5c648b4bf7fcc9e8f624f1380f566dd0b4.

jethrokuan:
Does this break existing configurations that set the path explicitly? Also there are some who prefer not to make the executable available anywhere in PATH, so I think having the defcustom point to the path of the executable is my preference.

zaeph:
> Does this break existing configurations that set the path explicitly? Also there are some who prefer not to make the executable available anywhere in PATH, so I think having the defcustom point to the path of the executable is my preference.

Are we OK with just the following error message, then?
> Cannot find executable to generate the graph.  Please adjust ‘org-roam-graph-executable’

zaeph:
6547a222d183a521c3e0dec8aac9e47e5b561f8c reverts the check to the definition.

I still need to iron out a problem when the exec is just the name and can't be found in `PATH`, since we get an uninformative message:

> let*: Searching for program: No such file or directory, foo
 
I'll user a `condition-case` to catch it.

jethrokuan:
> Cannot find executable to generate the graph. Please adjust ‘org-roam-graph-executable’

Yeah, this is fine I think.

Alternatively, we can do a

```
(or org-roam-graph-executable (executable-find org-roam-graph-executable))
```

if people just set it to the name of the executable.

zaeph:
> Alternatively, we can do a
> 
> ```
> (or org-roam-graph-executable (executable-find org-roam-graph-executable))
> ```
> 
> if people just set it to the name of the executable.

Did something in the same spirit, with an error message that tries to be as relevant as it can be.

zaeph:
94cb85455adc784f41d4f71ed752b4196ff9d5de changes the default value, but it won't cause any problem for people who still have the old one in their config.

org-roam-graph.el
```diff
@@ -219,19 +219,23 @@ into a digraph."
...
+                          "Please adjust `org-roam-graph-executable'")))
+    (let* ((node-query (or node-query
+                           `[:select [file titles]
+                             :from titles
+                             ,@(org-roam-graph--expand-matcher 'file t)]))
+           (graph (org-roam-graph--dot node-query))
+           (temp-dot (make-temp-file "graph." nil ".dot" graph))
+           (temp-graph (make-temp-file "graph." nil ".svg")))
+      (call-process exec nil 0 nil temp-dot "-Tsvg" "-o" temp-graph)
+      temp-graph)))
```
progfolio, line 238:
```suggestion
  (unless (stringp org-roam-graph-executable)
    (user-error "`org-roam-graph-executable' must be a string."))
  (unless (executable-find org-roam-graph-executable)
    (user-error "Cannot find executable \"%s\" to generate the graph.
Please adjust `org-roam-graph-executable'"
                org-roam-graph-executable))
  (let* ((node-query (or node-query
                         `[:select [file titles]
                                   :from titles
                                   ,@(org-roam-graph--expand-matcher 'file t)]))
         (graph      (org-roam-graph--dot node-query))
         (temp-dot   (make-temp-file "graph." nil ".dot" graph))
         (temp-graph (make-temp-file "graph." nil ".svg")))
    (call-process exec nil 0 nil temp-dot "-Tsvg" "-o" temp-graph)
    temp-graph))
```
This will cover cases where `org-roam-graph-executable` is the wrong type.
It will also cover cases where the executable is not found.
The only corner-case is when it is set to the empty string. Though, the wording of the suggestion ("Please adjust...") seems clear enough to me.

PR #580, opened 2020-05-07
(fix): make graph--build error on non-string

zaeph:
Follow up to #579 to improve the validation.

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -219,14 +219,13 @@ into a digraph."
 
 (defun org-roam-graph--build (&optional node-query)
   "Generate a graph showing the relations between nodes in NODE-QUERY."
-  (let* ((name org-roam-graph-executable)
-         (exec (ignore-errors (executable-find name))))
-    (unless exec
-      (user-error (concat "Cannot find executable "
-                          (when name
-                            (format "\"%s\" " name))
-                          "to generate the graph.  "
-                          "Please adjust `org-roam-graph-executable'")))
+  (let* ((name org-roam-graph-executable))
+    (unless (stringp name)
+      (user-error "`org-roam-graph-executable' must be a string"))
+    (unless (executable-find org-roam-graph-executable)
+      (user-error (concat "Cannot find executable \"%s\" to generate the graph.  "
+                          "Please adjust `org-roam-graph-executable'")
+                  name))
     (let* ((node-query (or node-query
                            `[:select [file titles]
                              :from titles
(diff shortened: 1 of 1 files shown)


org-roam-graph.el
```diff
@@ -219,22 +219,21 @@ into a digraph."
...
-                            (format "\"%s\" " name))
-                          "to generate the graph.  "
-                          "Please adjust `org-roam-graph-executable'")))
+  (let ((name org-roam-graph-executable))
+    (unless (stringp name)
+      (user-error "`org-roam-graph-executable' must be a string"))
+    (unless (executable-find org-roam-graph-executable)
+      (user-error (concat "Cannot find executable \"%s\" to generate the graph.  "
+                          "Please adjust `org-roam-graph-executable'")
+                  name))
```
progfolio, line 228: Do we need the store org-roam-graph-executable in `name`?

zaeph: We re-using it three times in the script, I think that's enough to warrant a local variable.

zaeph: https://github.com/jethrokuan/org-roam/blob/ac444501f37e6deeacf6b6d1bfdfd2789f83bb63/org-roam-graph.el#L223

https://github.com/jethrokuan/org-roam/blob/ac444501f37e6deeacf6b6d1bfdfd2789f83bb63/org-roam-graph.el#L228

https://github.com/jethrokuan/org-roam/blob/ac444501f37e6deeacf6b6d1bfdfd2789f83bb63/org-roam-graph.el#L236

progfolio: I suggest changing "must be a string" to "is not a string" as well per:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Signaling-Errors.html

> The message should state what is wrong (“File does not exist”), not how things ought to be (“File must exist”)

zaeph: Addressed in ab99164dc1d15e5184a22f10f5c16729f449ab59.  Anything else or can I merge?

progfolio: The suggestion for `name` is that it's unnecessary as we can just refer to `org-roam-graph-executable` in those three spots. Unless you think it helps with readability to have a local alias, I would argue that introducing a lexical-binding alongside the global binding introduces more cognitive overhead than it saves typing. Either way it's a short function, so it's not too hard to reason about. Just a suggestion/preference of mine. 

>  Anything else or can I merge?
Looks good to me!

PR #584, opened 2020-05-08
(fix): allow group templates in capture-templates

### LINKED ISSUE #583: org-capture-set-target-location: Invalid capture target specification: nil
### Description

Using the standard `org-mode` hierarchical capture templates doesn't appear to work in `org-roam`.

#### Steps to Reproduce

1. Customize `org-roam-capture-templates` as per normal but with the two-element tree variant.

```elisp
      '(
        ("i" "Inbox" plain (function org-roam--capture-get-point) "%?" :file-name "${slug}" :head "#+TITLE: ${title}")
        ("p" "Projects")
        ("pa" "Project A" plain (function org-roam--capture-get-point) "%?" :file-name "a/${slug}" :head "#+TITLE: ${title}")
        ("pb" "Project B" plain (function org-roam--capture-get-point) "%?" :file-name "b/${slug}" :head "#+TITLE: ${title}")
        ("pc" "Project C" plain (function org-roam--capture-get-point) "%?" :file-name "c/${slug}" :head "#+TITLE: ${title}")
        ))
```

2. Run `org-roam-find-file`.
3. Insert any new title.
4. Select a capture template:
```
Select a capture template
=======================

[i]     Inbox
[p]     Projects
------------------------------------------------------------------
[C]     Customize org-capture-templates
[q]     Abort
```
5. Select `p` for menu

#### Expected Results

Expecting `p...` in the menu rather than just `p`, and then being able to select a, b or c from there.

#### Actual Results

Get error `org-capture-set-target-location: Invalid capture target specification: nil`.

### Versions
- Emacs (`C-h v emacs-version`): v26.3
- Org-roam version: 20200507.1838 from MELPA

zaeph:
Fixes #583.

@progfolio Do you think the pase is warranted, here?  Ever since you've shown me how to use `pred`, I want to use it all the time, even when its use isn't warranted.  But here, it allows to check `(str str)` quite nicely:
https://github.com/jethrokuan/org-roam/blob/cb448b6a325eb7c19574cf0e06d166bc6b1fe630/org-roam-capture.el#L288-L289

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -289,7 +289,7 @@ This function is used solely in Org-roam's capture templates: see
            org-roam-plist
            key
            val)
-       ;;put positional args on converted template
+       ;; Convert positional args before taking care of the plist-args
        (dotimes (_ 5)
          (push (pop copy) converted))
        (while (setq key (pop copy)
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -281,7 +281,7 @@ This function is used solely in Org-roam's capture templates: see
 (defun org-roam-capture--convert-template (template)
   "Convert TEMPLATE from Org-roam syntax to `org-capture-templates' syntax."
   (pcase template
-    (`(,key ,description) template)
+    (`(,_key ,_description) template)
     (`(,key ,description ,type ,target . ,rest)
      (let ((converted `(,key ,description ,type ,target
                              ,(unless (keywordp (car rest)) (pop rest))))
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -283,30 +283,38 @@ This function is used solely in Org-roam's capture templates: see
 Return t if the TEMPLATE is well-formed, nil otherwise.
 As a special case, return 'group if TEMPLATE is a template-group.
 See `org-roam-capture-templates' for details."
-  (condition-case err
-      (pcase template
-        (`(,(pred stringp) ,(pred stringp))
-         'group)
-        ((pred (lambda (x)
-                 (>= (length x) 5)))
-         (pcase (cl-subseq template 0 5)
-           (`(,(pred stringp)
-              ,(pred stringp)
-              ,(pred symbolp)
-              ,(pred listp)
-              ,(pred stringp))
-            t)
-           (malformed-template
-            (signal 'wrong-type-argument
-                    `((list stringp stringp symbolp listp stringp …)
-                      (list ,@malformed-template …))))))
-        ((pred (lambda (x) (not (listp x))))
-         (signal 'wrong-type-argument `((listp) ,template)))
-        (_
-         (signal 'wrong-type-argument `((,@template)))))
-    (wrong-type-argument
-     (user-error "Malformed template in `org-roam-capture-templates: %s"
-                 (error-message-string err)))))
+  (let ((correct-format '(list stringp stringp symbolp listp stringp …)))
+    (condition-case err
+        (pcase template
+          ;; Error if TEMPLATE is not a list
+          ((pred (lambda (x) (not (listp x))))
+           (signal 'wrong-type-argument `((listp) ,template)))
+          ;; Check if TEMPLATE is a special group-template
+          (`(,(pred stringp) ,(pred stringp))
+           'group)
+          ;; Validate elements in TEMPLATE
+          ((pred (lambda (x)
+                   (>= (length x) 5)))
+           (pcase (cl-subseq template 0 5)
+             (`(,(pred stringp)
+                ,(pred stringp)
+                ,(pred symbolp)
+                ,(pred listp)
+                ,(pred stringp))
+              t)
+             ;; Error if elements in TEMPLATE are not the right type
+             (malformed-template
+              (signal 'wrong-type-argument
+                      `(,correct-format
+                        (list ,@malformed-template …))))))
+
+          ;; Catch-all
+          (_
+           (signal 'wrong-type-argument `(,correct-format
+                                          (list ,@template)))))
+      (wrong-type-argument
+       (user-error "Malformed template in `org-roam-capture-templates: %s"
+                   (error-message-string err))))))
 
 (defun org-roam-capture--convert-template (template)
   "Convert TEMPLATE from Org-roam syntax to `org-capture-templates' syntax."
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -33,6 +33,7 @@
 (require 'org-capture)
 (require 'dash)
 (require 's)
+(require 'seq)
 
 ;; Declarations
 (defvar org-roam-encrypt-files)
@@ -291,7 +292,7 @@ See `org-roam-capture-templates' for details."
      'group)
     ((pred (lambda (x)
              (>= (length x) 5)))
-     (pcase (cl-subseq template 0 5)
+     (pcase (seq-subseq template 0 5)
        (`(,(pred stringp)
           ,(pred stringp)
           ,(pred symbolp)
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -310,7 +310,7 @@ See `org-roam-capture-templates' for details."
   (pcase (org-roam-capture--validate-template template)
     ('group
      template)
-    (t
+    (_
      (let ((copy (copy-tree template))
            converted
            org-roam-plist
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -281,26 +281,20 @@ This function is used solely in Org-roam's capture templates: see
 (defun org-roam-capture--convert-template (template)
   "Convert TEMPLATE from Org-roam syntax to `org-capture-templates' syntax."
   (pcase template
-    (`(,(pred stringp) ,(pred stringp))
-     template)
-    (_
-     (let ((copy (copy-tree template))
-           converted
+    (`(,key ,description) template)
+    (`(,key ,description ,type ,target . ,rest)
+     (let ((converted `(,key ,description ,type ,target
+                             ,(unless (keywordp (car rest)) (pop rest))))
            org-roam-plist
-           key
-           val)
-       ;; Convert positional args before taking care of the plist-args
-       (dotimes (_ 5)
-         (push (pop copy) converted))
-       (while (setq key (pop copy)
-                    val (pop copy))
-         (if (member key org-roam-capture--template-keywords)
-             (progn
-               (push val org-roam-plist)
-               (push key org-roam-plist))
-           (push key converted)
-           (push val converted)))
-       (append (nreverse converted) `(:org-roam ,org-roam-plist))))))
+           options)
+       (while rest
+         (let* ((key (pop rest))
+                (val (pop rest))
+                (custom (member key org-roam-capture--template-keywords)))
+           (push val (if custom org-roam-plist options))
+           (push key (if custom org-roam-plist options))))
+       (append converted options `(:org-roam ,org-roam-plist))))
+    (_ (user-error "Invalid capture template format: %s" template))))
 
 (defun org-roam-capture--find-file-h ()
   "Opens the newly created template file.
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -278,32 +278,35 @@ This function is used solely in Org-roam's capture templates: see
     (widen)
     (goto-char (point-max))))
 
-(define-error 'org-roam-malformed-template
-  "Malformed template in `org-roam-capture-templates'"
-  'org-roam-errors)
-
 (defun org-roam-capture--validate-template (template)
   "Validate TEMPLATE by checking its elements.
 Return t if the TEMPLATE is well-formed, nil otherwise.
 As a special case, return 'group if TEMPLATE is a template-group.
 See `org-roam-capture-templates' for details."
-  (pcase template
-    (`(,(pred stringp) ,(pred stringp))
-     'group)
-    ((pred (lambda (x)
-             (>= (length x) 5)))
-     (pcase (cl-subseq template 0 5)
-       (`(,(pred stringp)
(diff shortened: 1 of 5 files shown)


zaeph:
I've realised I could convolute the code a little better to not have the `let`-form at the start create useless variables.

zaeph:
> I've realised I could convolute the code a little better to not have the `let`-form at the start create useless variables.

Working on a better version, now.

zaeph:
Came up with a massively over-engineered solution in c83d37729300fbd06d674f6cd5da5a53f0fa9e9b.

To make the error message as informative as possible, I've defined a new error:
https://github.com/jethrokuan/org-roam/blob/469b58206f966b093521b602b5e0b1afe7957fcd/org-roam-capture.el#L280-L282

Not sure if this is good style or not.  Worst case scenario, we could `catch` and `throw` and output the same error:
https://github.com/jethrokuan/org-roam/blob/c83d37729300fbd06d674f6cd5da5a53f0fa9e9b/org-roam-capture.el#L304

zaeph:
@progfolio You can take a peek now!

org-roam-capture.el
```diff
@@ -277,25 +277,54 @@ This function is used solely in Org-roam's capture templates: see
...
+     'group)
+    ((pred (lambda (x)
+             (>= (length x) 5)))
+     (pcase (cl-subseq template 0 5)
+       (`(,(pred stringp)
+          ,(pred stringp)
+          ,(pred symbolp)
+          ,(pred listp)
+          ,(pred stringp))
+        t)
```
zaeph, line 301: I really thought that `_` would save me from having to `cl-subseq` the list, i.e., by only considering the the first 5 members of the list, but it didn't.  I actually think adding a `_*` expansion in `pcase` would be useful.

…Or maybe I'm just missing something really obvious.

org-roam-capture.el
```diff
@@ -277,25 +277,54 @@ This function is used solely in Org-roam's capture templates: see
     (widen)
     (goto-char (point-max))))
 
+(define-error 'org-roam-malformed-template
+  "Malformed template in `org-roam-capture-templates'"
+  'org-roam-errors)
```
zaeph, line 283: I define a new error to make it obvious that it comes from `org-roam-capture-templates`, which wouldn't be the case with a mere `wrong-type`.

org-roam-capture.el
```diff
@@ -277,25 +277,54 @@ This function is used solely in Org-roam's capture templates: see
...
+          ,(pred stringp)
+          ,(pred symbolp)
+          ,(pred listp)
+          ,(pred stringp))
+        t)
+       (malformed-template
+        (signal 'org-roam-malformed-template
+                `((list stringp stringp symbolp listp stringp …)
+                  (list ,@malformed-template …))))))
+    (_ (user-error "Malformed template in `org-roam-capture-templates'"))))
```
zaeph, line 306: If we don't go for a new type of error, we can just `catch` and `throw` to here.

org-roam-capture.el
```diff
@@ -277,25 +277,54 @@ This function is used solely in Org-roam's capture templates: see
...
     (goto-char (point-max))))
 
+(define-error 'org-roam-malformed-template
+  "Malformed template in `org-roam-capture-templates'"
+  'org-roam-errors)
+
+(defun org-roam-capture--validate-template (template)
+  "Validate TEMPLATE by checking its elements.
+Return t if the TEMPLATE is well-formed, nil otherwise.
+As a special case, return 'group if TEMPLATE is a template-group.
```
zaeph, line 288: Maybe that behaviour would warrant another name than `validate`, but since `'group` is still `t`, I think it works.

org-roam-capture.el
```diff
@@ -244,7 +244,7 @@ If the search is via title, it is assumed that the file does not
 yet exist, and Org-roam will attempt to create new file.
 
 If the search is via daily notes, 'time will be passed via
-`org-roam-capture--info'. This is used to alter the default time
+`org-roam-capture--info'.  This is used to alter the default time
```
jethrokuan, line 248: Can we disable double spacing haha

org-roam-capture.el
jethrokuan: think this is okay, it's a bit more explicit, but it seems that makes `user-error` message inaccurate, since malformed templates should signal the new error

org-roam-capture.el
zaeph: #blamecheckdoc

I'll revert it.

zaeph: Reverted in 24b9e2f5a4464cb1e97121e06f7c9a8340d24732.

zaeph:
Working on your comment; got a pretty nice solution.

zaeph:
> think this is okay, it's a bit more explicit, but it seems that makes user-error message inaccurate, since malformed templates should signal the new error

Done in 1d779ad4eca55125810a63f5ce8b5d10ece59ac3.  I think it's cleaner now.

jethrokuan:
I feel like the solution should be simpler. Org-roam's expansion should just expand certain elements of org-capture templates, and return the templates with best-effort, then validations should be done by org-capture.

zaeph:
> I feel like the solution should be simpler. Org-roam's expansion should just expand certain elements of org-capture templates, and return the templates with best-effort, then validations should be done by org-capture.

Yeah, I agree.  I was just excited about writing an error-handler.  I'll see what I can do.

zaeph:
3597893d6928a9df7fda4d6931995190048c18a5 lets `org-capture` handle the error checking.  It isn't super informative, so maybe I could use a `condition-case` to append something to the error message.

jethrokuan:
>  It isn't super informative, so maybe I could use a `condition-case` to append something to the error message.

Yep, I think that's a good idea. Catch the error thrown by org-capture, and let users know the error comes from `org-roam-capture-templates` (or `org-roam-ref-capture-templates`), rather than from their org-capture templates.

zaeph:
> Yep, I think that's a good idea. Catch the error thrown by org-capture, and let users know the error comes from `org-roam-capture-templates` (or `org-roam-ref-capture-templates`), rather than from their org-capture templates.

Done in fd4d16870a6006f6f23a3aa8f1b4b05fd6a486e3.

It prints error-messages like this one:
> user-error: Invalid capture target specification: nil.  Please adjust ‘org-roam-capture-templates’

org-roam-capture.el
```diff
@@ -279,23 +280,27 @@ This function is used solely in Org-roam's capture templates: see
...
+         (push (pop copy) converted))
+       (while (setq key (pop copy)
+                    val (pop copy))
+         (if (member key org-roam-capture--template-keywords)
+             (progn
+               (push val org-roam-plist)
+               (push key org-roam-plist))
+           (push key converted)
+           (push val converted)))
+       (append (nreverse converted) `(:org-roam ,org-roam-plist))))))
```
progfolio, line 303:
```suggestion
  (pcase template
    (`(,key ,description) template)
    (`(,key ,description ,type ,target . ,rest)
     (let ((converted `(,key ,description ,type ,target
                             ,(unless (keywordp (car rest)) (pop rest))))
           org-roam-plist
           key
           val)
       (while (setq key (pop rest)
                    val (pop rest))
         (if (member key org-roam-capture--template-keywords)
             (progn (push val org-roam-plist)
                    (push key org-roam-plist))
           (push key converted)
           (push val converted)))
       (append converted `(:org-roam ,org-roam-plist))))
    (_ (user-error "Invalid capture template format: %s" template))))
```
A pcase is a good choice for a structure like an org-capture-template.
I wouldn't worry too much about type checking every element of the template, though, because beyond simple checks it's enough work to become its own package (that's a large part of what doct does).
Note in the second case the use of the dotted notation to capture the rest of the list. It's a useful pattern.

progfolio: Actually, I neglected to consider the order of the options when appending that, one second...

zaeph: > Note in the second case the use of the dotted notation to capture the rest of the list. It's a useful pattern.

Aw, I knew there was a way to do this!  Thanks a bunch for this, although I wish I could have found it on my end, especially since I'm supposed to know how lists work in a Lisp. \*sigh\*

Let me review your code and implement it.

zaeph: Your solution is much more elegant than mine, and you're making a better use of `pcase`.  It's always a pleasure to study your code and learn new ways of using it.  Thanks!

zaeph: Applied in 7107ce4efa658ba44f0294c914c9d4b7b4df5039.

org-roam-capture.el
```diff
@@ -327,7 +327,10 @@ GOTO and KEYS argument have the same functionality as
     (when (= (length org-capture-templates) 1)
       (setq keys (caar org-capture-templates)))
     (add-hook 'org-capture-after-finalize-hook #'org-roam-capture--save-file-maybe-h)
-    (org-capture goto keys)))
+    (condition-case err
+        (org-capture goto keys)
+      (error (user-error "%s.  Please adjust `org-roam-capture-templates'"
```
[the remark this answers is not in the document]
zaeph: Yeah, you're right, it's used by other commands and functions.  I think we can drop it.  If you run an Org-roam command, and get an error pointing out a problem with templates, chances are you're going to assume it's Org-roam's templates, not `org-capture`'s.

zaeph:
The linter doesn't seem to be understanding the `pcase` assignments.  I initially thought it was a problem with `pcase` not being required, but it wasn't it.

org-roam-capture.el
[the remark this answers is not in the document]
zaeph: I've got an idea about writing a wrapper that would take the name of the relevant template, wrap a form in `condition-case`, and format the an informative `error` with the template.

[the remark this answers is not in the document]
zaeph: Prototyped in d55f658f4a60580f33095a168ecf83693cd09a7d.

org-roam.el
```diff
@@ -441,6 +441,18 @@ Examples:
...
+  "Eval BODY, and point user to TEMPLATES on error.
+
+\(fn TEMPLATES BODY...)"
+  (declare (debug (form body)) (indent 1))
+  `(condition-case err
+       ,@body
+     (error (user-error "%s.  Please adjust `%s'"
+                        (error-message-string err)
+                        ,templates))))
+
```
progfolio, line 455: Should this live in org-roam-macs.el?

zaeph: Definitely.

zaeph: Done in2deb02fdfe6295fd3ddb85ad03353c7fe9d97e0c.

zaeph:
> The linter doesn't seem to be understanding the `pcase` assignments. I initially thought it was a problem with `pcase` not being required, but this wasn't it.

@progfolio Any idea on how to get CI to cooperate?

progfolio:
looks like `org-roam-dailies.el` needs to `(require 'org-roam-macs)`

zaeph:
> looks like `org-roam-dailies.el` needs to `(require 'org-roam-macs)`

That might be it, trying it out in b0cf07c6cd90561e6ccbc1ed3793860039c76ac4.

Pretty sure the CI output wasn't informative, but next time I'll make sure to check the `flycheck` errors.

zaeph:
> That might be it, trying it out in [b0cf07c](https://github.com/org-roam/org-roam/commit/b0cf07c6cd90561e6ccbc1ed3793860039c76ac4).

Welp, that wasn't it.  Back to hunting, then.

org-roam-capture.el
```diff
@@ -279,23 +280,21 @@ This function is used solely in Org-roam's capture templates: see
...
-                 val (pop copy))
-      (if (member key org-roam-capture--template-keywords)
-          (progn
-            (push val org-roam-plist)
-            (push key org-roam-plist))
-        (push key converted)
-        (push val converted)))
-    (append (nreverse converted) `(:org-roam ,org-roam-plist))))
+  (pcase template
+    (`(,key ,description) template)
```
progfolio, line 284:
```suggestion
    (`(,_key ,_description) template)
```
silence byte-compiler warning about unused lexical args

zaeph: Same thing with the line below, I suppose.

progfolio: We're using the variables in the second case, so that should be fine.

org-roam-capture.el
```diff
@@ -281,7 +281,7 @@ This function is used solely in Org-roam's capture templates: see
 (defun org-roam-capture--convert-template (template)
   "Convert TEMPLATE from Org-roam syntax to `org-capture-templates' syntax."
   (pcase template
-    (`(,key ,description) template)
+    (`(,_key ,_description) template)
     (`(,key ,description ,type ,target . ,rest)
```
zaeph, line 285: You beat me to it, but shouldn't we do the same thing for the next line?

zaeph: Apparently not.  I'm dumb, it's because they're clearly unused, here.

zaeph: Couldn't we just use `_`?

progfolio: They're used in `converted`, so byte compiler should not warn.

zaeph: > Couldn't we just use `_`?

Trying it out in a6029564e041495ef432d8e326866781d366e7a2.

zaeph: 🎉

progfolio: We could, prefixing them with the `_` just signals to the byte-compiler that they're ignored variables. The names just help the reader understand the intent.

zaeph: Sure thing, I'll revert.

jethrokuan:
thanks all!

zaeph:
Thanks!

BrettWitty:
Thanks! Amazingly swift resolution! You rock!

PR #586, opened 2020-05-08
(fix): do not store empty-string refs in db

### LINKED ISSUE #585: peculiar error: "UNIQUE constraint failed: refs.ref"
#### Steps to Reproduce

Probably only on my machine with my database, but:

1. org-roam-insert
2. In the capture buffer, finish with C-c C-c.
3. New title does not get inserted at the previous location and the error `peculiar error: "UNIQUE constraint failed: refs.ref"` shows up. However, the new note is created, and when triggering org-roam-insert again and selecting the note in the list, the link is inserted as expected.

I'm happy to help debugging, but I have no experience with emacs/elisp etc. I use ROAM_ALIAS quite a bit, maybe I have a duplicate entry there?

### Versions
- Emacs (`C-h v emacs-version`): 28.0.50
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/d0f5477c323ef6600e29d22b75d876c49916b3f4 (I pulled this from `~/.emacs.d/.cache/quelpa/build`, I hope that's correct)

zaeph:
Fixes #585.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -384,6 +384,7 @@ current buffer is used."
   "Extract the ref from current buffer and return the type and the key of the ref."
   (pcase (cdr (assoc "ROAM_KEY"
                      (org-roam--extract-global-props '("ROAM_KEY"))))
+    ('nil nil)
     ((pred string-empty-p)
      (user-error "ROAM_KEY cannot be empty"))
     (ref
--- a/org-roam.el
+++ b/org-roam.el
@@ -382,15 +382,16 @@ current buffer is used."
 
 (defun org-roam--extract-ref ()
   "Extract the ref from current buffer and return the type and the key of the ref."
-  (if-let* ((ref (cdr (assoc "ROAM_KEY"
-                             (org-roam--extract-global-props '("ROAM_KEY")))))
-            (not-empty (not (string-empty-p ref)))
-            (type (org-roam--ref-type ref))
+  (pcase (cdr (assoc "ROAM_KEY"
+                     (org-roam--extract-global-props '("ROAM_KEY"))))
+    ((pred string-empty-p)
+     (user-error "ROAM_KEY cannot be empty"))
+    (ref
+     (let* ((type (org-roam--ref-type ref))
             (key (cond ((string= "cite" type)
                         (s-chop-prefix (org-roam--cite-prefix ref) ref))
                        (t ref))))
-      (cons type key)
-    nil))
+       (cons type key)))))
 
 (defun org-roam--ref-type (ref)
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -382,12 +382,14 @@ current buffer is used."
...
   "Extract the ref from current buffer and return the type and the key of the ref."
-  (if-let ((ref (cdr (assoc "ROAM_KEY" (org-roam--extract-global-props '("ROAM_KEY"))))))
-      (let* ((type (org-roam--ref-type ref))
-             (key (cond ((string= "cite" type)
-                         (s-chop-prefix (org-roam--cite-prefix ref) ref))
-                        (t ref))))
-        (cons type key))
+  (if-let* ((ref (cdr (assoc "ROAM_KEY"
+                             (org-roam--extract-global-props '("ROAM_KEY")))))
+            (not-empty (not (string-empty-p ref)))
```
zaeph, line 387: I don't think it's particularly good style to define a variable in an `if-let` form which is just meant to short-circuit the condition.  I could probably rewrite it with a `catch` and a thread (`->`), but I don't think it's worth it.

jethrokuan: I think the correct approach is to not have the `#+ROAM_KEY` property at all in the Org file.

zaeph: I thought the  same.  Maybe we could just `user-error` on the empty-string, then?

jethrokuan: yeah, that sounds like a better approach. Could probably add that as one of the checks later on.

jethrokuan: Or catch the `peculiar error` on inserting the ref, Saying that the ref `""` already exists is quite telling.

zaeph: Could work.  I'll try things out when I'm done eating.

zaeph:
> This looks fine.

I'm not a fan of the `('nil nil)`, though:
https://github.com/jethrokuan/org-roam/blob/b81f807300ff28016cce2064dc3ae1b6e1660f16/org-roam.el#L387

jethrokuan:
> > This looks fine.
> 
> I'm not a fan of the `('nil nil)`, though:
> https://github.com/jethrokuan/org-roam/blob/b81f807300ff28016cce2064dc3ae1b6e1660f16/org-roam.el#L387

It's explicit. The alternative is to do a when-let on the `cdr assoc`, which results in an implicit nil.

zaeph:
> It's explicit. The alternative is to do a when-let on the `cdr assoc`, which results in an implicit nil.

I guess that's not a blocker, then.  Let's merge.

Issue #510, opened 2020-04-22
Autocomplete and create new pages


mimischi:
### Brief Abstract

I would like to be able to type `[[`, have it autocomplete to `[[_]]` (`_` is the cursor) and type any page name. If it does not exist, it should be created. Similar to how RoamResearch does it.

### Long Description

RoamResearchs allows to type `[[` and autocompletes it to `[[_]]` (with `_` being your cursor). It then autocompletes a) existing page names and b) allows to create a new page.

(a) Existing pages can already be referenced using [`company-org-roam`](https://github.com/jethrokuan/company-org-roam) just by typing any name. This was addressed in #421. Although this currently does _not_ work by typing `[[`.

(b) If a page does not exist, I have to trigger `org-roam-insert` first. If I am in the `insert` mode of `evil`, I first have to exit to `normal` mode and then type `SPC n i` (with `doom-emacs`) to insert a new link (even to non-existent files). This slows down my typing and I would love to be able to insert pages just as fluently as with RoamResearch.

### Proposed Implementation

As @zaeph has mentioned in the [Slack thread](https://orgroam.slack.com/archives/CV20S23C0/p1587539981023500), one needs to keep track of files that have a) no content and b) are not referenced any longer. Creating a link by typing `[[ foo` (autocompleting to `[[foo]]`) should first create a `foo.org` file. If we then change the link to `[[bar]]`, it should delete `foo.org` (if it has no content) and create `bar.org`. As far as I am aware, this would replicate the behaviour of RoamResearch.

### Please check the following:
- [x] No similar feature requests

zaeph:
> As @zaeph has mentioned in the [Slack thread](https://orgroam.slack.com/archives/CV20S23C0/p1587539981023500), one needs to keep track of files that have a) **no content** and b) are not referenced any longer. [emphasis added]

Just to be clear, I'm not suggesting that we remove files without any content.  We'd have to create a file-metadata tag like `#+ROAM_AUTOCREATED: t` which would allow us to tag those entries, and remove them if they're not referenced anywhere else.

A few points on implementations:
1. Right now, it's possible to have an `org-capture` return immediately with `:immediate-finish t` when using `org-roam-insert`.  This prevents the `org-capture` window from spawning, and reduces the bump in the workflow.
2. We could have a custom link-handler which would create the file upon formatting.  Then, we'd clean up the orphans in `before-file-save`.
3. If we do it like this, I foresee a lag happening upon link creation, which wouldn't be too good.
[...]

PR #592, opened 2020-05-10
(feat): Add custom org-mode link, roam:

alan-carroll:
This is my first ever PR, and my first serious use of git abilities, so let me know if I can improve this whole thing in any way 😄 

###### Summary
Adds custom org-mode link, roam:, which uses org-roam note TITLEs or ALIASes as link path instead of filenames. Allows lazy note links to as-yet-non-existent files and less friction when writing!

[See Discourse topic](https://org-roam.discourse.group/t/custom-roam-style-link/39)
Also #510 

###### Key features
- Show/hide brackets around link (or use a separate face) to differentiate roam-link vs. other links
- Use TITLEs instead of filenames
- Case-insensitive
- Lazy creation of notes -- TITLEs do not need to exist before creating roam-link
- Link face changes if TITLE points to existing / non-existing file
- Update org-roam--extract-links to update org-roam database with custom roam-link backlinks
- Ability to convert file-to-roam or roam-to-file links. Can be done individually on link (default key-binding `M-f` to convert to file and `M-r` to convert to roam), or can be done on the entire buffer at once using `org-roam-convert-buffer-links`. Possible to use on a hook on save/load.
- All roam-link notes in current buffer without existing corresponding files in org-roam database can be automatically created on calling `org-roam-auto-create-buffer` (calling with `C-u` prefix or optional argument allows selecting template per link, but this is ... buggy)
- Add option to insert roam-links instead of file-links when calling `org-roam-insert`
- Completions for roam-links include in-buffer notes that haven't been created yet
- Minibuffer message when point is on roam-link indicating filepath / existence
- Shortcut functions to insert roam-link with completions, or just roam-link syntax at point. Unbound by default (I set mine to `M-r` and `M-[` since these are fairly easy and don't seem to be a default binding for anything else).
- Link works anywhere in org-mode generally (although the org-roam database will not be updated outside of org-roam files).

company-org-roam PR coming soon, I just have to move the advice functions I made in my config into non-advice modifications of the core code!

###### Possible improvements to make
- Update org-roam linter to work with roam-links
- Allow completion of non-existing roam-link TITLEs *across* buffers/files (currently only the active buffer is parsed)
- Add advice/hook on TITLE/ALIAS change to update existing roam-links in other org-roam files
- Add function to scan org-roam files for links to current buffer, and update database (motivation: someone can save a series of notes containing links to note with NEW-TITLE, but without creating the actual note. When they finally do create it, those links DO NOT exist in the database).
- Lazy / nested note creation in org-capture (see #527 #532)
- Intuitive/better/working manual selection of capture templates for different notes during `org-roam-auto-create-buffer`
- Ensuring existence of 'automatic' template when calling `org-roam-auto-create-buffer` without prefix -- current implementation requires a default 'automatic' capture template under 'a' key. Perhaps a variable containing a user-supplied template, and a permanent, basic fall-back template if no user-supplied template exists?
- Add fun roam-link key-map commands, like jumping to next existing/non-existing/any roam-link, or turning word-at-point into a link.
- Enforce TITLE / ALIAS uniqueness? Current schema would require enforcement in org-roam code rather than database itself
- Add a store-link function?
- Using a custom org-mode link allows for custom mouse-over tool-tips and other custom functionality that you otherwise wouldn't implement on top of standard file-links. Any ideas/requests?


org-roam.el
```diff
@@ -251,12 +259,17 @@ it as FILE-PATH."
...
                                   (require 'org-ref nil t)
                                   (-contains? org-ref-cite-types type))
                                  "cite")
-                                (t nil))))
-          (when link-type
+                                (t nil)))
+               (is-title (if (string= type "roam") t nil))
+               (roam-file (if is-title (org-roam--get-file-from-title path) t)))
+          ;; Ignore link if it is a roam-link and roam-file doesn't exist
+          (when (and link-type roam-file)
```
jethrokuan, line 272: can be simplified to 

```
(when (and link-type (string= link-type "roam"))
 ...
```

org-roam.el
```diff
@@ -251,12 +259,17 @@ it as FILE-PATH."
                (link-type (cond ((and (string= type "file")
                                       (org-roam--org-file-p path))
                                  "roam")
```
jethrokuan, line 261: might want to change this away from "roam" into something like "org", just to prevent confusion down the road.

org-roam.el
```diff
@@ -821,6 +851,321 @@ for Org-ref cite links."
...
+  "Call `org-roam-capture' with a template using :immediate-finish t.
+TITLE is the title for the file to be created.
+MANUAL is boolean allowing manual selection of capture template(s)."
+  (let ((org-roam-capture--info (list (cons 'title title)
+                                      (cons 'slug (org-roam--title-to-slug title))))
+        (org-roam-capture--context 'title))
+    (if manual (org-roam-capture--capture)
+      (org-roam-capture--capture :keys "a"))))
+
+(defun org-roam-auto-create-buffer (&optional manual)
```
jethrokuan, line 1061: Same as above.

org-roam.el
```diff
@@ -821,6 +851,321 @@ for Org-ref cite links."
...
+             (type (org-element-property :type elem)))
+        (if (and (string= file-or-roam "roam")
+                 (string= type "file"))
+            (org-roam--convert-file-to-roam-link elem)
+          (when (and (string= file-or-roam "file")
+                     (string= type "roam"))
+            (org-roam--convert-roam-to-file-link elem)))
+        (org-next-link)))))
+
+(defun org-roam--auto-create-file (title &optional manual)
```
jethrokuan, line 1051: This function should be renamed, it's not immediately clear what auto-create-file does.

org-roam.el
```diff
@@ -821,6 +851,321 @@ for Org-ref cite links."
...
+             (when (string= type "roam")
+               (cons path res)))))
+       (-flatten)
+       (-distinct)))
+
+(defun org-roam--roam-link-completion (&optional _arg)
+  "Completion for roam-links in `org-mode'.
+ARG is optional prefix supplied through `org-mode'"
+  (let ((completions (--> (org-roam--get-title-path-completions)))
+        (in-buffer-completions (--> (org-roam--current-buffer-roam-link-titles))))
```
jethrokuan, line 1093:
```suggestion
  (let ((completions (org-roam--get-title-path-completions))
        (in-buffer-completions (org-roam--current-buffer-roam-link-titles)))
```

Unused threading macros here.

org-roam.el
```diff
@@ -821,6 +851,321 @@ for Org-ref cite links."
...
+            (run-with-idle-timer 0.5 t #'org-roam--roam-link-message))))
+
+(defun org-roam-cancel-link-messages ()
+  "Disable minibuffer status message for roam-links."
+  (interactive)
+  (cancel-timer org-roam-link-message-timer)
+  (setq org-roam-link-message-timer nil))
+
+;; Turn roam-link status message on by default
+(org-roam-show-link-messages)
```
jethrokuan, line 1132: I'm not sure we need this timer machinery here. We also have a variable `org-roam-verbose`, which we should use instead to control whether the messages are shown.

org-roam.el
```diff
@@ -821,6 +851,321 @@ for Org-ref cite links."
...
+ :keymap (let ((map (copy-keymap org-mouse-map)))
+           (define-key map (kbd "M-f") 'org-roam-convert-roam-to-file-link)
+           (define-key map (kbd "<tab>") 'org-roam--exit-roam-link)
+           map))
+
+(org-link-set-parameters
+ "file"
+ :keymap (let ((map (copy-keymap org-mouse-map)))
+           (define-key map (kbd "M-r") 'org-roam-convert-file-to-roam-link)
+           map))
```
jethrokuan, line 1167: We should move this such that this is activated within `org-roam-mode`, and reset when the mode is deactivated, rather than make a lasting change when requiring org-roam.

org-roam.el
jethrokuan: #593

org-roam.el
alan-carroll: This is required to ensure the roam-link points to an existing file, since db enforces 'file' to be non-null:
`(roam-file (if is-title (org-roam--get-file-from-title path) t)))`

`(is-title (if (string= type "roam") t nil))` was partly to reduce verbosity, and line 288 uses it to differentiate filename processing between roam- and file-links (which are both passed as link-type "roam").

org-roam.el
alan-carroll: I'm not sure I understand the change you're suggesting

org-roam.el
alan-carroll: The timer is lifted straight from org-ref and how it persistently shows the bibtex entry in the minibuffer on a cite-link (or a note that is doesn't exist). Without it, the link message disappears from the minibuffer fairly quickly. It seemed desirable to have it be persistent while on a link if you were using the message to resolve a filename, or a note title if you added a link description.

If we do remove the timer I would suggest changing/removing the link message functionality also.

I'll look into `org-roam-verbose`!

org-roam.el
alan-carroll: Tried to simplify by introducing "org" type. Updated org-roam-db in two places to work with the change. Don't think it affects anything else?

org-roam.el
alan-carroll: Updated

org-roam.el
alan-carroll: Updated

org-roam.el
alan-carroll: Updated

org-roam.el
alan-carroll: The timer is now started on `org-roam-mode` activation when `org-roam-verbose` is `t`.

org-roam.el
alan-carroll: Moved everything into `org-roam-mode`, but I don't know if I'm not understanding how mode activation/de-activation works. On my Doom setup everything loads like I would expect even in just a normal org file. On manual de-activation the settings are reset as expected, but manaully re-enabling `org-roam-mode` DOES NOT re-activate `roam:` links with all of its faces and functions.

zaeph:
Just a note to let you know that I'll take the time to review everything next week.

org-roam.el
alan-carroll: Updated again to what I think is more along the lines of what you meant with `org-roam-verbose`

progfolio:
The bulk of this could probably stand to be organized into its own module. Something like `org-roam-link.el`

org-roam.el
jethrokuan: ah I see, I thought it was a simple message, in which case it's fine, but also good to probably disable that if `org-roam-verbose` is nil.

TimQuelch:
Those gifs in the discourse look really good and this looks like a fairly useful improvement to the UI.

One thing that I am confused about is why do we actually need the new `roam:` link type
- What does the new link type do that we can't do with `file:` links?
- If we can convert freely between `roam:` links and `file:` links, couldn't we just use `file:` links in the first place and save complexity and compatibility?
- Doesn't `org-roam-insert` already allow the insertion of links to files that don't exist yet? And with a capture template with `immediate-finish` set it can create this link in the background.

I haven't really been following the threads on this and I think I'm misunderstanding something about what this is trying to achieve. There's obviously a reason behind this (it seems like everyone is onboard) so I was just wondering if you could help my misunderstanding.

org-roam.el
```diff
@@ -823,6 +853,301 @@ for Org-ref cite links."
...
+
+(defun org-roam-convert-roam-to-file-link ()
+  "Convert a custom roam-link to standard org file-link."
+  (interactive)
+  (let ((elem (org-element-context)))
+    (when (and (eq 'link (car elem))
+               (string= "roam" (org-element-property :type elem)))
+      (org-roam--convert-roam-to-file-link elem))))
+
+(defun org-roam-convert-file-to-roam-link ()
```
rsperko, line 1024: I was thinking of this work as primarily a way to lazy create links, much like a true Wiki. When would someone want to change an instantiated link back into a roam link?

alan-carroll:
> Those gifs in the discourse look really good and this looks like a fairly useful improvement to the UI.
> 
> One thing that I am confused about is why do we actually need the new `roam:` link type
> 
> * What does the new link type do that we can't do with `file:` links?
> * If we can convert freely between `roam:` links and `file:` links, couldn't we just use `file:` links in the first place and save complexity and compatibility?
> * Doesn't `org-roam-insert` already allow the insertion of links to files that don't exist yet? And with a capture template with `immediate-finish` set it can create this link in the background.
> 
> I haven't really been following the threads on this and I think I'm misunderstanding something about what this is trying to achieve. There's obviously a reason behind this (it seems like everyone is onboard) so I was just wondering if you could help my misunderstanding.

You are correct that much of this is otherwise syntactic sugar and is not adding something you could not inherently do with existing org-roam functions. It is trying to remove some of the friction associated with those functions though, and more closely mimic the sheer ease of inserting new/existing page links in actual Roam.

The `roam:` link means you can semantically insert and change links as you write, because they use the `TITLE` rather than the filename. There is no way to achieve this with the `file:` link without something that would transform text as you write (and then there is no way to change it without simply deleting the link). Inserting / deleting / re-inserting multiple file-links is a point of friction. It might be relatively small, but nonetheless you cannot write out and subsequently edit a sentence/paragraph with multiple existing and new links without interrupting your actual writing process quite a bit. Some people I'm sure don't care or notice, YMMV.

The conversion between `roam:` and `file:` is there because it's easy to implement and I can foresee use-cases where someone wants to write in `roam:` but save/export in `file:`. In the spirit of plain-text, your links are not at the mercy of continuing to use org-roam and its database! However there's no way to extrapolate this to make file-links more convenient within org-roam. It would be possible to write something that immediately switched `roam:` to `file:` on link completion, and something that switched `file:` to `roam:` on edit, but that still requires existence of the new link-type (and requires editing the functionality of file-links, whereas the conversion I've written is just a keyboard shortcut you can override).

`org-roam-insert` does happily let you create new notes, and adding `:immediate-finish` does reduce writing interruption, but it is still an interruption nevertheless. Depending on how much editing a user does it also risks polluting their note-space with empty notes that they ended up removing the links to or changing the `TITLEs` for. Also, although the functionality is really not implemented yet, using lazily created `roam:` links allows for the possibility of batch-creating notes using different templates for each of the notes (and additional nested captures eventually). In this way you can write out a paragraph without interrupting your thought process, and only when you're done then switch to focusing on exactly how you want to create/elaborate on new notes.

Although I don't have any immediate examples of what or why, creating a new org link type means you can add on new functionality without disrupting the functionality of the built-in `file:` type (what org-ref does for example). Basically it opens the door for exploration and expansion, while file-links will always need to simply be file-links.

I made it for myself because I wanted to, but enough people liked it I made the PR 🙂 It shouldn't interfere with simply using `file:` links like org-roam already does in any way though, if that's what you prefer!

org-roam.el
alan-carroll: For easy editing. It's likely a narrow use case, but there's I don't think prescribing others' usage is good practice when the functionality is so easy to include. If it can convert one way, it should be able to convert back.

org-roam.el
```diff
@@ -823,6 +853,301 @@ for Org-ref cite links."
        :link        (format "file:%s" (abbreviate-file-name buffer-file-name))
        :description title))))
 
+;;; Custom org-link, roam:
+(defcustom org-roam-show-roam-brackets nil
+  "Whether to show brackets around [[roam:]] links."
```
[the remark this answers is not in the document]
alan-carroll: It's possible. This was a personal aesthetic and so far no one has said they like it too. Would you still want it moved into a separate module if the bulk of the PR was moved to a new file like `org-roam-link.el`?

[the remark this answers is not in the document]
rsperko: I am unlikely to use it, think of org-roam as a  zettelkasten extension of org more than an implementation of roam. I can understand the desire though.

TimQuelch:
> The `roam:` link means you can semantically insert and change links as you write, because they use the `TITLE` rather than the filename. There is no way to achieve this with the `file:` link without something that would transform text as you write (and then there is no way to change it without simply deleting the link). Inserting / deleting / re-inserting multiple file-links is a point of friction. It might be relatively small, but nonetheless you cannot write out and subsequently edit a sentence/paragraph with multiple existing and new links without interrupting your actual writing process quite a bit. Some people I'm sure don't care or notice, YMMV.

Oh right that makes a lot of sense. 

> `org-roam-insert` does happily let you create new notes, and adding `:immediate-finish` does reduce writing interruption, but it is still an interruption nevertheless. Depending on how much editing a user does it also risks polluting their note-space with empty notes that they ended up removing the links to or changing the `TITLEs` for. Also, although the functionality is really not implemented yet, using lazily created `roam:` links allows for the possibility of batch-creating notes using different templates for each of the notes (and additional nested captures eventually). In this way you can write out a paragraph without interrupting your thought process, and only when you're done then switch to focusing on exactly how you want to create/elaborate on new notes (you also have the option to NOT create them right away -- if I know I'm still working on something but want to put it away for the day, I can save it without worrying about creating/deleting notes until I want to).

Ah I see. I didn't realise that with this new link that now files wouldn't be created until they were actually visited. That is definitely very useful.

Apologies if my previous comment came across as a bit "what's the point", I just misunderstood. I can definitely see the value this feature would add!

alan-carroll:
No problem! They're good questions 😃

alan-carroll:
Yes the linter is very upset 😄 I'll fix it tomorrow afternoon, it's late here!

jethrokuan:
@alan-carroll do you mind me committing on top of this? I'll leave the core functionality alone, just need to do some misc cleanups.

alan-carroll:
@progfolio I updated the function names, let me know if things still look out of order
@jethrokuan Feel free to!

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+
+(defun org-roam-link-convert-roam-to-file-link ()
+  "Convert a custom roam-link to standard org file-link."
+  (interactive)
+  (let ((elem (org-element-context)))
+    (when (and (eq 'link (car elem))
+               (string= "roam" (org-element-property :type elem)))
+      (org-roam-link--convert-roam-to-file-link elem))))
+
+(defun org-roam-link-convert-file-to-roam-link ()
```
progfolio, line 212: These functions should have tests implemented if we're mutating user's data.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+                     (string= type "roam"))
+            (org-roam-link--convert-roam-to-file-link elem)))
+        (org-next-link)))))
+
+(defun org-roam-link--auto-create-roam-link-file (title &optional manual)
+  "Call `org-roam-capture' with a template using :immediate-finish t.
+TITLE is the title for the file to be created.
+MANUAL is boolean allowing manual selection of capture template(s)."
+  (let ((org-roam-capture--info (list (cons 'title title)
+                                      (cons 'slug (org-roam--title-to-slug title))))
```
progfolio, line 246: backquote/dotted pair notation would be good here.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+        (org-roam-capture--context 'title))
+    (if manual (org-roam-capture--capture)
+      (org-roam-capture--capture :keys "a"))))
+
+(defun org-roam-link-auto-create-roam-links-in-buffer (&optional manual)
+  "Create all non-existant roam-link files in current buffer.
+MANUAL is boolean which allows manual selection of `org-roam-capture'
+templates for each file created.
+If called with PREFIX `C-u' then manual is non-nil."
```
progfolio, line 255:
```suggestion
  "Create all non-existent roam-link files in current buffer.
If MANUAL is non-nil, prompt for template with `org-roam-capture'."
```
No need to mention the universal arg in this case.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+                  (path (org-element-property :path link))
+                  res)
+             (when (string= type "roam")
+               (cons path res)))))
+       (-flatten)
+       (-distinct)))
+
+(defun org-roam-link--completion (&optional _arg)
+  "Completion for roam-links in `org-mode'.
+ARG is optional prefix supplied through `org-mode'"
```
progfolio, line 281: if `arg` is not being used, it can be ignored/renamed to `_` and removed from docstring.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+  (let ((completions (org-roam--get-title-path-completions))
+        (in-buffer-completions (org-roam-link--current-buffer-roam-link-titles)))
+    (format "roam:%s" (completing-read "Roam note: "
+                                       (-union (map-keys completions) in-buffer-completions)))))
+
+(defun org-roam-link--status-message ()
+  "Send roam-link status message to minibuffer."
```
progfolio, line 288:
```suggestion
  "Print roam-link status message in the minibuffer."
```

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+                  (when abs-file-path
+                    (string-match (format "%s\\(.*\\)" org-roam-directory) abs-file-path)))
+                 (file-path (when rel-file-path
+                              (concat "~org-roam-dir~" (match-string 1 abs-file-path))))
+                 (raw-link (org-element-property :raw-link elem)))
+            (if file-path
+                (org-roam-message "file: %s → %s" file-path raw-link)
+              (org-roam-message "No file found in db → %s" raw-link))))))))
+
+(defun org-roam-link-show-link-messages ()
```
progfolio, line 307:
```suggestion
(defun org-roam-link-show-messages ()
```

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+  "Enable minibuffer status message for roam-links.
+Follows example of `org-ref' and displays on idle timer."
+  (interactive)
+  (when org-roam-verbose
+    (or org-roam-link-message-timer
+        (setq org-roam-link-message-timer
+              (run-with-idle-timer 0.5 t #'org-roam-link--status-message)))))
+
+(defun org-roam-link-cancel-link-messages ()
```
progfolio, line 316:
```suggestion
(defun org-roam-link-cancel-messages ()
```

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+  "Shortcut to insert roam-link with standard completion prompt."
+  (interactive)
+  (insert "[[" (org-roam-link--completion) "]]"))
+
+(defun org-roam-link-insert-syntax ()
+  "Shortcut to insert roam-link syntax without completion.
+Move point inside brackets, ready for roam title entry."
+  (interactive)
+  (insert "[[roam:]]")
+  (backward-char 2))
```
progfolio, line 332: These two commands could stand to be combined into one (`org-roam-link-insert`) which changes behavior if called with a universal argument. The common case should be the default, the less common case should require the universal argument.

e.g. 
```emacs-lisp
(defun org-roam-link-insert (&optional completion)
    "Insert a :roam link.
If COMPLETION is non-nil,..."
; etc
)
```

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+
+(defun org-roam-link-exit-link ()
```
progfolio, line 334: Maybe a better name for this function. "Exit" is a little ambiguous, and the docstring suggests it advances point forward.

org-roam.el
```diff
@@ -535,15 +566,28 @@ If DESCRIPTION is provided, use this as the link label. See
                            (if filter-fn
                                (funcall filter-fn it)
                              it)))
-         (title (org-roam-completion--completing-read "File: " completions
-                                                      :initial-input region-text))
+         (in-buffer-completions (--> (org-roam-link--current-buffer-roam-link-titles)
+                                     (if filter-fn
+                                         (funcall filter-fn it)
+                                       it)))
```
progfolio, line 572:
```suggestion
         (in-buffer-completions (funcall (or filter-fn #'identity)
                                  (org-roam-link--current-buffer-roam-link-titles)))
```

org-roam.el
```diff
@@ -535,15 +566,28 @@ If DESCRIPTION is provided, use this as the link label. See
...
+    (if org-roam-link-use-roam-links
+        (progn
+          (when region ;; Remove previously selected text.
+            (delete-region (car region) (cdr region)))
+          (insert (format "[[roam:%s]]" title)))
+      (if (and target-file-path
+               (file-exists-p target-file-path))
         (progn
           (when region ;; Remove previously selected text.
             (delete-region (car region) (cdr region)))
```
progfolio, line 593: a `cond` may be clearer here

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+         (match-end 1)
+         (if org-roam-link-show-brackets
+             '(face org-roam-link-brackets invisible nil)
+           '(face org-roam-link-brackets invisible t)))
+        (add-text-properties
+         (match-beginning 3)
+         (match-end 3)
+         (if org-roam-link-show-brackets
+             '(face org-roam-link-brackets invisible nil)
+           '(face org-roam-link-brackets invisible t)))
```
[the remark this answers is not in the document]
alan-carroll: I'll double-check that it fails, but I remember trying to set `invisible` with a boolean var and it absolutely refused to work.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+             (match-end 2)
+             '(invisible t))
+          ;; If descriptive, hide roam: and path. Optionally show single brackets
+          (progn (if org-roam-link-show-brackets
+                     (add-text-properties (+ start 1) (- end 1) '(invisible t))
+                   (add-text-properties start end '(invisible t)))
+                 (save-match-data
+                   (goto-char start)
+                   (re-search-forward org-link-bracket-re end t)
+                   (add-text-properties (match-beginning 2) (match-end 2) '(invisible nil)))))))))
```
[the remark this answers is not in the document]
alan-carroll: Yes at this point the link MUST be a description link with full brackets. Start/end are directly passed by org to cover the link element, and it uses org's built-in regexp `org-link-bracket-re` so a match is guaranteed. I just need it for the regexp grouping.

org-roam-link.el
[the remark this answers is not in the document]
progfolio: Note that the list is backquoted, so that `org-roam-link-show-brackets' is evaluated in that list.

Similarly you could try something like this in a buffer:
```emacs-lisp
(let ((undelined nil))
  (save-excursion)
  (goto-char (point-max))
  (insert (propertize "test" 'face `(:underline ,undelined))))
```
swapping the value of `underlined` will change the `:undeline` property.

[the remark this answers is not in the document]
alan-carroll: "Today I learned ... "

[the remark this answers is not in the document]
progfolio: It's a very useful feature:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Backquote.html#Backquote

The list splicing syntax in that manual node is handy, too.

org-roam-link.el
alan-carroll: Any suggestions? `forward` in function names seems to typically imply it will also move to the next instance, which this does not. 

How about `org-roam-link-advance`?

progfolio: Looking at it, I'm not sure we need this command.
What's the expected use case? Is it to be used after filling out a link's body?
If so, we should just have the command that inserts the link use a prompt and move point forward after the insertion.

org-roam-link.el
```diff
@@ -0,0 +1,341 @@
...
+  "Shortcut to insert roam-link syntax without completion.
+Move point inside brackets, ready for roam title entry."
+  (interactive)
+  (insert "[[roam:]]")
+  (backward-char 2))
+
+(defun org-roam-link-exit-link ()
+  "When point is in roam-link, advance point forward out of the link."
+  (interactive)
+  (goto-char (org-element-property :end (org-element-context))))
```
progfolio, line 337:
```suggestion
(defun org-roam-link-insert (&optional title)
  "Insert an org-roam link with TITLE."
  (interactive)
  (insert  "[[" (or title (org-roam-link--completion)) "]]")
  (goto-char (org-element-property :end (org-element-context))))
```
This allows for programatically creating links (by passing the title argument).
It also obviates the need for the exit function.
The user is prompted the same way they would be for `org-roam-insert`.
They can then choose a completion candidate or insert text that does not match a candidate to create a new link.

org-roam-link.el
alan-carroll: So the one that uses a prompt already does that. This is purely convenience for moving the point out of a link if you aren't entering it through a prompt, which I do quite often.

It really is just a personal use case, so I'll remove it from the code here.

progfolio: See the suggested change I made to those three commands.
It will make `org-roam-insert` and `org-roam-insert-link` consistent.

org-roam-link.el
alan-carroll: This is fine. I'll make my own shortcuts since they really don't belong in the source code. I had to add an `if` clause though since `org-roam-link--completion` returns with `roam:`, but `title` still needs it.

With `insert` the `goto-char` is unnecessary also.

progfolio:
Looks good to me so long as CI passes after resolving conflicts.

progfolio:
@alan-carroll: I took care of resolving the conflicts w master branch for you. Just please double check to make sure I caught everything, as a lot had changed w `org-roam.el` between start of this pull and current master. In the future, you can always rebase your branch on master as you work on it. That way you catch all the conflicts as they happen instead of having to deal w them all at the end. Makes it easier when there is a lot of simultaneous development.

Thanks again for your work on this. Looking forward to seeing it land.

alan-carroll:
Changing:
```
(let ((org-roam-capture--info (list (cons 'title title)
                                    (cons 'slug (org-roam--title-to-slug title))))
```
to:
```
(let ((org-roam-capture--info `(('title . ,title)
                                ('slug  . ,(org-roam--title-to-slug title))))
```
is producing unexpected behavior with `org-roam-link--find-file` and `org-roam-insert` for titles that don't have existing files (`org-roam-find-file` appears to work fine). 

The expected behavior is to open capture template selection, then fill `TITLE` section with the input title (either prompt in `org-roam-insert` or the `roam:` path with `org-roam-link--find-file`).

The actual behavior is template selection, then a prompt for `slug: ` followed by a prompt for `title: ` which overrides the original TITLE input.

#### Minimal example:
With the back-quote change in-place, and `test-file.org` doesn't exist, run:
`(org-roam-link--find-file "test-file")`
Select `d` for `default` template. Instead of opening `org-capture` with the `TITLE` filled, it prompts for `slug: ` and then `title: `

Revert back to `(list (cons ...` version and run `(org-roam-link--find-file "test-file")`
Select `d` for `default` template. You now get the expected behavior.

With `(setq org-roam-link-enabled nil)` to insert `file:` links, call `org-roam-insert` and at the prompt pass `test-file` (assuming test-file doesn't exist). It presents prompts for `slug: ` and `title: ` also.

alan-carroll:
I'm not understanding this problem. I don't have the same issue on the current master branch with `org-roam-insert`, so it is something here. I'll be away for the next couple of days; if no one figures out what's going on, I'll keep poking at it after I get back 😄 

Otherwise I think everything else is good. I still need to test `org-roam-link-auto-create-roam-links-in-buffer` with the changes, but that depends on fixing the issue above.

On the same note, `org-roam-link--auto-create-roam-link-file` presumes an 'auto' capture template under the "a" key, which I have added by default. If a user changes the templates and keys, it will have problems or unexpected behavior. I think the ability to create all of the links in a file is desirable, but obviously this needs to allow user-customization without breaking. I should also remove `roam` from the function names.

Any suggestions? I'm thinking just define a customizable var the user can set for which key their preferred 'auto' template is under, and throw some helpful errors if they don't have it set properly when calling the function.

progfolio:
Just eyeballing it:
``` emacs-lisp
(let ((org-roam-capture--info `((title . ,title)
                                (slug  . ,(org-roam--title-to-slug title))))
```
Symbols in car position were quoted within a quoted list.
Will test when at computer

org-roam-capture.el
```diff
@@ -80,7 +80,13 @@ note with the given `ref'.")
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")
 
 (defvar org-roam-capture-templates
-  '(("d" "default" plain (function org-roam-capture--get-point)
+  '(("a" "Automatic" plain (function org-roam--capture-get-point)
```
jethrokuan, line 83: the automatic template might want to be moved out of `org-roam-capture-templates`, since it's unlikely the user wants to use the automatic template in a normal insert/find-file command.

alan-carroll: Yes I agree. That’s why I think it would be good to just let a user define their own template key for auto-creation and throw errors if they don’t have one set when calling the auto-create functions. 

It may also be possible to move this default auto-template to its own var and use it for auto-creation unless a user sets their own key. 

Either is fine with me!

alan-carroll:
@jethrokuan I think this kind of modularity in org-roam sounds great, and would arguably significantly reduce overall maintenance work down the road.

It also sounds like it should make adding in links to headlines as a feature much simpler. Possibly reduce the threshold for others to contribute new ideas, too. 

As arguments against refactoring, though, I am mostly out of ideas for what next new ‘link’ feature to add 😂 I’m sure something will come up! I’m also inexperienced with modular design and elisp generally, so I don’t know how much effort will be needed to make truly extending caching ‘easy’. 

I’m open to it though!

org-roam-capture.el
alan-carroll: See the new commits

alan-carroll:
I've updated `org-roam-link-auto-create-links-in-buffer` and friends so that:
- A default "auto" template exists outside of the normal `org-roam-capture-templates`. This means the function will work no matter what a user does with `org-roam-capture-templates`. The "auto" template also does not appear in the normal template selection. A user can also customize the "auto" template to their liking
- A var `org-roam-link-auto-template-key` can hold a template key to be used with `org-roam-capture-templates` instead of using the default "auto" template. If a user has an auto template in their regular templates they would like to use, this allows them to specify it without duplicating the template in two places.

Selecting different templates for different files while allowing for manual input during capture was something I wanted to allow, but it is way more difficult than I was hoping (and the potential solutions would likely interfere with nested captures if that feature is eventually finished). 

Essentially it asks for all of the template selections up front, and then only the initial capture is actually saved (same problem as nested capture). It doesn't seem like elisp is very accommodating for pausing function execution to wait on an org-capture hook, so the only thing I can think of is a capture-buffer local hook calling an iterator until it reaches the end of the link list and removes itself (or some mash up of global variables doing something similar).

This would certainly have problems with nested capture. If energy were to go into this problem at all I think it would be best spent trying to create an aggregation of captures to perform, and then let the user inspect and execute them at their leisure.

I've left in manual template selection with `C-u` (since it **does** work if the templates selected use `:immediate-finish t`). It recenters and highlights the link being used for capture so the user can determine which template to use.

Let me know if anyone has thoughts on these things.

@jethrokuan Unless you want to move forward with the modular refactor first, I think I am out of things to work on here! Next up for me is to either help with modularizing where I can, or continuing with link-tags.

samspo:
I don't know if this idea is abandoned. I really like the possibility of having a specific link to roam. It allows elegant access to roam files from outside roam (but also inside roam:  roam:MyFile is nicer that file:MyFile.txt ). I also think we should have the choice to link to the title, an alia or the filename. Ideally, we could also choose the name of the link (to use wiki: for example instead of roam, which would facilitate compatibility with other previously used programs).

zaeph:
It's not abandoned, and it was held up because we wanted to get headlines support first.  I'll finally be able to review this next week

jethrokuan:
Superceded by #1105

PR #593, opened 2020-05-10
(internal): add `org-roam-message`

jethrokuan:

This macro simplifies message printing, so we can respect
`org-roam-verbose` in more places. Also silences more messages when
`org-roam-verbose` set to false.

### DIFF
--- a/org-roam-macs.el
+++ b/org-roam-macs.el
@@ -33,6 +33,8 @@
 ;;; Code:
 ;;;; Library Requires
 
+(defvar org-roam-verbose)
+
 (defmacro org-roam--with-temp-buffer (&rest body)
   "Execute BODY within a temp buffer.
 Like `with-temp-buffer', but propagates `org-roam-directory'."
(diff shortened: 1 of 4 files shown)


org-roam-macs.el
```diff
@@ -56,7 +58,10 @@ to look.
                         (error-message-string err)
                         ,templates))))
 
-
+(defmacro org-roam-message (msg)
+  "Message MSG when `org-roam-verbose' is true."
+  `(when org-roam-verbose
+     (message msg)))
```
zaeph, line 64: I think you need to expand `msg` with `,` here.

zaeph:
Could we maybe add a prefix to all those messages to identify them more easily?
```el
(message "(org-roam) %s" msg)
```

org-roam-macs.el
jethrokuan: i made that change locally, not sure why it didn't show up here!

zaeph:
LGTM!

PR #595, opened 2020-05-10
"Refactor org-roam-message as a function"

progfolio:
Add to commentary that org-roam-macs may contain other utility
functions (similar to org-macs).

Unless we need to use a macro, we should not.

### DIFF
--- a/org-roam-macs.el
+++ b/org-roam-macs.el
@@ -62,7 +62,7 @@ to look.
 (defun org-roam-message (format-string &rest args)
   "Pass FORMAT-STRING and ARGS to `message' when `org-roam-verbose' is t."
   (when org-roam-verbose
-    (apply #'message `(,(concat "(org-roam) " format-string) ,@args))))
+    (funcall #'message (concat "(org-roam) " format-string) args)))
 
 (provide 'org-roam-macs)
 
--- a/org-roam-macs.el
+++ b/org-roam-macs.el
@@ -27,7 +27,7 @@
 
 ;;; Commentary:
 ;;
-;; This library implements macros used throughout org-roam
+;; This library implements macros and utiltiy functions used throughout org-roam
 ;;
 ;;
 ;;; Code:
@@ -58,11 +58,10 @@ to look.
                         (error-message-string err)
                         ,templates))))
 
-(defmacro org-roam-message (format-string &rest args)
-  "Message MSG with ARGS when `org-roam-verbose' is true."
-  (declare (indent 0) (debug t))
-  `(when org-roam-verbose
-     (message (concat "(org-roam) " ,format-string) ,@args)))
+(defun org-roam-messgae (format-string &rest args)
+  "Pass FORMAT-STRING and ARGS to `message' when `org-roam-verbose' is t."
+  (when org-roam-verbose
+    (message (concat "(org-roam) " format-string args))))
 
 (provide 'org-roam-macs)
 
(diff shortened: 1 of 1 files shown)


org-roam-macs.el
```diff
@@ -58,11 +58,10 @@ to look.
                         (error-message-string err)
                         ,templates))))
 
-(defmacro org-roam-message (format-string &rest args)
-  "Message MSG with ARGS when `org-roam-verbose' is true."
-  (declare (indent 0) (debug t))
```
zaeph, line 63: Any reason to delete the `declare` form?  Was it because it was the default?

zaeph: Yeah, I think it's the default, so we can avoid it indeed.

progfolio: `org-roam-message` is a function now. No need for the declare form.

progfolio:
It's a slippery word!

zaeph:
Not sure what's going on with CI and... Oh.

zaeph:
There seems to be some problems with the new definition of `org-roam-message`.

zaeph:
And here I was rg'ing to see what we might have broken.  Should have just looked at the macro instead.  Oh well. ¯\_(ツ)_/¯
I think we can merge.

progfolio:
Sorry, thought I could handle this with my phone. Not suited for programming.

zaeph:
> Sorry, thought I could handle this with my phone. Not suited for programming.

It's in those moments that you realise how convenient it is to have a mode that handle the parens for you in Emacs.

progfolio:
It's the way args is being applied. Will fix in a minute.

org-roam-macs.el
```diff
@@ -62,7 +62,7 @@ to look.
 (defun org-roam-message (format-string &rest args)
   "Pass FORMAT-STRING and ARGS to `message' when `org-roam-verbose' is t."
   (when org-roam-verbose
-    (message (concat "(org-roam) " format-string) args)))
+    (apply #'message `(,(concat "(org-roam) " format-string) ,@args))))
```
zaeph, line 65: Couldn't we do a `funcall` instead of `,@`?

zaeph: Done in 0807252d64230840faaa8687fee8399a08e3f8b4.

zaeph:
Welp.  I think we can go back to `apply` then, but it's weird.

progfolio:
`&rest` packs the args into a list.
`funcall` would then pass that list as single argument which would get packed into another list by `message`.

progfolio:
so message receives: `(message "the string" '((args)))`

zaeph:
Right, it makes total sense.  Are we good on the merge?

progfolio:
When we squash commits before merging, we should clean up commit message to avoid the noise of "fix typo" etc. Lends itself to cleaner history.

zaeph:
Yeah, we haven't been paying too much attention to those, since they're just in the commit-message.  Honestly, at the beginning here, I wanted to rewrite the history for all the PRs to clean them up and have informative commit messages, but I gave up along the way.  I'll try to be more diligent in the future.

jethrokuan:
> When we squash commits before merging, we should clean up commit message to avoid the noise of "fix typo" etc. Lends itself to cleaner history.

Yeah, you can just do so in the GH interface. Typically my first commit has a more meaningful commit message, and I just keep that one.

PR #600, opened 2020-05-12
(fix): handle creation of org-roam-index-file

### LINKED ISSUE #597: org-roam-jump-to-index not working properly
#### Steps to Reproduce
1. Basic org-roam settings:
```lisp
(setq org-roam-directory "~/source/Org/Roam" 
          org-roam-index-file "index.org")
```

2. `M-x` `org-roam-jump-to-index`

>   Would you like to create it? (y or n)
Enter "y", press `C-c C-c`

- Close buffer

3. `M-x` `org-roam-jump-to-index`

>   Would you like to create it? (y or n)

#### Expected Results
The 2nd time around the file should be found. It exists, it's placed in `~/source/Org/Roam/index.org`

It can be found with `M-x` `org-roam-find-file` `Index`

#### Actual Results
I get told the file does not exist yet sand needs to be created every time.

### Versions
- Emacs (`C-h v emacs-version`): 28.0.50

Don't know how to get that easily.
I use straight.el with use-package and there are created several directories in various places for each package.
In `~/.config/emacs/elpa/` it shows `org-roam-20200510.2317`

zaeph:
Clearly not the best implementation we could to, but it works.

Fixes #597.

-----

More work to come, especially there's no freedom in the creation of the index file, even if it's supposed to be a one-time transaction.

We're investigating either to replace the `:file-name` parameter in all templates to allow user freedom, or if we should rethink the way `org-roam-capture--info` works.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -596,16 +596,8 @@ command will offer you to create one."
              (file-exists-p index))
         (find-file index)
       (when (y-or-n-p "Index file does not exist.  Would you like to create it? ")
-        (let ((org-roam-capture--context 'title)
-              (org-roam-capture-templates `(("i" "index" plain
-                                             (function org-roam-capture--get-point)
-                                             "%?"
-                                             :file-name ,(-> org-roam-index-file
-                                                             (file-name-nondirectory)
-                                                             (file-name-sans-extension))
-                                             :head "#+TITLE: Index\n\n"
-                                             :unnarrowed t))))
-          (org-roam-capture--capture))))))
+        (find-file (org-roam--get-index-path))
+        (insert "#+TITLE: Index\n\n")))))
 
 ;;;; org-roam-find-ref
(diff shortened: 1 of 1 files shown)


zaeph:
> This looks like the simplest solution.

I agree.  I'll use this PR to also support `org-roam-file-extensions` by making `org-roam-index-file` extension-agnostic, and then we can merge.

org-roam.el
```diff
@@ -596,7 +596,8 @@ command will offer you to create one."
              (file-exists-p index))
         (find-file index)
       (when (y-or-n-p "Index file does not exist.  Would you like to create it? ")
-        (org-roam-find-file "Index")))))
+        (find-file (org-roam--get-index-path))
+        (insert "#+TITLE: Index\n\n")))))
```
theianjones, line 600:
```suggestion
        (insert "#+title: Index\n\n")))))
```

Should this be lowercased a la #770?

Issue #572, opened 2020-05-06
Support tags as first-class values


Wetlize:
### Brief Abstract
There's use cases when you need to interact with notes for meta specific purposes, where title alone end-up insufficient, e.g. as a meta-data for an an extra taxonomy. Right now sub-directories or custom file prefixes kind of allow you to do this, but they aren't as flexible as tags are (e.g. storing one note under 2 different directories is a PITA experience) and involve a higher level of friction.

### Long Description
I should clarify that the purpose (or at least focus) of tags that I'm describing here is not for the formation of new notes that would create backlinks in them (nor empty files). The backlinks here (at least for me) don't really carry much value, *however*, if you feel like backlinks should be allowed to work as tags too, feel free to do so.

At the very basic level (see "Assignment of tags" section for more details), tags would be assigned as simple strings to a dedicated org-mode keyword in files, that would not form any backlinks, so they will not create any extra notes (nor empty files) with corresponding to them names.

The purpose of tags that described here is to provide more value at a meta level instead. For example, they can be utilized by the graph, completion system in Emacs, exporting/publishing systems, or organizational purposes. As a more concrete example, take the completion system: just like we do right now with sub-directories, tags could format titles and assign themselves as part of the title.

![Example 1](https://user-images.githubusercontent.com/17055067/81127726-c9d74600-8f47-11ea-9728-30cf1fa4ee20.png) 
*(here, `bib` displayed as tag and it isn't part of the `#+TITLE:`, instead it's just formatted to be  displayed this way along with title)*

In this case we are able to use tags to narrow down to a specific set of candidates, which is very useful when you don't remember specific title but can quickly recap (or guess) its taxonomy.

#### Assignment of tags
This would be done programmatically, so the user would be able to specify how tags should be set based on specific conditions. He could write and add his own rules and/or use rules that we would supply ourselves as part of the default "library". Here's some basic examples of what can be supplied to the user as part of the library:

- Extract tags based on specific org-mode keyword in files. It was suggested in Slack by @zaeph that `#+ROAM_TAGS` would be a good candidate for this.
  - It was also noted in Slack that we shouldn't reuse `#+FILETAGS:` for this purpose, because its semantic is a bit different from what is proposed here.
  - As values accepts:
    - comma or space separated in-buffer strings, e.g. `#+ROAM_TAGS: tag, another tag`
[...]

Issue #588, opened 2020-05-09
Backlinks that are in bullet points should show nested bullet points as context.


spacemacadet:
### Brief Abstract

Links that are in (parent) bullet points should show nested (child) bullet points as part of context in the backlinks buffer. This could also be the case for links in headings and their respective nested subheadings. This gives more utility for the backlinks buffer, which is currently just a list of links with the setence as the context.  A bigger context, or an option for one, would mean that the backlinks buffer can be used for a contextual but more informative and heirarchical preview of the linked file in it's correct context. 

### Long Description

To be more clear, suppose I make a link in my org file but it is part of a bullet point and I then make nested bullet points under this. When I look for the backlinks, for this link, where-ever the link is part of nested bullet points it should show the bullet points below it.  Currently the only context of the link is the paragraph or sentence it is in. 

Perhaps this could be also done for nested sub-headings. 

**Pros:**
- Firslty, this allows for the backlinks buffer to be more useful than just a list of links, I can have partial view of the linked file with more context than just the sentence it is in, but also to allow for heirarchical context structures should the user need to. 
- Secodnly, this allows for the flexibility for a "outliner" type organization system, for users who prefer this. 

### Proposed Implementation (if any)

I don't know any Lisps, but on the surface given that currently the context of the backlinks are sentences already, how difficult would it be to include a bigger context?

### Please check the following:
- [ x ] No similar feature requests

zaeph:
@Wetlize mentioned those breadcrumbs in [this comment](https://github.com/org-roam/org-roam/issues/245#issuecomment-597374337) on #245.

We could implement this without too much trouble by modifying `org-roam--extract-links`.  However, since we're planning a major revamp of `org-roam-buffer`, this is going to be a low-priority for now.

Issue #587, opened 2020-05-09
Option to remove one-file-one-citation workflow


stites:
### Brief Abstract
The current org-roam workflow preferences one-file / one-citation workflow. This workflow is a little prescriptive and users should have the ability to opt-out if they want to work with more atomic notes with the same ROAM_KEY. 
 
### Long Description
I am transitioning a backlog of a physical zettelkasten which worked under the principle that each index card is a stand-alone idea or thought. This idea conflicts with the 1-1 note-to-citation correspondence (you can't fit a textbook in a 6x4" card!). In org-roam, I believe this workflow resembles an [ORB-based workflow](https://github.com/zaeph/org-roam-bibtex) using [structure notes for each reference](https://zettelkasten.de/posts/three-layers-structure-zettelkasten/).

I'm of the opinion that there are some benefits to this level of granularity. One motivating example might be starting with the note "theorem 1 from textbook A" which is linked to several papers. At some later point I might update this note so that the reference has better provenance. With the proposed workflow I would not have to update any linked documents and can do the same with other theorems in the textbook.

A potential pain point of this workflow is that atomic notes might require more context from the original reference via a page-number ([a la org-ref support](https://www.reddit.com/r/orgmode/comments/90sk11/how_do_you_cite_pages_for_quotes_with_orgref/e2tckcu/)) -- this is considered out-of-scope.

### Proposed Implementation (if any)
Proposed by @zaeph, this feature would be flagged by a variable that would remove the UNIQUE key constraint in the refs.ref table. This would require a user to clear and rebuild their current DB cache. ref-backlinks would also have to be deduped via data structure (ie: use a set) or algorithmically. 

### Similar requests
- #585 is the expected error you would see if you do not adhere to the one-note/one-reference rule.

zaeph:
Thanks for filing the issue.

Implementing this in a sensible way would also imply to modify existing commands acting on refs.  For instance, `org-roam-find-ref` should be a two-step completion: 1) the current behaviour of picking the `ROAM_KEY`, then 2) completion with the titles of all the entries which have that ROAM_KEY.  Note that this behaviour should also be used by `org-ref` when opening a link.

Cite-backlinks should be all right, with probably just an adjustment in the way they're constructed in the `org-roam-buffer`.

However, this seems to be the tip of the iceberg, and I expect plenty of things to go wrong along the way.  I'm setting a low priority on this, and I don't expect to get back to it within the fortnight, but if anyone wants to tackle it, please do.

[...]

Issue #548, opened 2020-05-01
[FEATURE] Link to org-mode Headlines


wakatara:
### Brief Abstract

Extend org-roam to allow linking to a org-mode headline within a file rather than just a file name. 

### Long Description

Org-mode's greatest asset is the representation of hierarchical information defined by headlines. Most files are organized in a way to make that the quantum of information and both org-mode links and similar allow direct linking and follow to those. 

Why? 
This would allow both more relevant information used by other programs (org-contacts etc) to be utilized more effectively, For example, structured formats which are used by other programs to provide functionality could be folded in and org-roam used as a semantic layer above that 
(for example: CRM. org-contact used to provide structured info, thta is linked to a page for each person... even more interestingly, perhaps programatically info could be embedded in an org-roam page from those sources via some sort of literate programming - for example, I'd love to see the contact info represented in a nice form, but say it have a complete log of all interactions with them across various backlinks and dates. 

### Proposed Implementation (if any)

As an example of how it might work with ivy completion in org-roam, I'd look at the org-mode has a Refile functionality (C-c C-w on my machine) which allows reflling of headlines both in pages and *under* headlines. Roam would be amazingly better, if rather than using the file as the quantum of info, it used the headline.

### Please check the following:
- [X ] No similar feature requests
I did search in the Issues but saw nothing

alan-carroll:
FWIW at first I wanted the same functionality, but the more I have been thinking about it I am not certain that it is a good idea. It would be a lot of code and logic to add and rewrite and doesn't fit well with the intention of the package (as I see it, at least), when the alternative it to just use a standard org-link in the places you want it.

In your contacts example, your contacts.org file could sit in your org-roam dir as an 'index' note that both had org-contact structural information AND an org-roam link to a separate per-person note. The backlink from the per-person note already links back to the content in contacts.org (if you click it I think -- perhaps this could be made easier). If you need something more specific still, it is trivial to add a standard header link in either location.

If you still need more aggregation, as you say you can do a literate code block or even write an elisp function that you can call through an org link. The backlinks buffer would already populate with all references from daily notes/interactions you record in org-roam on the per-person note; it would be easy to filter these out in some way as a list of results under their contact header in your contacts.org. I think adding some documentation and/or helper functions for these types of actions would be helpful for users uncomfortable with digging through code themselves.

[...]

PR #604, opened 2020-05-12
(feat): Add a tagging system

jethrokuan:

Adds a tagging system to Org-roam. Addresses #572.

Tags are obtained in 2 ways:

1. Via sub-directories: "a/b/c/foo.org" will have tags `a, b, c`
2. Via `#+ROAM_TAG`: `#+ROAM_TAG: foo bar` will assign tags `foo bar`

These 2 means are freely combined.

@zaeph this PR removes your stuff on directory specific filtering, in favour of assigning tags based on directories, functionally they are equivalent, but this is more general.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -134,7 +134,7 @@ space-delimited strings.
            (symbol)))
   :group 'org-roam)
 
-(defcustom org-roam-tag-sources '(directories global-prop)
+(defcustom org-roam-tag-sources '(prop)
   "Sources to obtain tags from."
   :type '(repeat symbol))
 
@@ -359,16 +359,24 @@ If NESTED, return the first successful result from SOURCES."
           (cl-return))))
     coll))
 
-(defun org-roam--extract-tags-directories (file)
-  "Extract tags from using the directory path FILE."
+(defun org-roam--extract-tags-all-directories (file)
+  "Extract tags from using the directory path FILE.
+All sub-directories relative to `org-roam-directory' are used as tags."
   (when-let ((dir-relative (file-name-directory
                             (file-relative-name file org-roam-directory))))
     (f-split dir-relative)))
 
-(defun org-roam--extract-tags-global-prop (_file)
+(defun org-roam--extract-tags-last-directory (file)
+  "Extract tags from using the directory path FILE.
+The final directory component is used as a tag."
+  (when-let ((dir-relative (file-name-directory
+                            (file-relative-name file org-roam-directory))))
+    (last (f-split dir-relative))))
+
+(defun org-roam--extract-tags-prop (_file)
   "Extract tags from the current buffer's \"#ROAM_TAGS\" global property."
-  (let* ((props (org-roam--extract-global-props '("ROAM_TAGS"))))
-    (org-roam--str-to-list (cdr (assoc "ROAM_TAGS" props)))))
+  (let* ((prop (org-roam--extract-global-props '("ROAM_TAGS"))))
+    (org-roam--str-to-list (cdr (assoc "ROAM_TAGS" prop)))))
 
 (defun org-roam--extract-tags (&optional file)
   "Extract tags from the current buffer.
--- a/org-roam.el
+++ b/org-roam.el
@@ -155,7 +155,9 @@ extraction methods:
     Extract the last directory relative to `org-roam-directory'.
     That is, if a file is located at relative path foo/bar/file.org,
     the file will have tag \"bar\"."
-  :type '(repeat symbol))
+  :type '(set (const :tag "#+ROAM_TAGS" prop)
+              (const :tag "sub-directories" all-directories)
+              (const :tag "parent directory" last-directory)))
 
 ;;;; Dynamic variables
 (defvar org-roam-last-window nil
--- a/org-roam.el
+++ b/org-roam.el
@@ -395,6 +395,19 @@ The final directory component is used as a tag."
   (let* ((prop (org-roam--extract-global-props '("ROAM_TAGS"))))
     (org-roam--str-to-list (cdr (assoc "ROAM_TAGS" prop)))))
 
+(defcustom org-roam-tag-sort nil
+  "When non-nil, sort the tags in the completions.
+When t, sort the tags alphabetically, regardless of case.
+`org-roam-tag-sort' can also be a list of arguments to be applied
+to `cl-sort'.  For example, these are the arguments used when
+`org-roam-tag-sort' is set to t:
+    \('string-lessp :key 'downcase)
+Only relevant when `org-roam-tag-sources' is non-nil."
+  :type '(choice
+          (boolean)
+          (list :tag "Arguments to cl-loop"))
+  :group 'org-roam)
+
 (defun org-roam--extract-tags (&optional file)
   "Extract tags from the current buffer.
 If file-path FILE, use it to determine the directory tags.
@@ -403,10 +416,20 @@ Tags are obtained via:
 1. Directory tags: Relative to `org-roam-directory': each folder
    path is considered a tag.
 2. The key #+ROAM_TAGS."
-  (let ((file (or file (buffer-file-name (buffer-base-buffer)))))
-    (mapcan (lambda (source)
-              (funcall (intern (concat "org-roam--extract-tags-" (symbol-name source))) file))
-            org-roam-tag-sources)))
+  (let* ((file (or file (buffer-file-name (buffer-base-buffer))))
+         (tags (mapcan (lambda (source)
+                         (funcall (intern (concat "org-roam--extract-tags-"
+                                                  (symbol-name source)))
+                                  file))
+                       org-roam-tag-sources)))
+    (pcase org-roam-tag-sort
+      ('nil tags)
+      ((pred booleanp) (cl-sort tags 'string-lessp :key 'downcase))
+      (`(,(pred symbolp) . ,_)
+       (apply #'cl-sort (push tags org-roam-tag-sort)))
+      (wrong-type (signal 'wrong-type-argument
+                          `((booleanp (list symbolp …))
+                            ,wrong-type))))))
 
 (defun org-roam--ref-type-p (type)
   "Return t if the ref from current buffer is TYPE."
--- a/org-roam.el
+++ b/org-roam.el
@@ -140,19 +140,21 @@ space-delimited strings.
 It should be a list of symbols representing any of the following
 extraction methods:
 
-1. 'prop: This extracts tags from the =#+ROAM_TAGS=
-   property.  Tags are space delimited, and can be multi-word
-   using double quotes.
+  `prop'
+    Extract tags from the #+ROAM_TAGS property.
+    Tags are space delimited.
+    Tags may contain spaces if they are double-quoted.
+    e.g. #+ROAM_TAGS: tag \"tag with spaces\"
 
-2. 'all-directories: All sub-directories relative to
-   =org-roam-directory= are extracted as tags.  That is, if a file
-   is located at relative path =foo/bar/file.org=, the file will
-   have tags =foo= and =bar=.
+  `all-directories'
+    Extract sub-directories relative to `org-roam-directory'.
+    That is, if a file is located at relative path foo/bar/file.org,
+    the file will have tags \"foo\" and \"bar\".
 
-3. 'last-directory: Extracts the last directory relative to
-   =org-roam-directory= as the tag.  That is, if a file is located
-   at relative path =foo/bar/file.org=, the file will have tag
-   =bar=."
+  `last-directory'
+    Extract the last directory relative to `org-roam-directory'.
+    That is, if a file is located at relative path foo/bar/file.org,
+    the file will have tag \"bar\"."
   :type '(repeat symbol))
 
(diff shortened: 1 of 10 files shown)


bdarcus:
> Tags are obtained in 2 ways:
> 
> 1. Via sub-directories: "a/b/c/foo.org" will have tags `a, b, c`
> 2. Via `#+ROAM_TAG`: `#+ROAM_TAG: foo bar` will assign tags `foo bar`

Nice!

zaeph:
> @zaeph this PR removes your stuff on directory specific filtering, in favour of assigning tags based on directories, functionally they are equivalent, but this is more general.

No problem whatsoever, thanks for the refactoring.

wakatara:
I love the tagging idea but would respectifully suggest
1. Sub directories will most likely have incidental effects with other org-mode programs - I'd avoid the subdirectories idea (too much like folders anyway which tags are more flexible than)
2. While I can definitely see advantages to explicit `#+ROAM_TAG:` is there a functional limitation which prevents native org-mode tags from being used in the form `:project:` etc?
3. Tags should have a summary buffer with created or virtual where you can navigate to all pages that are linked to that tag (much like you see in the Roam product - it is a very cool feature.).

Wetlize:
Thanks for the implementation. Hope I'll test drive this soon enough.

> 1. Sub directories will most likely have incidental side effects with other org-mode programs or organization - I'd avoid the subdirectories idea (too much like folders anyway which tags are more flexible than and I sort of perceive org-roam as a virtual "superstructure" over that sort of organization cutting "across" and "through" it if you will.)
> 2. While I can definitely see advantages to an explicit #+ROAM_TAG: property is there a functional limitation which prevents native org-mode tags from being used in the form :project: etc? - at least at the top level of the page?

1. There's can be cases where virtual structure would be more clunky or insufficient approach than sub-directories. I personally don't think we should completely ignore the file system.
2. If you mean `#+FILETAGS` they were mentioned in #572. If you need a tl/dr: they semantic is a bit different. Or you only mean the values should be assigned to `#+ROAM_TAG:`, just like they're assigned for the `#+FILETAGS` and headings, surrounding them by colons? If so, then why bother? This is rather clunky.

This essentially boils down to what I've proposed that tags should be programmable in a way how they should be collected; users just have different personal preferences with how tags should be resolved (e.g. how I disagree about sub-directories, because I probably have different ideas in mind). 

With this implementation the only way to change how tags can be extracted would by writing an advice for `org-roam--extract-tags` function, which is potentially unstable. I personally can build a needed for me ways to collect tags (in my config) from this implementation, but I still wonder if there was a specific reason why the proposal didn't make it into the current implementation? Do you think that would be too error prone for an average user? @jethrokuan

Wetlize:
> A killer additional feature would be that tags should have a summary buffer created or virtually where you can navigate to all pages that are linked to that tag (much like you see in the Roam product - it is a very cool feature.) especially if to level properties could be displayed there much as we do with backlinks.

I think this is connected with #588 and [this post](https://github.com/org-roam/org-roam/issues/587#issuecomment-626224306). We could end-up later generating virtual nodes, that would act similarly to what you suggest.

zaeph:
> * Sub directories will most likely have incidental effects with other org-mode programs - I'd avoid the subdirectories idea (too much like folders anyway which tags are more flexible than)

Subdirs-as-tags was already present before, albeit a toggle with `org-roam-title-include-subdirs`.  We could implement the same toggle if you only want to tag files explicitly.

> * While I can definitely see advantages to explicit `#+ROAM_TAG:` is there a functional limitation which prevents native org-mode tags from being used in the form `:project:` etc?

Yes.  Here, we are implementing file-tags, not heading-tags.  We can have this discussion when #548 is implemented.

> * Tags should have a summary buffer with created or virtual where you can navigate to all pages that are linked to that tag (much like you see in the Roam product - it is a very cool feature.).

It sounds like something we could do with a revamped `org-roam-buffer`.

> With this implementation the only way to change how tags can be extracted would be by writing an advice for `org-roam--extract-tags` function, which is potentially unstable. I personally can build a needed for me ways to collect tags (in my config) from this implementation, but I still wonder if there was a specific reason why the proposal didn't make it into the current implementation? Do you think that would be too error prone for an average user? But then I would imagine that most Emacs users are power users. In any case, if that so, then maybe a middle ground would be to have at least a set of expected options, instead of an API dedicated functions. @jethrokuan

I'm on the fence about this.  I understand your point about having customisable retrieval methods, but since we're trying to stay true to as much Org-mode internals as possible for compatibility, I'm not sure if we want to change the meaning of tags, whether they be at file-level or heading-level.

Could you give me some use-cases about how you would go about programmatically defining tags?  That'd help me figure out the realm of possibility.  I have a vague recollection of you describing some in an earlier issue.

Wetlize:
> we're trying to stay true to as much Org-mode internals as possible for compatibility, I'm not sure if we want to change the meaning of tags, whether they be at file-level or heading-level.

If you mean in terms of what can possibly represent tags, then this legit concern, fair enough.

> Could you give me some use-cases about how you would go about programmatically defining tags? That'd help me figure out the realm of possibility. I have a vague recollection of you describing some in an earlier issue. 

Not really. My desire was mostly driven by making tags more granular and less coupled, so they would be able to satisfy an arbitrary workflow for users whenever they would need them for a specific kind of reason. The idea was the next: as long as the process of collection yields tags to an API in expected to it format, it shouldn't really matter how they're obtained and what defines them. Since sub-directories driven tags *with* the `#+ROAM_TAG` keyword satisfy all of my needs (as for now) I cannot really see myself too what this realm of possibilities can be. 

Don't get me wrong though, I'm not pushing for this hard; only proposing, and in case of decline I want to see *at most* some sort of feedback in terms of response (the reason on top of this post is more than enough, btw). If you can see that this would be potentially harmful (especially in future), than of course this is legit enough reason to push a no-no button.

TimQuelch:
Considering that there is some disagreement whether directories should be mapped to tags; would it be worth including a customization option to toggle this behaviour?

e.g. `org-roam-use-subdirectories-as-tags` if not nil, tags come from subdirs and `ROAM_TAG`, if nil tags only come from `ROAM_TAG`

wakatara:
> Considering that there is some disagreement whether directories should be mapped to tags; would it be worth including a customization option to toggle this behaviour?
> 
> e.g. `org-roam-use-subdirectories-as-tags` if not nil, tags come from subdirs and `ROAM_TAG`, if nil tags only come from `ROAM_TAG`

Yes, Feature flag-ing that option seems like a great idea. Good call.

zaeph:
> Considering that there is some disagreement whether directories should be mapped to tags; would it be worth including a customization option to toggle this behaviour?
> 
> e.g. `org-roam-use-subdirectories-as-tags` if not nil, tags come from subdirs and `ROAM_TAG`, if nil tags only come from `ROAM_TAG`



> Yes, Feature flag-ing that option seems like a great idea. Good call.

Yes, that's what I was suggesting when I said this:

> Subdirs-as-tags was already present before, albeit a toggle with org-roam-title-include-subdirs. We could implement the same toggle if you **only** want to tag files explicitly.  [emphasis added]

jethrokuan:
> With this implementation the only way to change how tags can be extracted would be by writing an advice for org-roam--extract-tags function, which is potentially unstable. I personally can build a needed for me ways to collect tags (in my config) from this implementation, but I still wonder if there was a specific reason why the proposal didn't make it into the current implementation?

@Wetlize I've added `org-roam-tag-sources` which defaults to `'(directories global-prop)`. If you want to add a new tag source, you can add a symbol `foo` and define a function `(org-roam--extract-tags-foo file)`.

jethrokuan:
okay I have no clue why the tests are failing, things are working ):

zaeph:
Re CI: ¯\_(ツ)_/¯

jethrokuan:
The failure was expected, I fixed it eventually.

org-roam.el
```diff
@@ -135,7 +135,24 @@ space-delimited strings.
...
+
+2. 'all-directories: All sub-directories relative to
+   =org-roam-directory= are extracted as tags.  That is, if a file
+   is located at relative path =foo/bar/file.org=, the file will
+   have tags =foo= and =bar=.
+
+3. 'last-directory: Extracts the last directory relative to
+   =org-roam-directory= as the tag.  That is, if a file is located
+   at relative path =foo/bar/file.org=, the file will have tag
+   =bar=."
```
TimQuelch, line 155: I think it would be a good idea to add info on how users can add their own sources in this documentation.

e.g. You can add your own source `'foo` by defining the function `org-roam--extract-tags-foo`

zaeph: #blamejethro 😄
I'm off to bed now, I'll check this in the morning.

jethrokuan:
Re double spaces, it's failing because you set didn't set double spacing to nil in your own Emacs config. I've done so in the CI.

progfolio:
>  set double spacing to nil in your own Emacs config. I've done so in the CI.

If that's the case we should probably make it a local setting via `org-roam-dev.el`.

org-roam.el
```diff
@@ -134,6 +134,27 @@ space-delimited strings.
...
```
progfolio, line 155:
```suggestion
  `prop'
    Extract tags from the #+ROAM_TAGS property.
    Tags are space delimited.
    Tags may contain spaces if they are double-quoted.
    e.g. #+ROAM_TAGS: tag \"tag with spaces\"

  `all-directories'
    Extract sub-directories relative to `org-roam-directory'.
    That is, if a file is located at relative path foo/bar/file.org,
    the file will have tags \"foo\" and \"bar\".

  `last-directory'
    Extract the last directory relative to `org-roam-directory'.
    That is, if a file is located at relative path foo/bar/file.org,
    the file will have tag \"bar\"."
```

Elisp convention is to quote symbols in docstrings.

```

Comment (progfolio) [reply org-roam.el:155]: I'd also consider changing `all-directories` to `sub-directories` and `last-directory` to `parent-directory`.

File: org-roam.el [hunk 3]
```diff
@@ -134,6 +134,27 @@ space-delimited strings.
...
+  :type '(repeat symbol))
```
progfolio, line 156:
```suggestion
   :group 'org-roam
  :type '(set (const :tag "#+ROAM_TAGS" prop)
              (const :tag "sub-directories" all-directories)
              (const :tag "parent directory" last-directory)))
```
This will give users the specific symbol options in the customization system.

tests/roam-files/tags/tag.org
```diff
@@ -0,0 +1,4 @@
+#+ROAM_TAGS: "t1" "t2"
```
progfolio, line 1:
```suggestion
#+ROAM_TAGS: "t1" "t2 with space"
```

tests/test-org-roam.el
```diff
@@ -141,6 +141,72 @@
...
+      ((test (fn file)
+             (let* ((fname (test-org-roam--abs-path file))
+                    (buf (find-file-noselect fname)))
+               (with-current-buffer buf
+                 (funcall fn fname)))))
+    (it "extracts from prop"
+      (expect (test #'org-roam--extract-tags-prop
+                    "tags/tag.org")
+              :to-equal
+              '("t1" "t2"))
```
progfolio, line 162:
```suggestion
              '("t1" "t2 with space"))
```

tests/test-org-roam.el
```diff
@@ -141,6 +141,72 @@
...
+              :to-equal
+              '("deeply")))
+
+    (describe "uses org-roam-tag-sources correctly"
+      (it "'(prop)"
+        (expect (let ((org-roam-tag-sources '(prop)))
+                  (test #'org-roam--extract-tags
+                        "tags/tag.org"))
+                :to-equal
+                '("t1" "t2")))
```
progfolio, line 202:
```suggestion
                '("t1" "t2 with space")))
```

tests/test-org-roam.el
```diff
@@ -141,6 +141,72 @@
...
+      (it "'(prop all-directories)"
+        (expect (let ((org-roam-tag-sources '(prop all-directories)))
+                  (test #'org-roam--extract-tags
+                        "tags/tag.org"))
+                :to-equal
+                '("t1" "t2" "tags"))))))
```
progfolio, line 208:
```suggestion
                '("t1" "t2 with space" "tags"))))))
```

zaeph:
LGTM!

laheadle:
Ok, I understand how to create tags. The next question is: In what context can I get a list of notes with a given tag? That is, How can I find tagged notes?

TimQuelch:
> Roam takes any tag, creates a page with that title, and backlinks all notes that have the #tag in them to that tag’s page. When they let you narrow a group of notes to those with 2 tags, it is just a search for notes that link to both of the 2 tags’ pages.

You can do this by just creating a normal roam link with `org-roam-insert`. If I am wanting to add keywords to a specific note though, I use `org-roam-insert` within the note to create a new note with that name (or link to a previously created one.

The tags added by this PR are more for organisational tags. I have all my daily notes in a subdir `dailies` and all my literature notes in a subdir called `lit`. This means when I am `org-roam-find-file`ing I can type "lit' and all my literature notes will show up in helm. You could also write filter functions to filter the completions of `org-roam-insert` and `org-roam-find-file` to exclude links to specific tags

PR #607, opened 2020-05-12
Refactor org-roam-dev to utilize dir-local variable

progfolio:
Sorry! Missed this refactor before the merge.
I forgot Emacs has directory local variables.
Using this prevents us from having to add it to file-local variable line of each file.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,4 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+((emacs-lisp-mode . ((eval . (require 'org-roam-dev)))))
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -1,4 +1,4 @@
-;;; org-roam-buffer.el --- Roam Research replica with Org-mode -*- coding: utf-8; lexical-binding: t; eval: (require 'org-roam-dev); -*-
+;;; org-roam-buffer.el --- Roam Research replica with Org-mode -*- coding: utf-8; lexical-binding: t; -*-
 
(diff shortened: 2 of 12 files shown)


jethrokuan:
This causes a prompt saying that `(require 'org-roam-dev)` is unsafe, on requiring any of org-roam's files. This means users who are calling any of the functions will see this prompt, and I don't see a good way around it.

progfolio:
When loading the file or when opening the file to edit? Should only be when opening the source file.

zaeph:
> This causes a prompt saying that `(require 'org-roam-dev)` is unsafe, on requiring any of org-roam's files. This means users who are calling any of the functions will see this prompt, and I don't see a good way around it.

How about hiding this require behind a toggle?  Then, we ask in `CONTRIBUTE.md` for people to set the variable to `t` so that everything is formatted properly.

zaeph:
> When loading the file or when opening the file to edit? **Should only be when opening the source file.**  [emphasis added]

Yeah, it doesn't load the local variable in the `init.el` `require` for me, only when opening the source.

jethrokuan:
oh hmm strange, I ran an interactive command and saw that, but if it's not the case then that's fine.

progfolio:
If you can reproduce, I'd be curious to see it. I don't use directory local variables much (mostly file-local variables) and this is Emacs, so there could be some corner case I'm unaware of.

progfolio:
Looks like el-get andy magit employ something similar:
https://github.com/dimitri/el-get/blob/master/.dir-locals.el
https://github.com/magit/magit/blob/master/.dir-locals.el

Issue #605, opened 2020-05-12
Developer coding standards.


progfolio:
I'm not proposing anything overly formal, but we should automate some of the simple settings for an agreed upon style standard for the code base. We could implement this in a module that is loaded when the source files are edited.

progfolio:
Leaving this open for now if anyone else has suggestions for what sort of standards we could/should automate.

jethrokuan:
I think we've got a good thing going here, that we can add to later if need be. Closing until something else pops up.

PR #609, opened 2020-05-12
(refactor): programatically print org-roam-db stats

progfolio:
Build stats message programatically.
Easier to maintain as elements are added to stats list.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -398,17 +398,18 @@ If FORCE, force a rebuild of the cache from scratch."
        [:insert :into refs
         :values $v1]
        all-refs))
-    (let ((stats (list :files (length all-files)
-                       :links (length all-links)
-                       :titles (length all-titles)
-                       :refs (length all-refs)
-                       :deleted (length (hash-table-keys current-files)))))
-      (org-roam-message "files: %s, links: %s, titles: %s, refs: %s, deleted: %s"
-                        (plist-get stats :files)
-                        (plist-get stats :links)
-                        (plist-get stats :titles)
-                        (plist-get stats :refs)
-                        (plist-get stats :deleted))
+    (let ((stats `(:files   ,(length all-files)
+                            :links   ,(length all-links)
+                            :titles  ,(length all-titles)
+                            :refs    ,(length all-refs)
+                            :deleted ,(length (hash-table-keys current-files)))))
+      (org-roam-message (replace-regexp-in-string
+                         ",$" ""
+                         (mapconcat (lambda (element)
+                                      (if (keywordp element)
+                                          (concat (substring (symbol-name element) 1) ":")
+                                        (format "%s," element)))
+                                    stats " ")))
       stats)))
 
 (provide 'org-roam-db)


zaeph:
I don't think I can push to your branch, but I've fixed-up the indent:
```diff
From 60aa9c23c968034617cbab98f4f9ca8cec1feff4 Mon Sep 17 00:00:00 2001
From: Leo Vivier <leo.vivier+dev@gmail.com>
Date: Tue, 12 May 2020 23:25:45 +0200
Subject: [PATCH] Fix indent

---
 org-roam-db.el | 23 ++++++++++++-----------
 1 file changed, 12 insertions(+), 11 deletions(-)

diff --git a/org-roam-db.el b/org-roam-db.el
index 127bc35..4a31867 100644
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -399,17 +399,18 @@ If FORCE, force a rebuild of the cache from scratch."
         :values $v1]
        all-refs))
     (let ((stats `(:files   ,(length all-files)
-                            :links   ,(length all-links)
-                            :titles  ,(length all-titles)
-                            :refs    ,(length all-refs)
-                            :deleted ,(length (hash-table-keys current-files)))))
-      (org-roam-message (replace-regexp-in-string
-                         ",$" ""
-                         (mapconcat (lambda (element)
-                                      (if (keywordp element)
-                                          (concat (substring (symbol-name element) 1) ":")
-                                        (format "%s," element)))
-                                    stats " ")))
+                   :links   ,(length all-links)
+                   :titles  ,(length all-titles)
+                   :refs    ,(length all-refs)
+                   :deleted ,(length (hash-table-keys current-files)))))
+      (org-roam-message
+       (replace-regexp-in-string
+        ",$" ""
+        (mapconcat (lambda (element)
+                     (if (keywordp element)
+                         (concat (substring (symbol-name element) 1)":")
+                       (format "%s," element)))
+                   stats " ")))
       stats)))
 
 (provide 'org-roam-db)
-- 
2.26.2
```

progfolio:
The indentation is the (unfortunate) way that emacs-lisp plists are indented.
There was a patch that was submitted to the emacs-dev mailing list, but it looks as if it has stalled.
Previous discussion: 

- https://www.reddit.com/r/emacs/comments/d7x7x8/finally_fixing_indentation_of_quoted_lists/
- https://stackoverflow.com/questions/22166895/customize-elisp-plist-indentation

If we manually fix the indentation, it will be overridden by elisp-mode's automatic indentation when using things like `indent-region`. I have a number of functions that normalize the buffer's style that run during `before-save-hook` that delete trailing whitespace, properly indent, etc. Manual indentation will get clobbered by those, too. 

One solution is to put in a leading comment like so:
```emacs-lisp
`(; comment for indentation
  :foo t
  :bar t)
```
But it seems like more noise than just dealing with the first line being out of alignment.

BTW: instead of pushing directly or commenting with a patch, you can use the github suggestion feature. It makes it easier to co-develop without pushing concurrent commits.

zaeph:
> The indentation is the (unfortunate) way that emacs-lisp plists are indented.

I know, and it's a big point of contention between Lisp and CL.

I personally use a patched `lisp-indent-function` which does the job:
```el
(use-package lisp-mode
  :custom
  (lisp-indent-function #'zp/lisp-indent-function)
  :config
  ;; Taken from https://github.com/CeleritasCelery/emacs.d/blob/master/emacs.org#indent
  ;; TODO: Use el-patch to catch upstream modifications
  (defun zp/lisp-indent-function (indent-point state)
    "Override `lisp-indent-function' to properly handle plists. See the original function fo full description"
    (let ((normal-indent (current-column))
          (orig-point (point)))
      (goto-char (1+ (elt state 1)))
      (parse-partial-sexp (point) calculate-lisp-indent-last-sexp 0 t)
      (cond
       ;; car of form doesn't seem to be a symbol, or is a keyword
       ((and (elt state 2)
             (or (not (looking-at "\\sw\\|\\s_"))
                 (looking-at ":")))
        (if (not (> (save-excursion (forward-line 1) (point))
                    calculate-lisp-indent-last-sexp))
            (progn (goto-char calculate-lisp-indent-last-sexp)
                   (beginning-of-line)
                   (parse-partial-sexp (point)
                                       calculate-lisp-indent-last-sexp 0 t)))
        ;; Indent under the list or under the first sexp on the same
        ;; line as calculate-lisp-indent-last-sexp.  Note that first
        ;; thing on that line has to be complete sexp since we are
        ;; inside the innermost containing sexp.
        (backward-prefix-chars)
        (current-column))
       ((and (save-excursion
               (goto-char indent-point)
               (skip-syntax-forward " ")
               (not (looking-at ":")))
             (save-excursion
               (goto-char orig-point)
               (looking-at ":")))
        (save-excursion
          (goto-char (+ 2 (elt state 1)))
          (current-column)))
       (t
        (let ((function (buffer-substring (point)
                                          (progn (forward-sexp 1) (point))))
              method)
          (setq method (or (function-get (intern-soft function)
                                         'lisp-indent-function)
                           (get (intern-soft function) 'lisp-indent-hook)))
          (cond ((or (eq method 'defun)
                     (and (null method)
                          (> (length function) 3)
                          (string-match "\\`def" function)))
                 (lisp-indent-defform state indent-point))
                ((integerp method)
                 (lisp-indent-specform method state
                                       indent-point normal-indent))
                (method
                 (funcall method indent-point state)))))))))
```

I'm not sure if we want to go down that route, though.

progfolio:
Agreed. It is ugly to look at, but I've learned to accept it for now.

Here's the bug thread: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=21922
The best solution is probably to try and make some progress there.

zaeph:
> BTW: instead of pushing directly or commenting with a patch, you can use the github suggestion feature. It makes it easier to co-develop without pushing concurrent commits.

I'd seen you do it, but I could never figure out where the button was.  Finally took the time to check it, so I'll do that next time.  I think we've only had one issue so far with commits crossing themselves with messy merging, but I think we did a good job so far of letting the other one know when we were working on the branch.

progfolio:
Another idea that occurred to me:

Could your indentation patch be enabled directory-local via `org-roam-dev.el`?

I'm not opposed to something like that so long as it only adjusts org-roam's source files.
#605

zaeph:
> Could your indentation patch be enabled directory-local via `org-roam-dev.el`?

That's actually what I was suggesting by mentioning the function.  I haven't vetted the code in any way, I just know that it works on my end.  If we get around to it, I wouldn't mind adding it there.

progfolio:
Holding off on merge for now. We can definitely implement `org-roam-db-build-cache` in a cleaner way.

PR #611, opened 2020-05-13
(refactor): org-roam-db--clear pull tables from schemata

progfolio:
Programatically pull tables from schemata, rather than manually
specifying each table.


zaeph:
LGTM!

PR #613, opened 2020-05-13
(refactor): org-roam-db--clear-file pull tables from schemata

progfolio:
Programatically pull tables from schemata, rather than manually
specifying.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -192,25 +192,15 @@ the current `org-roam-directory'."
     (org-roam-db-query [:delete :from links])
     (org-roam-db-query [:delete :from refs])))
 
-
 (defun org-roam-db--clear-file (&optional filepath)
   "Remove any related links to the file at FILEPATH.
 This is equivalent to removing the node from the graph."
-  (let* ((path (or filepath
-                   (buffer-file-name)))
-         (file (file-truename path)))
-    (org-roam-db-query [:delete :from files
-                        :where (= file $s1)]
-                       file)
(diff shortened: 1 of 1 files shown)


org-roam-db.el
```diff
@@ -192,25 +192,14 @@ the current `org-roam-directory'."
...
-    (org-roam-db-query [:delete :from links
-                        :where (= from $s1)]
-                       file)
-    (org-roam-db-query [:delete :from titles
-                        :where (= file $s1)]
-                       file)
-    (org-roam-db-query [:delete :from refs
-                        :where (= file $s1)]
-                       file)))
+  (let ((file (file-truename (or filepath (buffer-file-name)))))
```
zaeph, line 198: Maybe we need to include indirect buffers as well.
```suggestion
  (let ((file (file-truename (or filepath
                                 (buffer-file-name)
                                 (buffer-file-name (buffer-base-buffer))))))
```

progfolio: Good catch. No need for the first call to `(buffer-file-name)` though. `(buffer-base-buffer)` will return nil if there is no base-buffer, which will result in current-buffer being used in `(buffer-file-name)`.

progfolio: `(buffer-file-name (buffer-base-buffer))` is correct for both cases.

zaeph: Oh yeah, you're right.  I'm pretty sure I did exactly the same mistake in a previous PR.

PR #615, opened 2020-05-13
(internal): remove assignment in issue templates

jethrokuan:

Used to be able to keep up with all the issues, now I'm finding it a bit of a struggle :sweat_smile:


zaeph:
Oof, you tell me.

I'll just assign myself to whatever I able to do, as I've done in the past.  I'll make sure to pester you in case of doubt. 🤡

Issue #614, opened 2020-05-13
[RFC] Rethinking testing org-roam


jethrokuan:
Right now tests are feeling very brittle, and I'd like to rethink the approach towards testing org-roam functionality. 

The fundamental difficulty with testing org-roam is that it's a highly interactive application, and the behaviour varies a lot based on customization options, and user input. Typically, what we want to do is to test *observable user behaviour*, which is the current approach taken. We simulate calls to interactive commands such as `org-roam-insert`, simulate user input, and test the expected result (changes in buffer-string etc.)

While working on #604 , these tests broke, because vanilla Emacs completions require exact matching from user input from start to end. The proportion of users using vanilla Emacs completion is likely to be small, so we're actually testing for behaviour that few people will see. The alternative then is run these tests with other completion mechanisms such as `ivy` enabled, but that makes org-roam tests subject to upstream changes to the completion mechanisms, and less isolated, among other downsides.

This has also led to some neglect in adding tests into the test suite. For the large part, a lot of the functionality added has not been with new interactive commands: they have been small, incremental improvements to existing features. We need to pick up on these and start adding regression testing as org-roam grows more complex.

What I'm suggesting here may go against typical testing recommendations, but here's what I was thinking we should do:

1. Test the methods directly instead simulating. Add tests for all functions, even the low level ones. This gives us the confidence to modify their behaviour without causing regressions. 
2. Test the typical behaviour. The test org directory we currently have is extremely contrived, and does not represent the typical slip-box. The test directory should also serve as an example of what a slip-box should look like, guiding the use of org-roam.
3. Move the interactive tests and continue to run them, but allow CI to pass regardless of whether they pass or fail. They currently do not inspire much confidence.

jethrokuan:
cc @progfolio @zaeph

zaeph:
> This has also led to some neglect in adding tests into the test suite. For the large part, a lot of the functionality added has not been with new interactive commands: they have been small, incremental improvements to existing features. We need to pick up on these and start adding regression testing as org-roam grows more complex.

Sounds good to me, and I agree that now is perfect time to do this before the code-base grows to a hardly-manageable level.

I've got very little experience with testing besides a tryst with Org-mode's ERT facility, but I could learn the ropes with this project.

PR #625, opened 2020-05-14
(refactor): refactor title extraction

jethrokuan:

Introduces `org-roam-title-sources,` which can be a list of lists,
infinitely deep. Each symbol corresponds to a function, that when called
returns a list of titles. If the element is a list, use the first
function in the list that returns a successful result.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -207,7 +207,9 @@ Take for example the following org file:
 | ='alias=    | '("WWII" "World War II") |
 
 One can freely control which extraction methods to use by customizing
-=org-roam-title-sources=: see the doc-string for the variable for more information.
+=org-roam-title-sources=: see the doc-string for the variable for more
+information. If all methods of title extraction return no results, the file-name
+is used in place of the titles for completions.
 
 ** File Refs
 
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -174,16 +174,40 @@ Org-mode. However, to support additional functionality, Org-roam adds
 several Org-roam-specific keywords. These functionality are not crucial
 to effective use of Org-roam.
 
-** File Aliases
+** File Titles
 
-Suppose you want a note to be referred to by different names (e.g.
-"World War 2", "WWII"). You may specify such aliases using the
-=#+ROAM_ALIAS= attribute:
+To easily find a note, a title needs to be prescribed to a note. A note can have
+many titles: this allows a note to be referred to by different names, which is
+especially useful for topics or concepts with acronyms. For example, for a note
+like "World War 2", it may be desirable to also refer to it using the acronym
+"WWII".
+
+Org-roam calls =org-roam--extract-titles= to extract titles. It uses the
+variable =org-roam-title-sources=, to control how the titles are extracted. The
+title extraction methods supported are:
+
+1. ='title=: This extracts the title using the file =#+TITLE= property
+2. ='headline=: This extracts the title from the first headline in the Org file
+3. ='alias=: This extracts a list of titles using the =#ROAM_ALIAS= property.
+   The aliases are space-delimited, and can be multi-worded using quotes
+
+Take for example the following org file:
 
 #+BEGIN_SRC org
   #+TITLE: World War 2
   #+ROAM_ALIAS: "WWII" "World War II"
+
+  * Headline
 #+END_SRC
+
+| Method      | Titles                   |
+|-------------+--------------------------|
+| ='title=    | '("World War 2")         |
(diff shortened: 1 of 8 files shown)


jethrokuan:
I've gone ahead and added the kind of tests I was hoping to shift towards in #614 , let me know what you think.

doc/org-roam.org
```diff
@@ -174,17 +174,41 @@ Org-mode. However, to support additional functionality, Org-roam adds
...
+like "World War 2", it may be desirable to also refer to it using the acronym
+"WWII".
+
+Org-roam calls =org-roam--extract-titles= to extract titles. It uses the
+variable =org-roam-title-sources=, to control how the titles are extracted. The
+title extraction methods supported are:
+
+1. ='title=: This extracts the title using the file =#+TITLE= property
+2. ='headline=: This extracts the title from the first headline in the Org file
+3. ='alias=: This extracts a list of titles using the =#ROAM_ALIAS= property.
```
zaeph, line 191: We might want to specify that if all extraction methods fail, we revert to using the file-name, which was the case before this PR and this is the case now.

org-roam.el
```diff
@@ -106,6 +106,34 @@ ensure that."
...
 
+(defcustom org-roam-title-sources '((title headline) alias)
+  "The list of sources from which to retrieve a note title.
+Each element in the list is either:
+
+1. a symbol -- this symbol corresponds to a title retrieval
+function, which returns the list of titles for the current buffer
+2. a list of symbols -- symbols in the list are treated as
+with (1). The return value of this list is first symbol in the
+list returning a non-nil value.
```
zaeph, line 117:
```suggestion
with (1). The return value of this list is the first symbol in the
list returning a non-nil value.
```
I think the fill-column is respected, so no need to refill.

org-roam.el
```diff
@@ -349,25 +377,49 @@ current buffer is used."
...
-                        (org-no-properties (org-element-property :raw-value h)))
-                      :first-match t)))
-         (alias-list (org-roam--aliases-str-to-list aliases)))
-    (if title
-        (cons title alias-list)
-      alias-list)))
+(defun org-roam--extract-titles-title ()
+  "Return title from \"#+TITLE\" of the current buffer."
+  (let* ((props (org-roam--extract-global-props '("TITLE")))
+         (title (cdr (assoc "TITLE" props))))
```
zaeph, line 383:
```suggestion
  (let* ((prop (org-roam--extract-global-props '("TITLE")))
         (title (cdr (assoc "TITLE" prop))))
```
Not sure if we need the plural here.

org-roam.el
```diff
@@ -349,25 +377,49 @@ current buffer is used."
...
+    (when title
+      (list title))))
+
+(defun org-roam--extract-titles-alias ()
+  "Return the aliases from the current buffer.
+Reads from the \"ROAM_ALIAS\" property."
+  (let* ((props (org-roam--extract-global-props '("ROAM_ALIAS")))
+         (aliases (cdr (assoc "ROAM_ALIAS" props))))
```
zaeph, line 391:
```suggestion
  (let* ((prop (org-roam--extract-global-props '("ROAM_ALIAS")))
         (aliases (cdr (assoc "ROAM_ALIAS" prop))))
```
Idem.

org-roam.el
```diff
@@ -349,25 +377,49 @@ current buffer is used."
...
+  "Extract the titles from current buffer using SOURCES.
+If NESTED, return the first successful result from SOURCES."
+  (let (coll res)
+    (cl-dolist (source (or sources
+                           org-roam-title-sources))
+      (cond
+       ((symbolp source)
+        (setq res (funcall (intern (concat "org-roam--extract-titles-" (symbol-name source))))))
+       (t ; list of sources
+        (setq res (org-roam--extract-titles source t))))
```
zaeph, line 415: I think we can refactor that with an `if`:
```suggestion
      (setq res (if (symbolp source)
                    (funcall (intern (concat "org-roam--extract-titles-"
                                             (symbol-name source))))
                  (org-roam--extract-titles source t)))
```

org-roam.el
```diff
@@ -349,25 +377,49 @@ current buffer is used."
...
+      (when res
+        (if nested
+            (progn
+              (setq coll res)
+              (cl-return))
+          (setq coll (nconc coll res)))))
```
zaeph, line 421: Let's golf!  4 chars saved!
```suggestion
        (if (not nested)
            (setq coll (nconc coll res)
          (setq coll res)
          (cl-return)))))
```

PR #628, opened 2020-05-14
 (docs): Fix org-roam-buffer-no-delete-other-windows

ahendriksen:
It appears the option has changed name, but this was not yet reflected
in the documentation.


zaeph:
Thanks for spotting this.  It seems like it's only in the doc, so no need to create an alias for it.

PR #629, opened 2020-05-15
(internal): org-roam-dev -> sentence-end-double-space nil

progfolio:
Required by CI.


progfolio:
@jethrokuan correct setting/value?

jethrokuan:
Yep, thanks!

Issue #631, opened 2020-05-15
Update `org-roam-find-ref`


zaeph:
`org-roam-find-file` was updated in #604, and we had to circumvent issues with `completing-read` by using a hastable instead of an alist.  We need to do the same with `find-ref` for consistency.

I'm ranking it as high-priority because it  breaks some of our commands in ORB, tracked in org-roam/org-roam-bibtex/issues/43.

zaeph:
Actually, the comparison should have been made between `org-roam--get-title-path-completions` and `org-roam--get-ref-path-completions`, because that's where the discrepancy in behaviour now lies.

PR #632, opened 2020-05-15
(fix): replace `if` with `when`

zaeph:
Spotted this I was working on #631.


zaeph:
4e6ee04e0433c5b359d1fc9a9fc8cab6c0b98411 is another brilliant feat of golfing. ⛳

PR #633, opened 2020-05-15
(fix): refactor get-ref-path-completions

### LINKED ISSUE #631: Update `org-roam-find-ref`

zaeph:
Fixes #631.

I've preserved the conditional behaviour with `org-roam-include-type-in-ref-path-completions`.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -696,8 +696,7 @@ included as a candidate."
   (let ((rows (org-roam-db-query [:select [type ref file] :from refs]))
         (ht (make-hash-table :test 'equal))
         (include-type (and interactive
-                           org-roam-include-type-in-ref-path-completions))
-        candidates)
+                           org-roam-include-type-in-ref-path-completions)))
     (dolist (row rows)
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -694,23 +694,26 @@ takes three arguments: the type, the ref, and the file of the
...
-          (let ((candidate (cons (if include-type
-                                     (format "%s:%s" type ref)
-                                   ref)
-                                 file)))
-            (push candidate candidates)))))))
+          (let ((k (concat
+                    (when include-type
+                      (format "(%s) " type))
+                    ref))
+                (v (list :path file-path :type type :ref ref)))
```
zaeph, line 714: Not entirely sure we need to add `type` to the hash-table, but it might be useful for filtering in ORB.

PR #642, opened 2020-05-16
(fix): reimplement mtime sorting

zaeph:
Broken by the switch from alist to hash-tables #604, although it was probably janky before as well.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -322,7 +322,7 @@ connections, nil is returned."
     (org-roam-db-query [:delete :from files
                         :where (= file $s1)]
                        file)
-    (org-roam-db--insert-meta file (list :atime atime :mtime 0))))
+    (org-roam-db--insert-meta file (list :atime atime :mtime mtime))))
 
 (defun org-roam-db--update-titles ()
   "Update the title of the current buffer into the cache."
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -384,7 +384,6 @@ If FORCE, force a rebuild of the cache from scratch."
   (org-roam-db) ;; To initialize the database, no-op if already initialized
   (let* ((org-roam-files (org-roam--list-all-files))
          (current-files (org-roam-db--get-current-files))
-         (time (current-time))
          all-files all-links all-titles all-refs all-tags)
     (dolist (file org-roam-files)
       (let* ((attr (file-attributes file))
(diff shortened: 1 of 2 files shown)


org-roam-db.el
```diff
@@ -121,7 +121,7 @@ SQL can be either the emacsql vector representation, or a string."
   '((files
      [(file :unique :primary-key)
       (hash :not-null)
-      (last-modified :not-null)])
+      (meta :not-null)])
```
jethrokuan, line 124: Bump the db version

org-roam-db.el
```diff
@@ -306,6 +313,17 @@ connections, nil is returned."
...
+(defun org-roam-db--update-meta ()
+  "Update the metadata of the current buffer into the cache."
+  (let* ((file (file-truename (buffer-file-name)))
+         (attr (file-attributes file))
+         (atime (file-attribute-access-time attr))
+         (mtime (file-attribute-modification-time attr)))
+    (org-roam-db-query [:delete :from files
+                        :where (= file $s1)]
+                       file)
+    (org-roam-db--insert-meta file (list :atime atime :mtime 0))))
```
jethrokuan:
```suggestion
    (org-roam-db--insert-meta file (list :atime atime :mtime mtime))))
```

org-roam-db.el
zaeph: Done in acaa3b4cd7c764c1c2dff71dca63e96fe2dbf5e4.

org-roam-db.el
zaeph: Fixed in b33ec3ee522603e1659a0aa499609d582d046bda.

org-roam-db.el
```diff
@@ -206,6 +206,13 @@ This is equivalent to removing the node from the graph."
                          file))))
 
 ;;;;; Insertion
+(defun org-roam-db--insert-meta (file hash meta)
```
jethrokuan, line 209: Maybe make file optional, default to buffer-file-name to be in line with the signatures of the prev functions

zaeph: 👍

zaeph: Actually, it's already done in `org-roam-db--update-meta`:
https://github.com/jethrokuan/org-roam/blob/46af880fbbe575d446fc8576a1b7ad374d578842/org-roam-db.el#L318

PR #644, opened 2020-05-16
(fix): modularise get-completions

zaeph:


org-roam-macs.el
```diff
@@ -64,6 +64,21 @@ to look.
   (when org-roam-verbose
     (apply #'message `(,(concat "(org-roam) " format-string) ,@args))))
 
+(defmacro org-roam--get-completions (query &rest processing)
```
jethrokuan, line 67: can this be a function?

also, i'm a bit worried about bugs surfacing from the `completions` implicitly defined and returned here

org-roam-macs.el
```diff
@@ -64,6 +64,21 @@ to look.
...
+  "Return an alist for completion from QUERY and PROCESSING.
+QUERY is the string to be passed to `org-roam-db-query'.
+PROCESSING should contain the forms to be applied on every
+element matched by QUERY."
+  (declare (indent 1))
+  `(let* ((rows (org-roam-db-query ,query))
+          completions)
+     (seq-sort-by (lambda (x)
+                    (plist-get (nth 3 x) :mtime))
```
jethrokuan, line 76: this is dependent on meta always being the third column, seems brittle.

zaeph: Oh, hadn't noticed, that's true indeed.  Let me see if I can keep the macro useful whilst modularising the sorting as well.

org-roam-macs.el
zaeph: You're right.  This feels incredibly brittle.

zaeph:
Let's hold off on this for now.  I still think it'd be nice to have a helper function to create those queries, but we need to come up with a better way to abstract it.

PR #648, opened 2020-05-16
(fix): also modify links inside moved file

### LINKED ISSUE #647: Links in File are Not Updated When It is Moved to a Different Directory
### Description

When renaming an org-roam file within emacs to move it to a subfolder of its current directory (i.e. so moving it form `roam/` to `roam/subfolder`), the backlinks to the moved file get updated, but the links within the file itself are not updated to reflect the change in the relative location of the linked-to file. As a result, all the links in the moved file are broken.

#### Steps to Reproduce

1. Open an org-roam file within Emacs that links to another org-roam file in the same directory
2. Use emacs to move the file containing the link to a different directory
3. Observe that the link within the file has not been updated with the new relative path to the linked file.

#### Backtrace

N/A

#### Expected Results

Link within the moved file still works

#### Actual Results

Link within the moved file is not updated, and is now broken.

### Versions

- Emacs (`C-h v emacs-version`): v26.3
- Org-roam commit: https://github.com/org-roam/org-roam/commit/265182a698be6babcbb11718c2821c747b1cff52

zaeph:
Fixes #647.

I still need to iron out a few kinks, like making sure that we're killing the buffers that have been created.


zaeph:
I've tried it on my biggest file, and whilst it slows down the renaming a little bit, all the magic in the background is working properly.  🧙✨

zaeph:
I could probably optimise this by only modifying the links when the subdir has changed.

zaeph:
> I could probably optimise this by only modifying the links when the subdir has changed.

Done in 9d73874c12e5626de07497085512c1daafe0291e.

progfolio:
Have any tests been implemented for this?

zaeph:
Zilch.  I'll try writing some later.

org-roam.el
```diff
@@ -1025,6 +1025,37 @@ Otherwise, behave as if called interactively."
             (while (re-search-forward named-regex nil t)
               (replace-match (format "[[file:%s][\\1]]" relative-path))))
           (org-roam-db--update-file file-from)))
+      (unless (string= (file-name-directory path)
```
jethrokuan, line 1028: I feel we should split this advice up into multiple functions.

zaeph: Yeah, same, but I figured we'd do it once we implement links to headlines.

jethrokuan: let me pull this down

zaeph:
My implementation is not replacing absolute links right now.  Whilst we should not touch absolute links pointing _outside_ `org-roam-directory`, if they're pointing  _inside_, we could just turn them into relative links.

Those absolute links are created when you use `org-store-link` on a file that is in a directory above the one in which you `org-insert-link`.

zaeph:
I'm going to take a couple of days off, so feel free to merge this.

Next week, I'll fix the absolute links, I'll write some tests, and I'll move on to #548.

jethrokuan:
can't seem to push to this branch, so I pushed changes up to https://github.com/org-roam/org-roam/tree/fix-renames

jethrokuan:
Superceded by #660

PR #653, opened 2020-05-17
(feature): org-roam-diagnostics command

progfolio:
Automates collection of user environment information.

An idea I had to help users provide us with useful debugging information.
Right now it just grabs `emacs-version`, the full `org-version` and prompts whether or not they are using a framework like Doom or Spacemacs. 

We might be able to grab org-roam git hash programatically as well, but I left that out for now.
Could easily be expanded to include more info  if needed.

Good idea?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -36,8 +36,8 @@ Example:
 
 <!-- Example: File A is missing -->
 
-### Versions
+### Environment
 
-- Emacs (Paste `M-x emacs-version` output here) 
-- Org   (Paste `M-x org-version` output here) 
+<!-- Please M-x org-roam-diagnostics and paste results here -->
+
 - Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere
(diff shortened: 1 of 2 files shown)


jethrokuan:
Beyond that I also want to implement an `org-roam-version` somehow, I've been looking through how it's done in magit, but there seems to be a lot of instrumentation around it, but yes, def a good idea. I think Nix does this and it's useful.

progfolio:
> Beyond that I also want to implement an org-roam-version somehow,
The naive solution could be as simple as locating the library file and pulling from the package header.
Not as good as a commit, but it could be useful.

jethrokuan:
> The naive solution could be as simple as locating the library file and pulling from the package header.

yeah, simple to implement. Would be more useful for sure if we make releases more frequent (they should be).

progfolio:
This seems to work for the simple case:
```emacs-lisp
(defun org-roam-version (&optional message)
  "Return `org-roam' version.
Interactively, or when MESSAGE is non-nil, show in the echo area."
  (interactive)
  (let ((version (with-current-buffer (find-file-noselect (locate-library "org-roam.el"))
                   (save-match-data
                     (goto-char (point-min))
                     (if (re-search-forward "\\(?:;; Version: \\([^z-a]*?$\\)\\)" nil nil)
                         (substring-no-properties (match-string 1))
                       "N/A")))))
    (if (or message (called-interactively-p 'interactive))
        (message "%s" version)
      version)))
```

jethrokuan:
Works for me!

progfolio:
Will add `org-roam-version` info to the command before merging.

PR #657, opened 2020-05-17
(fix): Require seq.el

progfolio:
seq.el is is not guaranteed to be loaded otherwise.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -42,6 +42,7 @@
 (require 's)
 (require 'f)
 (require 'cl-lib)
+(require 'seq)
 
 ;;;; org-roam features
 (require 'org-roam-compat)


zaeph:
I was going to say that it was my fault, but someone else had used `seq`-functions before me.  🤡

jethrokuan:
compiler wasn't complaining, guess it was loaded transitively somehow.

zaeph:
I wish there was a way to create an arborescence of requirements to know which library loaded which other library, something like an advice to `require` that could also be generated locally.  That'd goa  long way towards helping debugging those pesky dependency errors.

progfolio:
https://yoo2080.wordpress.com/2014/05/16/how-to-list-emacs-package-dependencies/
may be of interest to you.
I think there was another package that attempted to do something like that, but its name is escaping me right now.

Issue #651, opened 2020-05-16
Failing to run org-roam-capture: Wrong type argument: char-or-string-p, nil


beetleb:
### Description

I just installed org-roam and wanted to try it out. I ran org-mode-capture with the default template. What happens is that I get a prompt `File:`, and I see a list of all my existing files in the org roam directory (or rather, a list of their titles). If I try to type anything else, I get the `Wrong type argument: char-or-string-p, nil` error. If I select one of the existing files I get a capture buffer with that existing file in it.

Perhaps I'm misunderstanding the point of capture, but I was thinking I could use it to:

1. Create a new file (with a given title)
2. Allow me to customize elements that are in the file (e.g. date, etc) in the usual org capture way.

I know I can create a new file using `org-roam-find-file`, but it doesn't let me use a template.

(Note: I'm using ido, with ido-everywhere and ido-ubiquitous. If I disable all of these, the problem is the same: I can only type in the names of existing files).

Am I doing things the wrong way?

#### Backtrace

```
Debugger entered--Lisp error: (wrong-type-argument char-or-string-p nil)
  ucs-normalize-NFD-string(nil)
  #f(compiled-function (--cl-nonspacing-mark-p-- s) #<bytecode 0x122b99d>)(#f(compiled-function (char) #<bytecode 0x122b985>) nil)
  org-roam--title-to-slug(nil)
  org-roam-capture()
  funcall-interactively(org-roam-capture)
  #<subr call-interactively>(org-roam-capture record nil)
  apply(#<subr call-interactively> org-roam-capture (record nil))
  call-interactively@ido-cr+-record-current-command(#<subr call-interactively> org-roam-capture record nil)
  apply(call-interactively@ido-cr+-record-current-command #<subr call-interactively> (org-roam-capture record nil))
  call-interactively(org-roam-capture record nil)
  command-execute(org-roam-capture record)
  execute-extended-command(nil "org-roam-capture")
[...]

PR #662, opened 2020-05-17
(fix): org-roam-capture--capture shadow org-capture-templates-contexts

progfolio:
Lexically bind org-capture-templates-contexts to nil before calling org-capture
with org-roam-capture-templates. Otherwise, contexts may shadow
org-roam-templates which have same keys as org-capture-templates.

mentioned in #651


jethrokuan:
Nice catch, thanks!

PR #664, opened 2020-05-18
Add rg and find support for org-roam--list-files

siawyoung:

Adds `ripgrep` and `find` support for `org-roam--list-files`, checking and using each in order, if found, and falling back to elisp otherwise.

Implements #104.


org-roam.el
```diff
@@ -231,15 +231,52 @@ If FILE is not specified, use the current buffer's file-path."
...
         (push file result)))))
 
+(defun org-roam--list-files (dir)
+  "Return all Org-roam files located recursively within DIR.
+Ignores hidden files and directories.
+org-roam--list-files will try rg, find, and fall back to the elisp implementation if none of the external tools are found."
+(cond
+ ((executable-find "rg") (org-roam--list-files-rg dir))
+ ((executable-find "find") (org-roam--list-files-find dir))
+ (t (org-roam--list-files-elisp dir))))
```
progfolio, line 278: We should not impose the order of the commands on the user.
We could, instead, implement a user-customizable variable (e.g. `org-roam-find-file-function`) which will allow them (or a third party package) to specify the function that will return the files. We could also allow this to be a list of function symbols which will be tried in order similar to what you've written here. This also has the added benefit of letting the user select only one find command if they know that's all they'll use. (Though the default elisp command should be used as a fail-safe fallback)

org-roam.el
```diff
@@ -231,15 +231,52 @@ If FILE is not specified, use the current buffer's file-path."
...
+E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
+  (append
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext) "\"")) exts)
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext ".gpg") "\"")) exts)))
+
+(defun org-roam--list-files-rg (dir)
+  "Return all Org-roam files located recursively within DIR, using ripgrep.
+Note that this function does not first check if ripgrep is available."
+(let* (
+    (globs (org-roam--list-files-search-globs org-roam-file-extensions))
```
progfolio, line 249:
```suggestion
(let* ((globs (org-roam--list-files-search-globs org-roam-file-extensions))
```
Please join lines like these.

org-roam.el
```diff
@@ -231,15 +231,52 @@ If FILE is not specified, use the current buffer's file-path."
          (f-descendant-of-p (file-truename path)
                             (file-truename org-roam-directory))))))
 
-(defun org-roam--list-files (dir)
-  "Return all Org-roam files located within DIR, at any nesting level.
-Ignores hidden files and directories."
+(defun org-roam--shell-command-files (cmd)
+  "Run CMD in the shell and return a list of files. If no files are found, an empty list is returned."
+  (seq-filter #'s-present? (split-string (shell-command-to-string cmd) "\n")))
```
progfolio, line 236: Probably want to `shell-quote-argument` before executing `cmd`

siawyoung:
> Thanks for the contribution. Hows the performance for this vs. the elisp implementation?

I was too lazy to generate a test directory, so I cloned @jethrokuan's braindump to run some benchmarks (406 files):

```
(benchmark 1000 '(org-roam--list-files-rg "./jethrokuan/braindump"))
"Elapsed time: 9.012230s (0.399595s in 5 GCs)"

(benchmark 1000 '(org-roam--list-files-find "./jethrokuan/braindump"))
"Elapsed time: 5.543965s (0.318566s in 4 GCs)"

(benchmark 1000 '(org-roam--list-files-elisp "./jethrokuan/braindump"))
"Elapsed time: 55.781495s (3.220956s in 41 GCs)"
```

org-roam.el
```diff
@@ -231,15 +231,51 @@ If FILE is not specified, use the current buffer's file-path."
...
+E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
+  (append
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext) "\"")) exts)
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext ".gpg") "\"")) exts)))
+
+(defun org-roam--list-files-rg (dir)
+  "Return all Org-roam files located recursively within DIR, using ripgrep.
+Note that this function does not first check if ripgrep is available."
+(let* ((globs (org-roam--list-files-search-globs org-roam-file-extensions))
+    (command (s-join " " `("rg" ,dir "--files"
```
jethrokuan, line 249: use `(executable-find "rg")` here.

org-roam.el
```diff
@@ -231,15 +231,51 @@ If FILE is not specified, use the current buffer's file-path."
...
+(defun org-roam--list-files-search-globs (exts)
+  "Given EXTS, return a list of search globs.
+E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
+  (append
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext) "\"")) exts)
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext ".gpg") "\"")) exts)))
+
+(defun org-roam--list-files-rg (dir)
+  "Return all Org-roam files located recursively within DIR, using ripgrep.
+Note that this function does not first check if ripgrep is available."
```
jethrokuan, line 247: Might want to warn if rg is not found. Something like:

```
(lwarn '(org-roam) :error
                    "Cannot find \"rg\" executable. Make sure it is available, or tweak `org-roam-find-file-functions'")
```

org-roam.el
```diff
@@ -231,15 +231,51 @@ If FILE is not specified, use the current buffer's file-path."
...
+(let* ((globs (org-roam--list-files-search-globs org-roam-file-extensions))
+    (command (s-join " " `("rg" ,dir "--files"
+                            ,@(mapcar (lambda (glob) (concat "-g " glob)) globs)))))
+    (org-roam--shell-command-files command)))
+
+(defun org-roam--list-files-find (dir)
```
jethrokuan, line 253: Same comments as above.

org-roam.el
jethrokuan: :+1: Look at:

https://github.com/org-roam/org-roam/blob/78b518efd32eeb5168ac96516328a4d21ad7f104/org-roam.el#L138

and 

https://github.com/org-roam/org-roam/blob/78b518efd32eeb5168ac96516328a4d21ad7f104/org-roam.el#L416

for some inspiration.

dit7ya:
> ###### Motivation for this change
> 
> Adds `ripgrep` and `find` support for `org-roam--list-files`, checking and using each in order, if found, and falling back to elisp otherwise.
> 
> Implements #104.

If you are using `ripgrep` already, why not also use [fd](https://github.com/sharkdp/fd) which is also written in Rust and is generally faster than `find`?

jethrokuan:
>  If you are using ripgrep already, why not also use fd which is also written in Rust and is generally faster than find?

Good idea, but doesn't have to be in the same PR, we can always add it later on.

org-roam.el
siawyoung: I think it's not relevant to this use case, since I'm sending characters that I need the shell to actually treat as special characters (`"*.org"`) and not be escaped. When I use `shell-quote-argument`, the function doesn't work as expected.

org-roam.el
```diff
@@ -231,15 +252,55 @@ If FILE is not specified, use the current buffer's file-path."
...
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext) "\"")) exts)
+   (mapcar (lambda (ext) (s-wrap (concat "*." ext ".gpg") "\"")) exts)))
+
+(defun org-roam--list-files-rg (dir)
+  "Return all Org-roam files located recursively within DIR, using ripgrep."
+  (let* ((globs (org-roam--list-files-search-globs org-roam-file-extensions))
+         (executable (executable-find "rg"))
+         (command (s-join " " `(,executable ,dir "--files"
+                            ,@(mapcar (lambda (glob) (concat "-g " glob)) globs)))))
+    (if executable
```
jethrokuan, line 272: Will the function call ever reach this branch? It looks like in the main entry point you now already check for the executable using `executable-find`, so perhaps this isn't necessary anymore.

org-roam.el
```diff
@@ -160,6 +160,27 @@ extraction methods:
...
 
+(defcustom org-roam-list-files-functions '(find rg)
+  "Commands that will be used to find Org-roam files.
+
+It should be a list of symbols representing any of the following supported file
+search methods.
+
+These commands will used in order, falling back to the next one if an executable
+for a particular method is not found. If none of the methods are available, the
+default Elisp implementation will be used.
```
progfolio, line 171:
```suggestion
The methods will be tried in order until an executable for a method is found.
The  Elisp implementation is used if no method in the list is found.
```

org-roam.el
```diff
@@ -231,15 +252,55 @@ If FILE is not specified, use the current buffer's file-path."
...
       (when (and (file-readable-p file) (org-roam--org-file-p file))
         (push file result)))))
 
+(defun org-roam--list-files (dir)
+  "Return all Org-roam files located recursively within DIR.
+It will use external shell functions if defined in `org-roam-list-files-functions'."
+  (let ((valid-command (seq-find (lambda (s) (executable-find (symbol-name s))) org-roam-list-files-functions)))
+    (if valid-command
+        (funcall (intern (concat "org-roam--list-files-" (symbol-name valid-command))) dir)
+      (org-roam--list-files-elisp dir))))
```
progfolio, line 302:
```suggestion
(defun org-roam--list-files (dir)
  "Return all Org-roam files located recursively within DIR.
Use external shell functions if defined in `org-roam-list-files-functions'."
  (let* ((found (seq-find (lambda (s) (executable-find (symbol-name s))) org-roam-list-files-functions))
         ((fn (intern (concat "org-roam--list-files-" (symbol-name found))))))
    (if (not found)
        (org-roam--list-files-elisp dir)
      (unless (fboundp fn) (user-error "%s is not an implemented search method."))
      (funcall fn dir))))
```

We need to account for the case where the symbol is a valid executable, but does not have an implemented search function.

progfolio: Another potential downside to the approach of using a symbol to map to a function name is that the user cannot specify the path to the executable. One possible solution is to accept either a symbol or consed pair. e.g.
```emacs-lisp
(setq org-roam-find-file-commands '(find (rg . "/path/to/rg")))
```
Then you could check the type in the lambda passed to seq-find. If it's a pair, grab the cdr and `excutable-find` that. If it's a symbol, use the strategy you've already implemented.

jethrokuan:
Thanks!

PR #675, opened 2020-05-19
Update org-roam-bibtex link in docs

crbinz:
This should point to the official repo, not a fork (right?)


progfolio:
That is correct. Thank you!

Issue #666, opened 2020-05-18
Race condition between graph building and opening


matrixfang:
### Description
I don't know why my org-roam graph generated by org-roam-graph is empty, here's my settings

(use-package org-roam
  :ensure t
      :hook
      (after-init . org-roam-mode)
      :custom
      (org-roam-directory "\~/org/")
     (org-roam-graph-viewer "/usr/bin/open")
     (org-roam-graph-executable "/usr/local/bin/dot")
      (org-roam-index-file "\~/org/roam-index.org")
      :bind (:map org-roam-mode-map
              (("C-c n l" . org-roam)
               ("C-c n f" . org-roam-find-file)
               ("C-c n j" . org-roam-jump-to-index)
               ("C-c n b" . org-roam-switch-to-buffer)
               ("C-c n g" . org-roam-graph))
              :map org-mode-map
              (("C-c n i" . org-roam-insert))))

After take some notes and add links in them, I ran "org-roam-graph", but I get an empty graph

My default svg-opening  app is chrome, and i got this

<img width="831" alt="Screen Shot 2020-05-18 at 5 52 13 PM" src="https://user-images.githubusercontent.com/28817186/82199787-c0c77b00-9930-11ea-85c7-715fcd7ba0f8.png">

I'm a fresher about org-mode, and I don't know what's wrong

jethrokuan:
Do backlinks show in your org-roam buffer?

When you generate the graph, there should be a corresponding `.dot` file, which you can obtain by replacing the `.svg` with `.dot`. Can you share that here, or just check if it's empty?

matrixfang:
Yes, I can provied the .dot file.
Looks like the .svg and .dot was generated well.
I have to click file then the chrome will show the right .svg file, but when I run "org-roam-graph", the chrome can only show this
<img width="568" alt="Screen Shot 2020-05-18 at 10 20 19 PM" src="https://user-images.githubusercontent.com/28817186/82224625-a6a09380-9956-11ea-85d3-6888429ae1f6.png">



[Archive.zip](https://github.com/org-roam/org-roam/files/4644659/Archive.zip)

jethrokuan:
Downloading the archive, I can view the image in the folder and in the web browser, so I think your web browser is misconfigured, and can't preview svg files, not sure why.

d12frosted:
I also encounter this problem in Safari. Refreshing page helps. Could it be that `open` is executed before `svg` is fully generated?

mchowning:
[...]

PR #679, opened 2020-05-21
(fix): org-roam-graph--build in seperate process

progfolio:
org-roam-graph--build accepts a callback function which is passed the
resultant graph file as its sole argument. This prevents a race between
graph building and opening.

See: #666


progfolio:
@jethrokuan @zaeph:
I don't have a slow graph build time on my systems. Tested this by making a shell script named `delayed-dot` in my PATH:
```bash
#!/bin/bash
printf "sleeping for 5s\n"
sleep 5;
dot "$@"

And then setting `org-roam-graph-executable` to `delayed-dot`. Seems to work for me, but I'd like to see it tested with an actual slow graph build.

PR #687, opened 2020-05-23
(internal): org-roam.el formatting

progfolio:
Organize org-roam.el:
- Move functions so they are defined before they are referenced
- Move defcustoms to appropriate section, sort alphabetically
- Move interactive commands to appropriate section, autoload, sort.

Should not change any behavior, but make review easier.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -33,34 +33,38 @@
 ;;
 ;;
 ;;; Code:
-;;;; Library Requires
+;;;; Dependencies
 (require 'org)
 (require 'org-element)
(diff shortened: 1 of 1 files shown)


jethrokuan:
Did you run a command to do this?

progfolio:
Nope. Just moved everything as I was reviewing for other issues/cleanups.
Though we could automate some of this to ensure consistency in modules,
I think it's good to comb through the code manually every now and again and do something like this.
I found a dozen or so other improvements we can make which I probably wouldn't have otherwise.

Issue #682, opened 2020-05-21
Need maintainable Windows 10 install


notuntoward:
I'd like to try out `org-roam` but the [Windows 10 install instructions](https://org-roam.readthedocs.io/en/master/installation/) are formidable, and will have to be repeated every time I update the org-roam package.  Can the Windows 10 install please be made simpler?

The problem is with `emacsql-sqlite3`:  I already have the two most popular unix tool sets installed on my machine (`msys2 `and `choco`) but both have only have options to install `sqlite3`.  By default, org-roam expects plain `sqlite`, meaning that I'll need to edit the `org-roam` source code in three places and then compile; every time the `org-roam` package is updated, I'll need to redo these steps.

Is there a way that `org-roam` can detect which version of `sqlite` is installed and  then just do the right thing?

visuelledata:
These install instructions aren't even feasible for people who want to have an easily replicable setup. Any idea what the latest version of org-roam that is Windows-compatible is?

jethrokuan:
> Is there a way that org-roam can detect which version of sqlite is installed and then just do the right thing?

That's not something org-roam can control. That just means `emacsql-sqlite` is not particularly platform-agnostic, and doesn't work out-of-the-box. By default, it expects that you have the necessary tools to compile the custom sqlite executable. You can try pointing it to the path to your local sqlite installation: https://github.com/skeeto/emacsql/blob/master/emacsql-sqlite.el#L30

```
(setq emacsql-sqlite-executable-path "...")
```

That said I'm happy to accept any reasonable changes to get better out-of-the-box windows support. I have no windows machines at all, and cannot test and work on this scenario.

jueqingsizhe66:
Windows10  platform with newest version of org-roam
（org-roam works well in the older version, now just have no idea to fix it）  
The bugs just right below by executing `org-roam-insert `

```
Debugger entered--Lisp error: (emacsql-error "no such table: tags")
  signal(emacsql-error ("no such table: tags"))
  #f(compiled-function (_ code message) "Get condition for MESSAGE provided from SQLite." #<bytecode 0x5082cd9>)(#<emacsql-sqlite-connection emacsql-sqlite-connection-51037d8> 1 "no such table: tags")
  apply(#f(compiled-function (_ code message) "Get condition for MESSAGE provided from SQLite." #<bytecode 0x5082cd9>) #<emacsql-sqlite-connection emacsql-sqlite-connection-51037d8> (1 "no such table: tags"))
[...]

PR #696, opened 2020-05-24
(feat): Add support for emacs-sqlite3 #682

nobiot:
It would be great if you could consider and review this.
Happy to have constructive critiques, or accept decline if that is the conclusion of the contributors.

At the moment, I have not removed my plugin to work with markdown files -- this is my normal usage and I have only one or two test `.org` files. I understand if this might be considered interfering with proper testing. It would be great if other Windows users, Linux and macOS users to test.

Motivation: issue #682 

For Windows users, requiring emacs-sqlite is an entry barrier as
there is no out-of-the-box way to compile it when launching
org-roam for the first time.

Add customizing variable `org-roam-db-use-sqlite3`.
Set it non-nil tells org-roam to use emacs-sqlite3 instead of
the default emacs-sqlite.

####### Some test results on my Windows 10 machine
As this is critical to existing users, I tried to focus more on regression tests. 
Regression on Linux or macOS is not done; I don't have access to them.

For each case, I manually deleted `org-roam.db` before starting.

## 1. Initial condition
- Removed compiled `emacsql-sqlite.exe`
- Set `org-roam-db-use-sqlite3` to `nil`
- Installed `emacsql-sqlite3` package
- Result as expected
image

## 2. Emacsql-sqlite3 enabled
- Removed compiled `emacsql-sqlite.exe`
- Set `org-roam-db-use-sqlite3` to `t`
- Installed `emacsql-sqlite3` package
- Result as expected (with the sqlite3 specific message from the update)
image

## 3. Emacsql-sqlite3 disabled, standard emacsql-sqlite used
- Compiled `emacsql-sqlite.exe`
- Set `org-roam-db-use-sqlite3` to `nil`
- Installed `emacsql-sqlite3` package
- Result as expected
image

## 4. Emacsql-sqlite3 uninstalled, standard emacsql-sqlite used
- Compiled `emacsql-sqlite.exe`
- Set `org-roam-db-use-sqlite3` to `nil`
- Purged `emacsql-sqlite3` package 
- Result as expected
image

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -33,6 +33,7 @@
 ;;;; Library Requires
 (eval-when-compile (require 'subr-x))
 (require 'emacsql)
+(require 'emacsql-sqlite3 nil t)
 (require 'emacsql-sqlite)
 (require 'org-roam-macs)
 
@@ -57,11 +58,20 @@ when used with multiple Org-roam instances."
   :type 'string
   :group 'org-roam)
 
+(defcustom org-roam-db-use-sqlite3 nil
+  "Set the value to non-nil if Org-roam is to use sqlite3.
+This is mainly for Windows users, who do not have out-of-the-box
+support for compiling emacs-sqlite."
+  :type 'boolean
+  :group 'org-roam)
+
 (defconst org-roam-db--version 5)
 (defconst org-roam-db--sqlite-available-p
-  (with-demoted-errors "Org-roam initialization: %S"
-    (emacsql-sqlite-ensure-binary)
-    t))
+  (if org-roam-db-use-sqlite3
+      (message "Emacs-sqlite3 is used")
+    (with-demoted-errors "Org-roam initialization: %S"
+      (emacsql-sqlite-ensure-binary)
+      t)))
 
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
   "Database connection to Org-roam database.")
@@ -87,7 +97,8 @@ Performs a database upgrade when required."
     (let* ((db-file (org-roam-db--get))
            (init-db (not (file-exists-p db-file))))
       (make-directory (file-name-directory db-file) t)
-      (let ((conn (emacsql-sqlite db-file)))
+      (let ((conn (if org-roam-db-use-sqlite3 (emacsql-sqlite3 db-file)
+                    (emacsql-sqlite db-file))))
         (set-process-query-on-exit-flag (emacsql-process conn) nil)
         (puthash (file-truename org-roam-directory)
                  conn


jethrokuan:
I looked through `emacsql-sqlite3`'s code and I don't see it doing anything much different from the original `emacsql-sqlite` code, except that it uses `(executable-find "sqlite3")` as the binary. This is also possible with `emacsql-sqlite`: instead of using the default compiled binary, you can point to your installed binary.

```
(use-package emacsql-sqlite
  :custom (emacsql-sqlite-executable-path (executable-find "sqlite3"))

(use-package org-roam...)
```

Can you confirm that simply setting the binary path does not work? Would prefer not to pull in a new package.

nobiot:
@jethrokuan, tried the variable as you suggested. 

Negative.
Do you think `emacsql-sqlite-executable-path` could work with `sqlite3.exe`, instead of `emacs-sqlite.exe` ?
Do you think it's worth trying to see if I can run `sqlite3` from within Emacs?
See detail below. 

I also tried this below, with `.exe` added (this seems to be the way Emacs on Windows), with the same negative result.
```
(use-package emacsql-sqlite
  :custom (emacsql-sqlite-executable-path (executable-find "sqlite3.exe")))
```

The error message you get in the end is:
`Query timeout error: "Query timed out", 30`.

This is probably because it takes virtually forever for `sqlite3.exe` command to return when run from within Emacs's `shell` or `eshell` (both tried).

I can run `sqlite3` with no problem directly from within `PowerShell` or `cmd`.

Regardless of the issue of `sqlite3` taking forever to return, though...

See this documentation below. Notice the docstring "this is not the sqlite3"...
image

Both variables shown are set by the customization.

In addition, looking at `defvar`, `emacsql-sqlite-executable-path` is for `emacsql-sqlite.exe`, not for `sqlite3.exe`.
```
(defvar emacsql-sqlite-executable-path
  (if (memq system-type '(windows-nt cygwin ms-dos))
      "sqlite/emacsql-sqlite.exe"
    "sqlite/emacsql-sqlite")
  "Relative path to emacsql executable.")
```
`emacs-sqlite.exe` is the one that needs to be compiled from the source.

What do you think?
Do you think it's worth trying to see if I can run `sqlite3` from within Emacs?

nobiot:
My bad. I will correct my statement about "sqlite3 takes forever" in `shell` and `eshell` within Emacs.
I just read [this](http://sqlite.1065341.n5.nabble.com/sqlite-in-a-shell-td5985.html).

I tried `sqlite3` in both Emacs shells, followed by `.help`. See the screen excerpts below.
So... `sqlite3.exe` is running.

I think `sqlite3.exe` is not meant for `emacsql-sqlite-executable-path`, then?

image
**Figure 1.** eshell

image

**Figure 2.** shell

notuntoward:
@nobiot and @jethrokuan, in #682, I also found that just pointing to the sqlite binary doesn't work on Windows.

jethrokuan:
Thanks for confirming that just changing the binary path doesn't work.

> Since sqlite3 exists on Windows, Linux and MacOS, I wonder if the simplest solution wouldn't be to just use sqlite3 for all platforms?

Yes, I think that would be the ideal. Let me check if `emacsql-sqlite3` works out of the box on my machine.

jethrokuan:
Superceded by #701

Issue #458, opened 2020-04-14
Multiple slipboxes


chartoin:
Thank you for this awesome project. Before org-roam, I was manually organizing slip boxes and indices with org files and links. This project reduces a lot of effort. On top of that, the org graph viewer feature is amazing! It gives me a clear view of my notes. I am sure others feel the same way. I have a small feature request. I have added my reasoning behind it.

### Brief Abstract
Configurable graph node coloring based on sub-directory.

### Long Description
Feature:
1. Ability to color graph nodes based on which sub-directory they belong to or based on a file tag.
2. Ability to filter out graph nodes based on sub-directory or tag. This could be done similarly to how org agenda is done with the ability to have a filtered view. 

Reasoning:
I find it very useful to separate my thought process and my notes on other people's opinions. At the moment I maintain two different slipboxes in the form of two sub-directories. One of them, [main], contains my ideas or my understanding of concepts. The other one, [bibl], contains notes on papers/books/articles. There are clearly links between the two slipboxes, but I would like to differentiate between them when I view the graph and maybe even hide the [bibl] notes when I don't need to see them.

### Please check the following:
- [x] No similar feature requests

jethrokuan:
We have tag support now, so we should also add support for filtering out nodes based on tags.

qlibp:
> We have tag support now, so we should also add support for filtering out nodes based on tags.

Though org roam have tagging system right now, I can't see any difference on ui. Is there a place that could tell me what effect it brings up with the new tagging system?

jethrokuan:
> Though org roam have tagging system right now, I can't see any difference on ui

It's possibly because you haven't any tags. The documentation is here: https://org-roam.github.io/org-roam/manual/Tags.html#Tags

If you use subdirectories, and set your tag sources as such:

qlibp:
I mis-spell the #+ROAM_TAGS, sorry for that. And quite looking forward to the feature of filtering via tag! Thanks!

progfolio:
> mis-spell the #+ROAM_TAGS
@jethrokuan: another candidate for `org-roam-doctor`.

TimQuelch:
Yes I spent an unreasonable amount of time trying to figure out why my brand new shiny `#+ROAM_TAG:` keywords weren't working. Then I realised i needed an 'S' :(

zoechi:
The naming is a bit inconsistent. Both `ROAM_TAGS` and `ROAM_ALIAS` allow multiple values.

PR #703, opened 2020-05-26
(feat): doctor: add check for misspelled roam properties

jethrokuan:

Addresses https://github.com/org-roam/org-roam/issues/458#issuecomment-633658391


TimQuelch:
Is there a reason `ROAM_KEY` isn't included in the supported properties constant?

jethrokuan:
oversight on my end, thanks for pointing it out!

PR #708, opened 2020-05-27
(fix): doctor: add ROAM_KEY to supported props

jethrokuan:

h.t. @TimQuelch


jgoerzen:
It does, thanks!  Sorry I missed that.

Issue #706, opened 2020-05-27
org-roam fails on startup


Silipwn:
### Description
Org-roam fails on startup

#### Steps to Reproduce
1. Load Emacs (Results in a read error)
2. Run `org-roam-build-cache` (Results read error)

#### Backtrace
`startup`

```
Debugger entered--Lisp error: (file-error "Read error" "Is a directory" "/home/silipwn")
  insert-file-contents("")
  org-roam-db-build-cache()
  org-roam-mode()
  run-hooks(org-load-hook)
  byte-code("\300\301\302\"\210\303\304!\210\305\306!\207" [add-hook org-mode-hook #f(compiled-function () #<bytecode 0x1ded9b9>) provide org run-hooks org-load-hook] 3)
  (org-clock-persistence-insinuate)
  eval-buffer(#<buffer  *load*-159782> nil "/home/silipwn/.doom.d/config.el" nil t)  ; Reading at buffer position 3068
  load-with-code-conversion("/home/silipwn/.doom.d/config.el" "/home/silipwn/.doom.d/config.el" t t)
  load("/home/silipwn/.doom.d/config" t nomessage)
  (let (file-name-handler-alist) (load (expand-file-name "config" doom-private-dir) t (quote nomessage)))
  (condition-case e (let (file-name-handler-alist) (load (expand-file-name "config" doom-private-dir) t (quote nomessage))) ((debug doom-error) (signal (car e) (cdr e))) ((debug error) (doom--handle-load-error e (expand-file-name "config" doom-private-dir) doom-private-dir)))
  (if no-config-p nil (maphash (doom-module-loader doom-module-config-file) doom-modules) (run-hook-wrapped (quote doom-init-modules-hook) (function doom-try-run-hook)) (condition-case e (let (file-name-handler-alist) (load (expand-file-name "config" doom-private-dir) t (quote nomessage))) ((debug doom-error) (signal (car e) (cdr e))) ((debug error) (doom--handle-load-error e (expand-file-name "config" doom-private-dir) doom-private-dir))) (load custom-file (quote noerror) (not doom-debug-mode)))
[...]

Issue #688, opened 2020-05-23
org-roam--list-files is not stripping out ANSI color codes


sunnyhasija:
### Description
When using the latest commit with DOOM emacs org-roam and org-roam-server do not load.

```Warning (initialization): An error occurred while loading ‘/home/cantos/.emacs.d/init.el’:

Error in private config: config.el, (file-missing Opening input file No such file or directory /home/cantos/[�[1;32mI�[0m] �[1;31mtheme�[0m: Set theme to �[1;37mcolors.json�[0m.)

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace
```

#### Steps to Reproduce

1. Load DOOM Emacs
2. Run `org-roam-server`

#### Backtrace
```
Debugger entered--Lisp error: (file-missing "Opening input file" "No such file or directory" "/home/cantos/[\033[1;32mI\033[0m] \033[1;31mtheme\033[0m: Set theme to \033[1;37mcolors.json\033[0m.")
  insert-file-contents("[\033[1;32mI\033[0m] \033[1;31mtheme\033[0m: Set theme to \033[1;37mcolors.json\033[0m.")
  org-roam-db-build-cache()
  org-roam-mode()
  run-hooks(org-load-hook)
  byte-code("\300\301\302\"\210\303\304!\210\305\306!\207" [add-hook org-mode-hook #f(compiled-function () #) provide org run-hooks org-load-hook] 3)
  require(org)
  byte-code("\300\301!\210\300\302!\210\300\303!\210\300\304!\210\300\305!\210\300\306!\210\300\307!\210\300\310!\210\300\311!\207" [require simple-httpd json subr-x org ox ox-html org-roam org-roam-graph org-roam-buffer] 2)
  require(org-roam-server nil nil)
  eval-buffer(# nil "/home/cantos/.doom.d/config.el" nil t)  ; Reading at buffer position 7691
  load-with-code-conversion("/home/cantos/.doom.d/config.el" "/home/cantos/.doom.d/config.el" t t)
  load("/home/cantos/.doom.d/config" t nomessage)
  (let (file-name-handler-alist) (load (expand-file-name "config" doom-private-dir) t (quote nomessage)))
[...]

PR #709, opened 2020-05-27
(fix): org-roam-db-build-cache filter non-file input

progfolio:
If a shell command is used to find files, it may include
output from another command.

see: #706, #688


progfolio:
@jethrokuan: This is the simplest solution I could think of. Note it does not stop invalid paths from slipping through, just input that is not a file.

jethrokuan:
instead of using `f-file?`, we can use `org-roam--org-roam-file-p`, which should hopefully take care of the weird cases.

PR #711, opened 2020-05-28
(fix): fix support for Emacs 26 built-in org

CeleritasCelery:
`org-link-make-string` is not defined until org version `9.3`. Emacs 26 ships with org `9.1`, so this function is undefined if the user is using the built-in org version. Backporting this function to support Emacs 26, Since that is the minimum supported version.

### DIFF
--- a/org-roam-compat.el
+++ b/org-roam-compat.el
@@ -80,6 +80,9 @@
 (define-obsolete-function-alias 'org-roam-db--maybe-update 'org-roam-db--update-maybe
   "org-roam 1.1.0")
 
+(when (version< (org-version) "9.3")
+  (defalias 'org-link-make-string 'org-make-link-string))
+
 ;;;; Variables
 (define-obsolete-variable-alias 'org-roam-graphviz-extra-options
   'org-roam-graph-extra-config "org-roam 1.0.0")


progfolio:
Thank you for the pull request. 
This should be moved to the `org-roam-comapt.el` file.

CeleritasCelery:
I moved it to `org-roam-compat` and changed it to satisfy the linter.

progfolio:
That's the appropriate place, but changing the function name will not work. That will define a new function that isn't referenced in the source anywhere. The original function name should be used.

progfolio:
I'll work on a fix for the linter.

CeleritasCelery:
turns out that the function was just renamed in org `9.3` so we can just map the old name to the new one. That makes this change much simpler.

org-roam-compat.el
```diff
@@ -80,6 +80,9 @@
 (define-obsolete-function-alias 'org-roam-db--maybe-update 'org-roam-db--update-maybe
   "org-roam 1.1.0")
 
+(when (version< (org-version) "9.3")
+  (fset 'org-link-make-string 'org-make-link-string))
```
progfolio, line 84:
```suggestion
(when (version< (org-version) "9.3")
  (defalias 'org-make-link-string 'org-link-make-string nil))
```
defalias is the proper choice here because the user will be able to see the aliasing in the functions' help entries.

CeleritasCelery: Okay. So long as you take care of the `package-lint` error. You could just exclude the whole file,

progfolio: @jethrokuan: 
makem.sh does have an --exclude option which would allow us to circumvent the CI for cases such as these, but I don't foresee us needing these types of changes too often. I think it's wiser to just manually review/accept the changes than turn off linting for `org-roam-compat.el`.

Thoughts?

CeleritasCelery: Perhaps I am misunderstanding. Won't this change cause CI to fail for all future PR's as well? I assume that `pacakge-lint` is run on the whole repo.

jethrokuan: I think it's alright to turn off linting for `org-roam-compat.el` actually.

Issue #446, opened 2020-04-12
No links in org roam buffer 


RCoeurjoly:
### Description
No links appear in org roam buffer, even after executing:
`org-roam-db-build-cache`
And getting the message:

> files: 3, links: 1, titles: 3, refs: 0, deleted: 0

#### Steps to Reproduce

1. Clone https://github.com/RCoeurjoly/babel_sandbox
2. Load Emacs
3. Open file babel_sandbox/20200412134256-literate_programming.org
4. Run `org-roam-db-build-cache`
5. Run `org-roam`

#### Expected Results
Backlinks in *org-roam* buffer
#### Actual Results
Empty *org-roam* buffer

### Versions
- Emacs (`C-h v emacs-version`): v26.3
- Org-roam commit: org-roam-20200411.1041

RCoeurjoly:
Doing some queries in sqlite3, I get the following results:

> sqlite> select * from files;
"/home/rcl/babel_sandbox/babel_sandbox.org"|"5317e809b74f0d0633f4ec20454bfddd772f7330"|(24211 5156 318286 602000)
"/home/rcl/babel_sandbox/20200412134256-literate_programming.org"|"af9d8cdcf355072d9e261e7d2e8de21a54d6d3cf"|(24211 5156 318286 602000)
"/home/rcl/babel_sandbox/2020-04-12.org"|"0eb34f0df452163bd82effb7378f4d20db41c58c"|(24211 5156 318286 602000)
sqlite> select * from files;
"/home/rcl/babel_sandbox/babel_sandbox.org"|"5317e809b74f0d0633f4ec20454bfddd772f7330"|(24211 5156 318286 602000)
"/home/rcl/babel_sandbox/20200412134256-literate_programming.org"|"af9d8cdcf355072d9e261e7d2e8de21a54d6d3cf"|(24211 5156 318286 602000)
"/home/rcl/babel_sandbox/2020-04-12.org"|"0eb34f0df452163bd82effb7378f4d20db41c58c"|(24211 5156 318286 602000)
sqlite> select * from links;
"/home/rcl/babel_sandbox/2020-04-12.org"|"/home/rcl/babel_sandbox/20200412134256-literate_programming.org"|"roam"|(:content "Trying org-roam\n[[file:20200412134256-literate_programming.org][literate programming]]" :point 22)
sqlite> select * from refs;
sqlite> select * from titles;
"/home/rcl/babel_sandbox/babel_sandbox.org"|
[...]

Issue #116, opened 2020-02-17
Backlinks don't always work


honza:
I checked the graph, and everything is connected the way I'd expect. However, someones the `org-roam` buffer has no backlinks where there should be some.

alanz:
That could be prior to the fix landed at #107 for #102

honza:
I'm running the latest `develop` 2d206134fde7158d95e49343d776dbedda1769c8

But I'll read through those issues just in case. Thanks

alanz:
Well, they would have left junk in your db prior to the fix landing. It would create duplicate files for the same reference. So the backlink resolves to only one of them, and it seems the others have been missed.

honza:
Does `org-roam` create some sort of database that's persisted to disk? I assume it was all in-memory.

honza:
I cleared the cache, and restarted Emacs just to be sure but the issue persists.

jethrokuan:
> Does org-roam create some sort of database that's persisted to disk? I assume it was all in-memory.

It's all in memory.

I'd need more info about what backlinks are missing in order to look into this

brabalan:
I also have one backlink that does not work. This is what happened:
- I created a file using `org-roam-insert`
- noticing that the file name was not what I expected (recent changes on automatic naming), I deleted the file, changed my configuration, and did an `org-roam-insert` with the same title, that created a different file
- the backlink is not present

I can provide more debugging information, but I don't know how.

jethrokuan:
> I deleted the file

The cache doesn't yet track file deletions and renames.It does however, rebuild backlinks for a given file on save.

> did an org-roam-insert with the same title, that created a different file
> the backlink is not present

I just tested this scenario where I:

1. created a file using `org-roam-insert`
2. deleted it
3. created another file using `org-roam-insert`
4. *saved the file*
5. visit new file

The backlinks are there as expected.

honza:
Here is what I did:

* Delete all files on disk
* Quit Emacs
* Update org-roam to latest
* Start emacs
* Manually call the async cache building function
* `org-roam-find-file` to create a new named note
* Add a toplevel heading (i.e. `*`), and in that heading call `org-roam-insert` to reference a yet non-existent note
* Save the file
* Navigate to the reference in the heading, hit enter
* The linked reference opens in a new split.
* Switch to the new buffer
* Run `M-x org-roam`
* The popup buffer is empty.

jethrokuan:
@honza can I also know:

[...]

PR #712, opened 2020-05-28
(fix): autoloaded commands load org-roam-mode

progfolio:
Users should not have to explicitly enable the minor mode before using
autoloaded commands.

see #446,  #116 and I'm sure others.
I have hit this issue myself on occasion.


jethrokuan:
Lgtm once lint errors are fixed

progfolio:
Taking care of that now. I made these changes in a personal branch awhile ago and almost forgot about them.

PR #714, opened 2020-05-28
(fix): don't print no citelinks if org-ref is not present

jethrokuan:

Change confusing behaviour where org-roam buffer prints "no citelinks"
even when org-ref is not present.


TimQuelch:
Might it be worth not printing cite links at all of no `ROAM_KEY` is present?

jethrokuan:
oops @TimQuelch that makes sense too.

TimQuelch:
I saw the merge happen right as I was typing it out :o

jethrokuan:
Done in #717

PR #730, opened 2020-05-29
Properties in org-mode are case insensitive

pjones:

This change allows lowercase properties such as "#+title:" and "#+roam_ref:" for those of us who think it looks better.

If this change warrants an update to the changelog please let me know and I'll do that.


pjones:
@jethrokuan I actually no longer use org-roam.  I was doing some house cleaning and decided to send over these patches.  I'll keep this PR along with #731 open in case someone wants to pick them up and finish them.

jethrokuan:
Superceded by #757

PR #731, opened 2020-05-29
Remove reliance on the current buffer when opening the *org-roam* window

pjones:

If the `org-roam-buffer--get-create` function is called from a hook
then `get-buffer-window` will return `nil` since the current buffer
may not have been placed in a window yet.  This eventually leads to an
error.

This change moves some code around so that we can use
`save-selected-window` to allow temporary focus on the org-roam
side window so the resizing functions can work.

I also added another window parameter: `no-other-window`.  I'm not
100% sure about this because I don't have a good feel for how the
side window is used.  I want `other-window` to avoid it but I'm not
sure if others do.

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -244,26 +244,32 @@ Valid states are 'visible, 'exists and 'none."
 
 (defun org-roam-buffer--get-create ()
   "Set up the `org-roam' buffer at `org-roam-buffer-position'."
-  (let ((window (get-buffer-window))
-        (position
+  (let ((position
          (if (member org-roam-buffer-position '(right left top bottom))
              org-roam-buffer-position
            (let ((text-quoting-style 'grave))
              (lwarn '(org-roam) :error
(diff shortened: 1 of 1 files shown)


org-roam-buffer.el
```diff
@@ -244,26 +244,32 @@ Valid states are 'visible, 'exists and 'none."
...
-      ((or 'right 'left)
-       (org-roam-buffer--set-width  (round (* (frame-width)  org-roam-buffer-width))))
-      ((or 'top  'bottom)
-       (org-roam-buffer--set-height (round (* (frame-height) org-roam-buffer-height)))))
-    (select-window window)))
+           'right))
+        (params
+         (list
+          (cons 'no-other-window org-roam-buffer-no-delete-other-windows)
+          (cons 'no-delete-other-windows org-roam-buffer-no-delete-other-windows))))
```
jethrokuan, line 258: I think we can move all these into a single variable: `org-roam-buffer-window-parameters`, which would be a list of cons cells.

You may choose to do this in the same PR, or a different one.

PR #747, opened 2020-06-02
Balance files and tags table

goktug97:
When a new file is created using `org-roam-find-file` the empty tags are inserted to the database. This causes imbalance between `files` table and `tags` table. If we are going to insert `nil` tags to the `tags` table, we should do it from the start.

I have started working on this https://github.com/org-roam/org-roam-server/issues/24 issue to see if I can make it work.

```elisp
(let* ((node-query `[:select [titles:file titles tags] :from titles
                     ,@(org-roam-graph--expand-matcher 'file t)
                     :inner :join tags :on (= titles:file tags:file)]))
  (print (org-roam-db-query node-query)))
```


TimQuelch:
Is there a reason we would actually want entries in the tags table with no tags?

goktug97:
It would be a easy fix which would be more logical in terms of db structure but I wont be able to do join as I described above. Can you think of another efficient method to this? Matching with two loops might suffer with large db. what I want is “join two tables if row doesn’t exist insert nil”

TimQuelch:
Yeah left join is what you want

goktug97:
I will update the pull request to only change org-roam-db—update-tags when I have access to my computer. Current changes are not required as left join does exactly what I want.

org-roam-db.el
```diff
@@ -333,10 +333,10 @@ connections, nil is returned."
 
 (defun org-roam-db--update-tags ()
   "Update the tags of the current buffer into the cache."
-  (let* ((file (file-truename (buffer-file-name)))
-         (tags (org-roam--extract-tags)))
+  (when-let ((file (file-truename (buffer-file-name)))
+             (tags (org-roam--extract-tags)))
     (org-roam-db-query [:delete :from tags
-                        :where (= file $s1)]
+                                :where (= file $s1)]
```
jethrokuan, line 339: Can you undo this indentation change?

jethrokuan:
Also, would be good to have this rebased on top of master and force-pushed, there's a lot of commits here

goktug97:
#759

PR #759, opened 2020-06-04
change `org-roam-db--update-tags` to only add non-nil tags

goktug97:
Don't add nil tags to the tags table when creating a note

#747 Couldn't manage to do `git rebase`.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -332,8 +332,8 @@ connections, nil is returned."
 
 (defun org-roam-db--update-tags ()
   "Update the tags of the current buffer into the cache."
-  (let* ((file (file-truename (buffer-file-name)))
-         (tags (org-roam--extract-tags)))
+  (when-let ((file (file-truename (buffer-file-name)))
+             (tags (org-roam--extract-tags)))
     (org-roam-db-query [:delete :from tags
                         :where (= file $s1)]
                        file)


org-roam-db.el
```diff
@@ -332,8 +332,8 @@ connections, nil is returned."
 
 (defun org-roam-db--update-tags ()
   "Update the tags of the current buffer into the cache."
-  (let* ((file (file-truename (buffer-file-name)))
-         (tags (org-roam--extract-tags)))
+  (when-let ((file (file-truename (buffer-file-name)))
+             (tags (org-roam--extract-tags)))
```
TimQuelch, line 336: I don't think this will behave correctly when a file originally has a tag, and then they are later removed.

`org-roam--extract-tags` will return nil and then the entry will not be deleted from the db.

I think the db insert should be guarded by the when, but not the delete.

e.g.
`
  (let ((file (file-truename (buffer-file-name)))
        (tags (org-roam--extract-tags)))
    (org-roam-db-query [:delete :from tags
                        :where (= file $s1)]
                       file)
    (when tags (org-roam-db--insert-tags file tags))))
`

TimQuelch:
I addressed that small fix I mentioned in https://github.com/org-roam/org-roam/pull/760 because this is already merged

org-roam-db.el
goktug97: Yes you are right. I didn’t realize that.

PR #760, opened 2020-06-05
Fix update-tags to correct delete removed tags

TimQuelch:
see https://github.com/org-roam/org-roam/pull/759#discussion_r435704588


jethrokuan:
Thanks for the catch!

Issue #768, opened 2020-06-07
Prefer lower case letters for keywords


doolio:
### Brief Abstract
Prefer lower case letters for keywords such as `#+title:`

### Long Description
With this `org-mode` [commit](https://code.orgmode.org/bzg/org-mode/commit/13424336a6f30c50952d291e7a82906c1210daf0) lower case letters are preferred. I believe either case is acceptable so not raising as a bug.

### Please check the following:

- [X] No similar feature requests

jethrokuan:
Everything should be lowercased now!

doolio:
Thank you for the quick turnaround. And this [blog post](https://scripter.co/org-keywords-lower-case) describes  a function to lower case existing keywords if users want to update existing files.

PR #769, opened 2020-06-07
(feat): prefer lowercase versions of org in-file settings

progfolio:
See #768


progfolio:
> Do you think we should lower-case in the code (beyond docstrings) as well? 

Up to you. Can be easily done project-wide with a regex.

jethrokuan:
Merging because it's not a blocker.

PR #772, opened 2020-06-07
(fix): fix prop extraction failures

### LINKED ISSUE #771: `org-roam--extract-titles` ignore `title` and `roam_alias` properties and instead uses headline
### Description

Even if `#+title:` or `#+roam_alias` is used, `org-roam--extract-titles` returns the first headline.

I believe this problem was introduced in #770. `org-roam--extract-global-props` seems to be the relevant part of the code: https://github.com/org-roam/org-roam/blob/9d5a34d0f4660381272e8cc0ab3ba2bd2de41ff5/org-roam.el#L453 `(org-element-property :key kw)` looks like it returns `"TITLE"` or `"ROAM_ALIAS"` so it it will never satisfy `string-collate-equalp` for `"title` or `"roam_alias"`; it looks like `string-collate-equalp` has the [`ignore-case`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Text-Comparison.html) option set to `t` though so I'm confused why it's returning `nil`.

#### Steps to Reproduce

1. Load example org file:
```elisp
#+title: example-title
#+roam_alias: "example-alias"

* example-headline

#+begin_src emacs-lisp :tangle yes
  (org-roam--extract-titles)
#+end_src
```
2. Evaluate the `emacs-lisp` code block.

#### Backtrace

N/A? It doesn't throw an error, `org-roam--extract-titles-title` and `org-roam--extract-titles-alias` just return `nil`.

#### Expected Results

```elisp
#+RESULTS:
| example-title | example-alias |

```

#### Actual Results

```elisp
#+RESULTS:
| example-headline |
```

### Environment

- Emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin19.0.0, NS appkit-1894.10 Version 10.15.1 (Build 19B88))
 of 2019-11-17
- Framework: Spacemacs
- Org: Org mode version 9.3.7 (9.3.7-elpa @ /Users/santi/.emacs.d/elpa/26.3/develop/org-20200602/)
- Org-roam: 1.1.1
- Org-roam commit: https://github.com/org-roam/org-roam/commit/9d5a34d0f4660381272e8cc0ab3ba2bd2de41ff5

jethrokuan:

This fixes a bug introduced in #770. Interestingly,
`org-element-property` always returns the prop in allcaps, so a
string-equal would have been sufficient. This commit reverts all usages
of `org-roam--extract-global-props` to use the upcase version of the
properties.

Fixes #771

### DIFF
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -93,7 +93,7 @@ AST is the org-element parse tree."
     (org-element-map ast 'keyword
       (lambda (kw)
         (let ((key (org-element-property :key kw)))
-          (when (and (string-prefix-p "roam_" key t)
+          (when (and (string-prefix-p "ROAM_" key t)
                      (not (member key org-roam-doctor--supported-roam-properties)))
             (push
              `(,(org-element-property :begin kw)
--- a/org-roam.el
+++ b/org-roam.el
@@ -450,7 +450,7 @@ The search terminates when the first property is encountered."
     (dolist (prop props)
       (let ((p (org-element-map buf 'keyword
                  (lambda (kw)
-                   (when (string-collate-equalp (org-element-property :key kw) prop nil t)
+                   (when (string-equal (org-element-property :key kw) prop)
                      (org-element-property :value kw)))
                  :first-match t)))
         (push (cons prop p) res)))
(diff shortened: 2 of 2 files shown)


zaeph:
> Interestingly,
> `org-element-property` always returns the prop in allcaps

Have you filed an issue upstream for this?  Seems like it's something they'd need to address.

zaeph:
Found the culprit in `org-element-keyword-parser`:
```el
(upcase (match-string-no-properties 1))))
```

I'm going to investigate further to see if this has any undesirable effect, and I'll submit the patch upstream.

PR #774, opened 2020-06-07
(feat): org-roam-commit command

progfolio:
Add org-roam-commit to get current vc commit for org-roam.
Add org-roam-commit output to org-roam-diagnostics.

Should help with debugging issues

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -42,6 +42,7 @@
 (require 'cl-lib)
 (require 'dash)
 (require 'f)
+(require 'files) ; file-chase-links
 (require 's)
 (require 'seq)
 (eval-when-compile (require 'subr-x))
@@ -1114,6 +1115,23 @@ Otherwise, behave as if called interactively."
 (defalias 'org-roam 'org-roam-buffer-toggle-display)
 
 ;;;###autoload
+(defun org-roam-commit (&optional message)
+  "Return the current commit of `org-roam' repository, nil if not found.
+Interactively, or when MESSAGE is non-nil, show in the echo area."
+  (interactive)
+  (when-let ((git (executable-find "git"))
+             (library (ignore-errors (file-chase-links (locate-library "org-roam.el")))))
+    (let* ((libdir (file-name-directory library))
+           (gitdir (expand-file-name ".git" (file-name-directory libdir))))
+      (when (and (file-exists-p gitdir)
+                 (file-exists-p (expand-file-name "../org-roam.el" gitdir)))
+        (when-let ((default-directory libdir)
+                   (commit (string-trim (shell-command-to-string "git rev-parse HEAD"))))
+          (if (or message (called-interactively-p 'interactive))
+              (message "%s" commit)
+            commit))))))
+
+;;;###autoload
 (defun org-roam-diagnostics ()
   "Collect and print info for `org-roam' issues."
   (interactive)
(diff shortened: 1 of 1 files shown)


progfolio:
I tried to be fairly guarded in the approach (inspired by magit-version), but I have not tested this on Windows. If the command fails it should just return nil and its output should not be included in `org-roam-diagnostics` output.

nobiot:
@progfolio, happy to. Let me see if I can do this some time this week...

nobiot:
Hi @progfolio, I tried `org-roam-commit` and `org-roam-diagnostics`.
See the results for the commit, nothing seems to have happened.
`(executable-find "git") finds the executable file,  so that's not the issue here.

E790FC29-DBF6-4EB6-9EB7-07660A0E9F19
**Figure 1.** Git animation for `org-roam-commit` 

image
**Figure 2.** Result from `org-roam-diagnostics` (chose Doom)

progfolio:
Thank you for taking the time to test.
Are you familiar with using edebug to step through a function?

nobiot:
Yes, and happy to do that. 
Let me try tomorrow -- it's late in my time zone (23:00).

org-roam.el
```diff
@@ -1113,6 +1114,23 @@ Otherwise, behave as if called interactively."
...
+;;;###autoload
+(defun org-roam-commit (&optional message)
+  "Return the current commit of `org-roam' repository, nil if not found.
+Interactively, or when MESSAGE is non-nil, show in the echo area."
+  (interactive)
+  (when-let ((git (executable-find "git"))
+             (library (ignore-errors (file-chase-links (locate-library "org-roam.el")))))
+    (let* ((libdir (file-name-directory library))
+           (gitdir (expand-file-name ".git" (file-name-directory libdir))))
+      (when (and (file-exists-p gitdir)
```
nobiot, line 1126: It's this condition. I use Doom, so use `straight`, and with cloned git repo that resides in my local machine.
This means.... The `.git` file does not exist in this location that the previous part finds: `C:\Users\nobiot\.emacs.d\.local\straight\build\org-roam\.git`

nobiot: "this condition" is meant to be `(and file-exists-p gitdir) -- sorry if the way I did this above was not clear. I guess commenting without using the review function would have been easier. Windows seems to correctly locate where the `.el` file for the package is. The issue seems to be related to how I use Emacs and packages.

nobiot:
Sorry if the way I did this above was not clear. I guess commenting without using the review function would have been easier. Windows seems to correctly locate where the `.el` file for the package is.

org-roam.el
progfolio: Thanks for looking into this.

I use straight as well, but on Linux.
On *nix systems, straight's build directory symlinks *.el files back to the original repo, so we find the correct directory. IIRC, that is not the case on Windows.

progfolio: https://github.com/raxod502/straight.el/blob/a7f94876b2bf96d2595706270be6630ecc94f0d3/straight.el#L692-L695

Looks like for Windows, straight copies the source files into the build dir.

progfolio:
This will only work for a subset of users. Those using package.el will not have the git directory either.
Accounting for OS differences may be more trouble than it's worth, considering the limited audience.

@nobiot: Thanks for looking into this. I appreciate your help.

@jethrokuan: If having the commit is useful for us, the easiest way to implement something like this would involve us programmatically appending the hash to the Version string or storing it in our own metadata file that is guaranteed to ship w all package managers.

Going to put a freeze on this for now.

jethrokuan:
> @jethrokuan: If having the commit is useful for us, the easiest way to implement something like this would involve us programmatically appending the hash to the Version string or storing it in our own metadata file that is guaranteed to ship w all package managers.

Yes, I think Magit does this in a Makefile, I've looked into it, but it seemed like a lot instrumentation. I think this is still useful for those who don't use package.el.

progfolio:
Yeah, I was trying to think of  way to cleverly do it with a pre-commit hook.
I'll thnk on it some more.

PR #775, opened 2020-06-08
(internal): parse AST once

jethrokuan:
Parse the buffer once to obtain AST, to be used in downstream functions.

This makes Org-roam more performant, but the functions get ugly.

Another idea is to store the currently required AST in a global variable so they
don't have to be passed around as arguments, but this is another surface area
for bugs (maybe a macro `with-ast`, not sure)

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -277,8 +277,8 @@ is used in place of the titles for completions.
 
 If you wish to add your own title extraction method, you may push a symbol
 ='foo= into =org-roam-title-sources=, and define a
-=org-roam--extract-titles-foo= which accepts no arguments. See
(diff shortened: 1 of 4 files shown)


TimQuelch:
> Another idea is to store the currently required AST in a global variable

This sounds kinda gross and bugprone.

jethrokuan:
Both are ugly. A good solution is to make `org-element-parse-buffer` idempotent (not require another pass if the buffer string is the same). I might go with that approach (a LRU cache of limited size)

jethrokuan:
Update, it looks the cache actually leads to a slight slowdown.

jethrokuan:
Closing, as this doesn't seem to provide any perf improvements.

PR #776, opened 2020-06-08
(UX): Add more meaningful error message for sqlite3 

nobiot:
Context is [discussion in Discourse](https://org-roam.discourse.group/t/error-apply-cannot-find-executable-sqlite3/316).

When the computer is missing sqlite3 or the path is not correctly
set, you get an error message like this:

  `Error: apply: Cannot find executable “sqlite3”!`

The motivation of this simple change is to see whether a simple
change to the error message might not change UX, and better
guide end-users who may not be familiar with dealing with
these terse error messages.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -164,6 +164,15 @@ operating system. You can verify that this is the case by executing:
   (executable-find "sqlite3")
 #+END_SRC
 
+If you have ~sqlite3~ installed, and ~executable-find~ still reports ~nil~, then
+it is likely that the path to the executable is not a member of the Emacs
+variable ~exec-path~. You may rectify this by manually adding the path within
+your Emacs configuration:
+
+#+BEGIN_SRC emacs-lisp
+  (add-to-list 'exec-path "path/to/sqlite3")
+#+END_SRC
+
 * Getting Started
 
 This short tutorial describes the essential commands used in Org-roam, to help
--- a/doc/org-roam.texi
+++ b/doc/org-roam.texi
@@ -119,6 +119,7 @@ Graphing
 
 Roam Protocol
 
+* _::
 * Installation: Installation (1). 
 * The @samp{roam-file} protocol::
 * The @samp{roam-ref} Protocol::
@@ -180,7 +181,7 @@ Emacs is also a fantastic interface for editing text, and we can inherit many of
 @chapter A Brief Introduction to the Zettelkasten Method
 
 Org-roam provides utilities for maintaining a digital slip-box. This section
-aims to provide a brief introduction to the "slip-box", or "Zettelkasten"
+aims to provide a brief introduction to the ``slip-box'', or ``Zettelkasten''
 method. By providing some background on the method, we hope that the design
 decisions of Org-roam will become clear, and that will aid in using Org-roam
 appropriately. In this section we will also introduce terms commonly used within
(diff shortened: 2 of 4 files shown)


nobiot:
@progfolio 
Thank you for the review and comment. 
- Added `Cannot find executable “sqlite3”` as the first sentence. 
- Dropped "Please" to align the additional help message to the rest of the terse and neutral-tone error message convention of Elisp (my interpretation)

jethrokuan:
Can we move the check to the enabling of `org-roam-mode`? This function is called a lot, and a conditional check like this doesn't need to be called so often.

progfolio:
Good idea. Do you think we should point to external documentation or to the internal info docs in the message?

jethrokuan:
I'd prefer internal, actually, but I'm actually not sure how to do so.

progfolio:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html

>  To make a hyperlink to Info documentation, write the single-quoted name of the Info node (or anchor), preceded by ‘info node’, ‘Info node’, ‘info anchor’ or ‘Info anchor’. The Info file name defaults to ‘emacs’. For example,

          See Info node `Font Lock' and Info node `(elisp)Font Lock Basics'.

nobiot:
Thanks for approving.
I’d also like to know how to output a clickable link. [Neither echo nor minibuffer seem to allow for that](https://stackoverflow.com/questions/8440013/clickable-text-in-emacs-minibuffer) (unverified). I thought opening a child frame (or whatever that can have a buffer) would be overkill, but I’d love to know if there is a light alternative...

zaeph:
I'm also in favour of pointing to the internal documentation, and of running the check on `org-roam-mode`.

progfolio:
We could direct the user to the appropriate info node in the message:
e.g.
`M-x info for more information at Org-roam >  Installation > Post-Installation Tasks`

nobiot:
Let me step back to the basics. I don't see/don't know how to view Org-roam info when I do `M-x info`. I suspect I'd need to install documentation somehow? How do we do it?

zaeph:
> Let me step back to the basics. I don't see/don't know how to view Org-roam info when I do `M-x info`. I suspect I'd need to install documentation somehow? How do we do it?

Try to run `Info-menu` by pressing `m`, then you get a completion list where `Org-roam` should appear.

nobiot:
Thank you, @zaeph.
Clearly, I'm missing something. I guess I need to manually put `org-roam.texi` somewhere in my Emacs directory... Is this supposed to happen automatically?

image

zaeph:
Now that you mention it, I don't think our MELPA recipe is creating the `.texi` files for users.

Compare [Magit's recipe](https://github.com/melpa/melpa/blob/68635e3c52d12c234200680fd393a0b8748d2993/recipes/magit) to [ours](https://github.com/melpa/melpa/blob/68635e3c52d12c234200680fd393a0b8748d2993/recipes/org-roam).  I'll look into it later.

nobiot:
I followed this [manual ](https://www.gnu.org/software/emacs/manual/html_node/efaq/Installing-Texinfo-documentation.html).
1. Manually convert `org-roam.texi` to org-roam.info` (`M-x texinfo-format-buffer`)
   I had this error `texinfo-unsupported: @headitem is not handled by texinfo`, but managed to create the `.info` file.

2. I don't have `install-info`; that means I cannot "update the main Info directory menu"

3. Used manual way to open the info file via `C-u C-h i path/to/org-roam.info <RET>`
   (Notice the @noindent not converted correctly?)

image

zaeph:
We're already generating the `.info` in the repo, but I just think we're not installing it.  It should be trivial to fix.

nobiot:
For this PR, we are aiming to guide the beginner users of Emacs (probably, primarily Windows users, who have issues with installing `sqlite3`. Especially because of Point 2 above ( missing `install-inf`), I'm leaning towards external source...

nobiot:
> 
> 
> We're already generating the `.info` in the repo, but I just think we're not installing it. It should be trivial to fix.

Oh cool! If installation does not have to be done on the user side, then that's good. If installation needs to be done by the user within Emacs or commandline, I feel that it would defeat the original purpose of guiding novice users. What do you think?

zaeph:
No yeah, definitely.  The goal is for those pages to be installed with Org-roam, I'm not sure what went on here.  I've opened #782 to resolve it.

PR #790, opened 2020-06-10
Add -other-window versions of org-roam-{find-file,switch-to-buffer,jump-to-index}

philip-bl:

Emacs has some [nice keybindings for manipulating buffers in another window](https://www.gnu.org/software/emacs/manual/html_node/emacs/Pop-Up-Window.html). These keybindings start with `C-x 4`, e.g.
* `C-x 4 b` opens an existing buffer in another window.
* `C-x 4 f` opens a file in another window.

I wanted to have similar keybindings for org-roam functions: `org-roam-find-file`, `org-roam-switch-to-buffer`, `org-roam-jump-to-index`. So I added `other-window` optional argument to these functions and created `org-roam-find-file-other-window`, `org-roam-switch-to-buffer-other-window`, `org-roam-jump-to-index-other-window`. Now we can have org-roam related keybindings which start with `C-x 4`. This is how I use it:

```
(define-key org-roam-mode-map (kbd "C-x 4 C-c n f") #'org-roam-find-file-other-window)
(define-key org-roam-mode-map (kbd "C-x 4 C-c n j") #'org-roam-jump-to-index-other-window)
(define-key org-roam-mode-map (kbd "C-x 4 C-c n b") #'org-roam-switch-to-buffer-other-window)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Here's a sample configuration with using `use-package`:
       :bind (:map org-roam-mode-map
               (("C-c n l" . org-roam)
                ("C-c n f" . org-roam-find-file)
+               ("C-x 4 C-c n f" . org-roam-find-file-other-window)
                ("C-c n g" . org-roam-show-graph))
               :map org-mode-map
               (("C-c n i" . org-roam-insert))))
--- a/org-roam.el
+++ b/org-roam.el
@@ -1305,6 +1305,18 @@ If OTHER-WINDOW is non-nil, open the buffer in another window similarly to
           (org-roam-capture--capture))))))
 
 ;;;###autoload
+(defun org-roam-find-file-other-window (&optional initial-prompt completions filter-fn)
+  "Find and open an Org-roam file in another window.
+INITIAL-PROMPT is the initial title prompt.
+COMPLETIONS is a list of completions to be used instead of
+`org-roam--get-title-path-completions`.
+FILTER-FN is the name of a function to apply on the candidates
+which takes as its argument an alist of path-completions.  See
+`org-roam--get-title-path-completions' for details."
+  (interactive)
+  (org-roam-find-file initial-prompt completions filter-fn t))
+
+;;;###autoload
 (defun org-roam-find-directory ()
   "Find and open `org-roam-directory'."
   (interactive)
--- a/org-roam.el
+++ b/org-roam.el
@@ -1524,6 +1524,12 @@ linked, lest the network graph get too crowded."
 
 
 ;;;###autoload
+(defun org-roam-switch-to-buffer-other-window ()
+  "Switch to an existing Org-roam buffer in another window."
+  (interactive)
+  (org-roam-switch-to-buffer t))
+
+;;;###autoload
 (defun org-roam-version (&optional message)
   "Return `org-roam' version.
 Interactively, or when MESSAGE is non-nil, show in the echo area."
--- a/org-roam.el
+++ b/org-roam.el
@@ -1395,20 +1395,22 @@ If DESCRIPTION is provided, use this as the link label.  See
           (org-roam-capture--capture))))))
 
 ;;;###autoload
-(defun org-roam-jump-to-index ()
+(defun org-roam-jump-to-index (&optional other-window)
   "Find the index file in `org-roam-directory'.
 The path to the index can be defined in `org-roam-index-file'.
 Otherwise, the function will look in your `org-roam-directory'
 for a note whose title is 'Index'.  If it does not exist, the
-command will offer you to create one."
+command will offer you to create one.
+If OTHER-WINDOW is non-nil, open the buffer in another window similarly to
+`find-file-other-window`"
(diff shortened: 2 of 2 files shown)


philip-bl:
Alright, I can wait until #783 is merged. About having a variable `org-roam-open-in-other-window`, why do you think it's a good idea? With additional argument you just pass it, it's simple. But with a global variable, a function like `org-roam-find-file-other-window` will need to save its current state, then set it to `t`, then invoke `org-roam-find-file-impl` (or whatever it'll be called), then restore its previous state. This is difficult.

zaeph:
> But with a global variable, a function like `org-roam-find-file-other-window` will need to save its current state, then set it to `t`, then invoke `org-roam-find-file-impl` (or whatever it'll be called), then restore its previous state. And also make sure you restore its previous state if the execution is interrupted, or if there's some kind of error. This is difficult.

The variable would be defined globally so that user can use it to define their default behaviour, but alternative functions would be able to set it locally within a `let`-form to force a particular behaviour.

philip-bl:
@zaeph I don't think people generally want a default behaviour on this. The idea is that sometimes I want to open a buffer in the current window, and other times I want to open it in another window. So, it doesn't make sense to have a default behaviour.

philip-bl:
Another way to implement this would be to have `org-roam-find-file-implementation` taking an `other-window` argument, and then we have `org-roam-find-file` which passes `nil`, and `org-roam-find-file-other-window` which passes `t`. And the user can bind `org-roam-find-file` to one key and `org-roam-find-file-other-window` to another.

zaeph:
I'm not sure if you're understanding our suggestion.  The default behaviour would be the one that we already cover, i.e., finding in the current window.  Having it `funcall` a variable would provide the biggest amount of freedom, and would come at no cost to what's already in place.  I don't like changing the signature of a function when I can avoid it.

philip-bl:
Adding another optional variable doesn't break anyone's existing code, does it?

Anyway, I think how to implement this is not very important, and since you're maintaining the package, I am fine with whatever way you choose.

zaeph:
> Adding another optional variable doesn't break anyone's existing code, does it?

It doesn't, no.  Changing a function's signature does, however.  If I were to implement your recommendation, for instance, I'd need to go rewrite my wrappers in ORB.  Now imagine doing that every time someone wants to add an alternative behaviour.  You end up with functions with 5+ arguments, and when you only need the `n`th, you need to `nil` all the previous ones.

Also, giving the possibility to change the default behaviour is just to open a path of customisation for other people, even though it might not make sense to you right now.  I think that's a healthy way to handle modularity and alternative behaviours.

> Anyway, I think how to implement this is not very important, and since you're maintaining the package, I am fine with whatever way you choose.

#783 will be merged soon, but I'll work on other UX stuff before working on #790.

hubisan:
Would it not be better to use the prefix arg? Many functions are using this to let both the user or a calling function optionally change the behavior.

A good example for this is the function `org-roam-graph`:
https://github.com/org-roam/org-roam/blob/79c75ac17468b5e2e3df03ce0936ddd8d0e5e947/org-roam-graph.el#L277-L301

jethrokuan:
yup, this PR has the right intentions, but the implementation should be easier with #807

philip-bl:
@hubisan 

>Would it not be better to use the prefix arg? Many functions are using this to let both the user or a calling function optionally change the behavior.

1. I am used to the fact that, for each function which switches to a buffer or opens a new buffer bound to the key combination Y, in order to do the same in another window I can do `C-x 4` Y. For me, spending time to remember that org-roam has `C-u` Y bindings won't be worth it.
2. A command can only have one functionality switch bound to the prefix argument. If we use it now for this, we won't be able to use it for something else later if we come up with some new functionality switch.

hubisan:
@philip-bl 

The function to find in other-window would not really change compared to the current implementation an can be bound to `C-x 4 Y` as well. Let's assume C-u = '(4) stands for the other-window:

``` emacs-lisp
(defun org-roam-find-file-other-window (&optional initial-prompt completions filter-fn)
  (interactive)
  (org-roam-find-file-other-window '(4) initial-prompt completions filter-fn))
```

There is not only C-u and C-u C-u. You can also use C-u 1, C-u 12, etc. So there is a lot of switches. And you can always write a wrapping function like `org-roam-find-file-other-window` from above to bind it directly. 

It can of course also be done by using an explicit argument. The current implementation might lead to a lot of additional arguments tough. Just imagine I would love to have an `other-frame` version or `split-vertically`. Probably better to use a symbol or list and have a switch decide what to do (which more or less is what the arg does).

progfolio:
```emacs-lisp
(defun org-roam-find-file-other-window (&rest args)
  (interactive)
  (let ((org-roam-find-file-function #'find-file-other-window))
    (apply #'org-roam-find-file args)))

(define-key org-roam-mode-map (kbd "C-x 4 C-c n f") #'org-roam-find-file-other-window)
```

PR #791, opened 2020-06-11
(fix): Fix syntax for calling compiled versions of functions

zaeph:
I'm not one to `git blame`, which is why I won't say that eae97487dcbea2b09249455a390fa501154aec79 introduced it. 🤡

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -858,9 +858,9 @@ This face is used for links without a destination."
              map)
   (if org-roam-backlinks-mode
       (add-hook 'org-open-at-point-functions
-                'org-roam-open-at-point nil 'local)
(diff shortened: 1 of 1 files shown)


progfolio:
> I won't say that eae9748 introduced it. clown_face

I wouldn't say that because it didn't introduce it.
Read the diff. That commit simply moved pre-existing code. Nothing was altered other than position.
That function hasn't been sharp-quoted the whole time (introduced in de36d15).

The commit message is a bit inaccurate, too. Sharp quoting isn't syntax to call compiled versions of functions. It's

zaeph:
That's why I'm *glad* I haven't actually `git blame`'d you instead of digging further!

PR #796, opened 2020-06-12
Add documentation about multiple templates for capture

TimQuelch:
Add some documentation on how the template system can be used to create notes in sub-directories for tagging purposes. As well as how multiple different templates can be used.

This is important now that subdirectories for tagging is implemented.

I haven't done too much texinfo stuff so let me know if formatting/syntax is incorrect

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -352,7 +352,9 @@ org-roam's templating system works, it may be useful to look into basic usage of
 =org-capture=.
 
 Org-roam's templates can be customized by modifying the variable
-=org-roam-capture-templates=.
+=org-roam-capture-templates=. Just like the base =org-capture= this
+variable can contain multiple templates, in which case you will be prompted on which
+one to use when capturing a new note.
 
 ** Template Walkthrough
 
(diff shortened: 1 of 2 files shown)


TimQuelch:
Ah that makes a lot of sense. Should I commit the automatic changes to the .texi file? The export seems to have made changes other than the ones that I made in the org file. Quotation marks changed from ``_'' to "_" and some sections around the org-protocl moved around slightly

jethrokuan:
It's fine to change everything. The idea is to have a match between the org docs, and the texi docs, for that particular commit (which MELPA uses to generate info docs).

doc/org-roam.org
```diff
@@ -376,9 +378,12 @@ the default template, reproduced below.
...
    This template means don't insert any content, but place the cursor
    here.
-6. =:file-name= is the file-name template for a new note, if it doesn't
-   yet exist. This creates a file at path that looks like
-   =/path/to/org-roam-directory/20200213032037-foo.org=.
+6. =:file-name= is the file-name template for a new note, if it doesn't yet
+   exist. This creates a file at path that looks like
+   =/path/to/org-roam-directory/20200213032037-foo.org=. This template also
+   allows you to specify if you want the note to go into a
+   subdirectory. e.g. the template =private/${slug}= will create notes in
```
zaeph, line 385:
```suggestion
   subdirectory.  For example, the template =private/${slug}= will create notes in
```

doc/org-roam.org
```diff
@@ -376,9 +378,12 @@ the default template, reproduced below.
...
+   =/path/to/org-roam-directory/private=
```
zaeph, line 386:
```suggestion
   =/path/to/org-roam-directory/private=.
```

doc/org-roam.org
TimQuelch: https://github.com/org-roam/org-roam/pull/796/commits/a7b5e24425683b2dfd154b1842f67625ffa481c9

doc/org-roam.org
TimQuelch: https://github.com/org-roam/org-roam/pull/796/commits/a7b5e24425683b2dfd154b1842f67625ffa481c9

zaeph:
Thanks!

PR #805, opened 2020-06-12
(feat): Switch to a minor-mode for the dev-suite

zaeph:


org-roam-dev.el
```diff
@@ -36,4 +36,11 @@
 (emacsql-fix-vector-indentation)
 (provide 'org-roam-dev)
 
+(define-minor-mode org-roam-dev-mode
+  "Minor mode for setting the dev environment of Org-roam."
+  :lighter " ORD"
+  (when org-roam-dev-mode
+    (require 'org-roam-dev)
+    (setq-local sentence-end-double-space nil)))
+
```
progfolio, line 45:
```suggestion

;;;###autoload
(define-minor-mode org-roam-dev-mode
  "Minor mode for setting the dev environment of Org-roam."
  :lighter " ORD"
  (when org-roam-dev-mode
    (emacsql-fix-vector-indentation)
    (setq-local sentence-end-double-space nil)))

(provide 'org-roam-dev)
```

zaeph: Wouldn't it be better to keep `(emacsql-fix-vector-indentation)` outside of the minor-mode?  We only need to run it once to advise `calculate-lisp-indent`, so I think we could keep it outside and have it loaded on the `require`.

progfolio: emacsql looks like it's using the old `advice` vs the newer `nadvice`.
`emacsql-fix-vector-indentation` adds and activates the advice.
Reactivating the advice reloads the functions if they have changed:

https://gnu.huihoo.org/elisp-manual-21-2.8/html_node/elisp_217.html

> To activate advice for a function whose advice is already active is not a no-op. It is a useful operation which puts into effect any changes in that function's advice since the previous activation of advice for that function. 

Not that I plan on changing that function for any reason, but the cost of activating the device when toggling the mode is negligible and prevents us from having to think about cases where the code need only be run once or multiple times. Simplicity at the cost of negligible efficiency.

zaeph: Sounds good. 👍

zaeph:
Good on the merge?

progfolio:
Tested. Everything looks like it's working fine for me.

PR #807, opened 2020-06-12
(feat): org-roam-find-file-function customization

progfolio:
Allow users to specify a function for finding files in various commands.
Can be let-bound to define new commands that suit user preferences.

See: #790

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -49,6 +49,7 @@
 (declare-function org-roam--get-backlinks     "org-roam")
 (declare-function org-roam-backlinks-mode     "org-roam")
 (declare-function org-roam-mode               "org-roam")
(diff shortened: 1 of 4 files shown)


progfolio:
Note this does not add any extra `...other-window` commands as #790 does and does not address the case of `org-roam-switch-to-buffer`.

zaeph:
> and does not address the case of `org-roam-switch-to-buffer`.

I'll think about this tomorrow.

jethrokuan:
Feel free to rebase and merge here.

zaeph:
I will.  Just finishing up with the notifications and I'll get on it.

zaeph:
Yeah, sorry, I haven't had time to get back to this one.  I'll review this later.

PR #812, opened 2020-06-13
(fix): fix unlinked-references for older Emacs

jethrokuan:

Emacs' in-built rx.el was rewritten in Emacs 27, and the form `anychar`
only exists in later versions. We use the older form `anything` that has
support even in Emacs 26.

Emacs 26: https://github.com/emacs-mirror/emacs/blob/emacs-26/lisp/emacs-lisp/rx.el#L889

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1451,7 +1451,7 @@ linked, lest the network graph get too crowded."
            (file-loc (buffer-file-name))
            (buf (get-buffer-create "*org-roam unlinked references*"))
            (results (split-string (shell-command-to-string rg-command) "\n"))
-           (result-regex (rx (group (one-or-more anychar))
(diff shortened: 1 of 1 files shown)


zaeph:
Have you tested it?

jethrokuan:
not on older Emacs.

zaeph:
I'll build one real quick and test it out, then.

bepolymathe:
The error is found for me with a version 26.3

zaeph:
> The error is found for me with a version 26.3

Could you try the PR and tell us if it works for you?  I'm having trouble building older versions of Emacs on my end.

bepolymathe:
Yes of course. It works for me with the PR.  👍

zaeph:
Splendid, I'm merging, then.

PR #814, opened 2020-06-13
(feat): Implement ‘org-roam-insert-immediate’

### LINKED ISSUE #813: Allow the creation of file references without opening org capture
### Description
- Modify the `org-roam-insert` function, to allow the user to skip the org capture screen, and continue working on the current note
- This saves one step when you're creating a link.
- I feel this would be useful for people who are starting to use org-roam and lack many of the references that they may want to add, and as a result may spend a lot of time creating new files.

### Proposed Implementation
- Modify `org-roam-insert` to accept a optional variable, which determines whether or not we open org capture (defaults to opening org capture)
- Set `:immediate-finish`, when calling org capture [here](https://github.com/org-roam/org-roam/blob/master/org-roam-capture.el#L328), if required.

I haven't written elisp before, so take whatever I have written here with a grain of salt. That said, if this would be a useful feature, I'd like to take a stab at implementing this, as it's a good forcing function to learn it's basics.

zaeph:
Fixes #813.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 1.2.1 (TBD)
+
+### Breaking Changes
+
+### Features
+- [#814] Implement `org-roam-insert-immediate`
+
+### Bugfixes
+
 ## 1.2.0 (12-06-2020)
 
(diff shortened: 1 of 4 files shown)


zaeph:
Aye for discovery, not sure about `I`.  I use `I` in my [ORB](https://github.com/org-roam/org-roam-bibtex) setup to have an alternative completion-source.  I guess it's fine to mention it in the default config, though.  It's not like people couldn't change it easily.

zaeph:
Also, I think I've conflated two problems with this implementation:
1. Not asking which template to use.
2. Immediately validating the `org-capture` so as to not interrupt the workflow.

The problem with the current implementation is that it assumes that you know that the notes doesn't exist in your library.  Functionally, 2. could be addressed as an action with `ivy` and `helm`, and the default behaviour could be defined by a variable.

I need to think some more about this.

zaeph:
I'd like to think some more about this today, so I'll hold off on the merge for now.

zaeph:
…I guess I'll PR again, then. :p

PR #818, opened 2020-06-13
(ux): update the error message for org-roam-unlinked-references

jethrokuan:

`rg` might be unfamiliar to some, so let's mention the name of the full
executable name.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1438,7 +1438,7 @@ linked, lest the network graph get too crowded."
   (unless (org-roam--org-roam-file-p)
     (user-error "Not in org-roam file"))
   (if (not (executable-find "rg"))
-      (user-error "Cannot find the rigrep executable \"rg\". Check that it is installed and available on `exec-path'")
+      (error "Cannot find the rigrep executable \"rg\". Check that it is installed and available on `exec-path'")
     (let* ((titles (org-roam--extract-titles))
            (rg-command (concat "rg -o --vimgrep -P -i "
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -1438,7 +1438,7 @@ linked, lest the network graph get too crowded."
   (unless (org-roam--org-roam-file-p)
     (user-error "Not in org-roam file"))
   (if (not (executable-find "rg"))
-      (user-error "Cannot find the \"rg\" executable, aborting")
+      (user-error "Cannot find the rigrep executable \"rg\". Check that it is installed and available on `exec-path'")
```
zaeph, line 1441:
```suggestion
      (error "Cannot find the rigrep executable \"rg\". Check that it is installed and available on `exec-path'")
```

Not sure if it qualifies as a `user-error`.

org-roam.el
```diff
@@ -1438,7 +1438,7 @@ linked, lest the network graph get too crowded."
   (unless (org-roam--org-roam-file-p)
     (user-error "Not in org-roam file"))
   (if (not (executable-find "rg"))
-      (user-error "Cannot find the \"rg\" executable, aborting")
+      (error "Cannot find the rigrep executable \"rg\". Check that it is installed and available on `exec-path'")
```
sgepigon, line 1441: Small typo.
```suggestion
      (error "Cannot find the ripgrep executable \"rg\". Check that it is installed and available on `exec-path'")
```

PR #819, opened 2020-06-13
(feat): split org-roam-buffer-toggle into -activate and -deactivate

langston-barrett:
I want to create a hook for activating this buffer when I enter an org file and deactivating it when I leave.

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -266,14 +266,25 @@ Valid states are 'visible, 'exists and 'none."
          (org-roam-buffer--set-height
           (round (* (frame-height) org-roam-buffer-height))))))))
 
-(defun org-roam-buffer-toggle-display ()
-  "Toggle display of the `org-roam-buffer'."
+(defun org-roam-buffer-activate ()
+  "Activate display of the `org-roam-buffer'."
   (interactive)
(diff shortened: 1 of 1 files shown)


zaeph:
Doom did something like that.  You should look up how they did it on their end.

langston-barrett:
> Doom did something like that. You should look up how they did it on their end.

Here's that code: https://github.com/hlissner/doom-emacs/blob/da3c1cb78dd806cd45f8c8520ebaf8db5c031802/modules/lang/org/contrib/roam.el#L53

It's not that different from what's in this PR. In fact, the call to the internal function `(org-roam-buffer--get-create)` in that code could instead be a call to `(org-roam-buffer-activate)`, which this PR exposes as part of the "public API".

zaeph:
Sorry, hadn't realised that it was a PR.  I'll have a look tomorrow, but I'd like to make this behavior optional and off by default.  We'll need to think about this for the revamp.

langston-barrett:
@zaeph No worries! Also, just a note that this is fully backwards compatible, the behavior of `org-roam-buffer-toggle` stays exactly the same.

zaeph:
I don't see any problem with splitting the toggle into two functions.  Thanks!

Issue #803, opened 2020-06-12
Install not working


edoloughlin:
### Description

I installed org-roam in Emacs 26.3 on windows using the `use-package` instructions from the readme.

When I M-x org-roam I see an error:

    Cannot open load file: No such file or directory, org-roam
 
No additional buffer

#### Steps to Reproduce

Install in Emacs in windows, run M-x org-roam

#### Backtrace

No backtrace produced

### Environment

_Please M-x org-roam-diagnostics and paste results here_

org-roam-diagnostics doesn't exist.

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

zaeph:
Have you initialised the packages?  See [MELPA's 'Usage' section](https://github.com/melpa/melpa#usage).

progfolio:
Likely a duplicate of: #755, #697, #582 

Please search for similar issues when creating a new one.

edoloughlin:
Thanks for the suggestion, but these don't seem to help. I installed via use-package. I've used use-package for quite some time and install all my other packages with it, so I don't think initialisation is the problem.

progfolio:
Does adding an `:ensure t` to your use-package declaration help?

edoloughlin:
Yes it does! Thanks. Don't know why I didn't spot it.

PR #820, opened 2020-06-13
(doc): Add :ensure to use-package declaration

progfolio:
Ensure use-package installs org-roam.

See: #803


zaeph:
Isn't that linked to a global option for `use-package`, though?

PR #828, opened 2020-06-15
Fixes caar into caadr for PRAGMA "user_version"

rby:
When evaluating this code
```emacs-lisp
(let ((conn (emacsql-sqlite3 "~/path/to/org-roam.db")))
  (emacsql conn "PRAGMA user_version"))
;;; => ((user_version) (3) (Run Time: real 0.0 user 4.8e-05 sys 0.0))
```
if we use `caar` it returns `"user_version"` instead of `3` which is what we want.
Switching to `caadr` fixes this.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -92,7 +92,7 @@ Performs a database upgrade when required."
                  org-roam-db--connection)
         (when init-db
           (org-roam-db--init conn))
-        (let* ((version (caar (emacsql conn "PRAGMA user_version")))
+        (let* ((version (caadr (emacsql conn "PRAGMA user_version")))
                (version (org-roam-db--update-maybe conn version)))
           (cond
            ((> version org-roam-db--version)


jethrokuan:
3 is really really old, I suspect you're on a really old version of org-roam?

For me, I see:

```
(caar (emacsql (org-roam-db--get-connection) "PRAGMA user_version")) => 6
```

so `caar` seems to be correct.

rby:
Thanks for your reply,
Indeed, the version is old, and I got stuck with an error during the
upgrade:
`custom-initialize-reset: Wrong type argument: number-or-marker-p,
user_version`

Could this be an issue with `emacsql` ?

jethrokuan:
Do you have a `.sqliterc` in your computer, that sets a setting for headers?

rby:
That was it. 👌
My org-roam is working again!
Thanks! off-road is life changing.

Ramzi

Issue #781, opened 2020-06-09
Slug: Title Length


klorophatu:
### Brief Abstract
Choose the length of the title.
### Long Description
Choose the length of the title. For example, if you had typed the title "Irresistible: The Rise of Addictive Technology and the Business of Keeping Us Hooked" into org-roam-find-file, the filename would be 20200609193343-irresistible_the_rise.org because you specifically chose it to have 3 words at max.
### Please check the following:

- [X] No similar feature requests

zaeph:
We could make the slug facility more modular by having a variable which users can use to use their own function.  I'll look into it.

bdarcus:
I don't have a strong opinion about this, but if shortening slugs, should throw out "stop words" ("the", "of", etc), at least optionally and/or by default. 

For above example, that could yield "20200609193343-irresistible_rise_addictive.org".

Which maybe makes @zaeph's point about allowing a user function?

zackteo:
@klorophatu the variable `org-roam-title-to-slug-fn` has been added - so you can do something like

```el
(defun string-filter (title)
    (let ((list (split-string title "\_")))
      (string-join (butlast list (- (length list) 3)) "_")))

(defun my-cool-title-function (str)
  (string-filter (org-roam--title-to-slug str)))
```
```el
:config ; And have this for your org-roam use-package config
(setq org-roam-title-to-slug-function  #'my-cool-title-function)
```
`org-roam--title-to-slug` was the old function btw - which I just decided to add on, instead of recreating the whole thing :x  

If you want common words removed as suggested by @bdarcus maybe something like this?
```elisp
(defun string-filter (title)
  (let ((common-words '("the" "of" "a" "as"))
        (title-length 3)
        (title-word-list (split-string title "\_")))
    (let ((new-list (seq-remove (lambda (elt) (member elt common-words)) title-word-list)))
      (string-join (butlast new-list (- (length new-list) title-length)) "_"))))
```
Forgive me if the code is really bad >< have just started using elisp :x But yes hope it gives you an idea of what you might want

lanceberge:
After spending some time digging, this solution is now deprecated. `org-roam-title-to-slug-function` is now customizable with:
  ```
(cl-defmethod org-roam-node-slug ((node org-roam-node))
```

PR #833, opened 2020-06-16
(feat): add customizable org-roam-title-to-slug-fn

zackteo:
Choose the length of the title. #781


org-roam-capture.el
```diff
@@ -349,7 +349,7 @@ This uses the templates defined at `org-roam-capture-templates'."
          (title (or (plist-get res :title) title-with-keys))
          (file-path (plist-get res :path)))
     (let ((org-roam-capture--info (list (cons 'title title)
-                                        (cons 'slug (org-roam--title-to-slug title))
+                                        (cons 'slug (org-roam-title-to-slug-fn title))
```
jethrokuan, line 352:
```suggestion
                                        (cons 'slug (funcall org-roam-title-to-slug-fn title))
```

zaeph:
```suggestion
                                        (cons 'slug (funcall #'org-roam-title-to-slug-fn title))
```

org-roam-capture.el
```diff
@@ -44,7 +44,7 @@
 (declare-function  org-roam--file-path-from-id          "org-roam")
 (declare-function  org-roam--find-file                  "org-roam")
 (declare-function  org-roam--format-link                "org-roam")
-(declare-function  org-roam--title-to-slug              "org-roam")
+(declare-function  org-roam-title-to-slug-fn            "org-roam")
```
zackteo, line 47: Is this causing the linting error?

org-roam.el
```diff
@@ -1296,7 +1304,7 @@ which takes as its argument an alist of path-completions.  See
     (if file-path
         (org-roam--find-file file-path)
       (let ((org-roam-capture--info `((title . ,title-with-tags)
-                                      (slug  . ,(org-roam--title-to-slug title-with-tags))))
+                                      (slug  . ,(funcall org-roam-title-to-slug-fn title-with-tags))))
```
jethrokuan, line 1307:
```suggestion
                                      (slug  . ,(funcall #'org-roam-title-to-slug-fn title-with-tags))))
```

org-roam.el
```diff
@@ -1371,7 +1379,7 @@ If DESCRIPTION is provided, use this as the link label.  See
       (when (org-roam-capture--in-process-p)
         (user-error "Nested Org-roam capture processes not supported"))
       (let ((org-roam-capture--info `((title . ,title-with-tags)
-                                      (slug . ,(org-roam--title-to-slug title-with-tags))))
+                                      (slug . ,(funcall org-roam-title-to-slug-fn title-with-tags))))
```
jethrokuan, line 1382:
```suggestion
                                      (slug . ,(funcall #'org-roam-title-to-slug-fn title-with-tags))))
```

zackteo:
@jethrokuan I have made the changes you asked for in the recent commit. 
Except for `org-roam-title-to-slug-function` to `#'org-roam-title-to-slug-function` where doing so will only result in "Symbol's function definition is void: org-roam-title-to-slug-function". Unless I missed something out, this should be correct

org-roam.el
```diff
@@ -208,6 +207,14 @@ extraction methods:
               (const :tag "sub-directories" all-directories)
               (const :tag "parent directory" last-directory)))
 
+(defcustom org-roam-title-to-slug-function #'org-roam--title-to-slug
+  "Function to be used in converting a title to the filename slug.
+Function should return a filename string based on title.
+Otherwise, function will remain as default of
+`org-roam--title-to-slug'."
```
jethrokuan, line 214:
```suggestion
Function should return a filename string based on title."
```

jethrokuan:
LGTM otherwise!

xeruf:
Seems this feature was lost in v2?

kijowski:
I've probably followed similar path to you @xeruf so maybe my findings would benefit you or someone else in the future :) 
`org-roam-node-slug` seems unused, because it's not being called directly - rather it's being called as a part of template filling mechanism described [here](https://www.orgroam.com/manual.html#Org_002droam-Template-Expansion).

So to customize slug generation you have two options:
1. Overwrite `org-roam-node-slug` method with your own definition
2. Write your own `org-roam-node-X` method and use `${X}` in template definition. In my config I have following function which does what I want:
```
(cl-defmethod org-roam-node-dash-slug ((node org-roam-node))
  "Return the slug of NODE with dashes"
  (replace-regexp-in-string "_" "-" (org-roam-node-slug node)))
```

xeruf:
ah, I solved it with advice, with the same goal as you ;)

```sh
    (defun xf/dashify-slug (slug)
      (s-replace "_" "-" slug))
    (advice-add 'org-roam-node-slug :filter-return #'xf/dashify-slug)
```

PR #838, opened 2020-06-17
(fix): Check sqlite3’s executability before existence

### LINKED ISSUE #837: Silence the "can't find sqlite3" warning
### Brief Abstract

The "can't find sqlite3" warning fires, even if `emacsql-sqlite3-executable` is set.

### Long Description

I use nix, and build my emacs using a nix expression that (among other things) sets `emacsql-sqlite3-executable` to a specific sqlite3 binary.

This warning still fires on startup:
https://github.com/org-roam/org-roam/blob/8b16e5d5205849b93b0af333bd431857f319aaad/org-roam.el#L1232-L1235

I am currently silencing it using the following advice:

```elisp
(advice-add 'lwarn :before-until
  (lambda (&rest args)
    (and
     (equal '(org-roam) (car args))
     (s-contains-p "Cannot find executable 'sqlite3'." (caddr args)))))
```

Either:
* If `emacsql-sqlite3-executable` names an executable file, do nothing; else
* If `(find-executable "sqlite3")` finds something, do nothing; else
* Emit the warning.

Or:
* Provide a variable to suppress this warning (default: `nil`).

zaeph:
Fixes #837.

@endgame: Could you test the PR on your machine and tell us if it works?

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1229,7 +1229,8 @@ Otherwise, behave as if called interactively."
   :global t
   (cond
    (org-roam-mode
-    (unless (executable-find "sqlite3")
+    (unless (or (file-executable-p emacsql-sqlite3-executable)
+                (executable-find "sqlite3"))
       (lwarn '(org-roam) :error "Cannot find executable 'sqlite3'. \
 Ensure it is installed and can be found within `exec-path'. \
 M-x info for more information at Org-roam > Installation > Post-Installation Tasks."))


org-roam.el
```diff
@@ -1229,7 +1229,8 @@ Otherwise, behave as if called interactively."
   :global t
   (cond
    (org-roam-mode
-    (unless (executable-find "sqlite3")
+    (unless (or (file-executable-p emacsql-sqlite3-executable)
```
jethrokuan, line 1232: ah this variable was only recently added. Sure.

zaeph:
Waiting on someone using Nix to test it to merge.

jethrokuan:
This is less of a nix thing, than a recent addition to the emacsql-sqlite3 package: https://github.com/cireu/emacsql-sqlite3/commit/1e411fd38a0137553986db209642fe93cae96060

We should also support checking that variable either way, and fall back to executable-find in case it doesn't yet exist (package not updated).

PR #839, opened 2020-06-17
return selected file from org-roam-insert

d12frosted:

This is useful in scenarios when you want to automatically do something with the
context where the link was inserted. For example, if you are linking someone in
the task, you can add a special tag for this person.

```emacs-lisp
(defun +org-notes-insert ()
  "Insert a link to the note."
  (interactive)
  (when-let* ((res (org-roam-insert))
              (path (plist-get res :path))
              (title (plist-get res :title))
              (roam-tags (+seq-flatten
                          (+seq-flatten
                           (org-roam-db-query [:select tags
                                               :from tags
                                               :where (= file $s1)]
                                              path)))))
    (when (seq-contains-p roam-tags "People")
      (save-excursion
        (ignore-errors
          (org-back-to-heading)
          (let ((tags (org-get-tags)))
            (org-set-tags (seq-uniq (cons (+org-notes--title-to-tag title) tags)))))))))

(defun +org-notes--title-to-tag (title)
  "Convert TITLE to tag."
  (concat "@" (s-replace " " "" title)))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Features
 - [#814] Implement `org-roam-insert-immediate`
 - [#833](https://github.com/org-roam/org-roam/pull/833) Add customization of file titles with `org-roam-title-to-slug-function`. 
+- [#839](https://github.com/org-roam/org-roam/pull/839) Return selected file from `org-roam-insert`
+
 ### Bugfixes
 
 ## 1.2.0 (12-06-2020)
--- a/org-roam.el
+++ b/org-roam.el
@@ -1339,6 +1339,7 @@ included as a candidate."
 ;;;###autoload
 (defun org-roam-insert (&optional lowercase completions filter-fn description)
   "Find an Org-roam file, and insert a relative org link to it at point.
+Return selected file if it exists.
 If LOWERCASE, downcase the title before insertion.
 COMPLETIONS is a list of completions to be used instead of
 `org-roam--get-title-path-completions`.
@@ -1384,7 +1385,8 @@ If DESCRIPTION is provided, use this as the link label.  See
                                                                :link-description link-description
                                                                :capture-fn 'org-roam-insert))
         (org-roam--with-template-error 'org-roam-capture-templates
-          (org-roam-capture--capture))))))
+          (org-roam-capture--capture))))
+    res))
 
 ;;;###autoload
 (defun org-roam-insert-immediate (arg &rest args)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Features
 - [#814] Implement `org-roam-insert-immediate`
+- [#839] Return selected file from `org-roam-insert`
```
d12frosted, line 9: Should I use a link to this pull request?

org-roam.el
```diff
@@ -1378,7 +1379,8 @@ If DESCRIPTION is provided, use this as the link label.  See
                                                                :link-description link-description
                                                                :capture-fn 'org-roam-insert))
         (org-roam--with-template-error 'org-roam-capture-templates
-          (org-roam-capture--capture))))))
+          (org-roam-capture--capture))))
+    res))
```
d12frosted, line 1383: if the user didn't select existing note, the value will be `nil`. We could return `title-with-tags`, but the problem is that `org-roam-insert` finishes before capture process is done. So it's totally fine to have `nil`.

d12frosted:
Usage example:

auto-tag

zaeph:
Looks interesting.  Where do you insert your `when` block, though?  Did you patch it in directly in `org-roam-insert`?  We could implement a hook for you to do this more easily.

d12frosted:
No, I just have a wrapper for `org-roam-insert` - `+org-notes-insert` that I use instead of the former.

```emacs-lisp
(defun +org-notes-insert ()
  "Insert a link to the note."
  (interactive)
  (when-let* ((res (org-roam-insert))
              (path (plist-get res :path))
              (title (plist-get res :title))
              (roam-tags (+seq-flatten
                          (+seq-flatten
                           (org-roam-db-query [:select tags
                                               :from tags
                                               :where (= file $s1)]
                                              path)))))
    (when (seq-contains-p roam-tags "People")
      (save-excursion
        (org-back-to-heading)
        (let ((tags (org-get-tags)))
          (org-set-tags (seq-uniq (cons (+org-notes--title-to-tag title) tags))))))))
```

d12frosted:
I think the hook could work as well. Especially with the newly captured files, which would be handy I think.

zaeph:
Definitely a candidate for a hook.  I'll review your PR later today, and I'll probably create another one adding a hook

d12frosted:
@zaeph thank you :) feel free to close this PR in favour of something more generic. I went the simplest path ;)

zaeph:
Your code looks all right: I'd probably just rename `res` to something a little more meaningful.  If I find any problem with the code, I'll just push to your branch.  Thanks!

CHANGELOG.md
jethrokuan: Yup, that would be good!

d12frosted:
I've added a link to this PR to changelog. And also fixed the conflict with master ;) Thanks for the review!

d12frosted:
Thanks for merging! ❤️

Issue #834, opened 2020-06-16
Performance issue with org-roam


hungpham2511:
### Description

I have enjoyed using org-roam thus far and been using that quite extensively. Though I can't help noticing that the mode becomes sluggish very quickly when there are a few hundred notes. This is getting a bit annoying and I end up turning off org-roam more and more often.

It seems to me that org-roam is trying to rebuild the cache all the time at saving hook, and enforce auto-save after a few seconds. 
I wonder if there is any way to strip the hooks to a bare minimum or any algorithmic improvement to improve performance?

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/7ab67436a7517ee6be35c32b96c3c6d697181603

hungpham2511:
I might not be fully aware of the internal implementation. However, I find it quite strange that this update step should be expensive at all? Assuming one 1 file changes at save, we only need to update the db of the entries that are referenced to in this note, if there are any. In the current implementation, I notice that the delay occurs even when there are very few backlinks.

jethrokuan:
I have about 500 notes and I'm not seeing any noticeable lag at all. The db updates are incremental, and they are implemented how you describe it.

Can you use `M-x profiler-start` and go along your business (editing notes, saving them) and `M-x profiler-stop`, and then `M-x profiler-report`?

That would help me see where the delays are.

isolet:
I have a similar performance issue that `org-roam` calls lots of `file-truename` and causes slowness, though the original callee is `org-roam-bibtex`. Below is the profile of a single call of `save-buffer` of my bib file. The file contains only 1000 items and I have only 200 org file in my `org-roam-directory`. It took 5 to 10 seconds to save my bib file.


```
- command-execute                                                1846  99%
 - call-interactively                                            1846  99%
  - funcall-interactively                                        1846  99%
   - file-notify-handle-event                                    1741  93%
    - file-notify--callback-kqueue                               1741  93%
     - file-notify--handle-event                                 1741  93%
      - file-notify--call-handler                                1741  93%
       - #<compiled 0x1ff20ab7bbe1>                              1741  93%
        - bibtex-completion-candidates                           1741  93%
         - bibtex-completion-parse-bibliography                  1699  91%
          - bibtex-completion-prepare-entry                      1658  88%
           - mapcar                                              1372  73%
            - #<compiled 0x1ff20ab8272d>                         1372  73%
             - orb-find-note-file                                1243  66%
              - let*                                             1243  66%
[...]

PR #841, opened 2020-06-17
(tests): add performance tests to CI

### LINKED ISSUE #742: Run performance tests in CI
### Brief Abstract

Perform performance testing on blocking functions.

### Long Description

In Discourse (https://org-roam.discourse.group/t/what-does-it-feel-like-to-work-with-10-000-notes-in-org-roam-benchmarking-org-roams-search-methods/227/4), @nobiot performed some performance testing on knowledge bases of different scales.

I'd like to have these tests included somehow in the CI, and throw warnings when they become too slow.

Have an additional testing file that fetchs a large org-roam knowledge base and run tests based on it.

jethrokuan:

We really want to start tracking down and optimizing cache builds, because it's causing issues for some.

Fixes #742, ref #834

### DIFF
--- a/tests/test-org-roam-perf.el
+++ b/tests/test-org-roam-perf.el
@@ -1,0 +1,53 @@
+;;; test-org-roam-perf.el --- Performance Tests for Org-roam -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 Jethro Kuan
+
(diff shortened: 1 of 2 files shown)


jethrokuan:
We can sort of see how long it takes for 1000 files (with links, no headlines): (about 70s on CI, 65s on my machine)

https://github.com/org-roam/org-roam/runs/780020614#step:6:20

PR #844, opened 2020-06-17
s/org-roam-show-graph/org-roam-graph-show/ README

stig:

I used the exsisting example from the README and was told:

> ‘org-roam-show-graph’ is an obsolete command (as of org-roam 1.0.0); use ‘org-roam-graph-show’ instead.


jethrokuan:
Thanks!

PR #851, opened 2020-06-18
(feat): add 'first-directory option for org-roam-tag-sources

rasendubi:

I use a directory structure where each subdirectory of `org-roam-directory` is a separate slip-box (e.g., literature notes, work notes, etc).

I believe that's a common enough setup to add built-in support.


jethrokuan:
Nice, thanks! Could you add an entry to the changelog?

PR #852, opened 2020-06-22
(feat): add org-roam-find-file-immediate

jethrokuan:
##### Motivation for Change

Adds `org-roam-find-file-immediate`. Ref #852.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1408,12 +1408,24 @@ If DESCRIPTION is provided, use this as the link label.  See
   "Find an Org-roam file, and insert a relative org link to it at point.
 This variant of `org-roam-insert' inserts the link immediately by
 using the template in `org-roam-capture-immediate-template'. The
-interactive ARG and ARGS are forward to `org-roam-insert'.
+interactive ARG and ARGS are passed to `org-roam-insert'.
 See `org-roam-insert' for details."
   (interactive "P")
   (let ((args (push arg args))
         (org-roam-capture-templates (list org-roam-capture-immediate-template)))
     (apply #'org-roam-insert args)))
+
+;;;###autoload
+(defun org-roam-find-file-immediate (arg &rest args)
+  "Find and open an Org-roam file.
+This variant of `org-roam-find-file' uses the template in
+`org-roam-capture-immediate-template', avoiding the capture
+process. The interactive ARG and ARGS are passed to
+`org-roam-find-file'. See `org-roam-find-file' for details."
+  (interactive "P")
+  (let ((args (push arg args))
+        (org-roam-capture-templates (list org-roam-capture-immediate-template)))
+    (apply #'org-roam-find-file args)))
 
 ;;;###autoload
 (defun org-roam-jump-to-index ()


jethrokuan:
Was just trying out `forge`'s create PR from issue and it did weird stuff.

I'm not sure I like the proliferation of these "immediate" commands, if one only uses the immediate versions, the correct approach would be to change the template, adding `:immediate-finish t` to the default `org-roam-capture-template`.

AlexanderDavid:
@jethrokuan The problem I find with `:immediate finish` is that it doesn't pop up in a buffer, it just adds it as a file.

jethrokuan:
The addition of immediate finish (and having only 1 template) is the only thing that differentiates the normal and the immediate version of the commands, so once that is added they are functionally equivalent

jmburgos:
But I do have a single template that includes `:immediate finish`, and if I use `org-roam-find-file` or `org-roam-insert` and type the name of a new note, I get the "Capture buffer" temporary buffer and I need to do C-c C-c to close it and keep writting.  

I do have a different template in orb-template, though. Would that make the difference?

TimQuelch:
@jmburgos do you have it as `:immediate-finish t` in your template?

jmburgos:
@TimQuelch Yes, I do. This is my template:

```
(org-roam-capture-templates '(
			      ("d" "default" plain #'org-roam--capture-get-point
			       :file-name "notes/${slug}"
			       :head "#+TITLE: ${title} \n#+CREATED: %U\n#+EDITED: %U\n#+ROAM_ALIAS: \n#+ROAM_TAGS: \n"
			       "%?"
			       :unnarrowed t
			       :immediate-finish t)
			      ))
```
As as said, I also have a different template in orb-templates, but that should not be an issue:

```
  (orb-templates
      '(("r" "ref" plain (function org-roam-capture--get-point) ""
         :file-name "refs/${citekey}"
         :head "#+TITLE: ${title}\n#+ROAM_KEY: ${ref}\n#+CREATED: %U\n#+EDITED: %U\n#+ROAM_TAGS: \n* Notes on cite:${citekey}\n :PROPERTIES:\n :Custom_ID:  ${citekey}\n :PDF: /home/julian/Documents/Refs/${citekey}.pdf\n :END:\n%? "
        :unnarrowed t
        :immediate-finish t)))
```

jethrokuan:
@jmburgos you have an error in your template. the argument `"%?"` is positional, while the rest are key-value pairs. The positional argument must come first.

```
(setq org-roam-capture-templates '(
			      ("d" "default" plain #'org-roam--capture-get-point
                    "%?"
			       :file-name "notes/${slug}"
			       :head "#+TITLE: ${title} \n#+CREATED: %U\n#+EDITED: %U\n#+ROAM_ALIAS: \n#+ROAM_TAGS: \n"
			       :unnarrowed t
                   :immediate-finish t
			       )
			      ))```

jmburgos:
Wow! I have not noticed that.  I guess Emacs teaches us that even the small things matter.  With the correct template things work as they should, and we do not really need the "immediate" functions.  Thanks!  I am getting closer to note taking nirvana....

PR #863, opened 2020-06-24
(feat): Add header level to backlinks buffer

jethrokuan:

This simulates roam-research a bit more, adding heading levels to the backlinks buffer.


YigitDemirag:
I think with this update @jethrokuan , backlink buffer started to show some `Top` hierarchy in some backlinks which makes the buffer looks more complicated than necessary.

jethrokuan:
@YigitDemirag what would you suggest as an alternative? How to differentiate nested preview content vs non-nested?

PR #885, opened 2020-07-04
(bugfix): prevent file-exists-p opening tramp links

herbertjones:

Links like `/ssh:me@host:/` cause emacs to lock up asking for the password repeatedly due to using `file-exists-p` in a function passed to font lock.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -998,8 +998,8 @@ Applies `org-roam-link-current' if PATH corresponds to the
 currently opened Org-roam file in the backlink buffer, or
 `org-roam-link-face' if PATH corresponds to any other Org-roam
 file."
-  (cond ((or (file-remote-p path) ;; Prevent lockups opening Tramp links
-             (not (file-exists-p path)))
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -998,7 +998,8 @@ Applies `org-roam-link-current' if PATH corresponds to the
 currently opened Org-roam file in the backlink buffer, or
 `org-roam-link-face' if PATH corresponds to any other Org-roam
 file."
-  (cond ((not (file-exists-p path))
+  (cond ((or (file-remote-p path) ;; Prevent lockups opening Tramp links
```
jethrokuan, line 1001:
```suggestion
  (cond ((and (not (file-remote-p path)) ;; Prevent lockups opening Tramp links
```

This would prevent font-locking all remote paths as invalid links, and default to the `org-link` face.

zoechi:
Thanks a lot <3 Was wasting so much time trying to find the cause.

Issue #903, opened 2020-07-08
org-roam-capture--get-point returning strange error


thor314:
May be putting this in the wrong location. Point me in the right one if that's the case.
### Description
Could be that I'm just doing something wrong, but I appended the default, 

```
("d" "default" plain (function org-roam--capture-get-point)
     "%?"
     :file-name "%<%Y%m%d>-${slug}"
     :head "#+title: ${title}\n"
     :unnarrowed t)
```
to my org-capture-templates (the rest of which are working as expected), and I get the error message `Invalid org-roam-capture-context`. 
I've also tried to change the default filename with:
`(setq org-roam-capture--file-name-default "<%Y-%m%-%d>")`,
but that doesn't appear to change the default timestring captures. Am I misunderstanding something simple?

### Environment

- Emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109))
 of 2019-09-02
- Framework: N/A
- Org: Org mode version 9.3.7 (9.3.7-8-g3ed035-elpa @ /Users/thor/.emacs.d/elpa/org-20200629/)
- Org-roam: 1.2.0

jethrokuan:
Hi, the capture templates for org-roam don't work with regular org-capture.

The variable you need to modify is `org-roam-capture-templates`, and which are used by the interactive functions `org-roam-insert`, `org-roam-find-file`, and `org-roam-capture`.

thor314:
Alright, thanks. I'm having a different problem now with setting `org-roam-capture templates`. I visited [your dot file](https://github.com/jethrokuan/dots/blob/master/.doom.d/config.el?#L432)
and made sure to get the syntax right. I copied the following into my config:

```
 (setq org-roam-capture--file-name-default "<%Y-%m%-%d>")
 (setq org-roam-capture-templates
        '(("l" "lit" plain (function org-roam--capture-get-point)
           "%?"
           :file-name "lit/${slug}"
           :head "#+setupfile:./hugo_setup.org
#+hugo_slug: ${slug}
#+title: ${title}\n"
           :unnarrowed t)
))

 (setq org-roam-capture-ref-templates
        '(("r" "ref" plain (function org-roam-capture--get-point)
           "%?"
           :file-name "lit/${slug}"
           :head "#+setupfile:./hugo_setup.org
#+roam_key: ${ref}
#+hugo_slug: ${slug}
#+roam_tags: website
[...]

PR #904, opened 2020-07-09
(docs): Update org-roam-capture documentation

jethrokuan:
Make the incompatibility with regular org-capture explicit.


jethrokuan:
Ref #903

PR #908, opened 2020-07-10
(internal): normalize titles in database

jethrokuan:
Instead of storing titles as a list within in the Org-roam cache, e.g.
file, ("title1" "title2"). We normalize it and store it as:

file, "title1"
file, "title2"

We need to be able to find a file given a title. This is in preparation for resolving fuzzy links such as `[[Title]]`.

In the old format, we need to do some elisp post-processing, but once we normalize, we can find all files that have that given title.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.2.1 (TBD)
 
 ### Breaking Changes
+- [#908](https://github.com/org-roam/org-roam/pull/908) Normalized titles in database. May break external packages that rely on unnormalized titles.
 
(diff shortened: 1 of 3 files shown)


zaeph:
This will also impact `org-roam--get-ref-path-completions`.  I'll update it when we're ready to merge.

zaeph:
I'll adjust `org-roam--get-ref-path-completions`, then.

zaeph:
Oh, nevermind, you already took care of it in 4a66a2d48096119ca234ff94f0ace11e8257bded.  Thanks!

PR #910, opened 2020-07-10
(feat): support fuzzy links

jethrokuan:

This is an alternative implementation of #592, which I currently favour. Instead
of adding a `roam:` type link, we add support for linking org-roam files via
fuzzy links:

output-2020-07-10-15:29:22

TODOs:

- [x] Support ID headline links (existing only)
- [ ] Replace fuzzy links with actual `file:` links on save/navigation
- [ ] Smart delete/backspace: remove `file:` portion on link editing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,17 @@
 
 ## 1.2.2 (TBD)
 
+In this release we support fuzzy links of the form `[[Title]]`, `[[*Headline]]` and `[[Title*Headline]]`. Completion for these fuzzy links is supported via `completion-at-point`.
+
 ### Breaking Changes
+
+- [#910](https://github.com/org-roam/org-roam/pull/910) Deprecate `company-org-roam`, using `completion-at-point` instead. To use this with company, add the `company-capf` backend instead.
 
 ### Features
 
 - [#974](https://github.com/org-roam/org-roam/pull/974) Protect region targeted by `org-roam-insert`
 - [#994](https://github.com/org-roam/org-roam/pull/994) Simplify org-roam-store-link
+- [#910](https://github.com/org-roam/org-roam/pull/910) Support fuzzy links of the form [[Title]], [[*Headline]] and [[Title*Headline]]
 
 ### Bugfixes
 
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -152,7 +152,9 @@ For example: (setq org-roam-buffer-window-parameters '((no-other-window . t)))"
 (defun org-roam-buffer--insert-backlinks ()
   "Insert the org-roam-buffer backlinks string for the current buffer."
   (if-let* ((file-path (buffer-file-name org-roam-buffer--current))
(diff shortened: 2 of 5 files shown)


TimQuelch:
Interesting, I've been keen curious what the hold up for #592 was.

This implementation avoids the definition of a new link type which is a positive. How is the behaviour of these fuzzy links different from the roam links in the other PR? Can you easily switch between fuzzy links and full file links (I see you have a todo for this, is it trivial?)?

Looks very nice!

jethrokuan:
> How is the behaviour of these fuzzy links different from the roam links in the other PR? 

Behaviour-wise, likely to be exactly the same.

>  Can you easily switch between fuzzy links and full file links

My thought right now is the fuzzy links resolve to exact file links upon navigation, and file save.

For those who want the Roam-like link editing, we will instead modify the backspace/delete behaviour for org-roam files, ala `org-roam-backward-delete-char`. The behaviour I'm thinking of will change all Org-roam file links to fuzzy links on link editing:

e.g.
[[file:foo.org][Foo|]] -> [[Fo|]]

Let me know if you have better ideas.

jethrokuan:
Suggestion from @tiensonqin is to have:

- `[[Foo]]` resolve to files with corresponding title/alias as per `(org-roam--extract-titles)`
- `[[#Foo]]` resolve to headlines within the file

I like this distinction, and will likely move forward with this. Upon navigation or file save, the links will be resolved to links that work in vanilla Org (id, or file links).

We'll also most likely only support this form of headline autocompletion/insertion for headlines within the same file. Anything else that requires more precision would be done via `store-link` and `insert-link`.

wedens:
Is it possible to make an option to not create a file if it doesn't exist and only show a buffer with backlinks to the title?

jethrokuan:
preview video: https://www.youtube.com/watch?v=v4fEwppIPhY

jethrokuan:
> Is it possible to make an option to not create a file if it doesn't exist and only show a buffer with backlinks to the title?

@wedens unlikely to be implemented, we want the eventual file be a regular org file.

TimQuelch:
That demo video looks great. I like the change to '*' from '#'; it makes a lot more sense in the context of org headings

zaeph:
> That demo video looks great. I like the change to '*' from '#'; it makes a lot more sense in the context of org headings

Well, frankly, we were attached to the `#`, but it turns out `org-ref` had already claimed it, which caused some weird interactions.  @jethrokuan thought of `*` a little after, and it all made sense. :)

kidoln:
> preview video: https://www.youtube.com/watch?v=v4fEwppIPhY

Hi, may I know how to configure the Emacs to get the pop-up menu list during inserting a file(or heading) link?

jethrokuan:
```
(setq company-idle-delay 0
      company-minimum-prefix-length 0)
```

kidoln:
Fuzzy link doesn't work in Chinese headline. Would you double check it?

PR #913, opened 2020-07-10
(fix): Make org-roam-graph work with new schemata

### LINKED ISSUE #912: commit: efba3c2bf0c887 breaks org-roam-graph, bibtex and roam-server
### Description
commit: efba3c2bf0c887 changes the structure of the data base. As a result, queries that were valid before are not no more. More specifically, queries to `titles` now return an error.  I think consequential  changes like this should be placed in a different branch until they are tested so that they don't leak into melpa.

#### Steps to Reproduce
1. update org-roam
2. run org-roam-show-graph

#### Expected Results
Graph.

#### Actual Results
get error emacsql no titles.

### Environment

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/efba3c2bf0c887f9aa5c4984ffc0cb5fb0b4bd31

zaeph:
Caused by #908.

Fixes #912.

----

I haven’t looked at it too deeply, but 1fe704cf716608d3859faa6084a36b715c492da4 is not enough to address the issue.  I’ve hardly worked on the graph, so I’m not the best person to address it in a timely fashion.

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -232,7 +232,7 @@ CALLBACK is passed the graph file as its sole argument."
                         "Please adjust `org-roam-graph-executable'")
                 org-roam-graph-executable))
   (let* ((node-query (or node-query
-                         `[:select [file titles] :from titles
+                         `[:select [file title] :from titles
                            ,@(org-roam-graph--expand-matcher 'file t)]))
          (graph      (org-roam-graph--dot node-query))
          (temp-dot   (make-temp-file "graph." nil ".dot" graph))
@@ -269,7 +269,7 @@ CALLBACK is passed to `org-roam-graph--build'."
                         (org-roam-db--links-with-max-distance file max-distance)
                       (org-roam-db--connected-component file))
                     (list file)))
-         (query `[:select [file titles]
+         (query `[:select [file title]
                   :from titles
                   :where (in file [,@files])]))
     (org-roam-graph--build query callback)))


zaeph:
#914 took care of it.  Closing.

jethrokuan:
Superceded by #914

icyflame:
Thank you for the quick fix in #914! I just installed org-roam and ran into this error :sweat_smile: 

***

For posterity, I will post a summary of the issue and my investigation here.

## Error

The error that was being thrown:

```sh
(emacsql-error "near line 17: no such column: titles")
```

## Root cause

I ran `toggle-debug-on-error` and then ran `org-roam-graph-show`. This gave me the following stacktrace:

```lisp
Debugger entered--Lisp error: (emacsql-error "near line 17: no such column: titles")
  signal(emacsql-error ("near line 17: no such column: titles"))
  #f(compiled-function (conn) #<bytecode 0x171b455>)(#<emacsql-sqlite3-connection emacsql-sqlite3-connection>)
  apply(#f(compiled-function (conn) #<bytecode 0x171b455>) #<emacsql-sqlite3-connection emacsql-sqlite3-connection> nil)
  emacsql-parse(#<emacsql-sqlite3-connection emacsql-sqlite3-connection>)
  #f(compiled-function (connection sql &rest args) "Send SQL s-expression to CONNECTION and return the results." #<bytecode 0x171385d>)(#<emacsql-sqlite3-connection emacsql-sqlite3-connection> [:select [file titles] :from titles])
  apply(#f(compiled-function (connection sql &rest args) "Send SQL s-expression to CONNECTION and return the results." #<bytecode 0x171385d>) #<emacsql-sqlite3-connection emacsql-sqlite3-connection> [:select [file titles] :from titles])
  emacsql(#<emacsql-sqlite3-connection emacsql-sqlite3-connection> [:select [file titles] :from titles])
  apply(emacsql #<emacsql-sqlite3-connection emacsql-sqlite3-connection> [:select [file titles] :from titles] nil)
  org-roam-db-query([:select [file titles] :from titles])
  org-roam-graph--dot([:select [file titles] :from titles])
  org-roam-graph--build(nil org-roam-graph--open)
  org-roam-graph-show(nil)
  funcall-interactively(org-roam-graph-show nil)
  call-interactively(org-roam-graph-show nil nil)
  command-execute(org-roam-graph-show)
```

Clearly, the error is coming from the query that is being built by the `org-roam-graph--build` function. It is building the query `select [file titles] :from titles`

```lisp
  org-roam-graph--dot([:select [file titles] :from titles])
  org-roam-graph--build(nil org-roam-graph--open)
```

I checked the schema of the SQLite3 file created by `org-roam-db-build-cache`:

```sql
sqlite> .schema
CREATE TABLE files (file UNIQUE PRIMARY KEY, hash NOT NULL, meta NOT NULL);
CREATE TABLE headlines (id UNIQUE PRIMARY KEY, file NOT NULL);
CREATE TABLE links ("from" NOT NULL, "to" NOT NULL, type NOT NULL, properties NOT NULL);
CREATE TABLE tags (file UNIQUE PRIMARY KEY, tags);
CREATE TABLE titles (file NOT NULL, title );
CREATE TABLE refs (ref UNIQUE NOT NULL, file NOT NULL, type NOT NULL);
```

Clearly, the `titles` table has two columns and they are called `file` and `title`.

After running `(org-roam-version)`, I realized that I was running v1.2.0. At v1.2.0, this line was indeed constructing the incorrect query:

https://github.com/org-roam/org-roam/blob/v1.2.0/org-roam-graph.el#L233

### Resolution

Construct the query using the column name `title` instead of using the column name `titles`

https://github.com/org-roam/org-roam/commit/f1dbe3fdf9821805dfb4d4d037bd8c8bc2f96252#diff-79abd010ed52c8e2ae26068495893cc1L235-R235

Issue #905, opened 2020-07-09
org-roam not respecting `org-todo-keywords`


cantao:
#### Steps to Reproduce

1. Load Emacs
2. Open a roam file
3. Hit `org-todo` to change TODO to DEFERRED (from my `org-todo-keywords`)
4. org-roam pulls up another set of keywords (Doom Emacs default?)

#### Backtrace
No backtrace because it does not break things apart.

#### Expected Results

My `org-todo-keywords` appearing for selection. In not-roam org files it works perfectly.

#### Actual Results

Another set of keywords, probably Doom Emacs' default ones.

### Environment

- Emacs: GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20) of 2020-05-19
- Framework: Doom
- Org: Org mode version 9.4 (nil @ /home/cantao/.emacs.d/.local/straight/build/org-mode/)
- Org-roam: 1.2.0

Thank you very much!

jethrokuan:
I cannot reproduce (it's using my set of `org-todo-keywords`)

How are you customizing `org-todo-keywords`? Are they file specific?

cantao:
In .doom.d/config.el:

For org-mode:

```
(use-package! org
  :defer t
  :config
  (setq org-agenda-files (list my-org-dir my-roam-dir)
        ;; ... lots of org mode configuration ...
        org-todo-keywords '((sequence "TODO(t)" "NEXT(n)" "WAITING(w@)" "|" "DONE(d@)" "CANCELLED(c@)" "DEFERRED(r@)"))
        ;; ... lots more ....
))
```
For org-roam:
```
(use-package! org-roam
  :defer t
  :config
  (setq org-roam-directory my-roam-dir
        org-roam-index-file (concat my-roam-dir "roam-index.org")))
```
I have similar sections for `org-ref`, `org-roam-bibtex`, `deft` and so on. Variables starting with `my-` refer to directories or files.

zaeph:
Are you sure the problem is caused by Org-roam?  Could you comment out the `use-package` block for Org-roam and try again?

cantao:
Very strange. It must be something with the order packages are loaded. I tried commenting everything except org-mode and org-roam. Tried uncommenting one block at a time and now I can't reproduce the problem :-|

I'll make some more tests and close the bug if cannot reproduce it.

cantao:
[...]

Issue #909, opened 2020-07-10
org-roam backlinks linking to emacs backup files after editing 


npalermo10:
### Description
When I edit a file, everything that was linked to it instead links to the file's emacs backup (the filename ending in "~"). 

#### Steps to Reproduce
run org-roam-find to find a file with backlinks
edit the file and save it
all backlinks to that file dissappear
If you check any of the files that were previously linked they have all changed their target file to the emacs backup '~' of the file instead.

### Environment
- Emacs: GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.14)
 of 2020-03-26, modified by Debian
- Framework: N/A
- Org: Org mode version 9.3.6 (9.3.6-elpa @ /home/nick/.emacs.d/elpa/org-9.3.6/)
- Org-roam: 1.2.0
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

zaeph:
Someone mentioned a similar issue on Slack today.

As a temporary measure, you can start storing your backup files in a dedicated directory *outside* of the CWD:
```el
(setq backup-directory-alist '(("." . "~/.saves")))
```

@jethrokuan: Have we recently touched something that might cause that behaviour?  Mind you, I haven't tried reproducing it, yet.

npalermo10:
that temporary fix definitely works! thank you and thanks for all the hard work. I love this program.

jethrokuan:
Likely to be this: https://github.com/org-roam/org-roam/pull/894

zaeph:
I cannot reproduce this.

I've tried with the following settings, which should be close to default:
```el
;; Don’t clobber symlinks
(setq backup-by-copying nil)

;; Use versioned backups
(setq version-control nil)

;; Number of backups to keep
(setq kept-new-versions 10
      kept-old-versions 2
      delete-old-versions nil)

;; Backup directories
;; (setq backup-directory-alist '(("." . "~/.saves")))
(setq backup-directory-alist nil)

;; Also backup versioned files
(setq vc-make-backup-files nil)
```

I've also tried with my personal settings, minus
```el
;; Don’t clobber symlinks
(setq backup-by-copying t)

;; Use versioned backups
(setq version-control t)

;; Number of backups to keep
(setq kept-new-versions 10
      kept-old-versions 0
      delete-old-versions t)

;; Backup directories
[...]

PR #915, opened 2020-07-10
Exclude backup files from renaming advice

mxco86:
Exclude backup files from triggering the file/link renaming advice


zaeph:
Closes #905.

npalermo10:
Yeah the update fixes it. Thanks!

zaeph:
It hasn't fixed it for everyone apparently, though, cf #909.

zaeph:
Two persons have confirmed that this fix works for them, and I don't think this would cause problem elsewhere.  I'll investigate some more later, but it's not a blocker.

Merging.

zaeph:
Oh, and thank you!

Issue #922, opened 2020-07-11
Declare `org-roam-capture-templates` with `defcustom`


myshevchuk:
### Description

See discussion in the [Slack thread](https://app.slack.com/client/TTPSWL02H/CV160S8EL/thread/CV160S8EL-1594245213.449000)

@jethrokuan, is `plain`still the only entry type and `org-roam-capture--get-point` the only destination to use in `org-capture-templates`?

myshevchuk:
Also, are `:file-name` and `:header` and `:unnarrowed t` strictly required for `org-roam-caputre` to function properly?

These elements can be enforced in Customize, and such an enforcement will eliminate potential sources of confusion for those using the interface, while requiring zero changes to `org-roam-capture` internal logic.

myshevchuk:
This is an example of how the Customize buffer would look like for the default template
``` el
(("d" "default" plain (function org-roam-capture--get-point)
     "%?"
     :file-name "%<%Y%m%d%H%M%S>-${slug}"
     :head "#+title: ${title}\n"
     :unnarrowed t))
```
Note that `plain`, `(function org-roam-capture--get-point)` and `:unnarrowed t` are completely hidden from the user, while `:file-name` and `:header` are enforced. 

<img width="577" alt="Screenshot 2020-07-11 at 14 26 58" src="https://user-images.githubusercontent.com/24212814/87224075-97e8c280-c382-11ea-9dac-09cf73eaf727.png">

myshevchuk:
Or like this, which would look slightly better for template/header strings containing newlines.

<img width="637" alt="Screenshot 2020-07-11 at 14 50 00" src="https://user-images.githubusercontent.com/24212814/87224479-f82d3380-c385-11ea-878e-5c51ce730d74.png">

zaeph:
Closed by #924, and similar issues tracked in #895.

PR #924, opened 2020-07-11
defcustom `org-roam-capture-templates`

myshevchuk:
Declare `org-roam-capture-templates` with `defcustom` and update the
docstring

See #922

`org-roam-capture-templates` is a user variable and should be accessible from the Customize interface.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -82,7 +82,7 @@ note with the given `ref'.")
 (defconst org-roam-capture--template-keywords '(:file-name :head)
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")
 
-(defvar org-roam-capture-templates
+(defcustom org-roam-capture-templates
   '(("d" "default" plain (function org-roam-capture--get-point)
      "%?"
      :file-name "%<%Y%m%d%H%M%S>-${slug}"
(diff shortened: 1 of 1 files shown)


zaeph:
Thanks.  I'll review this later today.

myshevchuk:
I'll also do the same for `org-roam-capture-ref-templates` and `org-roam-capture-immediate-template`, so don't merge this yet. By the way, what was the rationale behind two additional templates variables? I didn't follow Org-roam closely for a while.

zaeph:
> By the way, what was the rationale behind two additional templates variables?

Users where bothered by the inability to nest `org-capture` sessions, and we've proposed those commands as an alternative which allowed to return directly from an `org-capture` session at the cost of being able to cancel it with `C-c C-k`.  It's a temporary fix which will have to do until we reconsider whether we actually need `org-capture`.

This is not to say that your work would be useless, by the way.  Presumably, we'd only be changing the plumbing, not the front-end for those templates, and it would be transparent for the users.

zaeph:
Thanks!

zaeph:
Ah, damn it, just remembered that you wanted to hold off until you'd addressed the other variables.  Sorry!

PR #927, opened 2020-07-11
(fix): Check if link-description is empty prior to inserting

### LINKED ISSUE #926: Org links of the type type:foo breaks the parser
### Description
As the title says, links `type:foo` break the parsing.

Sorry for not giving more details, it was already confirmed by @zaeph.

zaeph:
Caused problems with plain links, i.e., those not framed by
brackets (e.g. `file:foo.org` or `id:$uuid`).

Fixes #926.


zaeph:
The change is in 05a958dc4e4cbe25827ca86d29f7a01b5eb624a1; the rest is indent adjustment and testing.

PR #928, opened 2020-07-12
(fix): Stop resolving symlinks to their destinations

### LINKED ISSUE #919: org-roam does not work correctly if `org-roam-directory` points to a link
### Description

My `org-roam-directory` is pointing to `/home/myuser/org/org-roam-notes`, which is a symolic link to `/mnt/data/myuser/Dropbox/org/org-roam-notes`. Regular emacs functions such as `find-file` respect the value of a variable `find-file-visit-truename`, which in my case (default value) is set to `nil`, and will treat the link as a regular file (or folder). For instance, if I use `find-file` to open the `/home/myuser/org/org-roam-notes` folder then I will be in that folder, but if the `find-file-visit-truename` variable is non-nil then emacs would follow the link and dired would be in the `/mnt/data/myuser/Dropbox/org/org-roam-notes` folder instead.

The problem is that org-roam seems to always follow the symbolic links. That is, functions such as `org-roam-insert` or `org-roam-find-file` use the true name and location of the file instead of using the link name and its location. This causes several problems, including the backlinks not working. Also, when using `org-roam-insert` to create a link to some `note2.org` file in a `note1.org` file, the `note2.org` file is not recognized as being in my `org-roam-directory` and the created link is something such  as `file:../../../../../../home/myuser/org/org-roam-notes/note2.org` instead of just `file:note1.org`.

#### Steps to Reproduce

1. Set `org-roam-directory` to point to a link of a folder with some roam notes
2. Use `org-roam-find-file` to open a note
3. Use (for instance) `find-file` to find out the directory of the currently open file, or use `org-roam-insert` to link to a different note and check the link created by org-roam

#### Backtrace

Not available

#### Expected Results

org-roam should behave as if my `org-roam-directory` was not pointing to a link

#### Actual Results

Not available

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.21, cairo version 1.17.3)
 of 2020-07-02
- Framework: Spacemacs
- Org: Org mode version 9.3.7 (9.3.7-12-g5d6868-elpaplus @ /home/darlan/.emacs.d/elpa/28.0/develop/org-plus-contrib-20200706/)
- Org-roam: 1.2.0

zaeph:
Affects both `org-roam-find-file' and the db.

Fixes #919.

-----

This might cause problem with libs.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -471,7 +471,7 @@ Use external shell commands if defined in `org-roam-list-files-commands'."
 
 (defun org-roam--list-all-files ()
   "Return a list of all Org-roam files within `org-roam-directory'."
-  (org-roam--list-files (file-truename org-roam-directory)))
+  (org-roam--list-files org-roam-directory))
 
 ;;;; Org extraction functions
 (defun org-roam--extract-global-props (props)


zaeph:
Closed this by mistake, but #919 will probably require a more thorough solution anyway.

zaeph:
Closed this by mistake. #919 will probably require a more thorough solution than what is currently suggested.

emacs18:
I don't think this change addresses the root cause of the problem.  It may seem to resolve some problems, but I think it may make matters worse in other cases.  I think the most robust solution is to make sure that `file-truename` is called everywhere where a file is passed to any db related functions, i.e., rather than removing one use of `file-truename` as is proposed in this PR, i think we should go the other way by adding many more `file-truename` calls so that whenever a file name is passed into any db related function, then we are assured of using the true name of a file without any symlinks involved.

I'll try to find time to submit such a PR, but it may take some time.

jethrokuan:
> i think we should go the other way by adding many more file-truename calls so that whenever a file name is passed into any db related function, then we are assured of using the true name of a file without any symlinks involved.

`file-truename` calls are expensive, and I'd like to reduce this as much as possible. We should be able to call file-truename once, and specify which functions require the true names to be passed as arguments.

emacs18:
@zoechi For me this is whether org-roam is dropped like hot potatoes or not on first test drive.  I probably tried org-roam couple of times over the past year, but I couldn't figure out what this "backlink" stuff was, because all I saw was 'No backlinks" on the buffer that was supposed to show the back links.

I wonder how many other folks are on the same boat as me and drop org-roam after few minutes of use thinking that it was just unusable.

emacs18:
I stumbled onto `find-file-visit-truename` variable.  I'm hopeful that setting this to `t` will help greatly.

Issue #958, opened 2020-07-22
I cannot add new files that partially match the name of existing files


tonicebrian:
### Description
When I try to insert a new file using either `org-roam-find-file` or `org-roam-insert-immediate` if there is a match with an existing file I cannot create the new file, it always picks the match in the list.

#### Steps to Reproduce

- Create a file called `CoreNLP` and save it.
- Then try to create a file called `nlp`. Any of the methods `org-roam-find-file` or `org-roam-insert-immediate` will pick `CoreNLP` entry instead of creating a new file.

#### Expected Results

* I have the option to create new files even when the new name partially matches an existing one.

#### Actual Results

I cannot insert files with shorter names that match existing files in my folder.

### Environment

- Emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109))
 of 2019-09-02
- Framework: Doom
- Org: Org mode version 9.4 (nil @ /Users/cebriant/.emacs.d/.local/straight/build/org-mode/)
- Org-roam: 1.2.0

zaeph:
With `ivy`, you need to use `C-M-j` (i.e. `Ctrl+Alt+j`), cf. [their FAQ](https://github.com/abo-abo/swiper#frequently-asked-questions).  With `helm`, just move to the candidate with your current input, usually the first one which is prepended with `[?]`.

PR #959, opened 2020-07-22
(doc): Add a FAQ section

zaeph:
* README.md: Create FAQ section and mention `C-M-j` for ivy

Prompted by #958.


jethrokuan:
There is a corresponding section in the manual, which could use this info as well.

PR #966, opened 2020-07-24
(feat): enable nested captures

jethrokuan:
This PR enables the long-awaited nested-captures.

This PR introduces a number of changes:

1. Adds a hook to `org-capture-prepare-finalize-hook`, which installs `org-roam-capture--finalize` into `org-capture-after-finalize-hook` if the capture is an Org-roam capture. This function contains key functionality for Org-roam to Do The Right Thing after specific interactive functions, such as finding the file, or inserting a link.

2. A patch for `org-capture-finalize`. Specifically, we make `org-capture-plist` valid during `org-capture-finalize`.

3. Many hacks that were originally in place are now replaced with nicer alternatives.

nested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [#863](https://github.com/org-roam/org-roam/pull/863) Display outline hierarchy in backlinks buffer
 - [#898](https://github.com/org-roam/org-roam/pull/898) Add `org-roam-random-note` to browse a random note.
 - [#900](https://github.com/org-roam/org-roam/pull/900) Support and index all valid org links
+- [#966](https://github.com/org-roam/org-roam/pull/966) Enable nested captures
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


jethrokuan:
Supercedes #527

jethrokuan:
another issue: point isn't advanced after link insertion

zaeph:
I'll have a look in the morning to see if I can come up with a solution for the snags.

zaeph:
Re: 770a4a15554fecf4e177fad48fa50931d65e5770.
I've sent a patch upstream.  Let's keep the extra hook for now, and remove it if the patch gets merged.

zaeph:
49a478c8c1c9dce46f0dca5574fc7cc3bb8f7d92 was functional, but I didn't like the implementation.  Tomorrow, I'd like to our custom `get`/`put`, now that I've grokked how it works upstream.

zaeph:
[Here's the upstream patch](https://orgmode.org/list/87h7tv9pkm.fsf@hidden/T/#u).

Issue #895, opened 2020-07-06
Prefer defcustom over defvar where appropriate


progfolio:
Users are  unable to leverage the customization system for variables defined with defvar (e.g. org-roam-capture-templates, org-roam-capture-ref-templates, etc.).

myshevchuk:
Sorry, I didn't see this issue. Can #922 be merged here?

progfolio:
#924 does not address `org-roam-capture-ref-templates`.

zaeph:
Yes, sorry for that, should have double-checked.

progfolio:
No problem. Additionally: `org-roam-dailies-capture-templates` is still a defvar.

myshevchuk:
Yeah, so would it be possible to combine #922 and this, because they are the same issue.

zaeph:
Let's close #922, then, since the issue mentioned in the title has been addressed by #924.

myshevchuk:
But there are questions there that have not been answered.

myshevchuk:
I thought it was possible to somehow merge the issues, to transfer the discussion from one issue o another, bu probably I confuse it with something else.

zaeph:
I've checked, and it doesn't seem to be possible.  Could you migrate your question to a different issue, since they're somewhat out of the scope of creating `defcustom` templates for the template-variables?

myshevchuk:
The questions were about creating the defcusoms. It is possible to do this in several different ways, so I wanted to clarify which way was preferable. Basically, I wanted to know whether `plain`, `(function org-roam-capture—get-point)`, `:unnarrowed t` should be hidden from the user and whether `:file-name` and `:Head` are optional. This is purely for defcustom and doesn’t touch the variable itself.

zaeph:
Right, I'd misread them, then.  I still think that there are interesting ramifications to those questions, i.e., whether we actually need those in the first place.  It'll be interesting to gauge this when the time comes to reconsider `org-capture`.

myshevchuk:
Could someone please assign me to this issue, so I don't forget it please?

myshevchuk:
I see that definition of `org-roam-capture-immediate-template` has changed. I assume it's an internal variable now and does not require a `defcutom`?

```elisp
(defvar org-roam-capture-immediate-template
  (append (car org-roam-capture-templates) '(:immediate-finish t))
  "Capture template to use for immediate captures in Org-roam.
This is a single template, so do not enclose it into a list.
[...]

PR #968, opened 2020-07-25
customize settings for templates variables

myshevchuk:
- org-roam-capture-ref-templates
- org-roam-dailies-capture-templates

#895


myshevchuk:
Since a few list parameters are declared positionally dependant, in some situations Customize may report value mismatch if set from `setq`. I believe, it should be fine if user uses exclusively Customize.

myshevchuk:
@jethrokuan Sorry, `org-roam-capture-immediate-template` needs an adjustment. Don't merge it yet.

PR #974, opened 2020-07-26
Protect region targeted by `org-roam-insert`

zaeph:


zaeph:
The linter isn't happy about this:
https://github.com/org-roam/org-roam/blob/77e8174a4161549de63f4835ab000ba915bc7dee/org-roam.el#L1472-L1473

org-roam.el
```diff
@@ -1407,6 +1407,31 @@ included as a candidate."
...
+  (when region
+    (pcase-let ((`(,min . ,max) region))
+      (when unshield
+        (let ((inhibit-read-only t))
+          (remove-text-properties min max '(read-only t))))
+      (delete-region min max)
+      (when reinsert
+        (save-excursion
+          (goto-char min)
+          (insert (org-roam-capture--get :link-description))))
```
jethrokuan, line 1425: Does reinsert change the text? why not just not delete?

zaeph: It's to handle `org-not-abort` cases.  Originally, I only removed the `[[` and `]]`, but, it made it harder to modularise the code.  So, since we were already storing `:link-description`, and the region cannot be edited whilst the `org-roam-capture` process is active,  I figured it was fine to do it that way.  I could remove the arg and use `(when (org-note-abort) …) instead.

zaeph: > I could remove the arg and use (when (org-note-abort) …) instead.

Not too sure about this, actually.  I thought it'd hurt modularity, but since we're only using it `org-roam-insert`…

Not a fan of this function.

zaeph:
1235284864209068760a24b3b0265510e795141d addresses most of the grievances I had with the function.

PR #978, opened 2020-07-27
Improve org-roam-dailies

zaeph:
`org-roam-scratch` is a rewrite of `org-roam-dailies` which strives to implement useful behaviours from `org-journal`.  The rationale is that `org-journal` can be a little inflexible (cf. https://github.com/bastibe/org-journal/pull/278), and we wouldn't want to hijack its functions to do something that it wasn't designed to do, especially since we'd only be using a subset of the features.

Therefore, the project is indebted to @bastibe, and will strive to incorporate the _ephemeral scratchings_ as defined by @andymatuschak in his [Taxonomy of Note Types](https://notes.andymatuschak.org/Taxonomy_of_note_types), and as demonstrated in a [livestream](https://www.youtube.com/watch?v=DGcs4tyey18) he did.


zaeph:
Re: default variable change
https://github.com/org-roam/org-roam/blob/ed8ed27b31d039c73ad865f1fb31c2fb5227b385/org-roam-dailies.el#L42-L45

I think it makes more sense to have `org-roam-dailies-today` and co. be `'entry` templates by default in order to replicate the traditional behaviour associated with `org-capture`.  I'm going to write separate commands like `org-roam-dailies-find-today` to only go to the file, since I do not believe it's necessary to use `org-capture` in that scenario.

Mind you, it's still be possible to use `'plain` templates with `org-roam-dailies`, and to have them return immediately with `:immediate-finish t`, but placing the cursor at an appropriate place would be complicated.

zaeph:
PR is almost ready.  I still need to define a rule for considering a file as a daily note.  I'm thinking of using `org-roam-dailies--file-to-date` as a predicate.  This way, people could have daily notes in multiple directories, and wouldn't have to declare those directories explicitly.  There would be false positives with regular notes including a timestamp, but those would be limited.  Other option is to add a metadata-tag for dailies which would contain the date.

…I'll think about this tomorrow.

zaeph:
PR is almost done.  I just need to edit the changelog and a couple of other minor stuff.

org-roam-capture.el
```diff
@@ -414,27 +435,58 @@ aborted, we do the following:
...
+           (find-file file-path)
+           (insert (substring (org-capture-fill-template (concat roam-head "*"))
+                              0 -2))
+           (set-buffer-modified-p nil))
+         (org-capture-put :template org-template))
+        (_
+         (org-capture-put :template
+           ;; Fixes org-capture-place-plain-text throwing 'invalid search bound'
+           ;; when both :unnarowed t and "%?" is missing from the template string;
+           ;; may become unnecessary when the upstream bug is fixed
```
progfolio, line 484: This was fixed as of: https://code.orgmode.org/bzg/org-mode/commit/09086b7e75a4ccb27f141d2cd495c253f1b6b025

zaeph: Splendid: I'll amend the code.

zaeph: Addressed in 7325b0af7fd053be506dfb32762c8fb8730ad390.

org-roam-capture.el
```diff
@@ -41,6 +41,9 @@
 (defvar org-roam-directory)
 (defvar org-roam-mode)
 (defvar org-roam-title-to-slug-function)
+(defvar org-roam-dailies-capture--file-name-default)
```
jethrokuan, line 44: let's remove these, and throw an error if the template doesn't contain it.

zaeph:
Here's a quick demo of the new features:
https://youtu.be/1q9x2aZCJJ4

lhindir:
Anything holding this up? It's kind of weird to see one of the few demo videos and then find out it's for unmerged features :)

zaeph:
I'm still bogged down in work, and I'd rather not announce any other ETA that I won't be able to meet.  That being said, I'm growing more antsy with the days to come back and merge this, so it shouldn't be too-too long.

zaeph:
bac221d6154a35e321b1533bc39df37691f1dc78 is a rudimentary rebase of the PR on `master`, and most things seem to work.  I'll need to review a few things, but you can expect a merge soon.

zaeph:
I've fixed the `defcustom` template in 0f81f5a7e6312cda74dc0711d84b490d85dbef65.

Here are example templates:
```el
(setq org-roam-dailies-capture-templates

      '(("d" "default" plain
         #'org-roam-capture--get-point
         "%?"
         :file-name "daily/%<%Y-%m-%d>"
         :head "#+title: %<%Y-%m-%d>\n\n"
         :unnarrowed t)

        ("a" "alternative template" entry
         #'org-roam-capture--get-point
         "* %?"
         :file-name "daily/%<%Y-%m-%d>"
         :head "#+title: %<%Y-%m-%d>\n\n* Foo"
         :olp ("Foo"))))
```

There are subtle difference (notably `plain` vs `entry`), but I think it's fairly powerful.

@jethrokuan I've elected to keep `org-roam-dailies-capture--file-name-default` and `org-roam-dailies-capture--header-default` for now because also have `org-roam-capture--file-name-default` and `org-roam-capture--header-default`.  If we need to drop the former in former of a clever `pcase`, we might as well remove the others, and I'd rather do that in a separate PR, if need be.

zaeph:
I'll review everything in the morning, but I think we'll be good for a merge.

jethrokuan:
> @jethrokuan I've elected to keep org-roam-dailies-capture--file-name-default and org-roam-dailies-capture--header-default for now because also have org-roam-capture--file-name-default and org-roam-capture--header-default. If we need to drop the former in former of a clever pcase, we might as well remove the others, and I'd rather do that in a separate PR, if need be.

sure, I'll drop them in a later PR.

zaeph:
Code is ready in 4ee16069590e2e3431945fb4c81e9caa76d8034e.

I'm working on the doc to mention the new features (also those in #1239).

org-roam-capture.el
zaeph: Will be addressed in another PR.

Issue #947, opened 2020-07-18
graph links not working in windows


fkgruber:
### Description
Clicking on a node does not open the correct file. Instead it creates a new one.
#### Steps to Reproduce
I configured org-protocol with
```
REGEDIT4

[HKEY_CLASSES_ROOT\org-protocol]
@="URL:Org Protocol"
"URL Protocol"=""
[HKEY_CLASSES_ROOT\org-protocol\shell]
[HKEY_CLASSES_ROOT\org-protocol\shell\open]
[HKEY_CLASSES_ROOT\org-protocol\shell\open\command]
@="\"C:\\emax64\\bin\\emacsclientw.exe\" \"%1\""
```
But when I click on a link a new empty emacs buffer is created with file name:

![image](https://user-images.githubusercontent.com/4311763/87859365-9a0ecc00-c902-11ea-9c8b-34ec7817751d.png)

It seems it is adding file= to the name so it can't find it.

Diagnostic info:
- Emacs: GNU Emacs 26.2 (build 1, x86_64-w64-mingw32)
 of 2019-04-17
- Framework: N/A
- Org: Org mode version 9.3.7 (9.3.7-8-g3ed035-elpaplus @ c:/Users/fred/emax/elpa/org-plus-contrib-20200629/)
- Org-roam: 1.2.0

nobiot:
Hi @fkgruber, I suspect you are using Chrome, and the cause of the issue is in Org Mode (not Org-roam).

I have just written up an extensive instruction on how to set up org-roam-protocol and graph (and fix the issue I suspect you experience) in my guide: https://github.com/nobiot/Zero-to-Emacs-and-Org-roam/blob/main/90.org-protocol.md

Could you see if it addresses your issue?

fkgruber:
Thanks! that solved my problem. I was missing the last step. The patch that fixed the chrome issues.

aleneum:
Just fyi: I managed to get links working without the need of changing the windows registry. When using Firefox, you can add a custom protocol handler in `about:config`:
- Type `about:config` into your address bar, confirm the dialog
- enter `network.protocol-handler.expose.org-protocol` in the search bar
- Leave the option `boolean`, hit `+` and change the value to `false`
- Next time you try to open an `org-protocol` link, Firefow will ask you what program to use
- Choose `emacsclientw` and make this permanent (check 'remember my choice'/'alway open with')

nobiot:
This is awesomeeeee!!!

PR #980, opened 2020-07-28
Fixed org-protocol link in graph

odomanov:
I'm not sure this is correct, but I had to do this on Windows to get it working. Perhaps `org-roam-protocol.el` should also be adjusted.


odomanov:
Sorry for bothering. I've just found #947. Org mode patch solved the problem (although I use Firefox, not Chrome). So I'm canceling the PR.

PR #993, opened 2020-07-31
Emacs created in 76, not 85.

dangom:

https://en.wikipedia.org/wiki/Emacs
Emacs development began during the 1970s.


zaeph:
Thanks for catching that.  Merging.

Issue #982, opened 2020-07-28
Change the way `org-roam-store-link` shadows `org-store-link`


zaeph:
Brought up in #853.

We need to find a better way to shadow `org-store-link` with `org-roam-store-link` than merely rebinding to the same key.  The rationale was to not affect commands/functions which use `org-store-link`, i.e., `org-capture` with the template extension for context (`%a`), but I think this would be done by completely shadowing `org-store-link`.  However, this comes with problems, since if we shadow `org-store-link`, other packages would be unable to do the same.

I'll think about it when I'm done with `org-roam-dailies`.

PR #994, opened 2020-07-31
(fix): Simplify org-roam-store-link

### LINKED ISSUE #853: Use org-roam-store-link for org-capture annotations
`org-capture-templates` support `%a` for "annotations", which use `org-store-link` to capture a link to heading `org-capture` was called from.

Now that we have `org-roam-store-link` to handle links to headlines, it would be great if `org-roam-store-link` could be used for the captured annotation.

-----------------------------------------------

See relevant [Slack conversation](https://orgroam.slack.com/archives/CV20S23C0/p1592455046209400) with @zaeph , who thinks this should already be working (making it a bug). I don't see why, since the only related override is to remap the `org-store-link` key binding.

zaeph:
Drop the wrapper, and refactor as an org-store-link function.

Fixes #853, #982.

Right now, as long as you're in an Org-roam file, it'll create an ID when you're capturing on a heading.  I think it's going to be hard to make it cleaner, i.e., only creating the ID when necessary.  Right now, if you cancel the capture, the ID remains, which leads to ID rot.  I'm investigating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ### Features
 
--[#974](https://github.com/org-roam/org-roam/pull/974) Protect region targeted by `org-roam-insert`
+- [#974](https://github.com/org-roam/org-roam/pull/974) Protect region targeted by `org-roam-insert`
+- [#994](https://github.com/org-roam/org-roam/pull/994) Simplify org-roam-store-link
 
 ### Bugfixes
 
--- a/org-roam.el
+++ b/org-roam.el
@@ -1015,7 +1015,7 @@ for Org-ref cite links."
                      target))
 
 (defun org-roam-store-link ()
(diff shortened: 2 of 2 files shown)


jethrokuan:
> Right now, if you cancel the capture, the ID remains, which leads to ID rot. I'm investigating.

That also happens with regular org-capture, since `org-capture` calls `org-store-link`. That's okay though, I think.

zaeph:
> That also happens with regular org-capture, since `org-capture` calls `org-store-link`.

Yup, that's right.

> That's okay though, I think.

It is, especially since  I'd like to implement an `org-roam-doctor` check for stale IDs.

dangirsh:
After pulling this change and running `doom sync && doom build`, I'm now getting "%a" template expansions as org-ids. Thanks!!!

PR #1002, opened 2020-08-05
(feat) Implements link types

odomanov:
This implements link types in Org-roam. Types should be previously defined in Org, for example via `org-link-set-parameters`. Org-roam already recognizes `file:` and `cite:` link types, this code makes it possible to add more. Additional types should be declared in `org-roam-extra-link-types`. Link types are saved in the database and faces, the look of graph's edges etc. can be adjusted according to them.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -340,11 +340,12 @@ the capture)."
              (when region
                (delete-region (car region) (cdr region)))
              (let ((path (org-roam-capture--get :file-path))
+                   (type (org-roam-capture--get :link-type))
                    (desc (org-roam-capture--get :link-description)))
                (if (eq (point) (marker-position mkr))
-                   (insert (org-roam--format-link path desc))
+                   (insert (org-roam--format-link path desc type))
                  (org-with-point-at mkr
-                   (insert (org-roam--format-link path desc))))))))))
+                   (insert (org-roam--format-link path desc type))))))))))
     (when region
       (set-marker beg nil)
       (set-marker end nil))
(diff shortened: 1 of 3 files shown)


odomanov:
Check failed. The Lint error is strange for me. I can't understand what's wrong.

odomanov:
OK, I fixed it.

jethrokuan:
Rather than have an `extra-types`, can we instead cache *all* link types, and be able to query them from the cache and display them appropriately with the buffer revamp that will come later on. E.g. if org-brain is required, then org-brain links will be cached.

There is already a column in sqlite db for this.

This has been on the roadmap for a while, if you want you can choose to tackle it. The solution would then be much more general and require less configuration.

odomanov:
Yes, I thought about this, but I'm not sure how to implement it. We can of course store all link types in the database, but what I had in mind is the difference between links _inside_ and _outside_ of Org-roam. For example, the graph should probably contain not all links but only some part of them (which I would call "inside links"). So should we maintain this difference explicitly? Of course, we can count all links pointing to files within org-roam directory as inside links, but I'm not sure this works for `cite` links.

What definitely makes sense is to rename `org-roam-extra-link-types` to `org-roam-link-types` (meaning "links inside Org-roam") and integrate `file:` and `cite:` links into it. I can do this.

jethrokuan:
> We can of course store all link types in the database, but what I had in mind is the difference between links inside and outside of Org-roam. For example, the graph should probably contain not all links but only some part of them (which I would call "inside links").

This can be done via some sort of post filtering. (e.g. are the do the files/ids at the vertices of these edges both belong to Org-roam files?). I think it may also be interesting to see where notes lead to outside of Org-roam, so these are worth caching.

Either way, I don't think link-types is the right abstraction level to filter between inside and outside links. For example, a file link could link outside of Org-roam.

jethrokuan:
Superceded by #1009

PR #1004, opened 2020-08-06
(internal): remove with-template-error wrapper

jethrokuan:
This has caused more confusion than it has helped.


zaeph:
I guess it made more sense to have this when we weren't inserting ourselves so much in the capture buffer-creation.  Removing it is probably for the best.

PR #1009, opened 2020-08-07
Implement link types

odomanov:
I'm sorry for the mess, I'm not very familiar with pull requesting. This is a continuation of #1002. I guess I made a mistake requesting from the `master` branch, now I can't understand how I can push there. So this request is from a separate branch. This is a general approach as you said, storing all links in the database.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -319,7 +319,7 @@ Insertions can fail if the key is already in the database."
 If the file does not have any connections, nil is returned."
   (let* ((query "WITH RECURSIVE
                    links_of(file, link) AS
-                     (WITH filelinks AS (SELECT * FROM links WHERE \"type\" = '\"file\"'),
+                     (WITH filelinks AS (SELECT * FROM links WHERE NOT \"type\" = '\"cite\"'),
                            citelinks AS (SELECT * FROM links
                                                   JOIN refs ON links.\"to\" = refs.\"ref\"
                                                             AND links.\"type\" = '\"cite\"')
@@ -341,7 +341,7 @@ This includes the file itself. If the file does not have any
 connections, nil is returned."
   (let* ((query "WITH RECURSIVE
                    links_of(file, link) AS
-                     (WITH filelinks AS (SELECT * FROM links WHERE \"type\" = '\"file\"'),
+                     (WITH filelinks AS (SELECT * FROM links WHERE NOT \"type\" = '\"cite\"'),
                            citelinks AS (SELECT * FROM links
                                                   JOIN refs ON links.\"to\" = refs.\"ref\"
                                                             AND links.\"type\" = '\"cite\"')
--- a/org-roam.el
+++ b/org-roam.el
@@ -500,7 +500,8 @@ PATH should be the root from which to compute the relativity."
         ;; Strip 'file:'
         (setq link (match-string 2))
         ;; Delete relative link
-        (when (f-relative-p link)
+        (when (and (f-relative-p link)
+                   (org-roam--org-roam-file-p (expand-file-name link dir)))
           (delete-region (match-beginning 2)
                          (match-end 2))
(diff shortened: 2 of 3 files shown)


org-roam.el
```diff
@@ -1625,9 +1625,10 @@ included as a candidate."
   (find-file (seq-random-elt (org-roam--list-all-files))))
 
 ;;;###autoload
-(defun org-roam-insert (&optional lowercase completions filter-fn description)
+(defun org-roam-insert (&optional type lowercase completions filter-fn description)
```
jethrokuan, line 1628: can you place this argument at the end, and rename it to `link-type`? Not sure if other packages rely on this function.

odomanov: I did. 
I actually put it in the beginning to make keybinding to `org-roam-insert type` more convenient.

odomanov:
I'm not sure what you mean. ID links are already stored in the database -- with the type "id".

org-roam-db.el
```diff
@@ -319,7 +319,7 @@ Insertions can fail if the key is already in the database."
 If the file does not have any connections, nil is returned."
   (let* ((query "WITH RECURSIVE
                    links_of(file, link) AS
-                     (WITH filelinks AS (SELECT * FROM links WHERE \"type\" = '\"file\"'),
+                     (WITH filelinks AS (SELECT * FROM links WHERE NOT \"type\" = '\"cite\"'),
```
jethrokuan, line 322: That question is wrt to this query, where `filelinks` will then contain all sorts of other links (including id links).

I'm just curious as to what additional link types you needed support for.

odomanov: Any types defined in Org, actually. I can then show them differently on the graph, for example.

But what I need first are types I myself can define. Semantically, notes in Org-roam can be connected differently. A link can denote the continuation of a main theme, a sort of  detalization, or simple informational link, or whatever. Luhman in his Zettels differentiated them through numbering. So I can, for example, define linktypes `major:` and `minor:` and see the difference in connections visually if I adjust faces or colors on the graph. At the moment, I'm not sure what linktypes I (or other users) might need and I'd like to have some general mechanism for that.

By the way, I noticed an error in handling id links, I'll need to check it.

jethrokuan: > So I can, for example, define linktypes major: and minor: and see the difference in connections visually if I adjust faces or colors on the graph.

So my only concern with this now is that Org-roam isn't aware of what kind of link that is supposed to be. Is the path a file path, or an id, or something else entirely? Org-roam doesn't handle this well now, so this PR is only part of the solution.

odomanov: Yes, and this is why in the beginning I introduced the concept of "inner link types" which Org-roam knows how to handle. But I don't think this is a problem, because when Org-roam doesn't know the type, it passes it to Org (right?), which knows it. As for custom links, users should describe how to handle them. For example, I define `major:` and `minor:` links as virtually equal to `file:`, only faces are different.

odomanov: There is a problematic place though: `org-roam--expand-links` should expand only `file:`-like links. Or. maybe, only links, pointing to Org-roam files.

jethrokuan: Let's keep `org-roam--expand-links` such that it expands only `file:` links for now, and then in a later PR, we can introduce a `org-roam-link-types` alist that acts as a lookup so the different link types can be handled appropriately.

odomanov: What if we expand any link type but only if the path belongs to the org-roam-directory? This makes sense.

jethrokuan: This is fine too.

odomanov: ok, done.

odomanov: Tests failed by timeout.

jethrokuan: It looks like that last change made the cache build much slower.

PR #1030, opened 2020-08-11
Add custom var to disable headline linking

cpbotha:
This small PR adds a customization variable with which org-roam's new headline linking functionality can be disabled.

This  means that:

- org-roam will not assign an auto-id when you use `org-store-link`
- org-roam will not scan the headlines of each org file for IDs to save into its database.

# Motivation

I would prefer not to have an org-id automatically assigned whenever I do `org-store-link`.  This is especially troublesome when I have already assigned a `CUSTOM_ID` as I often do for normal orgmode use, and then this code assigns an automatic ID and links to that instead.

Furthermore, I don't make use of org-roam's headline linking functionality, preferring its initial approach of supporting exclusively compact, single-thought files, a la zettelkasten, and so having org-roam extract the IDs of all of the headlines in the file I'm working on and saving that to the database further increases the delay I see when saving org files.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -512,9 +512,10 @@ If FORCE, force a rebuild of the cache from scratch."
                       :values $v1]
                      (vector file contents-hash (list :atime atime :mtime mtime)))
                     (setq file-count (1+ file-count))
-                    (when-let ((headlines (org-roam--extract-headlines file)))
-                      (when (org-roam-db--insert-headlines headlines)
-                        (setq headline-count (1+ headline-count)))))
+                    (when org-roam-enable-headline-linking
+                      (when-let ((headlines (org-roam--extract-headlines file)))
+                        (when (org-roam-db--insert-headlines headlines)
+                          (setq headline-count (1+ headline-count))))))
                 (file-error
                  (setq org-roam-files (remove file org-roam-files))
                  (org-roam-db--clear-file file)
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -38,7 +38,7 @@
 (require 'org-roam-macs)
 
 (defvar org-roam-directory)
-(defvar org-roam-disable-headline-linking)
+(defvar org-roam-enable-headline-linking)
 (defvar org-roam-verbose)
 (defvar org-roam-file-name)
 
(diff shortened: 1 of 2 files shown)


zaeph:
Looks good on the paper, and I'll have to go through the codebase to make sure that this does not have side-effects elsewhere.  Thanks!

TimQuelch:
I don't like the mixing of negatives by having 'disable' in the name of the option. It leads to convoluted double negatives like `(when (not org-roam-disable-headline-linking))` or `(unless org-roam-disable-headline-linking)`.

I'd recommend changing the option name to `org-roam-enable-headline-linking` and set the default to be on.

TimQuelch:
It's probably also be a good idea to include it as a `defcustom` so it is exposed as a user variable

e.g. https://github.com/org-roam/org-roam/blob/master/org-roam.el#L247-L250

cpbotha:
> It's probably also be a good idea to include it as a `defcustom` so it is exposed as a user variable
> 
> e.g. https://github.com/org-roam/org-roam/blob/master/org-roam.el#L247-L250

Could you double-check the files of this PR? defcustom is the first thing I did. :)

cpbotha:
> I don't like the mixing of negatives by having 'disable' in the name of the option. It leads to convoluted double negatives like `(when (not org-roam-disable-headline-linking))` or `(unless org-roam-disable-headline-linking)`.
> 
> I'd recommend changing the option name to `org-roam-enable-headline-linking` and set the default to be on.

I'll flip it around like you suggest.

Just FYI, I started with `disable-` to make the new behaviour opt-in, but indeed that gives these mixed negatives. (purely as a curiosity, both my native language Afrikaans and  that of @zaeph have double negatives, so maybe it comes more naturally. ;)

cpbotha:
@zaeph The liter is erroring on other parts of the code, claiming that dependency on emacs 27.1 should be specified. It didn't do this with my earlier commits. Could you take a look?

jethrokuan:
It's due to 27.1 being newly released. You can ignore it for now, I'll fix it in a separate PR.

It looks like you've missed `org-roam-db-build-cache` in your changes: https://github.com/org-roam/org-roam/blob/22006be751804cd19f617a1433fc77ab98258916/org-roam-db.el#L513

cpbotha:
> It's due to 27.1 being newly released. You can ignore it for now, I'll fix it in a separate PR.
> 
> It looks like you've missed `org-roam-db-build-cache` in your changes:
> 
> https://github.com/org-roam/org-roam/blob/22006be751804cd19f617a1433fc77ab98258916/org-roam-db.el#L513

Thank you very much @jethrokuan -- I have just added the check also to `org-roam-db-build-cache`.

cpbotha:
@jethrokuan and @zaeph please let me know if there's anything else I can do here to help to wrap up this work.

jethrokuan:
I'm going to modify this a little.

jethrokuan:
On second thoughts, this is fine.

PR #1035, opened 2020-08-13
let org-roam-capture's arguments same as org-capture

lld2001:
I want to bind `org-roam-capture` to a key by using a template. And I do not want to select a template every time.
Keeping `org-roam-capture` arguments the same as `org-capture`, then you can designate template by using KEYS. Which is useful if there are more than one template in `org-roam-capture-templates`.


lld2001:
Sorry. I'm a github newbee. I don't know is the PR workflow OK.

PR #1053, opened 2020-08-20
(feat): add org-roam-delete-backward-char and org-roam-delete-char

jethrokuan:

Add smart delete, making links fuzzy upon edit.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -324,6 +324,50 @@ function are expected to catch the error."
                  (t
                   (signal 'wrong-type-argument `((stringp numberp symbolp) ,item))))) items))))
 
+;;;; Movement functions
+
+(defun org-roam-delete-backward-char (N)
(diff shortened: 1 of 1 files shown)


TimQuelch:
I very rarely actually start editing something by backspacing into it. I use evil, so most of my edits would be initiated with a `cw`, or maybe a `R` to enter replace, or a `r<letter>` to replace a single letter. Often I don't even want to delete part of it, I just want to insert text (`i` or `a`).

I can see myself becoming frustrated with needing to edit the link in exactly the right way for it to turn into a fuzzy link properly.

I wonder if there is some general way to catch it on *any* editing of the text in the link? This however could also be very expensive to be checking constantly

jethrokuan:
Ah yeah, I didn't think of that. Let me think about this abit more.

PR #1056, opened 2020-08-21
(feat): add case sensitivity for Org-roam completions

jethrokuan:
Completions are now case-insensitive by default.
`org-roam-completion-case-sensitive` can be set to `t` to allow
case-sensitive completions.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1211,8 +1211,6 @@ If USE-STACK, include the parent paths as well."
                        (or (find-buffer-visiting file)
                            (find-file-noselect file)))
                   (current-buffer)))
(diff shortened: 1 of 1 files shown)


BenSturmfels:
It looks as though the above variable doesn't exist anymore, so for anyone looking to make `org-roam-node-find` and `org-roam-node-insert` case-insensitive, you can add the following Emacs config. This temporarily sets `completion-ignore-case` before it runs the `org-roam-node-read` function:

```
(defun case-insensitive-org-roam-node-read (orig-fn &rest args)
  (let ((completion-ignore-case t))
    (apply orig-fn args)))

(advice-add 'org-roam-node-read :around #'case-insensitive-org-roam-node-read)
```

PR #1057, opened 2020-08-23
Disable latex preview and images for temp buffers

TimQuelch:
Generation of these preview is often expensive and completely unnecessary
when the buffer won't actually be viewed.

This will have the side effect of not showing latex or images when fixing something in a doctor run, but I think that is worth the sacrifice.

----

Should make DB builds and doctor runs faster when the user has `org-startup-with-latex-preview` or `org-startup-with-inline-images` set.

### DIFF
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -58,6 +58,7 @@
 
 (defvar org-roam-verbose)
 (defvar org-roam-mode)
+(defvar org-roam-doctor-inhibit-startup)
 
 (cl-defstruct (org-roam-doctor-checker (:copier nil))
   (name 'missing-checker-name)
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -294,7 +294,9 @@ If CHECKALL, run the check for all Org-roam files."
 (defun org-roam-doctor-start (files checkers)
   "Lint FILES using CHECKERS."
   (save-window-excursion
-    (let ((existing-buffers (org-roam--get-roam-buffers)))
+    (let ((existing-buffers (org-roam--get-roam-buffers))
+          (org-startup-with-latex-preview nil)
+          (org-startup-with-inline-images nil))
       (dolist (f files)
         (let ((buf (find-file-noselect f)))
           (org-roam-doctor--check buf checkers)
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -295,8 +295,7 @@ If CHECKALL, run the check for all Org-roam files."
   "Lint FILES using CHECKERS."
   (save-window-excursion
     (let ((existing-buffers (org-roam--get-roam-buffers))
-          (org-startup-with-latex-preview nil)
-          (org-startup-with-inline-images nil))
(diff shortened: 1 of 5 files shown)


org-roam-doctor.el
```diff
@@ -294,7 +294,9 @@ If CHECKALL, run the check for all Org-roam files."
 (defun org-roam-doctor-start (files checkers)
   "Lint FILES using CHECKERS."
   (save-window-excursion
-    (let ((existing-buffers (org-roam--get-roam-buffers)))
+    (let ((existing-buffers (org-roam--get-roam-buffers))
+          (org-startup-with-latex-preview nil)
```
jethrokuan, line 298: Looks like we can have a variable `org-roam-doctor-inhibit-startup` that we set `org-inhibit-startup` to. Similar to how `org-agenda` speeds up the buffer preparation.

org-roam-macs.el
```diff
@@ -55,7 +55,9 @@ If FILE, set `org-roam-temp-file-name' to file and insert its contents."
     `(let ((,current-org-roam-directory org-roam-directory))
        (with-temp-buffer
          (let ((org-roam-directory ,current-org-roam-directory)
-               (org-mode-hook nil))
+               (org-mode-hook nil)
+               (org-startup-with-latex-preview nil)
```
jethrokuan, line 59: simply setting `(org-inhibit-startup t)` should save more time:

https://github.com/emacs-straight/org-mode/blob/a1e5bee5cb9c34ceb8226597605a49638bee7cec/lisp/org.el#L4883

jethrokuan (changes requested):

org-roam-macs.el
TimQuelch: https://github.com/org-roam/org-roam/pull/1057/commits/b2de9f4b76423010aa1c8c2a00bb7d05e7fac9b3

org-roam-doctor.el
TimQuelch: https://github.com/org-roam/org-roam/pull/1057/commits/b2de9f4b76423010aa1c8c2a00bb7d05e7fac9b3

jethrokuan (approved):
Thanks!

spacekitteh:
This prevents the `vanilla` tag source from retrieving `#+FILETAGS` when the cache is first built. When the cache is updated after a save, though, it includes them for that specific file. Is there a way around this?

TimQuelch:
I had a quick look in `org-inhibit-startup` and I couldn't see anything which is the obvious cause for this.

As a quick hacky workaround you could enable startup when `'vanilla` is a tag source
```diff
@@ -56,7 +56,7 @@ If FILE, set `org-roam-temp-file-name' to file and insert its contents."
        (with-temp-buffer
          (let ((org-roam-directory ,current-org-roam-directory)
                (org-mode-hook nil)
-               (org-inhibit-startup t))
+               (org-inhibit-startup (not (member 'vanilla org-roam-tag-sources))))
            (org-mode)
            (when ,file
              (insert-file-contents ,file)
```

spacekitteh:
Hmm, that might be worth doing until a more principled approach is found.

Issue #940, opened 2020-07-15
Performance issues on large files


brettmilford:
### Description
Emacs hangs for a considerable time (~20 seconds) when saving a particular org-mode file and org-roam mode is enabled.
The file is ~1300 lines long, it was previously ~6k but I reduced it to see if it would improve the issue (it did, but only slightly)
Is there an expected file size, or known upper limit on what can be processed is a reasonable amount of time? Running profile seems to suggest its spending most of its time parsing the file.

#### Steps to Reproduce

1. Load an org file of >1300 lines with org-roam on
2. Edit and save the file

#### Expected Results

Emacs completes the save and returns input.

#### Actual Results

Emacs hangs for a considerable amount of time before returning.

#### Profile

```
- command-execute                                                1524  92%
 - call-interactively                                            1524  92%
  - funcall-interactively                                        1524  92%
   - evil-ex                                                     1422  86%
    - apply                                                      1422  86%
     - +ivy--inhibit-completion-in-region-a                      1422  86%
      - let                                                      1422  86%
       - apply                                                   1422  86%
        - #<compiled 0x41279e11>                                 1422  86%
         - evil-ex-execute                                       1413  85%
          - eval                                                 1413  85%
           - evil-ex-call-command                                1413  85%
            - call-interactively                                 1413  85%
             - funcall-interactively                             1412  85%
              - evil-write                                       1412  85%
               - save-buffer                                     1412  85%
                - basic-save-buffer                              1412  85%
                 - run-hooks                                     1408  85%
                  - org-roam-db--update-file                     1374  83%
                   - org-roam-db--update-links                   1214  73%
                    - org-roam--extract-links                    1211  73%
                     - org-element-map                           1187  71%
                      + #<compiled 0x437e6ef9>                   1187  71%
                     + org-element-parse-buffer                    23   1%
                     + file-truename                                1   0%
                    + org-roam-db-query                             2   0%
                    + file-truename                                 1   0%
                   + org-roam-db--update-titles                    47   2%
                   + org-roam-db--update-tags                      36   2%
                   + org-roam-db--update-headlines                 30   1%
                   + org-roam-db--update-refs                      25   1%
                   + org-roam-buffer--update-maybe                 14   0%
                   + org-roam-db--update-meta                       6   0%
                   + org-roam--org-roam-file-p                      2   0%
                  + undo-fu-session-save-safe                      30   1%
                  + org-encrypt-entries                             1   0%
[...]

PR #1061, opened 2020-08-24
Speed up extraction of global props, headlines, and title

kisaragi-hiu:

Org-roam takes a long time to process files. In addition, big files (10000+ lines) can take several seconds to process after it is modified. This becomes unbearable for eg. habit (`org-habit`) tracking files.

This PR introduces some optimizations:

- Replace unnecessary uses of `org-element-parse-buffer`
  - `org-roam--extract-headlines`: Use `org-map-region` and `org-entry-get` instead of `org-element-parse-buffer`.
  - `org-roam--extract-titles-headline`: Find the first headline with `org-outline-regexp-bol` instead of using `org-element-parse-buffer`.
- If it's available, use `org-collect-keywords` which is a lot faster as it also avoids `org-element-parse-buffer`. (As it is a relatively new function in Org, [added in April 2020](https://code.orgmode.org/bzg/org-mode/commit/b4e91b7e944c900db6b8217d78011afcd2c1e62c), I have kept the previous implementation if it's not present.)

## Performance comparison

`vocab.org` is an Org file with 14795 lines (1052 headings).

I'm running this on a Ryzen 1400 CPU.

```elisp
(with-current-buffer "vocab.org"
  (list
   (list 'total 'gc-count 'gc-time)
   (benchmark-run 1
     (org-roam--extract-headlines))
   (benchmark-run 1
     (org-roam--extract-titles-title))
   (benchmark-run 1
     (org-roam--extract-titles-headline))
   (benchmark-run 1
     (org-roam--extract-titles-alias))))
```

Before:

| total       | gc-count | gc-time            |
|-------------|----------|--------------------|
| 1.771265479 | 1        | 0.7655653740000048 |
| 0.932756371 | 0        | 0.0                |
| 1.645399532 | 1        | 0.7050031620000254 |
| 1.701283592 | 1        | 0.742148186999998  |

After:

| total       | gc-count | gc-time |
|-------------|----------|---------|
| 0.149613758 | 0        | 0.0     |
| 0.00223786  | 0        | 0.0     |
| 8.2539e-05  | 0        | 0.0     |
| 0.001961503 | 0        | 0.0     |

From 5 seconds down to 0.15 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ In this release we support fuzzy links of the form `[[Title]]`, `[[*Headline]]` 
 
 ### Features
 
+- [#1061](https://github.com/org-roam/org-roam/pull/1061) Speed up the extraction of file properties, headlines, and titles
 - [#1046](https://github.com/org-roam/org-roam/pull/1046) New user option to exclude files from Org-roam
 - [#1032](https://github.com/org-roam/org-roam/pull/1032) File changes are now propagated to the database on idle timer. This prevents large wait times during file saves.
 - [#974](https://github.com/org-roam/org-roam/pull/974) Protect region targeted by `org-roam-insert`
--- a/org-roam.el
+++ b/org-roam.el
@@ -495,16 +495,20 @@ Use external shell commands if defined in `org-roam-list-files-commands'."
 (defun org-roam--extract-global-props (props)
   "Extract PROPS from the current org buffer.
 The search terminates when the first property is encountered."
-  (let ((buf (org-element-parse-buffer))
-        res)
-    (dolist (prop props)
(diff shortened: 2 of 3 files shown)


jethrokuan:
The changes LGTM.

It looks from the tests like the headline extraction now returns the candidates in reverse order, so the tests also need to be fixed here.

kisaragi-hiu:
> It looks from the tests like the headline extraction now returns the candidates in reverse order, so the tests also need to be fixed here.

Ah, yes, because `push` adds to the front. I've made it so that the test uses `:to-have-same-items-as` to ignore the ordering of its items.

jethrokuan (approved):
LGTM, very nice :smile:

jethrokuan:
Related to #834 #940

PR #1065, opened 2020-08-25
(internal): use regexp-search for fuzzy link replacement

jethrokuan:
This should improve responsiveness on large files

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1344,18 +1344,13 @@ Three types of fuzzy links are supported:
 (defun org-roam--replace-all-fuzzy-links ()
   "Replace all fuzzy links in current buffer."
   (save-excursion
-    (let ((fuzzies (org-element-map (org-element-parse-buffer) 'link
-                     (lambda (l)
-                       (when (equal (org-element-property :type l)
-                                    "fuzzy")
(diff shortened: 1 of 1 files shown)


TimQuelch:
How does this interact with places where links cannot appear such as within source blocks? 

If I have,

    #+begin_src python
    a = [['some text']]
    #+end_src

will this change interpret this as a fuzzy link and attempt to replace this?

The reason I ask is I'm having trouble saving a note with a large number of source blocks with brackets everywhere and it is just hanging on save. I haven't done significant debugging yet but I thought that this might be the issue.

If this is the issue I think it would apply to #1066.

TimQuelch:
I just reverted to f6bf9d94019547436f565156bf7dc12f414cfa86 and it fixed the issue i'm having, so I do think it is these changes to regexp which is the problem. I might have time to do a full bisect later.

jethrokuan:
yes, this is also the cause of #1069, looks like replacing fuzzy link behaviour backfired a little.

jethrokuan:
The real offender is the unneeded `(goto-char (match-beginning 0))` which will cause the link replacement to fail when it does not find a match, but I fixed it all in one go in #1070 .

PR #1071, opened 2020-08-26
(feat): rename link descriptions on title changes

### LINKED ISSUE #1044: Update links when the note's title is changed
### Brief Abstract

Org-roam already updates links when filenames are renamed, but not when titles are changed. It would be great if org-roam updated the links' titles too.

### Long Description

Org-roam uses the linking mark-up implemented in org, which is `[[file:filename A][title A]]`. If you rename `filename A` to `filename B` through `dired` or `deft`, org-roam globally updates all links to the recently modified file, and the link becomes `[[file:filename B][title A]]`. However, suppose that you also want to change your target note's title to `title B`. When you do that, org-roam doesn't propagate the change to the links in other notes. The links remain as `[[file:filename B][title A]]`, even if your note is now titled `title B`.

Someone on the Discourse group suggested that if you changed the title before renaming the filename, the change would propagate, but it didn't work. I'm not sure such a feature is implemented yet.

I think the simplest way to implement this would be creating a function like `org-roam-change-title` which not only prompts the user for a new title but also visits each backlink and replaces the old title for the new one. There could be an option to prevent org-roam from behaving like this if the user wishes so.

jethrokuan:
Previously the rename file advice was responsible for trying to update
the link descriptions as well. This was brittle and didn't work in some
cases.

The rename-file-advice has now been altered to solely deal with breaking
links. We use the before and after-save-hooks to determine whether the
title has changed, and update the link descriptions accordingly.

Fixes #1044

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ In this release we support fuzzy links of the form `[[Title]]`, `[[*Headline]]` 
 
 ### Features
 
+- [#1071](https://github.com/org-roam/org-roam/pull/1071) Update link descriptions on title changes, and clean-up rename file advice
 - [#1061](https://github.com/org-roam/org-roam/pull/1061) Speed up the extraction of file properties, headlines, and titles
 - [#1046](https://github.com/org-roam/org-roam/pull/1046) New user option to exclude files from Org-roam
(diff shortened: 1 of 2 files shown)


lrustand:
This doesn't work. All existing links have still the old name after I changed the title of a note. Even manually syncing the org-roam db and killing + re-opening any note containing a link to the renamed note doesn't help. All links still show the old name.

PR #1073, opened 2020-08-27
(feat): rename file on title change

### LINKED ISSUE #939: Rename a file using the #+TITLE property
### Brief Abstract

It would be useful to have a function that renames a file using the `#+TITLE:` property.

### Long Description

I find myself constantly returning to my notes, extending them, adding references, refiling paragraphs to new notes and otherwise making them more concise. Sometimes this involves changing the title of the note. Whenever I do this, I also change the filename with `rename-file` or `doom/move-this-file` so that I can more easily identify the file in a file browser or on Android. This can be a bit of a pain.

What I would like to be able to do is run a function that gets the `#+TITLE` property from the file header, passes it to `org-roam--title-to-slug`, and then renames the file and updates all of the backlinks. Ideally, this would be run on the file in the currently open buffer, but I think having the option to pass the function a file path also makes sense.

I had a shot at putting this together but it is beyond me. What I tried to do was: fetch the title with `+org-get-global-property`, convert this to a slug with `org-roam--title-to-slug`, use this to create a 'new-path' variable, and then, use a modified version of the `doom/move-this-file` code to rename the file (which in turn calls `rename-file` and so updates the backlinks).

jethrokuan:
Adds `org-roam-title-change-hook`. This now contains two functions:

1. `org-roam--update-links-on-title-change`: updates the link
description of all files that link to the current file.
2. `org-roam--update-file-name-on-title-change`: updates the current
buffer's filename to reflect the updated title.

Resolves #939 .

rename

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ In this release we support fuzzy links of the form `[[Title]]`, `[[*Headline]]` 
 - [#910](https://github.com/org-roam/org-roam/pull/910) Deprecate `company-org-roam`, using `completion-at-point` instead. To use this with company, add the `company-capf` backend instead.
 
 ### Features
-
+ 
+- [#1073](https://github.com/org-roam/org-roam/pull/1073) Rename file on title change, when `org-roam-rename-file-on-title-change` is non-nil.
 - [#1071](https://github.com/org-roam/org-roam/pull/1071) Update link descriptions on title changes, and clean-up rename file advice
 - [#1061](https://github.com/org-roam/org-roam/pull/1061) Speed up the extraction of file properties, headlines, and titles
 - [#1046](https://github.com/org-roam/org-roam/pull/1046) New user option to exclude files from Org-roam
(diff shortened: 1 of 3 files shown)


oatmealm:
I didn't get to test it yet. Will it also change descriptions entered manually? Ie I often link back to notes from many different descriptions, not necessarily the title of the particular note I'm linking to...

jethrokuan:
> Will it also change descriptions entered manually?

if the description doesn't exactly match the title, it won't change the description.

oatmealm:
Great. Just got a chance to try. Looks great 👍 thanks.

TimQuelch:
Just for future reference for anyone: If you want to disable the renaming of link descriptions in other files on title change you can add the following to your config
```
(remove-hook 'org-roam-title-change-hook 'org-roam--update-links-on-title-change)
```

shlevy:
Possibly I had something set up wrong, but I just changed the #+title of a note and it renamed the file (which was nice) but didn't update the target of any links that point to the file (not so nice). Is that expected?

shlevy:
Note that I moved the old title into aliases.

PR #1076, opened 2020-08-29
Handle new file title change.

zaks:
Creating new files with org-roam-insert causes errors:

Wrong type argument: char-or-string-p, nil


jethrokuan (approved):

d12frosted:
Thanks for the fix!

PR #1085, opened 2020-09-04
(fix): fix link outline extraction

jethrokuan:
In addition to getting the correct element, obtaining the outline path
requires the point to be correctly set.

cc @Wetlize, thanks for testing.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -584,11 +584,10 @@ it as FILE-PATH."
   (let (links)
     (org-element-map (org-element-parse-buffer) 'link
       (lambda (link)
+        (goto-char (org-element-property :begin link))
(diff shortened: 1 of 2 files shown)


Wetlize:
Thanks for the fix, this seems to work fine.

A bit offtopic, but wouldn't it make sense to display outlines in the same order as they appear in each buffer? Right now it displays them bottom-up, instead of a top-down manner.

jethrokuan:
> I am not sure if this is off-topic, but wouldn't it make sense to display outlines in the same order as they appear in each buffer? Right now it displays them bottom-up, instead of a top-down manner.

Yeah it does, the sort should happen during display, rather than link extraction though.

PR #1088, opened 2020-09-06
(feat): Support quoting text when storing via org-roam-protocol

russell:

There is a difference documentation between the elisp file and the example in the online documentation. The online documentation contains less functionality than the example in the elisp file. There was also a bug in the example in the emacs lisp file.   

image

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -219,7 +219,7 @@ Template string   :\n%v")
 
 (defcustom org-roam-capture-ref-templates
   '(("r" "ref" plain (function org-roam-capture--get-point)
-     "%?"
+     "%i%?"
      :file-name "${slug}"
      :head "#+title: ${title}\n#+roam_key: ${ref}\n"
      :unnarrowed t))
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -79,7 +79,7 @@ note with the given `ref'.")
 
 (defcustom org-roam-capture-templates
   '(("d" "default" plain (function org-roam-capture--get-point)
-     "%i%?"
+     "%?"
      :file-name "%<%Y%m%d%H%M%S>-${slug}"
      :head "#+title: ${title}\n"
      :unnarrowed t))
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -79,7 +79,7 @@ note with the given `ref'.")
 
 (defcustom org-roam-capture-templates
   '(("d" "default" plain (function org-roam-capture--get-point)
-     "%?"
+     "%i%?"
      :file-name "%<%Y%m%d%H%M%S>-${slug}"
      :head "#+title: ${title}\n"
      :unnarrowed t))
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -63,14 +63,14 @@ It opens or creates a note with the given ref.
            (template (cdr (assoc 'template decoded-alist)))
            (org-capture-link-is-already-stored t))
       (let* ((title (cdr (assoc 'title decoded-alist)))
-            (url (cdr (assoc 'ref decoded-alist)))
-            (body (cdr (assoc 'body decoded-alist)))
-            (type (and url
-                       (string-match "^\\([a-z]+\\):" url)
-                       (match-string 1 url)))
-            (orglink
-             (if (null url) title
-               (org-link-make-string url (or (org-string-nw-p title) url)))))
+             (url (cdr (assoc 'ref decoded-alist)))
+             (body (or (cdr (assoc 'body decoded-alist)) ""))
+             (type (and url
+                        (string-match "^\\([a-z]+\\):" url)
+                        (match-string 1 url)))
+             (orglink
+              (if (null url) title
+                (org-link-make-string url (or (org-string-nw-p title) url)))))
         (when url
           (push (list url
                       title)
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -5,7 +5,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 1.2.1
-;; Package-Requires: ((emacs "27.1") (org "9.3"))
+;; Package-Requires: ((emacs "26.1") (org "9.3"))
 
 ;; This file is NOT part of GNU Emacs.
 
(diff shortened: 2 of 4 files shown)


russell:
ok, the extended example from the elisp file doesn't actually work, it's ignoring the data field.

I've done a bit of an investigation, and it looks like because the 'org-protocol` method uses the org-link's to pass data to the rendering template.  I have created an org-link with all the required fields and set the initial field as the value passed in as data, this allows passing of selections through to a new roam note

org-roam-protocol.el
```diff
@@ -5,7 +5,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 1.2.1
-;; Package-Requires: ((emacs "26.1") (org "9.3"))
+;; Package-Requires: ((emacs "27.1") (org "9.3"))
```
jethrokuan, line 8: is this change necessary?

org-roam-protocol.el
```diff
@@ -60,7 +60,26 @@ It opens or creates a note with the given ref.
     (let* ((org-roam-capture-templates org-roam-capture-ref-templates)
            (org-roam-capture--context 'ref)
            (org-roam-capture--info decoded-alist)
-           (template (cdr (assoc 'template decoded-alist))))
+           (template (cdr (assoc 'template decoded-alist)))
+           (org-capture-link-is-already-stored t))
+      (let* ((title (cdr (assoc 'title decoded-alist)))
+            (url (cdr (assoc 'ref decoded-alist)))
```
jethrokuan, line 66: indentation looks off here

jethrokuan (commented):
In the quoted elisp example, you can include `${body}` in your ref template to make use of the body field, was that what you were looking for?

org-roam-protocol.el
russell: It's because of these 

```
org-roam-protocol.el:73:16: error: You should depend on (emacs "27.1") if you need `org-link-make-string'.
org-roam-protocol.el:78:9: error: You should depend on (emacs "27.1") if you need `org-link-store-props'.
```

But I will make this more portable, I just did the minimum to get the PR green enough to make it worth your time :)

russell:
> In the quoted elisp example, you can include `${body}` in your ref template to make use of the body field, was that what you were looking for?

Yeah, so that works, but the problem with using the `${body}` variable is that if it's not passed through then the user will be prompted for it.

Let me expand what i have there and i'll describe my motivation a bit better.


Passing in quoted and unquoted org-protocol links is handled in a less than friendly way for `org-capture`. The way most people seem to solve this is that they have 2 different methods for handling this situation,

```
              ("p" "protocol - capture a link and quote" entry (function rs/goto-current-journal)
               "* %(format-time-string org-journal-time-format)%^{Title}\n   [[%:link][%:description]]\n\n   #+BEGIN_QUOTE\n   %i\n   #+END_QUOTE\n\n%?"
               :head "#+title: ${title}\n"
               :empty-lines 1)
              ("L" "protocol - capture a link" entry (function rs/goto-current-journal)
               "* %(format-time-string org-journal-time-format)%^{Title}\n   [[%:link][%:description]]\n\n%?"
               :head "#+title: $(format-time-string org-journal-date-format)\n"
               :empty-lines 1)
```

Then you have have to setup 2 bookmarklets or to use a browser extension like https://github.com/sprig/org-capture-extension and then use the correct one based on wheither your passing in quoted text or not.

Screenshot from 2020-09-07 08-00-28

You can see this behavior and how it's configured in this plugin here.



This seem a bit weird to me, i mean essentially we just want to provide a way to template the body on capture, because having 2 alternative endpoints for very similar data is unnecessary, and isn't a very good user experienc.

Why the initial value is better than the variable body, well basically because it won't prompt the user if it's missing.

Why use this org-link thing, yeah, not  that i'm surprised to see that `org-capture` is using global state to pass data through.  But i think it's good to use the same mechanism where it can be, if nothing else just to make sure that the `org-capture-templates` templating functionality is identical. these are the main variables that i can see use values from the org-link
```
  %i          Initial content, copied from the active region.  If
              there is text before %i on the same line, such as
              indentation, and %i is not inside a %(sexp), that prefix
              will be added before every line in the inserted text.
  %a          Annotation, normally the link created with ‘org-store-link’.
  %A          Like %a, but prompt for the description part.
  %l          Like %a, but only insert the literal link.
```

# Example

This is the output i get in a buffer with text selected:
```
#+title: (fix): Support quoting text when storing via org-roam-protocol by russell · Pull Request #1088 · org-roam/org-roam
#+roam_key: https://github.com/org-roam/org-roam/pull/1088

#+BEGIN_QUOTE
There is a difference documentation between the elisp file and the example in the online documentation. The online documentation contains less functionality than the example in the elisp file. There was also a bug in the example in the emacs lisp file.
#+END_QUOTE

```

If nothing is selected
```
#+title: (fix): Support quoting text when storing via org-roam-protocol by russell · Pull Request #1088 · org-roam/org-roam
#+roam_key: https://github.com/org-roam/org-roam/pull/1088

```

org-roam-protocol.el
jethrokuan: You may make use of these functions for backward-compatibility:

https://github.com/org-roam/org-roam/blob/7722b65aaaff985bb9764746bc0a7a68877eaa7c/org-roam-compat.el#L85

org-roam-capture.el
```diff
@@ -85,7 +85,7 @@ note with the given `ref'.")
 
 (defcustom org-roam-capture-templates
   '(("d" "default" plain (function org-roam-capture--get-point)
-     "%?"
+     "%i%?"
```
jethrokuan, line 88: It seems like the wrong template is modified here. Org-roam-ref uses the `org-roam-capture-ref-templates` variable.

org-roam-protocol.el
```diff
@@ -60,7 +67,28 @@ It opens or creates a note with the given ref.
...
            (org-roam-capture--context 'ref)
            (org-roam-capture--info decoded-alist)
-           (template (cdr (assoc 'template decoded-alist))))
+           (template (cdr (assoc 'template decoded-alist)))
+           (org-capture-link-is-already-stored t))
+      (let* ((title (cdr (assoc 'title decoded-alist)))
+             (url (cdr (assoc 'ref decoded-alist)))
+             (body (or (cdr (assoc 'body decoded-alist)) ""))
+             (quoted-text (unless (string-equal body "")
+                            (format org-roam-protocol-quote-template body)))
```
jethrokuan, line 76:
```suggestion
             (body (cdr (assoc 'body decoded-alist)))
             (quoted-text (when body (format org-roam-protocol-quote-template body)))
```

org-roam-protocol.el
```diff
@@ -60,7 +67,28 @@ It opens or creates a note with the given ref.
...
+             (type (and url
+                        (string-match "^\\([a-z]+\\):" url)
```
jethrokuan, line 78: I think you can use `org-link-types-re` here

bphenriques:
👋  I was looking at this exact thing yesterday through custom ref templates (my use case is different)! 😄 What are the results if the selected text is `%(EXP)` for example? I.e., a org template expression.

russell:
>  I was looking at this exact thing yesterday through custom ref templates (my use case is different)!  What are the results if the selected text is `%(EXP)` for example? I.e., a org template expression.
> 
> Edit: I was having troubling as `%EXP` was being evaluate when it shouldn't be.

Yep i see it here too

```
text is %![Error: (void-function EXP)] for example? I.e., a org template expression.
```

russell:
@jethrokuan thank you for rebasing :1st_place_medal:

russell:
I have added back the initial value, @jethrokuan would you like tests or anything else before merging?

jethrokuan:
The primary reason I haven't merged this is because I'm still thinking whether the quote-template should be part of Org-roam, or just something in the user config.

zaeph:
Superseded by #1239.

Thank you for your work!

Issue #1089, opened 2020-09-08
Include a vanilla org-mode tag `org-roam-tag-source`


spacekitteh:
Org-mode has a lot of features related to its tagging system, such as hierarchical tags, file and headline tags, tag searching, etc. These should be an easily used tool for `org-roam-tag-sources`.

### Proposed Implementation (if any)
A list of tags can be obtained with `(org-get-buffer-tags)`. This returns a table of the tags in the buffer, along with their properties. It should be straightforward to extract just the actual tags from that table.

jethrokuan:
Closed with #1093

PR #1093, opened 2020-09-10
Add a vanilla org-mode tag source

spacekitteh:
Addresses #1089.

#1089

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -197,7 +197,9 @@ extraction methods:
     Tags are space delimited.
     Tags may contain spaces if they are double-quoted.
     e.g. #+roam_tags: TAG \"tag with spaces\"
-
+  `vanilla'
+    Extract vanilla org-mode tags, including #+FILETAGS and
+    inherited tags.
   `all-directories'
     Extract sub-directories relative to `org-roam-directory'.
     That is, if a file is located at relative path foo/bar/file.org,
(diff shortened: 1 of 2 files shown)


jethrokuan:
I was looking at `org-get-buffer-tags`, it seems that it additionally pulls down tags for headlines within the file. Is that desired behaviour? I would think that parsing the `FILETAGS` file property would be the right choice.

spacekitteh:
It is, yes - at least until proper integration with headlines/`ID`s is worked out.

org-roam.el
```diff
@@ -726,6 +728,10 @@ tag."
                     (buffer-file-name)))
          nil)))))
 
+(defun org-roam--extract-tags-vanilla (_file)
+  "Extract vanilla `org-mode' tags."
```
jethrokuan, line 732: Can we add some details here about what tags are extracted?

This would be all tags in the file, including filetags, and tags in headlines within the file.

spacekitteh:
Sorry - I was busy all day yesterday! :)

real-or-random:
@spacekitteh 

Thanks for this feature. I just wanted to implement a tag source and then I figured out it's already there. :) 

Though what I had in mind was also that using  `FILETAGS` is more natural.
> It is desired, yes - at least until proper integration with headlines/`ID`s is worked out.

Can you elaborate?

(I'm just trying to understand your reasoning. Maybe it's just better to use `org-get-buffer-tags` only. If my use case is different, I can still implement and possibly PR the `FILETAGS` variant.)

PR #1103, opened 2020-09-14
possibly fix hang for windows

jethrokuan:
URL links such as `https://google.com/hi/` get split into link-type `https` and path `google.com/hi/`.

On Linux, the `f-relative-p` check passes since `google.com` is a valid relative path on unix filesystems. It is possible that this path is not valid on Windows, so the db build gets stuck forever in this function.

Not sure if this is the true cause, but I think it makes sense to me, will require some testing.

cc @nobiot @ahristow

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -335,12 +335,11 @@ function are expected to catch the error."
                  (t
                   (signal 'wrong-type-argument `((stringp numberp symbolp) ,item))))) items))))
 
-(defun org-roam--is-url-p (path)
+(defun org-roam--url-p (path)
   "Check if PATH is a URL.
 Assume the protocol is not present in PATH; e.g. URL `https://google.com' is
 passed as `//google.com'."
-
-  (s-starts-with-p "//" path))
+  (string-prefix-p "//" path))
 
 ;;;; File functions and predicates
 (defun org-roam--file-name-extension (filename)
@@ -629,7 +628,7 @@ it as FILE-PATH."
                            (org-ref-split-and-strip-string path))
                           ("fuzzy" (list path))
                           (_ (if (or (file-remote-p path)
-                                     (org-roam--is-url-p path))
+                                     (org-roam--url-p path))
                                  (list path)
                                (let ((file-maybe (file-truename
                                                   (expand-file-name path (file-name-directory file-path)))))
--- a/org-roam.el
+++ b/org-roam.el
@@ -335,6 +335,13 @@ function are expected to catch the error."
                  (t
                   (signal 'wrong-type-argument `((stringp numberp symbolp) ,item))))) items))))
 
+(defun org-roam--is-url-p (path)
+  "Check if PATH is a URL.
+Assume the protocol is not present in PATH; e.g. URL `https://google.com' is
+passed as `//google.com'."
+
+  (s-starts-with-p "//" path))
+
 ;;;; File functions and predicates
 (defun org-roam--file-name-extension (filename)
   "Return file name extension for FILENAME.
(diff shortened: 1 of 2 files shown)


nobiot:
@jethrokuan ty!

org-roam.el
```diff
@@ -622,7 +628,8 @@ it as FILE-PATH."
                            (setq type "cite")
                            (org-ref-split-and-strip-string path))
                           ("fuzzy" (list path))
-                          (_ (if (file-remote-p path)
+                          (_ (if (or (file-remote-p path)
+                                     (org-roam--is-url-p path))
```
pcrama: Suggestion: use something like `(and (string-match-p ffap-url-regexp type) (org-roam--is-url-p path))`

On windows, `\\network\share$\path\file.org` is a valid file name and emacs will replace `/` with `\` so that 
- https://google.com/ gets split into type="https" and path="//google.com/"
- type does not match any pattern and `(file-remote-p path)` returns nil (I just tried and `(file-remote-p (concat type ":" path))` is nil, too, so no help there)
- so the code ends up trying `(file-truename path)` and on my Windows systems (8.1 & 10), `(file-truename "//google.com/")` will take a very long time (probably a network timeout) because emacs tries to access "google.com" as if it was a file server

On Linux, the problem is (less ?) visible, because `//google.com/` is interpreted as `/google.com/` and unless you happen to have a `google.com` symbolic link in your root folder to somewhere else, nothing bad happens... but if there was such a symbolic link, a different path would get registered, I guess.

pcrama: My `org-roam--extract-links` would look like this:

```el
(defun org-roam--extract-links (&optional file-path)
  "Extracts all link items within the current buffer.
Link items are of the form:

    [from to type properties]

This is the format that emacsql expects when inserting into the database.
FILE-FROM is typically the buffer file path, but this may not exist, for example
in temp buffers.  In cases where this occurs, we do know the file path, and pass
it as FILE-PATH."
  (require 'org-ref nil t)
  (unless file-path
    (setq file-path (file-truename (buffer-file-name))))
  (save-excursion
    (let (links)
      (org-element-map (org-element-parse-buffer) 'link
        (lambda (link)
          (goto-char (org-element-property :begin link))
          (let* ((type (org-element-property :type link))
                 (path (org-element-property :path link))
                 (begin (or (org-element-property :content-begin element)
                            (org-element-property :begin element)))
                 (content (or (org-element-property :raw-value element)
                              (buffer-substring-no-properties
                               begin
                               (or (org-element-property :content-end element)
                                   (org-element-property :end element)))))
                 (content (string-trim content))
                 (content (org-roam--expand-links content file-path))
                 (properties (list :outline (mapcar (lambda (path)
                                                      (org-roam--expand-links path file-path))
                                                    (org-roam--get-outline-path))
                                   :content content
                                   :point begin))
                 (names (pcase type
                          ("id"
                           (list (car (org-roam-id-find path))))
                          ((pred (lambda (typ)
                                   (and (boundp 'org-ref-cite-types)
                                        (-contains? org-ref-cite-types typ))))
                           (setq type "cite")
                           (org-ref-split-and-strip-string path))
                          ((or "fuzzy"
                               ;; looked in ffap-url-regexp, but can't use
                               ;; that directly because it contains "file"
                               "https" "http" "ftp" "telnet"
                               "gopher" "www" "wais" "news"
                               "newspost")
                           (list path))
                          (_ (if (file-remote-p path)
                                 ;; avoid slowness for TRAMP paths like
                                 ;; "/ssh:user@host:path/to/file"
                                 (list path)
                               (let ((file-maybe (file-truename
                                                  (expand-file-name path (file-name-directory file-path)))))
                                 (if (f-exists? file-maybe)
                                     (list file-maybe)
                                   (list path))))))))
            (dolist (name names)
              (when name
                (push (vector file-path name type properties) links))))))
      links)))
```

pcrama: This still misses the fact that `file://network/share$/path` should be considered a remote path under Windows (what about WSL?)...

PR #1105, opened 2020-09-15
Move fuzzy links to roam: links

jethrokuan:


zaeph:
If I understand properly, the rationale behind this PR was to prevent bad interactions with other parts of org-mode.

I'm not sure if the `[[foo]]` syntax is a hill worth dying on, even though it seems to be a feature than many RoamResearch users crave.  I think this is something we'll need to rethink in the future.

Issue #1038, opened 2020-08-14
Building org-roam.db hangs on Windows


ahristow:
### Description
When I delete my org-roam.db or execute `org-roam-db-build-cache` under Windows 7, the process (and therefore Emacs) hangs.

#### Steps to Reproduce
1a. With Emacs closed, delete org-roam.db, the start Emacs.
1b. Alternatively, with Emacs open, issue `org-roam-db-build-cache`.

#### Backtrace

N/A
#### Expected Results
org-roam.db is built within a few seconds, maybe a minute (the file system on my Windows machine is rather slow).

#### Actual Results
An org-roam.db-journal file is created and, if org-roam.db was deleted, a new org-roam.db file is also created. The file system often, but not always, shows the journal file being updated after about 5 minutes. I have let the process run for as long as an hour before killing Emacs, but it never completes.

I do not observe this behavior on my Linux machine, which is running Emacs 26.3 under Ubuntu 20.04 LTS. On that machine, org-roam.db builds within a few seconds. I use the same `init.el` on both machines and the same package configuration, enforced through through the package freeze/thaw functionality of `straight.el`.

### Workaround
Downgrade org-roam to the last official release (org-roam commit: https://github.com/jethrokuan/org-roam/commit/89e9121f26cf03bb7820d257cd9d9106916cd1e9). With this version, org-roam.db successfully builds. I have not tested every commit since this one, so it may also work with other, later commits.

### Environment
- Emacs: GNU Emacs 26.3 (build 1, x86_64-w64-mingw32)
 of 2019-08-29
- Framework: N/A
- Org: Org mode version 9.3.7 (release_9.3.7-708-g5417e3 @ c:/Users/J0398597/AppData/Roaming/.emacs.d/straight/build/org/)
- Org-roam: 1.2.1

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/2081e1268ad57dfb94171d1a99e8f13126fbdaa5

beetleb:
I have the same problem. I did not try a prior version, though.

beetleb:
I tried cleaning out my files in the directory to see if I could find any pattern. Not 100% sure, but any file that has a http:// in it seems to slow it down (takes several seconds per such file).

It does eventually end, but it takes a long time. Several minutes to do 25 files that I have. Ouch.

jethrokuan:
> Not 100% sure, but any file that has a http:// in it seems to slow it down (takes several seconds per such file).

[...]

PR #1107, opened 2020-09-15
(fix): #1038 long db build when notes contain URLs with https://

nobiot:
See the detail in issue #1038 

With using [test files](https://github.com/nobiot/org-roam-performance-test-v1.2.1-Windows) of five (5) `.org` notes with `https` URL links, the following performance improvement is observed:

---

### Steps / code executed for benchmark

```
#+begin_src emacs-lisp
;;(org-roam-db--close-all)
;; delete the db file manually
(benchmark 1 '(org-roam-db-build-cache))
#+end_src
```

### Results
[2020-09-15 Tue]
Before change
Elapsed time: 123.473523s (0.506992s in 12 GCs)

After change
1st: Elapsed time: 0.253582s
2nd: Elapsed time: 0.240944s
3rd: Elapsed time: 0.568807s (0.043532s in 1 GCs)

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -532,8 +532,8 @@ PATH should be the root from which to compute the relativity."
       ;; Loop over links
       (while (re-search-forward org-roam--org-link-bracket-typed-re (point-max) t)
         (setq link (match-string 2))
-        (when (and (file-attributes link)
-                   (f-relative-p link))
+        (when (and (f-relative-p link)
+                   (file-attributes link))
           (save-excursion
             (goto-char (match-beginning 2))
             (delete-region (match-beginning 2)
--- a/org-roam.el
+++ b/org-roam.el
@@ -6,7 +6,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 1.2.1
-;; Package-Requires: ((emacs "26.1") (dash "2.13") (f "0.17.2") (s "1.12.0") (org "9.3") (emacsql "3.0.0") (emacsql-sqlite3 "1.0.0"))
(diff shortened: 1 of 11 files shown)


org-roam.el
```diff
@@ -532,8 +532,8 @@ PATH should be the root from which to compute the relativity."
       ;; Loop over links
       (while (re-search-forward org-roam--org-link-bracket-typed-re (point-max) t)
         (setq link (match-string 2))
-        (when (and (file-attributes link)
-                   (f-relative-p link))
+        (when (and (f-relative-p link)
+                   (file-attributes link))
```
jethrokuan, line 536: There's no need for the `file-attributes` check anymore, since we are no longer using it to determine whether the path is a URL.

org-roam.el
```diff
@@ -622,7 +622,8 @@ it as FILE-PATH."
                            (setq type "cite")
                            (org-ref-split-and-strip-string path))
                           ("fuzzy" (list path))
-                          (_ (if (file-remote-p path)
+                          (_ (if (or (file-remote-p path)
+                                     (s-starts-with-p "//" path))
```
jethrokuan, line 626: I think this is fine, but we should wrap it in a helper function ala `(org-roam--is-url-p path)` or add a comment, before we forget why we did this in the first place.

org-roam.el
nobiot: > we are no longer using it to determine whether the path is a URL.
Is this reflected in this fix/windows branch, or it's in the master?

Are you suggesting that we should simply remove the `and` condition like this?
`(when (f-relative-p link) ... [omitted] ... )`

jethrokuan: > Are you suggesting that we should simply remove the and condition like this?
(when (f-relative-p link) ... [omitted] ... )

yes. I added it in the `fix/windows` branch, because I thought it would be a fast way to not do processing on non file links.

PR #1109, opened 2020-09-16
Remove file-truename everywhere

kisaragi-hiu:
##### Motivation for this change

`file-truename` calls alone accounts for over 20% of CPU samples in `org-roam-db-build-cache`. It's expensive and unnecessary. To get an absolute path, `expand-file-name` alone is enough.

#1099 fixes org-roam-buffer by making it consistent with the DB. This PR should still keep them consistent while speeding up `org-roam-db-build-cache` by 40% (110s → 70s on my setup).

Org-roam shouldn't care if a file is a symlink or located in a symlinked directory, and therefore shouldn't be resolving symlinks. There is no point in canonicalizing everything: each Org-roam notes repository already has its own org-roam.db.

This can probably be considered an extended version of #928.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,12 @@
 
 In this release we support fuzzy links of the form `[[Title]]`, `[[*Headline]]` and `[[Title*Headline]]`. Completion for these fuzzy links is supported via `completion-at-point`.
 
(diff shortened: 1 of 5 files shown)


jethrokuan:
I like the direction of the change here, but there are a couple of issues to resolve now that we don't resolve symlinks.

If one symlinks a file within the Org-roam directory to another file within an Org-roam directory, a conflict will arise. Both files will be processed by Org-roam, as if they were different files. If the file contains an ID, it will also raise an error.

PR #1110, opened 2020-09-17
(fix): perform link expansion only on file links

jethrokuan:
This prevents ID links from being expanded in the Org-roam buffer.


jethrokuan:
I'm aware that this is a bit restrictive, later on we might need to register link types provided by external packages like `org-brain`.

PR #1114, opened 2020-09-18
improvement: skip check filepath when the path is remote style for wh…

carlos-wong:
…en network is not work or serve

It will cause tramp erro signal when update link with the filepath which remote server is down

Skip the remote path link update, because the server will be down sometimes. In my situation I closed some azure server, but the org file have link to the server which is closed

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1541,8 +1541,8 @@ update with NEW-DESC."
       (when-let ((link (org-element-lineage (org-element-context) '(link) t)))
         (let ((type (org-element-property :type link))
               (path (org-element-property :path link)))
-          (when (and (not (tramp-handle-file-remote-p path))
-                     (string-equal (file-truename path) old-path)a
+          (when (and (not (file-remote-p path))
+                     (string-equal (file-truename path) old-path)
                      (org-in-regexp org-link-bracket-re 1))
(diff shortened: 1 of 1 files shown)


carlos-wong:
@jethrokuan Hi, If there is something wrong pls let me know. I like this software and want to contribute to it.

org-roam.el
```diff
@@ -1541,7 +1541,8 @@ update with NEW-DESC."
       (when-let ((link (org-element-lineage (org-element-context) '(link) t)))
         (let ((type (org-element-property :type link))
               (path (org-element-property :path link)))
-          (when (and (string-equal (file-truename path) old-path)
+          (when (and (not (tramp-handle-file-remote-p path))
```
jethrokuan, line 1544: use `(file-remote-p path)`

org-roam.el
```diff
@@ -1541,7 +1541,8 @@ update with NEW-DESC."
...
+                     (string-equal (file-truename path) old-path)a
```
jethrokuan, line 1545: remove straggling a.

jethrokuan:
this change has been made unnecessary by the removal of `file-truename` throughout Org-roam in #1109

PR #1117, opened 2020-09-19
(ci): make CI pass on non-blocking errors

jethrokuan:


alphapapa:
@jethrokuan If you're interested, I think we could add a `warn` function that could be used instead of `error` and give similar output at the end of the script.  Of course, I do want you to be able to customize `makem.sh` according to your needs, but if we can minimize divergence, it will be easier for you to merge improvements later.

jethrokuan:
> If you're interested, I think we could add a warn function that could be used instead of error and give similar output at the end of the script.

Yeah, that's exactly what was needed here. Happy to use whatever makem implements upstream.

PR #1122, opened 2020-09-21
feat: extract idle timer interval as a customizable value

carlos-wong:
My org-roam buffer list is very long. The number is about 200 files, and in my workflow I aways refile heading. So when the auto save trigger the push queue function will stuck. And then the idle timer will trigger in 2 secs. So In may situation I like to customize this idle timer interval to bigger value

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -113,7 +113,7 @@ If nil, `find-file' is used."
   :type 'function
   :group 'org-roam)
 
-(defcustom org-roam-update-db-idletimer-by-seconds 2
+(defcustom org-roam-update-db-idle-seconds 2
   "Number of idle seconds before triggering an Org-roam database update."
   :type 'integer
   :group 'org-roam)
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -113,6 +113,11 @@ If nil, `find-file' is used."
   :type 'function
   :group 'org-roam)
 
+(defcustom org-roam-update-db-idle-delay 2
```
jethrokuan, line 116: rename to `org-roam-update-db-idle-seconds`, to make it more explicit that it is the number of seconds of idle time.

org-roam.el
```diff
@@ -113,6 +113,11 @@ If nil, `find-file' is used."
...
+  "The idle timer when org-roam update db index."
```
jethrokuan, line 117:
```suggestion
  "Number of idle seconds before triggering an Org-roam database update.
```

PR #1128, opened 2020-09-22
(fix) Add org-roam--completion-exit-fn for post-completion of [[Title]] to move cursor out of the brackets

nobiot:
Currently Org-roam's completion-at-point implements :exit-fn as
a lambda function. For some reason, it is not triggered at post-
completion; however, it has been observed that a named function is
triggered as expected. 

This commit implements a function, and
replaces the assignment of :exit-function with the newly
created function. This is intended to get the cursor (point)
out of the [[Title]] bracket for more intuitive writing experience.

At the moment, when you type [[Title]], accepts a company completion, the cursor remains within the square brackets, like this:

`[[Title|]]`
`-------^ here`

---

I realise this implementation is crude, and probably does not address the "tag" completion.
Please take this PR as a suggestion. 
It works...
company-fix

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1146,6 +1146,13 @@ This function hooks into `org-open-at-point' via
   :group 'org-roam
   :type 'boolean)
 
+(defun org-roam--completion-exit-fn (&rest _)
+  "After accepting complete-at-point, move out of `]]`.
+This is intended to be set to :exit-function within
+`org-roam-complete-at-point` function."
+  
+  (forward-char 2))
+
 (defun org-roam-complete-at-point ()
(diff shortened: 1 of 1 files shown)


nobiot:
If you like the idea, I'll be happy to refine the implementation 😄

jethrokuan:
Sometimes when I complete the title, I want to add a `*` to complete headlines right after. That's why I left it like that. Could definitely be a configurable option, but do wait for #1105 , the completion-at-point functions have been separated out.

Nothing is blocking that PR, I'm just using it now to see if it has any bugs.

nobiot:
#1133 supersedes this PR. Closing.

PR #1133, opened 2020-09-24
(feat): enable capt for [[|]]; add exit-fn

nobiot:
Please take this as a gentle inviting for a discussion with a concrete implementation.

The intention of this commit / PR is to suggest an alternative to the current
capf in org-roam-link.el. Currently, you need to type [[roam:]] for capf to
work for the link completion-at-point. This alternative enables capf to be
triggered without roam: for [[|]]. roam: is added automatically with
exit-fn (implemented as a lambda function as originally intended).

The keycord for `[[roam:]]` is not necessarily fluid for writing. I assumed this was a workaround, and ideally the way this alternative proposes is an originally intended behavior for completion-at-point for links with titles.
Up for discussion of course.

My initial testing shows that it works as intended.
In this animation, capf is not manually triggered (it's automatic).
org-roam-link-alt

### DIFF
--- a/org-roam-link.el
+++ b/org-roam-link.el
@@ -236,27 +236,28 @@ DESC is the link description."
   "Do appropriate completion for the link at point."
   (let ((end (point))
         (start (point))
-        (exit-fn (lambda (&rest _) nil))
+        (exit-fn (lambda (str &rest _)
+                   (delete-char (- (length str)))
+                   (insert "roam:" str)))
         collection)
     (when (org-in-regexp org-link-bracket-re 1)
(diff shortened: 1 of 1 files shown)


nobiot:
I just found out that completion for 'title+headlines' does not work with my suggestion (ouch 😓).  Nevertheless, would the Org-roam project approve of the direction: not needing the "roam:" prefix for capf?

jethrokuan:
I approve of this direction, and find it silly i never thought of this. I'll make a PR that allows for completion of links, but still support the `[[roam:|]]` completion.

PR #1136, opened 2020-09-25
how to install from git repository

plantarum:
Fixing an open TODO in the manual.


jethrokuan:
Lovely, thanks!

jethrokuan:
fyi regarding generating the texi file, editing the Org file should re-generate the texi file, there's a local file hook in the Org file that would run texinfo.

Issue #1129, opened 2020-09-22
org roam opens extra files alongside file


natask:
### Description
when opening org roam files with org roam enabled, other org roam files are opened with it.

Hard to debug because it doesn't happen to all files and the files it opens alongside change over time. With that said, I have noticed that opening files with many file links tends to activate this bug.

I tracked it down to the following command in `org-roam-mode`. https://github.com/org-roam/org-roam/blob/cd87cfdd58f402f687306bad082d1def3a1e3636/org-roam.el#L1730
 
I looked at this more. And I think it happens when opening files that have id links to other org roam files headlines.
#### Steps to Reproduce
1. Open a org roam file with many file links

#### Expected Results
It is the only file that opens.

#### Actual Results
Other files open with it.

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.22, cairo version 1.17.3)
 of 2020-08-20
- Framework: Spacemacs
- Org: Org mode version 9.4 (9.4-elpaplus @ ~/spacemacs/elpa/develop/org-plus-contrib-20200914/)
- Org-roam: 1.2.1

natask:
I suspect it is arsing from  (org-roam-id-find id) at https://github.com/org-roam/org-roam/blob/cd87cfdd58f402f687306bad082d1def3a1e3636/org-roam.el#L1481 and https://github.com/org-roam/org-roam/blob/cd87cfdd58f402f687306bad082d1def3a1e3636/org-roam.el#L1487.

natask:
the following fixes it up I think. Will try it out for a while. I think what is going on is that the `org-id-find-id-in-file` opens the file to find id and doesn't close it for some reason. closing a newly created buffer pointing to the file if it exists and if it indeed newly created by `org-id-find-id-in-file` seems to solve the problem. 
```
(defun org-roam-id-find (id &optional markerp strict)
  "Return the location of the entry with the id ID.
When MARKERP is non-nil, return a marker pointing to theheadline.
Otherwise, return a cons formatted as \(file . pos).
When STRICT is non-nil, only consider Org-roam’s database."
  (let ((file (or (caar (org-roam-db-query [:select [file]
                                                    :from headlines
                                                    :where (= id $s1)]
                                           id))
                  (unless strict
                    (org-id-find-id-file id)))))

    (when file ;; not needed because  org-id-find-id-in-file does its own checking but for efficiency sake
[...]

PR #1146, opened 2020-09-27
(fix): org-roam leaves files open when it shouldnt

natask:
`org-roam-db--update-file` and `org-roam-roam-find-file` open unvisited files
and leave them open while in most cases, unvisited files should be closed back
again.

this commit fixes that. it extends work in #1129, #1131, #1140.

### DIFF
--- a/org-roam-macs.el
+++ b/org-roam-macs.el
@@ -46,6 +46,17 @@
         (nconc new-lst (list separator it)))
       new-lst)))
 
+(defmacro org-roam--with-file (file &rest body)
+  "Execute BODY within a FILE.
+Closes the file if the file is not yet visited."
+  (declare (indent 1) (debug t))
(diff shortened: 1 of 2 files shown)


natask:
@jethrokuan #1131, #1140 broke `org-roam-db--update-file` update in  #1129. 

https://github.com/org-roam/org-roam/blob/be64f107e9d24cb457a4fb5d92e8aed26b3d06e7/org-roam.el#L1099 
logic here should be some form of unless.  

this pr is stale now that we are using with-temp-buffer instead of find-file-noselect. I had added more commits to merge the two but I think It would be better to rebase.

jethrokuan:
the changes here have no effective change right?

jethrokuan:
Superceded by #1154

PR #1147, opened 2020-09-27
(perf): disable org-agenda while building the cache

juergenhoetzel:
`org-mode` builds the `org-agenda-files` menu entry for every file
visited. This functionality isn't required and useful when building
the cache in temporary org buffers.

Performance improvement. Master:
```
ELISP> (benchmark 1 '(org-roam-db-build-cache 1))
"Elapsed time: 45.214780s (11.347281s in 219 GCs)"
```
This branch:
```
ELISP> (benchmark 1 '(org-roam-db-build-cache 1))
"Elapsed time: 10.499171s (2.793775s in 54 GCs)"
```
I have all my org-roam files (550 currently) in my agenda-files:

`(add-to-list 'org-agenda-files "~/shared/org/roam"))`

I was surprised that org makes 550 menu entries for every visited file, even if the buffer is only in temporary use (while building the cache).

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -478,6 +478,7 @@ If FORCE, force a rebuild of the cache from scratch."
   (org-roam-db--close) ;; Force a reconnect
   (org-roam-db) ;; To initialize the database, no-op if already initialized
   (let* ((gc-cons-threshold org-roam-db-gc-threshold)
+	 (org-agenda-files nil)
          (org-roam-files (org-roam--list-all-files))
          (current-files (org-roam-db--get-current-files))
          (file-count 0)


TimQuelch:
Wow nice catch

natask:
the core issue is that org-roam opens files using `find-file-noselect` when updating db, enabling whatever else the user has configured to run when files are open sans what org-roam ends up disabling.
 we should move towards a whitelist instead of a blacklist of executed commands by replacing `find-file-noselect` with `with-temp-buffer` + `insert-file-contents`. Then enabling functions to run as needed will alleviate future issues.

jethrokuan:
`org-roam-db-build-cache` has always read the file contents into a temp buffer, not sure why it's affected by `org-agenda`, so I'd like to do a deeper investigation before merging this.

juergenhoetzel:
> `org-roam-db-build-cache` has always read the file contents into a temp buffer, not sure why it's affected by `org-agenda`, so I'd like to do a deeper investigation before merging this.

`org-mode` itself triggers the building of `agenda-files-menu`:

```
(define-derived-mode org-mode outline-mode "Org"
...
  (org-load-modules-maybe)
  (org-install-agenda-files-menu)
...

```

natask:
you are right. my bad for not checking closely.
this is why it is affected (org-agenda). 
https://github.com/org-roam/org-roam/blob/be64f107e9d24cb457a4fb5d92e8aed26b3d06e7/org-roam-macs.el#L60-L70 in function https://github.com/org-roam/org-roam/blob/be64f107e9d24cb457a4fb5d92e8aed26b3d06e7/org-roam-macs.el#L49-L64

jethrokuan:
The correct solution is to design a version of `(org-mode)` that only enables the things that are needed (setting regexps, outline-level etc.).

I'll accept this as a hotfix, though.

natask:
you might want to propagate the hot-fix to `org-roam--with-temp-buffer` so that other usages of the function outside `org-roam-db-build-cache` can benefit from it as well.

PR #1150, opened 2020-09-28
(internal): replace find-file-noselect with with-temp-buffer where possible

jethrokuan:
`find-file-noselect` is affected by the user's init.el, and there is
little control over what is being activated. For database updates, read
access is all we need, so we use `with-temp-file`.


natask:
for maintainability sake, it is important to say where `org-roam-file-name` comes from. lets add to docs that `org-roam-file-name` is set by `org-roam-with-temp-buffer` within these functions.

jethrokuan:
It's already in the variable's docstring.

Issue #1124, opened 2020-09-21
following backlink entries that  only contain headlines with org links


natask:
### Description
when a backlink points to a line that  contains only an org link, it is not possible to follow the back-link to position in file.

if the org-link is on a headline, it is not possible to open the headline.

simple solution is to append a string to the backlink buffer entry as I have done in https://github.com/natask/org-roam/commit/7a2de12f1b8a58ee62f6e12db4acd70ad00edc90. Then clicking on the string with follow to position in file. 

Another solution is to change **org-roam-open-at-point** so that links follow to file where they are linked not the link they contain.

#### Steps to Reproduce
1. create two org roam files.
2. link one to the other (without adding any extra text to that line and as a headline) and run **org-roam-db-build-cache**,
example:
``` org
#+title: file1

```
``` org
#+title: file2
[[file:file1.org][this link can't be followed to line although pressing on top in backlink buffer will bring up the linking file]]
*[[file:file1.org][this link can't be followed at all]]
```

3. try to follow backlink entry, (in the example, link in file2), through org-roam backlink buffer.

#### Expected Results
follow backlink entry to position in linking file.

#### Actual Results
follows link, which opens location in current file. This isn't the desired behavior.

### Environment

Copy info below this line into issue:
- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.22, cairo version 1.17.3)
 of 2020-08-20
- Framework: Spacemacs
- Org: Org mode version 9.4 (9.4-elpaplus @ ~/spacemacs/elpa/develop/org-plus-contrib-20200914/)
- Org-roam: 1.2.1

natask:
org-roam-buffer-preview-function can do this
add > to org-roam-buffer-preview-function
either modify the function under the variable  org-roam-buffer-preview-function or do the following

(advice-add  org-roam-buffer-preview-function :filter-return (lambda (content) (concat "> " content)))

PR #1152, opened 2020-09-29
(fix): insert > to follow link only Backlinks.

natask:
Simple solution for #1124. Inserts => so that backlinks that don't contain any other text in line can be followed.


natask:
superseded by `org-roam-buffer-preview-function`
add > to org-roam-buffewr-preview-function
``` elisp
(advice-add  'org-roam-buffer-preview-function :filter-return (lambda (content) (concat "> " content)) 
```

PR #1154, opened 2020-09-30
(fix): fix org-roam-id-find buffer-kill logic

jethrokuan:
same as #1146 without introducing a new macro


Whil- (commented):
This here is probably the reason for #1209 

`existing-buf` is set by function `(find-buffer-visiting file)` and the only way to pass the when-condition starting at line 1099 is if `existing-buf` is `nil`. Which means `(find-buffer-visiting file)` will be (again...) `nil` and consequently `kill-buffer` will try to kill the current buffer, which in the context of #1209 is the org-roam-buffer.

PR #1155, opened 2020-09-30
(feat): remove IDs created during capture

jethrokuan:
Previously we had overwritten the Org's store link function for files,
to create IDs if the point was under the headline. This has several
issues:

1. It becomes impossible to store a link to the file using `org-store-link`
2. This override is global across Org, although we had a guard in
`org-roam-store-link`
3. IDs are created during an org-capture, because org-capture calls `org-store-link`

This is unnecessary. Org provides `org-id-store-link` instead for this
purpose. Instead, we advice `org-id-new`, to queue the file for a DB
update if an ID was created.

Also, setting `org-id-link-to-org-use-id` to `t` restores this behaviour.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -456,18 +456,21 @@ If the file does not exist anymore, remove it from the cache.
 If the file exists, update the cache with information."
   (setq file-path (or file-path
                       (buffer-file-name (buffer-base-buffer))))
-  (cond ((not (file-exists-p file-path))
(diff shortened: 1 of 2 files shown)


rtrppl:
Sorry to open this again but if I use IDs the backlinks to headings do not show up (it works fine without using `ord-id`). I've added this to my .emacs:

`(setq org-id-link-to-org-use-id 'create-if-interactive-and-no-custom-id)`

PR #1156, opened 2020-09-30
Use relative path in side buffer

kisaragi-hiu:

This allows faster copy-and-paste into one's notes to create an explicit list note.

Before:

Screenshot_20200930_154459

After:

Screenshot_20200930_154257


jethrokuan:
This is superceded by #1157 , the behaviour you want should be the default.

Issue #1172, opened 2020-10-08
Duplicate ID's when rebuilding db-cache


Whil-:
### Description
I get duplicate ID's when rebuilding db-cache with files where the first line also is a heading with an ID property.

Ex:
```
* Heading 1
:PROPERTIES:
:ID: smth
:END:
```
Will signal for duplicate ID's, while the following will not:
```
(empty line)
* Heading 1
:PROPERTIES:
:ID: smth
:END:
```

Seems `org-roam--extract-ids` is the culprit. Suggestion for fix on it's way...

PR #1173, opened 2020-10-08
(fix): Fix org-roam--extract-ids at outline level 0

Whil-:
* org-roam.el (org-roam--extract-ids): Only check for ID if first
character in file is at outline level 0. #1172

Issue #1172


Whil-:
Hmm, didn't test properly... New pull request coming in a moment..

zaeph:
Could you not reset this PR's branch?

Whil-:
I might have failed a bit here.. Instead of fixing this one I closed it and opened a new one, #1174. How that will work out.

PR #1174, opened 2020-10-08
(fix): Fix org-roam--extract-ids at outline level 0

Whil-:
* org-roam.el (org-roam--extract-ids): Only check for ID if first
character in file is at outline level 0. #1172

#1172

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -623,8 +623,9 @@ If FILE-PATH is nil, use the current file."
   (let (result)
       ;; We need to handle the special case of the file property drawer (at outline level 0)
       (org-with-point-at (point-min)
-        (when-let ((id (org-entry-get nil "ID")))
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -623,8 +623,9 @@ If FILE-PATH is nil, use the current file."
   (let (result)
       ;; We need to handle the special case of the file property drawer (at outline level 0)
       (org-with-point-at (point-min)
-        (when-let ((id (org-entry-get nil "ID")))
-           (push (vector id file-path (org-outline-level)) result)))
+        (when-let ((id (org-entry-get nil "ID"))
+                   (before-first-heading (= 0 (org-outline-level))))
```
jethrokuan, line 627: Might want to swap the order of checks, I think the outline level check is cheaper.

Whil-: True that, updated.

PR #1175, opened 2020-10-08
(fix): Make org-roam--extract-links less likely to throw error

Whil-:
* org-roam.el (org-roam--extract-links): Remove risk of
wrong-type-argument error in call to buffer-substring-no-properties.

I ran in to issues on my machine where `before` parameter was set to nil and `buffer-substring-no-properties` threw an error. The fixup in this pull request removes the risk of such errors when `element` doesn't have `:content-end` or `:end` properties (or where `element` by any reason is `nil` itself). While that in itself may warrant an investigation, this change cleans up the function a tad and adds a few guardrails to at least prevent this type of error in the future.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -588,12 +588,12 @@ it as FILE-PATH."
                  (element (org-element-at-point))
                  (begin (or (org-element-property :content-begin element)
                             (org-element-property :begin element)))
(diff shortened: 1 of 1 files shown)


jethrokuan:
Superceded by #1179, thanks!

Issue #1074, opened 2020-08-28
wrong-type-argument integer-or-marker-p nil


bbigras:
### Description

One weird thing is that I only seems to have this error when using `envrc`.

#### Steps to Reproduce

Not sure yet. I'll try to get a minimal example if it's not a trivial fix.

#### Backtrace

```
Debugger entered--Lisp error: (wrong-type-argument integer-or-marker-p nil)
  buffer-substring-no-properties(nil nil)
  #f(compiled-function (link) #<bytecode -0x1034ece9f6ae52bc>)((link (:type "https" :path "//www.sciencedaily.com/releases/2019/07/1907191735..." :format bracket :raw-link "https://www.sciencedaily.com/releases/2019/07/1907..." :application nil :search-option nil :begin 2 :end 127 :contents-begin 68 :contents-end 125 :post-blank 0 :parent (paragraph (:begin 2 :end 128 :contents-begin 2 :contents-end 128 :post-blank 0 :post-affiliated 2 :parent (section (:begin 2 :end 128 :contents-begin 2 :contents-end 128 :post-blank 0 :post-affiliated 2 :parent (org-data nil #43)) #27)) #1 #("\n" 0 1 (:parent #27)))) #("Take a bath 90 minutes before bedtime to get bette..." 0 57 (:parent #1))))
  #f(compiled-function (--data) #<bytecode -0xd4d286dd6ac900a>)((link (:type "https" :path "//www.sciencedaily.com/releases/2019/07/1907191735..." :format bracket :raw-link "https://www.sciencedaily.com/releases/2019/07/1907..." :application nil :search-option nil :begin 2 :end 127 :contents-begin 68 :contents-end 125 :post-blank 0 :parent (paragraph (:begin 2 :end 128 :contents-begin 2 :contents-end 128 :post-blank 0 :post-affiliated 2 :parent (section (:begin 2 :end 128 :contents-begin 2 :contents-end 128 :post-blank 0 :post-affiliated 2 :parent (org-data nil #43)) #27)) #1 #("\n" 0 1 (:parent #27)))) #("Take a bath 90 minutes before bedtime to get bette..." 0 57 (:parent #1))))
[...]

PR #1176, opened 2020-10-09
(fix): org-roam--extract-links to handle content boundaries (#1074)

d12frosted:

Explained in https://github.com/org-roam/org-roam/issues/1074#issuecomment-706055580

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.2.3 (TBD)
+
+## Bugfixes
+
+- [#1074](https://github.com/org-roam/org-roam/issues/1074) fix `org-roam--extract-links` to handle content boundaries
+
 ## 1.2.2 (06-10-2020)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.2.3 (TBD)
```
d12frosted, line 3: Not sure which version to pick 😸

org-roam.el
```diff
@@ -586,12 +586,12 @@ it as FILE-PATH."
           (let* ((type (org-roam--collate-types (org-element-property :type link)))
                  (path (org-element-property :path link))
                  (element (org-element-at-point))
```
d12frosted, line 588: as I explained in comment, it seems that `element` might be `nil`, though I am not sure if it's valid case or not. So in my local copy I monkey-patched this function by changing `let*` to `when-let*`. WDYT, should I include that change here as well?

jethrokuan:
Superceded by #1179, thanks!

PR #1179, opened 2020-10-10
(fix): fix org-roam--extract-links

jethrokuan:
Fix `org-roam--extract-links` erroring at content boundaries. Cherry-picked commits from both @d12frosted and @Whil-'s PR.


org-roam.el
```diff
@@ -588,12 +588,11 @@ it as FILE-PATH."
                  (element (org-element-at-point))
                  (begin (or (org-element-property :contents-begin element)
                             (org-element-property :begin element)))
+                 (end (or (org-element-property :content-end element)
```
Whil-, line 591: Should it be `:contents-end` instead of `:content-end` here maybe?

Issue #1149, opened 2020-09-28
There should be a command to add a link description to the destination's #+ROAM_ALIAS


spacekitteh:
### Brief Abstract
In order to more easily populate the `#+ROAM_ALIAS` field, there should be a command to add a link's description to the destination's alias field.

### Long Description
- Point should be in the link to an `org-roam` note
- The command should should copy the link's description
- If there is no `#+ROAM_ALIAS:` for the destination, create one. Note that for `id` links, this means there should be a `:ROAM_ALIAS:` property, treated just like the `#+ROAM_ALIAS` file header
- Add the copied link description to the destination's `ROAM_ALIAS` field, adding quotes if necessary

This should _not_ bring up the destination's buffer.

### Please check the following:

- [x] No similar feature requests

d12frosted:
🤔 I've spent some time trying to better understand your use case, so let me share my thoughts and please let me know if I missed the point 😸 

From my experience, I need an alias in two cases - when (1) I want to open a note or when (2) I link a note. Both of them rely on searching for a note, but use cases are different.

(1) So when I just want to open a note and I realise that it requires an alias, I open the file by original name and add an alias using `+org-notes-alias-add` (and `+org-notes-alias-delete` for removal). You can find their definition in this [gist][1] (it originally describes how to set tags). It doesn't support `:ROAM_ALIAS:` property for headers, and this is something I will add.

(2) When I want to link a note, if I realise that the note requires an alias, I usually stop linking process, open a note and the task boils down to the (1) scenario. But here comes your interesting proposal - to link using an old name and be able to add an alias for the linked note without opening it! It actually might be useful in other scenarios - even for long existing links. 

@spacekitteh Is that right? :) 

---

That being said, @jethrokuan would you be interested to get a PR with functions to add/remove roam tags/aliases? I believe it might be useful by itself and possibly for resolving this issue.

  [1]: https://gist.github.com/d12frosted/4a55f3d072a813159c1d7b31c21bac9a#gistcomment-3483253

spacekitteh:
That's pretty much it. For example, I have a lot of "definition" notes, and I like to add aliases for, e.g. plural forms of the title, the initial letter capitalised, adjective forms, synonyms, etc, so that links are suggested in more cases. A concrete example:

```org
#+TITLE: Topological solitons are solutions to a field theory homotopically distinct from the vacuum solution
#+FILETAGS: :definition-note:topology:
[...]

PR #1183, opened 2020-10-11
(feat): interactive functions for managing aliases and tags

d12frosted:

See my comment in https://github.com/org-roam/org-roam/issues/1149#issuecomment-706073127

See it in action:

org-roam-alias-tag-test

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -870,6 +870,26 @@ whose title is 'Index'."
         (concat (expand-file-name org-roam-directory) path)
       index)))
 
+;;;; dealing with file-wide properties
+(defun org-roam--set-global-prop (name value)
+  "Set a file property called NAME to VALUE.
+
+If the property is already set, it's value is replaced."
+  (save-excursion
+    (widen)
+    (goto-char (point-min))
+    (if (re-search-forward (concat "^#\\+" name ": \\(.*\\)") (point-max) t)
+        (replace-match (concat "#+" name ": " value) 'fixedcase)
+      (while (and (not (eobp))
+                  (looking-at "^[#:]"))
+        (if (save-excursion (end-of-line) (eobp))
+            (progn
+              (end-of-line)
+              (insert "\n"))
+          (forward-line)
+          (beginning-of-line)))
+      (insert "#+" name ": " value "\n"))))
+
 ;;;; org-roam-find-ref
 (defun org-roam--get-ref-path-completions (&optional arg filter)
   "Return an alist of refs to absolute path of Org-roam files.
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -870,6 +870,26 @@ whose title is 'Index'."
         (concat (expand-file-name org-roam-directory) path)
       index)))
 
+;;;; dealing with file-wide properties
+(defun org-roam--set-global-prop (name value)
```
d12frosted, line 874: I actually have an idea how we can remove this function in the future. Since 9.4 is out, and files also have properties, we can gradually migrate to storing `ROAM_xxx` as properties (aka `:ROAM_xxx:`) instead of 'settings' (aka `#+ROAM_xxx`).

org-roam.el
```diff
@@ -870,6 +870,26 @@ whose title is 'Index'."
...
+  "Set a file property called NAME to VALUE.
+
+If the property is already set, it's value is replaced."
+  (save-excursion
+    (widen)
+    (goto-char (point-min))
+    (if (re-search-forward (concat "^#\\+" name ": \\(.*\\)") (point-max) t)
```
d12frosted, line 881: It doesn't remove the whole property if the value is empty, which is totally fine in my opinion. Though if you wish, I can change the behaviour.

org-roam.el
```diff
@@ -1680,6 +1700,72 @@ command will offer you to create one."
       (when (y-or-n-p "Index file does not exist.  Would you like to create it? ")
         (org-roam-find-file "Index")))))
 
+;;;###autoload
+(defun org-roam-alias-add ()
```
d12frosted, line 1704: Not sure about naming of these functions, so if you have a better names in mind, please let me know 😸

org-roam.el
```diff
@@ -870,6 +870,26 @@ whose title is 'Index'."
...
+;;;; dealing with file-wide properties
+(defun org-roam--set-global-prop (name value)
+  "Set a file property called NAME to VALUE.
+
+If the property is already set, it's value is replaced."
+  (save-excursion
+    (widen)
+    (goto-char (point-min))
+    (if (re-search-forward (concat "^#\\+" name ": \\(.*\\)") (point-max) t)
+        (replace-match (concat "#+" name ": " value) t)
```
d12frosted: Fixed case here is required to avoid unwanted capitalisations.

```
#+ROAM_ALIAS: TBD
```

If you add an alias like 'lower', without fixed case it will be capitalised 😸

org-roam.el
```diff
@@ -1680,6 +1700,72 @@ command will offer you to create one."
...
+  "Delete an alias from Org-roam file."
+  (interactive)
+  (unless org-roam-mode (org-roam-mode))
+  (if-let ((aliases (org-roam--extract-titles-alias)))
+      (let ((alias (completing-read "Alias: " aliases nil 'require-match)))
+        (org-roam--set-global-prop
+         "ROAM_ALIAS"
+         (combine-and-quote-strings (delete alias aliases)))
+        (org-roam-db--update-file (buffer-file-name (buffer-base-buffer))))
+    (user-error "No aliases to delete")))
```
d12frosted, line 1732: Here I just want to avoid empty completion list. Plus `required-match`...

org-roam.el
```diff
@@ -1680,6 +1700,72 @@ command will offer you to create one."
...
+
+(defun org-roam-tag-add ()
+  "Add a tag to Org-roam file.
+
+Return added tag."
+  (interactive)
+  (unless org-roam-mode (org-roam-mode))
+  (let* ((all-tags
```
d12frosted, line 1740: Here I just want to provide completion from existing tags.

org-roam.el
```diff
@@ -870,6 +870,26 @@ whose title is 'Index'."
...
+;;;; dealing with file-wide properties
+(defun org-roam--set-global-prop (name value)
+  "Set a file property called NAME to VALUE.
+
+If the property is already set, it's value is replaced."
+  (save-excursion
+    (widen)
+    (goto-char (point-min))
+    (if (re-search-forward (concat "^#\\+" name ": \\(.*\\)") (point-max) t)
+        (replace-match (concat "#+" name ": " value) 'fixedcase)
```
d12frosted, line 882: Fixed case is required to avoid unwanted capitalisations.

```
#+ROAM_ALIAS: TBD
```

In this case if you add an alias like 'lower' without fixed case, it will be capitalised.

org-roam.el
jethrokuan: Yeah, this is fine, I think we can file this as a feature request (supporting fetching from property drawers).

org-roam.el
jethrokuan: :ok_hand:

org-roam.el
jethrokuan: :ok_hand:

org-roam.el
```diff
@@ -870,6 +870,26 @@ whose title is 'Index'."
...
+      ;; find the first line that doesn't begin with ':' or '#'
+      (let ((found))
+        (while (not (or found (eobp)))
+          (beginning-of-line)
+          (if (or (looking-at "^#")
+                  (looking-at "^:"))
+              (line-move 1 t)
+            (setq found t)))
```
jethrokuan, line 890:
```suggestion
(while (and (not (eobp))
            (looking-at "^[#:]"))
  (forward-line)
  (beginning-of-line))
```

org-roam.el
jethrokuan: Names are fine, I think.

org-roam.el
```diff
@@ -1680,6 +1700,72 @@ command will offer you to create one."
...
+(defun org-roam-alias-delete ()
+  "Delete an alias from Org-roam file."
+  (interactive)
+  (unless org-roam-mode (org-roam-mode))
+  (if-let ((aliases (org-roam--extract-titles-alias)))
+      (let ((alias (completing-read "Alias: " aliases nil 'require-match)))
+        (org-roam--set-global-prop
+         "ROAM_ALIAS"
+         (combine-and-quote-strings (delete alias aliases)))
+        (org-roam-db--update-file (buffer-file-name (buffer-base-buffer))))
```
jethrokuan, line 1731: you can leave out the argument, it defaults to the current buffer.

org-roam.el
jethrokuan: `(org-roam-db--get-tags)` gives you what you want, and also applies the `distinct` op via SQL, which should be faster.

org-roam.el
jethrokuan: Just for explain this diff:

1. `line-move` runs a lot of things, and depends on some user config options (like whether to create new lines). `forward-line` is the simplest way to go one line down

2. We need to explicitly check the boundary condition so we know when to create a new line, since `forward-line` does not error

org-roam.el
d12frosted: Oh wow, didn't know about this. Really works great and simplifies implementation 🙇

org-roam.el
d12frosted: Makes sense! Thank you.

d12frosted:
Updated PR with suggestions @jethrokuan :)

jethrokuan (approved):

TimQuelch:
I've noticed these functions use uppercase ROAM_TAGS and ROAM_ALIAS. Should these instead be lowercase to be more consistent?

I've got a branch on my fork where this is the case, I can submit a quick PR if you like

d12frosted:
Oh wow, didn't know that the new convention is to use lowercase. Will send PR to follow this convention in new functions.

PR #1189, opened 2020-10-14
(fix): allow title changes that don't modify filename

herbertjones:
- Prevent error when trying to move file to same name
- Prevent string matching in org from messing with the current re-search-forward
  position which has caused infinite broken search and replace in testing (might
  not affect all org versions or may be due to certain code paths)

I tried to change the title of a page from "Webassembly" to "WebAssembly", which caused an error due to renaming the file to the existing name.  After fixing that, auto-updating links began corrupting files in this case.  A file with a link to "Webassembly" began infinitely inserting (broken) links to the "webassembly.org" file at the beginning of the file, as something was evidently changing the match position and causing `replace-match` in `org-roam--replace-link` to replace the wrong area.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1329,6 +1329,19 @@ update with NEW-DESC."
                               old-path new-path old-desc new-desc))))
         (replace-match new-link)))))
 
+(defun org-roam--get-relative-link-replacement (old-path)
+  "Create file-relative link for link at point if needed.
+File relative links are assumed to originate from OLD-PATH. The
+replaced links are made relative to the current buffer."
+  (when-let ((link (org-element-lineage (org-element-context) '(link) t)))
+    (let ((type (org-element-property :type link))
+          (path (org-element-property :path link)))
+      (when (and (f-relative-p path)
+                 (org-in-regexp org-link-bracket-re 1))
+        (let* ((file-path (expand-file-name path (file-name-directory old-path)))
+               (new-path (org-roam-link-get-path file-path)))
+          (concat type ":" new-path))))))
+
 (defun org-roam--fix-relative-links (old-path)
   "Fix file-relative links in current buffer.
 File relative links are assumed to originate from OLD-PATH. The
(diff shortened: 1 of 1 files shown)


d12frosted:
Just curious, are you on macos?

herbertjones:
@d12frosted No, I'm on Debian Linux testing (Bullseye).

org-roam.el
```diff
@@ -1302,21 +1302,24 @@ If OLD-DESC is passed, and is not the same as the link
...
-          (when (and (string-equal (expand-file-name path) old-path)
-                     (org-in-regexp org-link-bracket-re 1))
-            (let* ((label (if (match-end 2)
-                              (match-string-no-properties 2)
-                            (org-link-unescape (match-string-no-properties 1))))
-                   (new-label (if (string-equal label old-desc)
-                                  new-desc
-                                label)))
-              (replace-match (org-roam-format-link new-path new-label type)))))))))
+  (cl-flet ((get-new-link ()
```
jethrokuan, line 1305: I don't think we need a cl-flet here. I think `org-in-regexp` butchers the old match data. Simpler to just wrap the `when-let` clause in a `save-match-data`, without introducing it as a function.

herbertjones:
Problem definitely occurs if I don't wrap `(link (org-element-lineage (org-element-context) '(link) t))` also.   Looks like `org-roam--fix-relative-links` may need to be fixed too possibly.

jethrokuan:
Have you tested that this works for you? What I meant was to `save-match-data` for everything within the original `when-let` clause, which is essentially what the original PR did, except with an additional named function. Now it's just saving for `org-element-lineage`.

herbertjones:
Yeah, I had both branches wrapped with `save-match-data` as they both needed it to work without erroring.  I've moved it to a separate named function in the change.  Also, since it was most likely a problem I updated the other possible problem loop similarly.

jethrokuan (approved):

PR #1193, opened 2020-10-17
(fix): do not kill temp buffer when extracting links for file

d12frosted:

Otherwise it breaks `org-roam-db-build-cache` as it relies on temporary buffer.
Without keeping the buffer, `org-roam--extract-links` kills it and any further
functionality of file processing breaks.


jethrokuan (approved):
Thanks!

PR #1195, opened 2020-10-19
(fix): fall back to org-id search ID face computation

### LINKED ISSUE #1191: Links using org-id are displayed as broken by org-roam
### Description
Current file is part of my agenda-files. The link is not broken, the issue is that in is displayed using org-roam-invalid-face.
org-roam-link-with-id-unvalid-face

Not sure if this is a bug or configuration issue on my part.
Any pointers reganding how to debug further are well appreciated.
Thanks.

### Environment
- Emacs: GNU Emacs 28.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.21, cairo version 1.17.3)
 of 2020-09-04

jethrokuan:
Fixes the scenario where the face is reported as invalid although it is part of Org's ID files (e.g. in the agenda).

Fixes #1191.

This can however slow face computation a lot (when there are many org id/agenda files). Here, we choose to focus on correctness. To mitigate this issue, we change the default scope of application of the face to Org-roam notes only.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1190,7 +1190,7 @@ This is active when `org-roam-completion-everywhere' is non-nil."
 ;;;; Function Faces
 ;; These faces are used by `org-link-set-parameters', which take one argument,
 ;; which is the path.
-(defcustom org-roam-link-use-custom-faces 'everywhere
+(defcustom org-roam-link-use-custom-faces t
   "Define where to apply custom faces to Org-roam links.
 
 Valide values are:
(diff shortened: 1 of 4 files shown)


org-roam.el
```diff
@@ -1202,9 +1202,9 @@ everywhere   Apply custom faces everywhere.
 
 Otherwise, do not apply custom faces to Org-roam links."
   :type '(choice
-	  (const :tag "Use custom faces inside Org-roam notes" t)
-	  (const :tag "Apply custom faces everywhere" everywhere)
-	  (const :tag "Do not apply custom faces" nil))
+          (const :tag "Use custom faces inside Org-roam notes" t)
+          (const :tag "Apply custom faces everywhere" everywhere)
```
d12frosted, line 1206: I didn't know about `everywhere` option. Thanks for adding information to docs!

Issue #1198, opened 2020-10-19
org-roam-buffer--find-file: Wrong type argument: window-live-p, nil


Whil-:
### Description
Sometimes when I'm following links from the org-roam backlinks buffer I get the error in the heading.

#### Steps to Reproduce
Not entirely sure how to reproduce it with another Org roam folder, but my steps are as follows:
- Open file which is referenced from other files in
- Run `org-roam` command
- Run `org-open-at-point` on link inside `*org-roam*` buffer
- Get error.

#### Backtrace
```
Debugger entered--Lisp error: (wrong-type-argument window-live-p nil)
  select-window(nil)
  (progn (with-selected-window org-roam-last-window (org-roam--find-file file)) (select-window org-roam-last-window))
  (if (and org-roam-last-window (window-valid-p org-roam-last-window)) (progn (with-selected-window org-roam-last-window (org-roam--find-file file)) (select-window org-roam-last-window)) (org-roam--find-file file))
  org-roam-buffer--find-file("c:/PIM/Notes/Systems.org")
  (progn (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t)
  (if p (progn (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t) nil)
  (let* ((file-from (and t (get-text-property (point) 'file-from))) (p (and file-from (get-text-property (point) 'file-from-point)))) (if p (progn (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t) nil))
  (if-let* ((file-from (get-text-property (point) 'file-from)) (p (get-text-property (point) 'file-from-point))) (progn (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t) nil)
  (if-let ((file-from (get-text-property (point) 'file-from)) (p (get-text-property (point) 'file-from-point))) (progn (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t))
  (when-let ((file-from (get-text-property (point) 'file-from)) (p (get-text-property (point) 'file-from-point))) (org-roam-buffer--find-file file-from) (goto-char p) (org-show-context) t)
[...]

PR #1200, opened 2020-10-19
(fix): let-bind org-roam-last-window (#1198)

Whil-:
* org-roam-buffer.el (org-roam-buffer--find-file): Make
org-roam-buffer--find-file insensitive to mutation of global
variables.

Issue #1198.


jethrokuan:
Thanks. I think I'll want to look into ways to remove the need for this window book-keeping, and rely on org-link-frame-setup.

Whil-:
@jethrokuan Sounds good! I didn't dig deeper than to solve the issue at hand. But there is something underneath that is not sound. Between this fix and now I got another windows-related error. This time related to non-existing window. Didn't have time to log it then though. But if the book-keeping can be removed altogether it's probably a good thing! =)

Thanks

PR #1204, opened 2020-10-20
(fix):  org-roam--extract-links leaves files open

### LINKED ISSUE #1203: org-roam--extract-links leaves files open
### Description
 org-roam--extract-links leaves files open

#### Steps to Reproduce
1. create a file with an id link to some other file or file's headline.
2. Run `org-roam--build-cache`

#### Expected Results
closes linked file intelligently (if it wasn't already open).

#### Actual Results
keeps linked file open.

#### Solution
```diff
 -                          (list (car (org-roam-id-find path nil nil 'keep-buffer))))
 +                          (list (car (org-roam-id-find path nil nil))))
```
there are no instances where the buffer in question should be kept open.

#### test clause
we need a test for functions that have open buffer side effects. there have been quite a few that leave side effects unhandled.
most notably ##1171, #1129 and #1146.
something like
``` elisp
   (let (before after)
   (setq before (length (buffer-list)))
   (function)
   (setq after (length (buffer-list)))
   (assert (= before after))
```

natask:
fixes #1203.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -598,7 +598,7 @@ it as FILE-PATH."
                                    :point begin))
                  (names (pcase type
                           ("id"
-                           (list (car (org-roam-id-find path nil nil 'keep-buffer))))
+                           (list (car (org-roam-id-find path nil nil))))
                           ("cite" (list path))
                           ("website" (list path))
                           ("fuzzy" (list path))


jethrokuan:
@d12frosted do you recall why you needed to change this?

I just took another look and it appears I'd overlooked something then, and `org-roam--extract-links` does not need to keep the org-id buffer open (since we're using pos in file, not markers).

d12frosted:
It was fixing another issue. If you use `id` type links (as opposed to `file` links), then `org-roam-db-build-cache` breaks. Don't remember all the details, but it tried to extract links in the invalid temp buffer.

d12frosted:
Oh, pressed wrong combination of keys, so sent too early.

I see two options:

1. To dig further to understand why `id` type of links break `org-roam-db-build-cache`.
2. Close buffers after building cache.

WDYT?

mchowning:
Although not a breakage, I did find that after the change to leave all the buffers open (https://github.com/org-roam/org-roam/pull/1193) the performance dramatically improved when extracting all the links from a file with many id-based links: https://github.com/org-roam/org-roam/issues/1208. With that said, I agree that just leaving all the buffers open isn't the right fix here.

jethrokuan:
Closed with #1212 , which no longer reads the linked files during link extraction.

PR #1210, opened 2020-10-24
(fix): update queue must store dir-local org-roam path

herbertjones:

When multiple org-roam directories are used, be sure to idle update the correct one.  Otherwise the main database will contain invalid paths and the secondary org-roam directories won't get the update.

Also, test the correct path variable in `org-roam--queue-file-for-update `.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1253,21 +1253,32 @@ file."
             (t
              'org-link)))))
 
+(defun org-roam--add-path-to-update-queue (file-path)
+  "Add FILE-PATH to `org-roam--file-update-queue.
+Stores possible local dir variables with path."
+  (add-to-list 'org-roam--file-update-queue
+               (list file-path org-roam-directory org-roam-db-location)))
+
 (defun org-roam--queue-file-for-update (&optional file-path)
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -1253,21 +1253,32 @@ file."
...
+      (org-roam--add-path-to-update-queue fp))))
 
 (defun org-roam--process-update-queue ()
   "Process files queued in `org-roam--file-update-queue'."
   (when org-roam--file-update-queue
-    (mapc #'org-roam-db--update-file org-roam--file-update-queue)
+    (dolist (info org-roam--file-update-queue)
+      (let ((org-roam-directory (elt info 1))
+            (org-roam-db-location (elt info 2)))
+        (org-roam-db--update-file (elt info 0))))
```
jethrokuan, line 1277: indentation is off here. Might also be cleaner to use `pcase-dolist`.

jethrokuan (changes requested):
small change here, otherwise lgtm.

jethrokuan:
i think #1212 makes this change impossible. I think we should add better support for owning multiple org-roam instances, treating each org-roam-directory as a project.

PR #1211, opened 2020-10-24
(fix): run title change hooks if title modified in org-capture buffer

herbertjones:

When I create a new org-roam entry, I don't always give it the exact title I want.  However inside the capture template, I may change the title after organizing my thoughts.  Unfortunately this doesn't change the filename when normally it would (when that feature is enabled).  Since now the filename is out of sync with the title, it won't be updated on any future changes either.

The filename for a capture is initially based off the given title.  However when that title is changed inside the template, the file is saved to the original name.  This just sets the local title variable so the title change hooks will run and the filename will be updated after saving as it normally would.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -385,6 +385,12 @@ The file is saved if the original value of :no-save is not t and
    ((and (not (org-roam-capture--get :orig-no-save))
          (not org-note-abort))
     (with-current-buffer (org-capture-get :buffer)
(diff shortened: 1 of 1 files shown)


herbertjones:
Ends up that the hook doesn't trigger correctly when inserting links into another page, so that this only works when creating new entries.  Perhaps there needs to be a post creation step that updates the title and filename and all variables that are returned.

PR #1212, opened 2020-10-25
(internal): simplify db update operations

jethrokuan:
Instead of maintaining a file queue to process for updating the Org-roam
database, we instead simply call `org-roam-db-build-cache` to rebuild
the db. `org-roam-db-build-cache` feels Fast Enough(TM), and basically
runs instantly if no files are modified. This greatly simplifies the
code, basically allowing to maintaining a single code path for db
operations.

This PR should also address the slowness wrt to ID links. Org-roam now
builds the headline information first, so there is no need to read the
file to check if the headline is there. The downside to this is that
this only works for IDs in Org-roam files.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -84,6 +84,9 @@ value like `most-positive-fixnum'."
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
   "Database connection to Org-roam database.")
 
+(defvar org-roam-db-dirty nil
+  "Whether the org-roam database is dirty and requires an update.")
+
 ;;;; Core Functions
 
 (defun org-roam-db--get-connection ()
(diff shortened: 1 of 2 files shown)


d12frosted:
> The downside to this is that this only works for IDs in Org-roam files.

Didn't get this part. Could you please elaborate?

jethrokuan:
>  Didn't get this part. Could you please elaborate?

Previously, to figure out which file the ID belongs to, Org-roam had to:

1. Query the Org-roam database, and if it doesn't exist, query the `org-id` database, or fall back to file of the current buffer
2. Search the file for the ID. If it is there, return the file, else, return nil (i.e. no ID for the file)

Now, Org-roam instead:

1. Always ensures that the ID database of Org-roam is up to date first. i.e. If the Org-roam database is queried for an ID, and there is a hit, it is guaranteed that the file contains the ID.
2. Instead of reading the file, now simply queries the database to find the ID.
3. *Optionally, we can fallback to `org-id`s cache, but we cannot provide the same guarantees, and this might be more confusing than it is worth*

Since the db only contains caches IDs for Org-roam files, this only works for Org-roam files.

PR #1213, opened 2020-10-25
(internal): rename link columns

### LINKED ISSUE #1207: Org-roam uses reserved keywords in the sqlite db
`from` is a column name in the `links` table, and is a reserved keyword. We should probably change the column names ("source", "dest" maybe).

jethrokuan:
from -> source
to -> dest

"from" is a reserved keyword in sqlite, so we avoid it.

Closes #1207

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -79,7 +79,7 @@ value like `most-positive-fixnum'."
   :type 'int
   :group 'org-roam)
 
-(defconst org-roam-db--version 9)
+(defconst org-roam-db--version 10)
 
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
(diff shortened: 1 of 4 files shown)


jethrokuan:
cc @org-roam/org-roam-server @org-roam/org-roam-bibtex in case your packages rely on querying the DB. Will merge once acked.

goktug97:
https://github.com/org-roam/org-roam-server/pull/104 Updated `org-roam-server`

myshevchuk:
Thanks!

alexkehayias:
This is worth a mention in the changelog. I have a custom export to markdown which relies on querying the sqlite DB directly (I suspect others do as well) which broke using latest melpa package.

Issue #1095, opened 2020-09-10
open org roam file when source links are captured


natask:
### Brief Abstract
When a specific or set of specific websites are captured through org-roam, open a corresponding org roam buffer.
### Long Description
This feature currently exists somewhat because org roam capture ref creates the same name for links when it creates corresponding org roam file. It also looks up if such file exists and opens it if it does.  but sometimes it fails. For example, if you capture this file while logged in, the name of the  file it creates is different than what it creates when signed out. Furthermore preserving data in the name of files is brittle because the user may change the name of the file without the intention of losing this feature. This is further exacerbated by the fact that names automatically generated by `org roam protocol open ref` are not palatable.
 A recent update to org roam added automatic renaming of files when the file title changes. This doesn't seem to work on files captured through `org roam protocol open ref`. It would be a sane system requirement to make features consistent across org roam.

### Proposed Implementation (if any)
Involves searching through files to check if such links exist. I will base it on  `- source ::` because that is what I have been using to keep source links but can make this another search term like `#+roam_tags:`. When capturing through org roam protocol open file, Search for this term within all org roam files and check if they contain the link, if so open it. If multiple files are found, pick the first one or open all. We can also display them and as the user to select. 
For the sake of preformace, add another column to the database that holds a list of source links and search the database instead, if there is an entry that holds the captured link. 
This can be abstracted to include ids or notmuch email entries as source links.
I will look if I can implement this feature this weekend.
### Please check the following:

- [X] No similar feature requests

jethrokuan:
Org-roam's capture ref protocol uses the ref key to reopen a note. When you use the bookmarklet, it uses the URL of the current page as the key. Org-roam looks up if there are any files with the URL as its ref. If there is, it will reopen the note, if there isn't it will create a new one.

So I think either your setup is incorrect, or your example is slightly misleading. I don't think GitHub uses different URLs when logged in or not, so the capture protocol should end up opening the same note.

> A recent update to org roam added automatic renaming of files when the file title changes. This doesn't seem to work on files captured through org roam protocol open ref. It would be a sane system requirement to make features consistent across org roam.

This is not true. The file will be renamed if as long as it is part of the Org-roam ecosystem. I've just tested it to be working.

natask:
@jethrokuan 
the url doesn't change but the name of the file created changes. 
[...]

PR #1215, opened 2020-10-25
(feat): Allow one file to have multiple roam_key statements

kisaragi-hiu:

I'd like to be able to associate multiple `ROAM_KEY` keywords to the same file. This will allow multiple bibliographical entries to share the same file.

My actual use case is that I want to add both a book and its chapters to my bibliographical database. If a file can only have one `ROAM_KEY` keyword, we have to create a note file for each chapter *and* the book. I'd like to have the choice of putting them all in one file.

Note that this is the inverse of #587. #587 wants one-key-to-many-files, while this implements many-keys-to-one-file.

This also allows setting both a URL and a cite key as referring to a file. For example:

```org
#+TITLE: swyx - Learn in public
#+ROAM_KEY: cite:shawnwang20180619
#+ROAM_KEY: https://www.swyx.io/learn-in-public/
```

###### Potential concerns

This PR makes `org-roam--extract-global-props` extract all matching props in the form `(("key" . "val1") ("key" . "val2"))` instead of only extracting the first match. I don't think this would be an issue, though, because `assoc` and other alist functions will simply take the first match themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - [#1183](https://github.com/org-roam/org-roam/pull/1183) add interactive functions for managing aliases and tags in Org-roam file, namely `org-roam-alias-add`, `org-roam-alias-delete`, `org-roam-tag-add`, and `org-roam-tag-delete`.
+- [#1215](https://github.com/org-roam/org-roam/pull/1215) Multiple `ROAM_KEY` keywords can now be specified in one file. This allows bibliographical entries to share the same note file.
 
 ### Bugfixes
 
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -47,7 +47,7 @@
 (defvar org-roam--org-link-bracket-typed-re)
 
 (declare-function org-roam-db--ensure-built   "org-roam-db")
-(declare-function org-roam--extract-ref       "org-roam")
+(declare-function org-roam--extract-refs      "org-roam")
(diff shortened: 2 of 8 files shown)


jethrokuan:
Looks like in this PR you're able to extract multiple refs from a single file, but you're not saving them all into the database. Any reason why?

kisaragi-hiu:
This PR does save all extracted refs into the database:

```elisp
(defun org-roam-db--insert-ref (&optional update-p)
  "Update the ref of the current buffer into the cache.
If UPDATE-P is non-nil, first remove the ref for the file in the database."
  (let ((file (or org-roam-file-name (buffer-file-name)))
        (count 0))
    (when update-p
      (org-roam-db-query [:delete :from refs
                          :where (= file $s1)]
                         file))
    (when-let ((refs (org-roam--extract-refs)))
      (dolist (ref refs) ; <= this should insert all refs from this file
        (let ((key (cdr ref))
              (type (car ref)))
...
```

although I just realized that the `--insert-ref` name is inconsistent now that it's using `org-roam--extract-refs`, so I've now renamed it to `org-roam-db--insert-refs`.

I kept `org-roam--extract-ref` around in case another package uses this private function.

jethrokuan (approved):

natask:
@kisaragi-hiu nice. I was looking for this in #1095. I opted for something more suited for my problem at the end [though](https://github.com/natask/org-roam-url).

PR #1218, opened 2020-10-26
(fix): fix id-face killing buffers

### LINKED ISSUE #1209: *org-roam* buffer crashes on while some org files opening
When run `org-roam-find-file` command for some org files *org-roam* buffer loading properly but for some org files I got `wrong-type-argument: window-live-p` error. 

output-2020-10-24-13:31:51

jethrokuan:

Prompted by https://github.com/org-roam/org-roam/issues/1209#issuecomment-716224554, I figured that font-locking was what was killing the org-roam buffer during backlink buffer population and causing `window-live-p` errors.

This PR changes that. `org-roam-id-get-file` no longer falls back onto the current buffer: It queries the org-roam database, and optionally `org-id-locations`, and declares that it does not exist if it isn't in either.

Fixes #1209

### DIFF
--- a/org-roam-buffer.el
+++ b/org-roam-buffer.el
@@ -167,53 +167,52 @@ ORIG-PATH is the path where the CONTENT originated."
                                                      "file")))
               (dolist (backlink bls)
                 (pcase-let ((`(,file-from _ ,props) backlink))
-                  (insert (propertize (org-roam-buffer-expand-links (plist-get props :content) file-from)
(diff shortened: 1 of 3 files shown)


oatmealm:
I'm still seeing this problem... also some corruption of content in buffers when following a new link created using `org-roam-insert` (which is probably another issue). Sometimes the backlinks buffer gets re-populated (from being empty) if I update the content of the org-roam note and save...

https://github.com/hlissner/doom-emacs/issues/4133

PR #1220, opened 2020-10-27
(internal): make sure there are no byte-comp warnings about undefined functions or unused lexical variables

kisaragi-hiu:

For undefined functions: for example, s.el and org-element should be `require`'d when their functions are used.

Unused lexical variables: if Org isn't loaded yet, dynamic variables defined in org.el would be treated as lexical and byte-comp would emit this warning. This is especially important in the future as native-comp / gccemacs will optimize away unused lexical variables, and we cannot rely on Org having been implicitly loaded before our modules are compiled.

(This is also the second culprit behind why the infinite loop happened in #1078: besides org-roam-mode being started in org-mode-hook, the code setting org-mode-hook to nil is optimized away.)

Explicitly stating in our modules that the variables are dynamic prevents that.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -35,13 +35,18 @@
 (require 'emacsql)
 (require 'emacsql-sqlite3)
 (require 'seq)
-(require 'org-macs)
-(require 'org-roam-macs)
+
+(eval-and-compile
+  (require 'org-roam-macs)
+  ;; For `org-with-wide-buffer'
+  (require 'org-macs))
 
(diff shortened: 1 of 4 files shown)


jethrokuan:
Interesting, there's actually a byte-compile check in the CI, that should error if there are byte-compile warnings. I suppose that the way gccemacs performs the byte-compilation is slightly different.

I'm on gccemacs now, and don't see any issues here, but this change is probably good.

Issue #1224, opened 2020-10-29
Moving a file containing id-links to a different directory improperly applies a "relative path" to the id in the link


mchowning:
### Description

When moving a file that contains an id-link to a new directory, the id-link is updated as though it was a file-path that needed to have the relative path updated. For example, if you move a file containing the link `[[id:a_link]]` from `/level_1` to `level_1/level_2` the id-link gets updated to be `[[id:../a_link][test_file]]`. Similarly, if you had instead started with the file in `/level_1/level_2/` and moved it to `/level_1/`, the id link would have become `[[id:level_2/a_link][test_file]]`.

I think this is occurring because [we search for and update the relative path for "any" link](https://github.com/org-roam/org-roam/blob/master/org-roam.el#L1336), instead of only doing it for file links.

#### Steps to Reproduce

1. Create a file that contains an id-link to another file, i.e., `[[id:some_id][test_file]]`
2. Rename the file from within emacs so that it is now one level deeper, i.e., in a subdirectory
3. Make sure the buffer of the moved file is up-to-date (close and reopen the file buffer or run `revert-buffer`)
4. Observe that the link is now `[[id:../some_id][test_file]]`

#### Backtrace

n/a

#### Expected Results

Moving a file should only update file-links and not update id-links.

#### Actual Results

Moving a file applies a relative path to id-links.

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30, cairo version 1.15.10)
 of 2020-08-26
- Framework: Spacemacs
- Org: Org mode version 9.4 (9.4-17-g9a6689-elpaplus @ /home/matt/.emacs.d/elpa/27.1/develop/org-plus-contrib-20201005/)
- Org-roam: 1.2.2
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/0d23568

PR #1226, opened 2020-10-30
(fix): only update relative path of file links

mchowning:

Addresses #1224 by only updating a relative link after we check that it is a file link.


d12frosted (commented):
this looks good! I also run into this issue, thanks for taking your time to send a PR :)

PR #1230, opened 2020-10-31
(fix): only replace path portion of relative link

### LINKED ISSUE #1227: Moving files with relative-path org-mode links converts the links into plain-text links
### Description

If `file1` contains a relative-file-path org-mode link to `file2` in the same directory (i.e., `[[file2][relative file link]]`), and `file1` is moved to a different directory, the link in `file1` is updated so that it is a plain-text link (i.e., `file:file2`) and the link description is lost. However, the path of the link is properly updated to reflect the new correct relative path.

#### Steps to Reproduce

1. Create two files in the same directory (`file1`, `file2`) with a relative-path org-mode link in `file1` to `file2` like this: `[[file2][link to file2]]`
2. From within emacs, move `file1` to a subdirectory
3. Ensure the `file1` buffer is up-to-date by either closing and reopening the buffer or running `revert-buffer`
4. Observe that the link to `file2` in `file1` is no longer an org-mode link and instead is just a pain-text link: "file:../file2"

#### Backtrace

n/a

#### Expected Results

When moving a file, relative-path org-mode links in that file should continue to work as org-mode links.

#### Actual Results

When moving a file, relative-path org-mode links in that file are converted to plain-text links.

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30, cairo version 1.15.10)
 of 2020-08-26
- Framework: Spacemacs
- Org: Org mode version 9.4 (9.4-17-g9a6689-elpaplus @ /home/matt/.emacs.d/elpa/27.1/develop/org-plus-contrib-20201005/)
- Org-roam: 1.2.2
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/56c47fb

mchowning:

Fixes #1227 to properly update relative links by updating the call to `replace-match` to only replace the path portion of the match instead of the entire link.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1336,7 +1336,7 @@ replaced links are made relative to the current buffer."
     (while (re-search-forward org-link-any-re nil t)
       (when-let ((new-link (save-match-data
                              (org-roam--get-relative-link-replacement old-path))))
-        (replace-match new-link nil t nil 1)))))
+        (replace-match new-link nil t nil 2)))))
 
 (defcustom org-roam-rename-file-on-title-change t
   "If non-nil, alter the filename on title change.


jethrokuan:
Superceded by #1233. The fix I applied does a bit less work, than the original implementation.

PR #1232, opened 2020-11-04
(fix): widen before title extraction

herbertjones:

Prevent renaming file and changing links incorrectly due to using `(org-narrow-to-subtree)` and saving.

The extract title functions will only use the narrowed contents when the buffer being read is narrowed.  These functions might ignore the `#+title` line, and use the first found subtree as the title instead.  Now all the links to the file and the filename has been changed, and won't be changed back unless narrowing is removed and another change is made and the file is saved again.  However if the file is reloaded and the title changed, it won't be since all the links no longer match the title, leaving the affected files with links that no longer match the title.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -669,18 +669,19 @@ Reads from the \"roam_alias\" property."
 (defun org-roam--extract-titles (&optional sources nested)
   "Extract the titles from current buffer using SOURCES.
 If NESTED, return the first successful result from SOURCES."
-  (let (coll res)
-    (cl-dolist (source (or sources
-                           org-roam-title-sources))
-      (setq res (if (symbolp source)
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -669,18 +669,20 @@ Reads from the \"roam_alias\" property."
...
-      (setq res (if (symbolp source)
-                    (funcall (intern (concat "org-roam--extract-titles-" (symbol-name source))))
-                  (org-roam--extract-titles source t)))
-      (when res
-        (if (not nested)
-            (setq coll (nconc coll res))
-          (setq coll res)
-          (cl-return))))
-    coll))
+  (save-restriction
```
jethrokuan, line 672: Nice catch. Maybe use `org-with-wide-buffer`?

jethrokuan (approved):

PR #1239, opened 2020-11-07
Allow org-roam-protocol to capture the webpage's selection

zaeph:
I believe we supported this feature a while ago, but an iteration must have overridden it.

You can now use `%i` in `org-roam-capture-ref-templates` to insert the selected text on the web-page you are capturing.  Note that it can either appear in `:head` (expanded once on file creation) or in the body of the template (expanded every time the ORP is called on a given webpage).  The latter makes it a good option for incremental reading.

Here's a use-case:
```el
(defun zp/org-protocol-insert-selection-dwim (selection)
  "Insert SELECTION as an org blockquote."
  (unless (string= selection "")
    (format "#+begin_quote\n%s\n#+end_quote" selection)))

(setq org-roam-capture-ref-templates
      '(("r" "ref" plain
         (function org-roam-capture--get-point)
         ""
         :file-name "web/${slug}"
         :head "#+title: ${title}
#+roam_key: ${ref}
#+created: %u
#+last_modified: %U

%(zp/org-protocol-insert-selection-dwim \"%i\")"
         :unnarrowed t)

        ("i" "incremental" plain
         (function org-roam-capture--get-point)
         "* %?\n%(zp/org-protocol-insert-selection-dwim \"%i\")"
         :file-name "web/${slug}"
         :head "#+title: ${title}
#+roam_key: ${ref}
#+created: %u
#+last_modified: %U\n\n"
         :unnarrowed t
         :empty-lines-before 1)))
```

You need to make sure that you have the proper bookmarklet (note the `template=r` to specify which template to use`):
```js
javascript:location.href = 'org-protocol://roam-ref?template=r&ref='%20+%20encodeURIComponent(location.href)%20+%20'&title='%20+%20encodeURIComponent(document.title)%20+%20'&body='%20+%20encodeURIComponent(window.getSelection())
```

### DIFF
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -39,6 +39,11 @@
 (require 'org-roam)
 (require 'ol) ;; for org-link-decode
 
+(defcustom org-roam-protocol-store-links nil
+  "Whether to store links when capturing websites with `org-roam-protocol'."
+  :type 'boolean
+  :group 'org-roam)
+
 ;;;; Functions
 (defun org-roam-protocol-open-ref (info)
   "Process an org-protocol://roam-ref?ref= style url with INFO.
@@ -79,6 +84,9 @@ It opens or creates a note with the given ref.
            (orglink
             (if (null ref) title
               (org-link-make-string ref (or (org-string-nw-p title) ref)))))
+      (when (and org-roam-protocol-store-links
+                 ref)
+        (push (list ref title) org-stored-links))
       (org-link-store-props :type type
                             :link ref
                             :description title
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -70,10 +70,8 @@ It opens or creates a note with the given ref.
              (title (or \.title ""))
              (body (or \.body ""))
              (orglink
-              (if (null ref) title
-                (org-link-make-string ref (or (org-string-nw-p title) ref)))))
-        (when (and org-roam-protocol-store-links
-                   ref)
+              (org-link-make-string ref (or (org-string-nw-p title) ref))))
+        (when org-roam-protocol-store-links
           (push (list ref title) org-stored-links))
         (org-link-store-props :type type
                               :link ref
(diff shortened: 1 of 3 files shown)


zaeph:
Using the `incremental` template to create a file almost works, but the lines in `:empty-lines-before` are inserted before `:head`.  There might be an easy way to fix this, but I think we're pushing the limits of our double-expansion system here.

jethrokuan:
See also #1088

zaeph:
Thanks for pointing it out, @jethrokuan, hadn't caught up on that PR yet.

Considering the wild variety of notes people are making, I think we shouldn't include `%i` in the default template:
- People have a set expectation of what is going to happen when they use the bookmarklet, and changing the default would clash with that.
- `%i` is a default `org-capture` feature, but modularising it inside a single template requires Elisp, which people aren't likely to have.
- It creates expectations for incremental reading (i.e. appending notes to a file that has already been created), and this isn't ready yet.

I vote for merging this now, and addressing the incremental reading stuff in another PR.

org-roam-protocol.el
```diff
@@ -58,6 +63,20 @@ It opens or creates a note with the given ref.
       (error "No ref key provided"))
     (when-let ((title (cdr (assoc 'title decoded-alist))))
       (push (cons 'slug (funcall org-roam-title-to-slug-function title)) decoded-alist))
+    (let-alist decoded-alist
```
jethrokuan, line 66: let-alist seems to be a package on elpa, not sure if it ships with Emacs by default.

Also from the docs of `let-alist`, it seems the backslashes aren't necessary.

zaeph: `let-alist` should be in core, but we might require `let-alist` for it to work.

The backslashes aren't necessary, but `lispy` is adding them for me.  I think that's a problem on their end, so I'll edit by hand and file a report there.

jethrokuan (approved):

PR #1244, opened 2020-11-10
(internal): remove unnecessary variables

jethrokuan:
For captures, enforce that `:file-name` needs to be specified, and
`:head` when not provided defaults to the empty string (i.e. no head)

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -410,7 +410,7 @@ aborted, we do the following:
 the file if the original value of :no-save is not t and
 `org-note-abort' is not t."
   (let* ((name-templ (or (org-roam-capture--get :file-name)
-                         (user-error "Template needs to have a file name.")))
+                         (user-error "Template needs to specify `:file-name'")))
          (new-id (s-trim (org-roam-capture--fill-template
                           name-templ)))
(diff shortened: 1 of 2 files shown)


zaeph:
LGTM!

We might have to think a little more about `:dir-name` for `org-roam-dailies`, but I need to think some more about this, so we'll keep it for later.

zaeph (approved):

org-roam-capture.el
```diff
@@ -410,7 +410,7 @@ aborted, we do the following:
 the file if the original value of :no-save is not t and
 `org-note-abort' is not t."
   (let* ((name-templ (or (org-roam-capture--get :file-name)
-                         (user-error "Template needs to have a file name.")))
+                         (user-error "Template needs to specify `:file-name'")))
```
zaeph, line 413: Good catch. 👍

jethrokuan:
There's no `:dir-name` in the template afaik...

zaeph:
> There's no `:dir-name` in the template afaik...

No, I elected to remove it, but it was in an earlier iteration of `org-roam-dailies`.  I'll find the commit.

zaeph:
There we go: f30104ad96744d5d1e232ea66142f659f31d3e9d.

PR #1265, opened 2020-11-12
(fix) Link to the documentation page

tskupinski:
I was setting up the org-roam today and saw that the current link in `README.md` points to the 404 page. This change should fix this.


jethrokuan:
Thanks! Missed this in my update

PR #1267, opened 2020-11-12
Allow timestamps as links and #+roam_keys

ricklupton:
With this change, timestamps are recorded as links of type "timestamp". Org active timestamps are recognised as ref keys. This means that e.g. daily note files can show all places where a matching org timestamp occurs in the backlinks buffer.

With daily notes, it would be nice to be able to easily see events/meetings/notes that are tagged as occuring on that day in the backlinks buffer. This is possible with a link to the daily note such as `[[file:2020-11-12.org][2020-11-12]]`, but since org already has a method for linking to dates -- timestamps -- it would be nice to be able to use that. Specifically, I often already have headlines with timestamps so they show up in org-agenda; it would be nice not to have to duplicate the information as a separate link to the daily note.

This means that if I have a file `2020-11-12.org` with this contents:
```org
#+title: 2020-11-12
#+roam_key: <2020-11-12 Thu>
```

and another file with this contents:
```org
#+TITLE: Temp

Working on this today [2020-11-12 Thu]
```

Then the backlink appears in the buffer for the daily note:
```org
* 1 Ref Backlink
** [[file:temp.org][Temp]]
Working on this today [2020-11-12 Thu]
```

###### Further work needed

If you think this is a reasonable addition, there are a couple more things to do:
- Add to changelog

And possible extensions:
- Allow customising behaviour -- should inactive, active or both types of timestamp be included?
- Optionally do something sensible when the ref timestamp falls within a date range
- Optional also link CLOCKED time

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -808,8 +808,8 @@ Each ref is returned as a cons of its type and its key."
           ((pred string-empty-p)
            (user-error "Org property #+roam_key cannot be empty"))
           ((pred (string-match org-link-plain-re))
-           (setq type (org-roam--collate-types (match-string 1 ref))
-                 path (match-string 2 ref)))
+           (setq type (org-roam--collate-types (match-string 1 roam-key))
+                 path (match-string 2 roam-key)))
           ((pred (string-match org-ts-regexp2))
            (setq type "timestamp"
                  path roam-key)))
--- a/org-roam.el
+++ b/org-roam.el
@@ -590,8 +590,9 @@ it as FILE-PATH."
                       org-roam-file-name
                       (buffer-file-name)))
   (save-excursion
-    (let (links)
-      (org-element-map (org-element-parse-buffer) 'link
+    (let ((parsed (org-element-parse-buffer))
+          links)
+      (org-element-map parsed 'link
         (lambda (link)
           (goto-char (org-element-property :begin link))
           (let* ((type (org-roam--collate-types (org-element-property :type link)))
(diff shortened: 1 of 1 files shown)


zaeph:
Thanks for this; it looks interesting.

I'll review it in the morning.

jethrokuan:
Think "ref backlinks" might be a bit overloaded here.

With the side buffer revamp, we should be able to add daily notes as another widget, much like how unlinked references is there now too.

ricklupton:
> Think "ref backlinks" might be a bit overloaded here.

I agree. I guess there are three directions:
1. Keep the current structure, but rename "ref backlinks" to something more general. To me, this use of `#+roam_key` is consistent with other existing uses. In every case, it says that the note is *about* another thing, but it isn't the other thing. For example, a bibliographic note is a note about a book/paper, but it isn't actually the paper. A note about a website isn't actually the website, and a daily note is about the day but it isn't actually the day itself. With all of these, you can see in the side buffer both direct backlinks to the note itself, and "ref backlinks" to the thing that the note is about.

2. Treat timestamps completely separately; add a new keyword `#+roam_timestamp` or similar and show separately as "Timestamp references" or similar in the side buffer.

3. Hybrid; keep the current internal structure but just alter the presentation of the side buffer by putting "timestamp references" in a different section.

Happy to adjust the PR when you have a chance to give feedback, or to hand it over to you.

Thanks,
Rick

jethrokuan:
Timestamps such as scheduled/deadline are captured in Org-roam v2.
Also, you can get Org-roam to cache timestamps by placing them in the PROPERTIES drawer, what you do with the cached data is then up to you.

PR #1272, opened 2020-11-13
documentation example for org-roam-dailies typo

brabalan:
The `olp` option in the lab notes example was swapped.


zaeph:
Good catch, thank you!

PR #1281, opened 2020-11-15
(fix): fix idle-timer not created on org-roam-mode

### LINKED ISSUE #1280: enabling org roam update on idle timer requires setting org-roam-db-file-update-timer unnecessarly
### Description
To enable org roam's feature to update, org-roam-db-file-update-timer needs to be set unnecessarily.
#### Steps to Reproduce
1. use the following as use-package
``` elisp
    (use-package org-roam
      :ensure t
      :hook
      (after-init . org-roam-mode)
      :custom
      (org-roam-db-update-method 'idle-timer)
      (org-roam-db-update-idle-seconds 0.5))
```
#### Expected Results
org roam updates on 0.5 second idle timer.

#### Actual Results
org roam doesn't update on timer at all.

### Solution
remove  org-roam-db-file-update-timer from
``` elisp
(if (and org-roam-db-file-update-timer
             (eq org-roam-db-update-method 'idle-timer))
```
in org-roam-mode.

or use
``` elisp
    (use-package org-roam
      :ensure t
      :hook
      (after-init . org-roam-mode)
      :custom
      (org-roam-db-update-method 'idle-timer)
      (org-roam-db-file-update-timer t)
      (org-roam-db-update-idle-seconds 0.5))
```

jethrokuan:
Fixes #1280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 
 ### Fixed
+- [#1281](https://github.com/org-roam/org-roam/pull/1281) fixed idle-timer not instantiated on `org-roam-mode`
 
 ## 1.2.3 (13-11-2020)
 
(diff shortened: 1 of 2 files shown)


shlevy:
Could we get a release with this? Or should users be running from master/melpa unstable for now?

PR #1320, opened 2020-11-22
update org-roam buffer fix for spacemacs

TS-CUBED:
Spacemacs does not update org-roam-buffer when changing roam files. This fixes that by adding the hook to `org-roam-buffer--update-maybe` to the autoload function.

It seems that hook get's removed on exiting org-roam-mode, but not getting added back when returning to an org-roam buffer in spacemacs (alternate-buffer switching).


jethrokuan:
Superceded by #1321

PR #1327, opened 2020-11-28
(fix): preserve existing description in a roam: link on replace

kisaragi-hiu:

Before, `[[roam:Abc][def]]` would be replaced with `[[file:abc.org][Abc]]`; the existing description "def" is replaced with the description fetched from abc.org.

This change makes it so that the existing description will be preserved:

- `[[roam:Abc][def]]` → `[[file:path/to/abc.org][def]]`
- `[[roam:Abc]]` → `[[file:path/to/abc.org][Abc]]`

This is useful for attaching links to existing text.

A word on retrieving the existing description: `org-element-link-parser` doesn't return the description itself, relying on the caller to be in the same buffer and retrieve with `buffer-substring-no-properties` themself. This is fine in `org-roam-link-replace-all`, where current buffer is known, but in `org-roam-link-follow-link` we're relying on Org not changing `current-buffer` during the link following process. I believe this is a reasonable trade off for not deleting existing text.

Edit: (feat) → (fix), as this is more like a fix than a feature (like #1252).

### DIFF
--- a/org-roam-link.el
+++ b/org-roam-link.el
@@ -73,16 +73,16 @@ noabbrev  Absolute path, no abbreviation of home directory."
 
 (defun org-roam-link-follow-link (path)
   "Navigates to location specified by PATH."
-  (pcase-let ((`(,link-type ,loc ,desc ,mkr) (org-roam-link--get-location path)))
+  (pcase-let ((`(,link-type ,loc ,desc ,mkr) (org-roam-link--get-location path t)))
     (when (and org-roam-link-auto-replace loc desc)
       (org-roam-link--replace-link link-type loc desc))
     (pcase link-type
-          ("file"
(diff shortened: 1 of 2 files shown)


org-roam-link.el
```diff
@@ -74,15 +74,26 @@ noabbrev  Absolute path, no abbreviation of home directory."
 (defun org-roam-link-follow-link (path)
   "Navigates to location specified by PATH."
   (pcase-let ((`(,link-type ,loc ,desc ,mkr) (org-roam-link--get-location path)))
```
jethrokuan, line 76: it probably makes more sense to change `org-roam-link--get-location` instead.

jethrokuan (changes requested):

PR #1345, opened 2020-12-15
(doc): Add async update to Getting Started

nobiot:
Clarification about the asycn db update seems helpful for beginners.

I tried my best to follow the writing style of the original. Please advise if this is appropriate.

Refer to this Slack thread
https://orgroam.slack.com/archives/CV160S8EL/p1608039505055000

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -358,9 +358,17 @@ nothing right now, since there are no notes in the directory. Entering the title
 of the note you wish to create, and pressing ~RET~ should begin the note
 creation process. This process uses ~org-capture~'s templating system, and can
 be customized (see [[*The Templating System][The Templating System]]). Using the default template, pressing
-~C-c C-c~ finishes the note capture. Running ~M-x org-roam-find-file~ again
-should show the note you have created, and selecting that entry will bring you
-to that note.
+~C-c C-c~ finishes the note capture.
+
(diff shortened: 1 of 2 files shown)


jethrokuan:
LGTM, this info is important to know for users.

jcguu95:
Hello,

Wanting the async feature, I took a look into the code.. but it doesn't appear to me that there's any. Would you mind pointing me out the entry point?

In particular, in `org-roam-db.el`, there is no `async` method called. It seems to be using `emacsql` fundamentally, but afaik it doesn't support async method either as in [this issue](https://github.com/skeeto/emacsql/issues/59).

jethrokuan:
I removed usage of `async`, because it's flaky and difficult to support. Updating on `idle-timer` is an alright alternative, and should be the default.

jcguu95:
That's sad to hear.. as my roam folder has grown to a point while each `db-sync` could freeze emacs up to 10 seconds. Sometimes I'd just disable `roam` because it's not clear to me when it's going to `db-sync`. 

It's a separate issue, but can I turn off auto-sync completely, and call it manually when I need?

jethrokuan:
> It's a separate issue, but can I turn off auto-sync completely, and call it manually when I need?

Not in v1, but will be made possible in v2.

PR #1348, opened 2020-12-16
(doc): Add footnote for Getting Started

nobiot:
Follow on to #1345.

The discussion in the Slack thread further clarified what was exactly
confusing for a beginner reader of this Getting Started. When a user is using
the built-in completion (no ivy, no ido), then `org-roam-find-file` does not
immediately show any files -- instead, the user has to press TAB. This is not
explicitly mentioned.

https://orgroam.slack.com/archives/CV160S8EL/p1608121018061700?thread_ts=1608039505.055000&cid=CV160S8EL

@jethrokuan , you might consider this to be too much of a detail for User Manual -- it could be distracting more experienced user. It could be a candidate for more beginner-oriented tutorial, rather than User Manual. Another view would be that it is in the Getting Started, so a little more of explanation for beginner users might be appropriate.

In either case, the use of footnote is intended to minimize the body of the text while adding additional clarification. Is this a good way? Or would you rather put something in the body of the text?

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -365,10 +365,10 @@ avoid getting in the way of writing. Org-roam queues updates to the files,
 waits for you to be idle for 2 seconds, and then automatically triggers
 updating the cache. After the cache has been updated, running ~M-x
 org-roam-find-file~ again should show the note you have created, and selecting
-that entry will bring you to that note [fn:1]. One can customize the waiting time by
-setting ~org-roam-db-update-idle-seconds~; or change the cache update to be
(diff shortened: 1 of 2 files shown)


jethrokuan:
>  @jethrokuan , you might consider this to be too much of a detail for User Manual -- it could be distracting more experienced user. It could be a candidate for more beginner-oriented tutorial, rather than User Manual. Another view would be that it is in the Getting Started, so a little more of explanation for beginner users might be appropriate.

I think I'm really not the best judge of what should go into the manual.

Footnotes appear a bit awkwardly on the single-page manual online, but that's not really a problem to me.

PR #1360, opened 2020-12-26
(feat) Prompt user to select a template for org-roam-capture.

zhoupingjay:
Currently `org-roam-capture` doesn't prompt for template keys like `org-capture`. If there are multiple org-roam-capture templates, user has to pass the selection as argument manually.

To make this more user-friendly, I'm adding a prompt menu to `org-roam-capture` for user to select the template. The screen is similar to `org-capture` - it includes a menu of templates (populated from `org-roam-capture-templates`), as well as a key "C" for customization.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -596,9 +596,10 @@ GOTO and KEYS argument have the same functionality as
       (funcall-interactively org-roam-capture-function))))
 
 (defun org-roam-capture-select-template (templates &optional keys)
-  "Prompt user to select a template for `org-roam-capture'.
-Argument TEMPLATES: org-roam-capture templates to be populated
-(`org-roam-capture-templates'). Argument KEYS: see `org-capture'."
+  "Prompt the user to select a template for `org-roam-capture'.
+Argument TEMPLATES: org-roam templates to be populated. Typically this
+should be `org-roam-capture-templates', but custom templates can be provided if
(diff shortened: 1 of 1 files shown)


jethrokuan:
The way org-roam-capture behaves is like this:

1. If only 1 template, no prompt
2. If more than 1 template, use `org-capture`'s prompting mechanism.

This is what I see with 2 templates:

image

zhoupingjay:
I see. So the user will be prompted for a template after entering the title or selecting a file. Just curious, what if user select an existing file and then select a template? Will the new content be appended to the selected file?

zhoupingjay:
Just tried a little bit with this scenario:
1. Capture an item with template A (using roam_tag "tag_a")
2. (org-roam-capture), select the existing item, and then select a different template B (which uses roam_tag "tag_b")
3. Enter something in the capture buffer, and then C-c C-c
4. The item was updated with new content, but roam_tag remained unchanged (tag_a).

What is the expected behavior in this case (org-roam-capture on an existing file but with a different template)? Should the metadata of new template be applied, in addition to the new content?

zhoupingjay:
More thoughts on this scenario:
How about we skip the template selection step on existing items (i.e. template selection only shows when capturing new items)?

jethrokuan:
Org-roam-capture has been overhauled, closing this PR.

PR #1362, opened 2020-12-28
(feat): sync and update desc for id and path.

natask:
extends update on title change to id links.
implements sync function that syncs current title as description of all links.


natask:
hold off on merging this. It has some bugs.

natask:
good to go after merging #1367

org-roam.el
```diff
@@ -1375,6 +1405,41 @@ respectively."
...
+ 'query        Ask before updating link.
+ 'force        Update link without asking.
+ 'nil          Do not update link description unless old description matches link description."
+  :type '(choice
+          (const :tag "Ask before updating link." 'query)
+          (const :tag "Update link without asking." 'force)
+          (const :tag "Do not update link unless old description matches link description." nil))
+  :group 'org-roam)
+
+(defun org-roam-sync-title-to-backlinks ()
```
jethrokuan, line 1427: This doesn't seem to be used anywhere.

org-roam.el
```diff
@@ -1447,7 +1530,8 @@ When NEW-FILE-OR-DIR is a directory, we use it to compute the new file path."
                              new-file
                            (car file)))
               (org-roam-with-file file nil
-                (org-roam--replace-link old-file new-file)
+                (org-roam--replace-link (list :path old-file :id file-id)
```
jethrokuan, line 1533: We still need to figure out how to pass the title change over.

org-roam.el
natask: The idea is that by running `org-roam-sync-title-to-backlinks`, I can update backlink discriptions *without* the chaining file name.

org-roam.el
natask: title change will be passed down when `org-roam--update-links-on-title-change` is called. Not efficient to update each affected file potentially two times but it works. 
Maybe you are thinking changing the file name directly should also update the title of a file which then propagates to file name to backlinks along with the file title?

org-roam.el
natask: actually, It is a premature implementation to separate the link between back-links, file name and file title that only works on files whose title doesn't currently match their file name. I will remove it.

natask: Need to call this in sync-title-to-filename

natask:
ready to merge.

org-roam.el
natask: removed `:id file-id`

natask (commented):
(unless new-buf
           (save-buffer))
       (setq res (progn ,@body))
       (unless (and new-buf (not ,keep-buf-p))
         (save-buffer)))
```
I am thinking of replacing with 

```
       (unless new-buf
           (save-buffer))
       (setq res (progn ,@body))
         (save-buffer))
```
because there are times where @body modifies the buffer. should this be handled by the caller?

**Comment by czikus** [comment]
What's holding this from being merged right now?

**Comment by jethrokuan** [comment]
I'm not making any changes to the current repo, am working on a new branch and don't want to detract from that.

**Comment by natask** [comment]
you can use my repo which has these updates in the mean time.
replace org-roam package declaration within configs with the following.
spacemacs/melpa
```(org-roam
                                       :location (recipe :fetcher github
                                                         :repo "natask/org-roam"
                                                         :branch  "my/latest"
                                                         :files ( "*.el")
                                                         ))
```
or on doom
``` elisp
(package! org-roam
  :recipe (:host github
           :repo "natask/org-roam"
           :branch  "my/latest"
           :files ( "*.el")
           ) 
```

jethrokuan:
The way links are captured in v2 is different, this will need some adapting for v2.

PR #1386, opened 2021-01-16
(doc) Update brew install script

rollacaster:

The API for Homebrew Cask changed (https://brew.sh/)


d12frosted (approved):
Yep, this is how it works now. 💯

PR #1393, opened 2021-01-19
Clarify behavior of capture functions

### LINKED ISSUE #1391: Org roam takes over original window after completion
### Description

Normally, when I use org-capture, saving and finalizing the capture closes the org-capture window and returns me to my prior window with the original buffer I had open in it.  I.e., if I have foo.txt open in a single window, when I'm done, foo.txt is still in the original window.

However, using org-roam, when I finalize the note, it closes the new window, but the note occupies my old window instead of the original buffer.  See below for an example of this behavior (assuming the animated gif works):

org-wrong-window

By contrast the following animated gif illustrates the ordinary behavior of org-capture, which I would have expected to carry over to org-roam

correct_behavior

I'm not sure if this is a bug or if it's just intended behavior.  If it's intended behavior, is there a setting to turn it off and get the normal org-capture behavior?

Caveats: 

1 I'm using evil

2. In the video/gifs I'm saving the buffer before finalizing/closing it with C-c C-c... not sure whether that's what one is supposed to do, or whether one should just close immediately.  HOWEVER, when I just finalize with C-c C-c without saving first, it does the same thing. 

#### Steps to Reproduce

1.  Create a new note with org-roam-find-file
2. Finalize the note

#### Backtrace

n/a no error given 

#### Expected Results

I'd land in my original buffer in my original window

#### Actual Results

the newly created note is in my original window

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.6.0, Carbon Version 162 AppKit 1894.6)
 of 2020-12-27
- Framework: N/A
- Org: Org mode version 9.4.4 (release_9.4.4-186-g7fa817 @ /Users/paulgowder/.emacs.d/straight/build/org/)
- Org-roam: 1.2.3

paultopia:

Makes clear the difference between org-roam-find-file and org-roam-capture.  Closes #1391

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -370,6 +370,12 @@ time by setting ~org-roam-db-update-idle-seconds~; or change the cache update
 to be triggered immediately after buffer save by setting
 ~org-roam-db-update-method~ to ~'immediate~.
 
+For experienced ~org-capture~ users, the behavior of ~M-x org-roam-find-file~ 
+may seem unfamiliar: after finishing a capture with ~C-c C-c~, you are returned  
+not to the original buffer from which you called ~M-x org-roam-find-file~, but 
+to a buffer pointing to the note you just created. For the usual ~org-capture~ 
+behavior you can call ~M-x org-roam-capture~ instead of ~M-x org-roam-find-file~. 
+
 Org-roam makes it easy to create notes, and link them together. To link notes
 together, we call ~M-x org-roam-insert~. This brings up a prompt with a list of
 title for existing notes. Selecting an existing entry will create and insert a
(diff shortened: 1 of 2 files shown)


doc/org-roam.org
```diff
@@ -370,6 +370,12 @@ time by setting ~org-roam-db-update-idle-seconds~; or change the cache update
 to be triggered immediately after buffer save by setting
 ~org-roam-db-update-method~ to ~'immediate~.
 
+For experienced ~org-capture~ users, the behavior of ~M-x org-roam-find-file~
```
d12frosted, line 373: Just reading this line makes we wonder, `find-file` is a common term in Emacs for finding and visiting file. Maybe the confusing part here is that it uses capture if file doesn't exist 😸

paultopia: Heh, for me, it's was mostly the surprise value of having `find-file` be the capture function described in the docs---there isn't currently any ref to the existence of a capture function that doesn't have find-file behavior, so I sort of assumed that the concept of find-file was being overloaded!  LOL...

d12frosted: Haha, indeed 😸 Confused me as well, when I used `org-roam-find-file` for the first time.

PR #1396, opened 2021-01-20
Feature/completing read tag position

marlinstrub:
Hi,

Thanks for your work on this great package.

This PR adds the `org-roam-file-completion-tag-position` option, which can
be used to control the position of `roam_tags` during file completion. It can
take one of the following values

  - `prepend`: Prepends tags to the file title
  - `append`: Appends tags to the file title
  - `omit`: Omits tags during completion.

I prefer to have the tags appended, but `prepend` is left as the default option.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -268,6 +268,13 @@ The currently supported symbols are:
            (symbol)))
   :group 'org-roam)
 
+(defcustom org-roam-file-completion-tag-position 'prepend
+  "Prepend, append, or omit tags from the file titles during completion."
+  :type '(choice (const :tag "Prepend" prepend)
+                 (const :tag "Append" append)
+                 (const :tag "Omit" omit))
+  :group 'org-roam)
+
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -852,6 +859,24 @@ file."
                  'face 'org-roam-tag))
    str))
 
+(defun org-roam--append-tag-string (str tags)
+  "Append TAGS to STR."
+  (concat
+   str
+   (when tags
+     (propertize (format " (%s)" (s-join org-roam-tag-separator tags))
+                 'face 'org-roam-tag))))
```
jethrokuan, line 868: Let's merge all these functions into `org-roam--add-tag-string`.

jethrokuan (commented):
Can you also add an entry into the changelog?

org-roam.el
marlinstrub: Done.

marlinstrub:
Sure. I've updated the changelog.

PR #1401, opened 2021-01-24
V2

jethrokuan:

See the [Discourse thread](https://org-roam.discourse.group/t/org-roam-major-redesign/1198)


org-roam-db.el
```diff
@@ -160,31 +138,24 @@ SQL can be either the emacsql vector representation, or a string."
...
+      mtime])
 
-    (ids
-     [(id :unique :primary-key)
+    (nodes
+     [(id :primary-key :not-null)
       (file :not-null)
-      (level :not-null)])
+      (level :not-null)
+      (pos :not-null)
```
d12frosted, line 148: Is it position in the buffer? Do I understand it correctly that for `(= level 0)` the `pos` is `(point-min)` and for `(> level 0)`, it is the point at the beginning of heading? It is nice to have it baked. I search for it time to time.

I must say, I love this structure, this has even more than I usually use:

```emacs-lisp
(cl-defstruct vulpea-note
  id
  title
  path
  tags
  level)
```

And with this structure, it will be much easier to query database 💯 

Side note, do you consider using structs for passing around data queried from `db`?

org-roam-db.el
```diff
@@ -160,31 +138,24 @@ SQL can be either the emacsql vector representation, or a string."
...
+      tags
+      title
+      ref])
 
     (links
-     [(source :not-null)
+     [(file :not-null)
+      (source :not-null)
       (dest :not-null)
       (type :not-null)
```
d12frosted, line 157: Is there a need for `type`? Or is it for migration?

org-roam-db.el
```diff
@@ -280,26 +231,56 @@ If UPDATE-P is non-nil, first remove the meta for the file in the database."
...
                           :where (= file $s1)]
                          file))
-    (org-roam-db-query
-        [:insert :into titles
-         :values $v1]
-        rows)
-    (length rows)))
+    (org-with-point-at 1
+      ;; First, we get the file-level ID
+      (when (setq id (org-id-get))
```
d12frosted, line 247: This could use `org-id-get-create` to enforce presence of `id`.

org-roam-db.el
```diff
@@ -280,26 +231,56 @@ If UPDATE-P is non-nil, first remove the meta for the file in the database."
...
+        (push (vector id file level pos tags title ref) nodes))
+      ;; Then we loop over all headlines
+      (org-map-entries
+       (lambda ()
+         (when (setq id (org-id-get))
+           (setq title (nth 4 (org-heading-components))
+              ;; TODO handle tags
+              tags nil
+              pos (point)
+              level (org-outline-level)
```
d12frosted, line 265: If you are using `org-heading-components`, you can get level from there, saves some parsing. Otherwise, instead of `org-heading-components` you can use `(org-entry-get nil "ITEM")`.

org-roam-db.el
```diff
@@ -280,26 +231,56 @@ If UPDATE-P is non-nil, first remove the meta for the file in the database."
...
+              ;; TODO handle ref
+              ref nil)
+        (push (vector id file level pos tags title ref) nodes))
+      ;; Then we loop over all headlines
+      (org-map-entries
+       (lambda ()
+         (when (setq id (org-id-get))
```
d12frosted, line 260: Same question about enforcement of `id`.

org-roam-db.el
```diff
@@ -280,26 +231,56 @@ If UPDATE-P is non-nil, first remove the meta for the file in the database."
...
+              ref nil)
+           (push (vector id file level pos tags title ref) nodes)))))
+    (when nodes
+      (org-roam-db-query
+       [:insert :into nodes
+        :values $v1]
+       nodes))))
+
+(defun org-roam-db-insert-links (&optional update-p)
+  (let ((file (or org-roam-file-name
```
d12frosted, line 276: It's not implemented yet, right? Maybe worth placing a TODO here 😸

org-roam.el
```diff
@@ -872,20 +872,12 @@ prepends TAGS to STR, appends TAGS to STR or omits TAGS from STR."
   "Return an alist for completion.
 The car is the displayed title for completion, and the cdr is a
 plist containing the path and title for the file."
-  (let* ((rows (org-roam-db-query [:select [files:file titles:title tags:tags files:meta] :from titles
-                                   :left :join tags
-                                   :on (= titles:file tags:file)
-                                   :left :join files
-                                   :on (= titles:file files:file)]))
+  (let* ((rows (org-roam-db-query [:select [file title tags pos] :from nodes]))
```
d12frosted, line 875: 🎉 this looks awesome now. Can't wait to simplify some of my queries like this 😸

d12frosted (commented):
I love how much code you remove and how things becoming simpler. I am looking forward to having this structure of db, would save time and effort not only for `org-roam` developers, but for developers of tools around `org-roam`. 💯 

P. S. How do you plan to test it? I can be a volunteer 😸

org-roam-db.el
jethrokuan: type is to retain support for other kinds of links (e.g. mail, cite)

org-roam-db.el
jethrokuan: Yeah, I'm thinking that for top level at least , we can use `(org-id-get-create)`.

org-roam-db.el
d12frosted: Ah. Good point. Don't use other types, so completely forgot that it might be useful outside of ID vs file links to notes.

org-roam-db.el
jethrokuan: Here I don't think we want to automatically create IDs. Then it's possible to create sub-headers that aren't part of Org-roam (e.g. a references section).

d12frosted: I wonder what UX would look like here. I like the idea of making possible not to have an ID generated automatically for selected headers, but how people would create an ID? By calling `org-id-get-create` (it's interactive function)?

Eldev
```diff
@@ -0,0 +1,29 @@
+; -*- mode: emacs-lisp; lexical-binding: t; no-byte-compile: t -*-
```
d12frosted, line 1: Oh nice. I like `eldev`. Using it for some time now and happy.

jethrokuan: yeah, it's pretty snappy.

jethrokuan:
Some roadbumps that need to be addressed at a later time:

1. foreign key doesn't work for me here in emacsql-sqlite3, altho emacsql advertises that it works. Maybe it works with emacsql-sqlite, but we made the decision to use emacsql-sqlite3 for good reason.
2. certain SQL functions don't play nice because of how objects are stored in the sqlite db. E.g. GROUP_CONCAT would've been really useful, but can't use it strings are stored as their elisp printed form in the database, rather than raw strings.

The most forward-looking way of addressing this is to move DB building and maintenance to a proper binary. This is looking a lot less hard now, but still requires some good amount of work.

d12frosted:
> foreign key doesn't work for me here in emacsql-sqlite3, altho emacsql advertises that it works.

Why do you need it? Just curious.

> we made the decision to use emacsql-sqlite3 for good reason.

Again, curiosity. Where can I read about that decision and the reasons behind? :) 

> certain SQL functions don't play nice because of how objects are stored in the sqlite db. E.g. GROUP_CONCAT would've been really useful, but can't use it strings are stored as their elisp printed form in the database, rather than raw strings.

Yes. I also had to write some very dirty and not-generally working queries because of this.

jethrokuan:
> Why do you need it? Just curious.

Just so deletes cascade down the different tables. E.g. if delete a node, its references and tags should also be deleted. The SQL queries are a lot cleaner.

> Again, curiosity. Where can I read about that decision and the reasons behind? :)

Because Windows users have great difficulty compiling the emacsql-sqlite binary (need mingw etc.)

d12frosted:
> Just so deletes cascade down the different tables. E.g. if delete a node, its references and tags should also be deleted. The SQL queries are a lot cleaner.

Makes sense. Thanks.

> Because Windows users have great difficulty compiling the emacsql-sqlite binary (need mingw etc.)

Alright. Windows is always a complex thing to consider.

org-roam-node.el
```diff
@@ -283,18 +294,61 @@ FILTER-FN is the name of a function to apply on the candidates
...
-      (let ((org-roam-capture--info `((title . ,node)
+          (find-file (org-roam-node-file node))
+          (goto-char (org-roam-node-point node)))
+      (let ((org-roam-capture--info `((title . (org-roam-node-title node))
                                       (slug  . ,(funcall org-roam-title-to-slug-function node))))
             (org-roam-capture--context 'title))
         (setq org-roam-capture-additional-template-props (list :finalize 'find-file))
         (org-roam-capture--capture)))))
 
+(defun org-roam-node-insert (&optional completionsn filter-fn)
```
Whil-: Potential typo here on `completionsn`?

Just trying to help. 😇

jethrokuan: Fixed, thanks!

theianjones:
For those of you wanting to see the discussion around this pr, [here's a good place to look](https://org-roam.discourse.group/t/org-roam-major-redesign/1198/168) 😄

kstuart:
- GNU Emacs 28.0.50 native-comp 686259e65aa7121683c0c65e45ce48adb08ddb58
- org-roam v2 3615969b355f1073c572894c204becaec61ba76f

When using selectrum and calling `org-roam-node-find` I get error from
`selectrum--update` **(wrong-type-argument org-roam-node nil)**.

The issue appears to be in function `org-roam-node--annotation` https://github.com/org-roam/org-roam/blob/3615969b355f1073c572894c204becaec61ba76f/org-roam.el#L641
When changing the line to `(tags (when node (org-roam-node-tags node))))` things work as expected, however as I'm new to emacs and common lisp I don't know if this is a _'correct'_ fix.

jethrokuan:
>     * GNU Emacs 28.0.50 native-comp 686259e65aa7121683c0c65e45ce48adb08ddb58
> 
>     * org-roam v2 [3615969](https://github.com/org-roam/org-roam/commit/3615969b355f1073c572894c204becaec61ba76f)
> 
> 
> When using selectrum and calling `org-roam-node-find` I get error from
> `selectrum--update` **(wrong-type-argument org-roam-node nil)**.
> 
> The issue appears to be in function `org-roam-node--annotation`
> 
> https://github.com/org-roam/org-roam/blob/3615969b355f1073c572894c204becaec61ba76f/org-roam.el#L641
> 
> When changing the line to `(tags (when node (org-roam-node-tags node))))` things work as expected, however as I'm new to emacs and common lisp I don't know if this is a _'correct'_ fix or the real issue lies elswhere.

Good job locating the cause! We actually don't use the annotation function to display tags anymore (since we want them searchable), so 624ac77 removes it, and should fix things.

org-roam.el
```diff
@@ -608,33 +608,35 @@ is the `org-roam-node'."
...
-                                    (complete-with-action action nodes string pred)))
-                                nil require-match initial-input)))
+         (node (completing-read
+                "Node: "
+                (lambda (string pred action)
+                  (if (eq action 'metadata)
+                      '(metadata
+                        (annotation-function . (lambda (title)
+                                                 (funcall org-roam-node-annotation-function
+                                                          (get-text-property 0 'node node-title))))
```
kstuart, line 629: looks like there's an issue with this change:
`(org-roam-node-find)` **Error in post-command-hook (selectrum--update): (void-variable node-title)**

jethrokuan: fixed, thanks

d12frosted:
@jethrokuan just curious, any roadmap on merging this? What is missing? 

I am asking this questions for three (highly subjective) reasons.

1. I see people referencing `v2` here and there, and this branch is kind of default now among users (don't have any numbers). And since `v2` is so much breaking for some people, it would not be nice to add extra effort for new users who still picked v1 by delaying migration to `v2`.
2. In some setups (e.g. when you are using `package.el` exclusively) it is hard to install `v2` branch.
3. I have migrated all code that depends on `org-roam` to `v2` and using this branch for almost a month and in my experience it is already stable 😸 

Since this branch is taking revolutionary approach instead of evolutionary, I keep asking myself, so when the revolution breaks out? 😸 

---

Side note. Have you considered releasing old version via MELPA (e.g. create a new branch called 'v1' and create a new package `org-roam-v1`), so people using `package.el` would easily stay at the old version untill they have time to migrate. Or maybe releasing `v2` as separate package? Again, since `v2` has many breaking changes, it might need some time to adapt to changes even with migration scripts (which I am sure will not cover all use cases anyways).

jethrokuan:
@d12frosted 

> @jethrokuan just curious, any roadmap on merging this? What is missing?

I'm thinking end of June. I've sort of come to terms with v2 never being able to be bug-free, so I'll likely release v2 sometime end June when a few more things are in place. I'm going to be putting in a few hours each day from now to end June on tidying up what's left.

- [ ] fully implement org-roam-graph (currently missing connected components)
- [ ] org-roam-doctor
- [ ] org-roam-dailies
- [ ] verify core functionality (e.g. org-roam-capture), org-roam-refile working fully
- [ ] documentation (migration instructions etc.)
- [ ] fixing support for multiple org-roam directories (might build that in first-class)

I also think v2 is more or less stable now, more so than v1 probably.

> Side note. Have you considered releasing old version via MELPA (e.g. create a new branch called 'v1' and create a new package org-roam-v1), so people using package.el would easily stay at the old version untill they have time to migrate. Or maybe releasing v2 as separate package? Again, since v2 has many breaking changes, it might need some time to adapt to changes even with migration scripts (which I am sure will not cover all use cases anyways).

Not sure if we want it to be easy to remain on v1, I'd like to prevent user-base fragmentation and v1 has some bugs that have remained unfixed for a while. I have little intention on supporting v1, if people prefer the v1 approach of using file links they have my permission to do the necessary (fork for MELPA).

doc/org-roam.org
```diff
@@ -1380,6 +1201,141 @@ NOTE: Since ripgrep cannot read encrypted files, this function cannot find
...
+the following functionalities:
+
+- Access to Org-roam's database
+- Usage/modification of Org-roam's interactive commands
+
+Org-roam provides no guarantees that extensions will continue to function as
+Org-roam evolves, but by following these simple rules, extensions can be made
+robust to local changes in Org-roam.
+
+1. Extensions should not modify the database schema. Any extension that requires
```
d12frosted, line 1250: I think that not all extensions make sense in Org-roam, because some applications have too specific needs. For one of my applications I've simply mimicked DB flow from Org-roam with custom scheme and extraction. And I decided not to go the length of building a generic tool for building custom schemes until I have another use case. And now I think that Org-roam could provide such functionality. E.g. as a user I simply provide a scheme and extractor function (that say takes an `id` and is executed in a buffer visiting note and returns a plist that needs to be inserted into db according to scheme) and Org-roam does all the lifting and maintenance. It should also provide versioning utilities via either migration function provided by user or full rebuild. This is something I've been thinking to implement, but now I wonder if there is interest in having it in Org-roam.

doc/org-roam.org
```diff
@@ -1380,6 +1201,141 @@ NOTE: Since ripgrep cannot read encrypted files, this function cannot find
...
+   the caching of additional data should make a request upstream to Org-roam.
+2. Extensions requiring access to the database should explicitly state support
```
d12frosted, line 1252: Do you mean that library authors should check for `org-roam-db--version` and if it's not of the right version, throw a `user-error` or degrade functionality? 🤔 This makes sense to me. The only think I would ask - make `org-roam-db--version` public (e.g. remove double dash).

doc/org-roam.org
```diff
@@ -1380,6 +1201,141 @@ NOTE: Since ripgrep cannot read encrypted files, this function cannot find
...
+
+The node interface is cleanly defined using ~cl-defstruct~. The primary
+method to access nodes is ~org-roam-node-at-point~ and ~org-roam-node-read~:
+
+- Function: org-roam-node-at-point &optional assert
+
+  Return the node at point. If ASSERT, throw an error if there is no node at
+  point.
+
+- Function: org-roam-node-read &optional initial-input filter-fn sort-fn
```
d12frosted, line 1281: BTW, I always was curious, and I think it's time to ask :) . Why `filter-fn` doesn't behave like `filter-fn` in `seq-filter` and instead of taking a `node`, it takes a list of nodes? I understand that this is so much more powerful, but I wonder how real is use-case where someone does some dirty stuff this the list of nodes (e.g. appends new elements).

doc/org-roam.org
```diff
@@ -1380,6 +1201,141 @@ NOTE: Since ripgrep cannot read encrypted files, this function cannot find
...
+
+
+Once you obtain the node, you can use the accessors for the node, e.g.
+~org-roam-node-id~ or ~org-roam-node-todo~.
+
+It is possible to define (or override existing) properties on nodes. This is
+simply done using a ~cl-defmethod~ on the ~org-roam-node~ struct:
+
+#+begin_src emacs-lisp
+(cl-defmethod org-roam-namespace ((node org-roam-node))
```
d12frosted, line 1298: I just realized that in the other comment I didn't understand that it's possible to use other prefix for this function 🤦 I think in this example you should not use `org-roam-` prefix, but something else, e.g. `my-org-roam-namespace`. Otherwise people might copy-paste it and risk a name clash. Because with `org-roam` prefix people extending Org-roam *need* to know all reserved/used names.

d12frosted (commented):
Nice development notes! Thanks for writing them 💯

doc/org-roam.org
jethrokuan: Note: everything in the properties drawer is cached, and you can define your own accessor using `cl-defmethod` and reading the cached properties: `(org-roam-node-properties node)`, which should fulfill most users needs (e.g. having a last modified property for each node)

doc/org-roam.org
jethrokuan: Yeah that's what I meant.

> The only think I would ask - make org-roam-db--version public (e.g. remove double dash).

I will.

doc/org-roam.org
jethrokuan: > Why filter-fn doesn't behave like filter-fn in seq-filter and instead of taking a node, it takes a list of nodes?

Probably an oversight of mine while trying to emulate v1's signature, will fix

doc/org-roam.org
jethrokuan: There's no `org-roam-namespace` now, so no worries about that. Also, I meant the `org-roam-node-` prefix, will fix the typo. Because of how it's currently implemented, implementing with the `org-roam-node-` prefix will make it available in the capture templates, which is the most common use-case.

Org-roam doesn't exactly reserve names, think the rule is not to override the existing interactive functions (e.g. `org-roam-node-read`), or accessors. But everything else is okay to override (e.g. `org-roam-node-slug`).

doc/org-roam.org
d12frosted: I am bad at explaining things :) So I simply mean that sometimes you need custom database for subset of your notes (or all notes). And if it contains data that is not cached by org-roam, the only solution is to mimic org-roam solution for db. Maybe it's a very rare case and only I faced it. In any case, it is irrelevant to your doc update :)

doc/org-roam.org
d12frosted: > There's no org-roam-namespace now, so no worries about that.

Yes, **now**. And that's the point. You never know what names users are using and you never know what names you will use in the future. Sharing namespace is begging for problems :) 

> Because of how it's currently implemented, implementing with the org-roam-node- prefix will make it available in the capture templates, which is the most common use-case.

Can you use full name of accessor in capture templates? That would solve the issue.

> Org-roam doesn't exactly reserve names

I mean, the way lisp is loaded, users might accidentally override functions defined in `org-roam`. 

> the rule is not to override the existing interactive functions

It is not future proof :)

Wetlize:
I'm not sure what the plans here for dailies, but wouldn't it make sense to decouple them from the notion of files? So instead of tracking dailies as files that live in some sort of special directory, with the current system, it can be implemented by just having some sort of special property that would mark the node as "daily":
```org
* Node
:PROPERTIES:
:ID: [id]
:ROAM_DAILY: [time stamp]
:END:
```
This way one can use an arbitrary structure for their dailies (e.g. one giant file, monthly file, per year file, the current structure, etc), instead of being forced to use a unique file per each day. This isn't something that's important for me, but I can see how this can be useful for others. WDYT?

Wetlize:
@jethrokuan Do you intentionally not set all the node fields in `org-roam-node--completions`? When using `FILTER-FN` in `org-roam-node-read` I would intuitively expect to have access to all the fields defined in `org-roam-node`:

https://github.com/org-roam/org-roam/blob/4fc346861c7d4aac83c91760c4a5281458a95436/org-roam.el#L413-L417

with the corresponding accessor functions, but `org-roam-node--completions` only sets a strict subset of all the available fields:

https://github.com/org-roam/org-roam/blob/4fc346861c7d4aac83c91760c4a5281458a95436/org-roam.el#L662-L676

and it looks like the avoided fields are duplicated under `properties` slot. If this is done intentionally, do the users then have to use something like `(assoc-string "TODO" (org-roam-node-properties node))` in order to access the missing fields?

jethrokuan:
> @jethrokuan Do you intentionally not set all the node fields in `org-roam-node--completions`?

Yes, but only because it's expensive to retrieve every property of the node from the database. TODO is easy to add in and shouldn't be much more expensive, but those that require table joins are issues. I'll do abit of profiling and see how far it can be taken.

d12frosted:
@jethrokuan in vulpea I am doing horrific joins and it works pretty fast, though I haven't really benchmarked it. I just know that compared to sequential population it works much faster. Example of the query:

```emacs-lisp
          (org-roam-db-query
           "select
  id,
  file,
  title,
  \"level\",
  '(' || group_concat(tags,
  ' ') || ')' as tags,
  aliases
from
  (
  select
    nodes.id as id,
    nodes.title as title,
    nodes.file as file,
    nodes.\"level\" as \"level\",
    tags.tag as tags,
    '(' || group_concat(aliases.alias, ' ') || ')' as aliases
  from nodes
  left join tags on tags.node_id = nodes.id
  left join aliases on aliases.node_id = nodes.id
  group by nodes.id, tags.tag )
group by id, aliases;")
```

AFAIK, this query can't be translated into proper `emacsql` expression, so you have to pass SQL as a string here. If you wish to make another join, it will become even more horrific 😨 Hope that helps.

Wetlize:
@jethrokuan Can you clarify please the next questions: 
1. During capture an ID is only automatically generated for files, correct? So, for example, if I have `:if-new (file+olp ...)` or `:if-new (file+datetree ...)` templates, only the file itself will automatically get a new ID? For everything else we supposed to produce ID by ourselves?
2. Same applies if the `type` of capture template set to `entry`? Only file will automatically get an ID and for the `entry` itself we supposed to set ID ourselves, either, interactively using `org-id-get-create` or by writing a capture template like that:
```emacs-lisp
("e" "Entry" entry
 ,(mapconcat #'identity
             '("* ${title}%?"
               ":PROPERTIES:"
               ":ID: %(org-id-new)"
               ":END:")
             "\n")
 :if-new (file+olp "file.org" ("Path 1" "Path 2" "Path 3")))
```
Am I following this right?

org-roam.el
```diff
@@ -644,40 +644,73 @@ Throw an error if multiple choices exist."
...
+     LEFT JOIN tags ON tags.node_id = nodes.id
+     LEFT JOIN aliases ON aliases.node_id = nodes.id
+     LEFT JOIN refs ON refs.node_id = nodes.id
+     GROUP BY nodes.id, tags.tag )
+   GROUP BY id, aliases, refs;")))
+    (cl-loop for row in rows
+             append (pcase-let* ((`(,id ,file ,title ,level ,todo ,pos ,priority ,scheduled
+                                        ,deadline ,properties ,olp ,atime ,mtime ,tags ,aliases, refs) row)
+                                 (all-titles (cons title aliases))
+                                 (nodes (mapcar (lambda (t)
```
Wetlize, line 695: This will error with `Attempt to set a constant symbol: t`, because `t` and `nil` are self-evaluating symbols in Elisp.

Wetlize (changes requested):

jethrokuan:
@Wetlize 

I don't plan to change make drastic changes to dailies, unless someone can bring forth a convincing case. 

> During capture an ID is only automatically generated for files, correct? So, for example, if I have :if-new (file+olp ...) or :if-new (file+datetree ...) templates, only the file itself will automatically get a new ID? For everything else we supposed to produce ID by ourselves? (for both, new entry itself and for outlines in the defined OLP).

Yes, that's the current behaviour. I'm undecided on whether the OLP should get IDs as well, and whether the entry should link to the OLP ID or the file ID. I don't use `olp` or `datetree`, so I guess this depends on which use-case is more common.

> Is the same applies if the type of capture template set to entry? Only file will automatically get an ID and for the entry itself we supposed to set ID ourselves, either, interactively using org-id-get-create or by writing a capture template like that:

I think we might want entries to get IDs, once I start building out the `org-roam-agenda` infrastructure it will become more useful.

3. Can we control whether files (say, in specific templates) will receive or not an automatic ID during capture? I can imagine there's can be situations where I would not need for a file to have an ID.

All functions in Org-roam assume you intend to create a node, and for now the node is the file. So unless any of the above behaviour changes, file node IDs will always be created.

---

I'm quite undecided on what the expected behaviour should be for these features, I don't use them. I'm personally sticking with a one-node-per-file infrastructure as it still lends the most flexibility. It'll be helpful if you can outline your use-case and the behaviour you want.

Wetlize:
@jethrokuan 

I've asked these questions about ID generation to ensure that the current behavior wasn't a bug, but a design decision, so I could try build on top some things accordingly :)

> It'll be helpful if you can outline your use-case and the behaviour you want.

I plan to try track the whole `org-directory` with v2, which would include a mixed bug of things, like personal task management, other projects related information (like contacts), personal journaling, note taking, and backlog of various resources. Such things would reside in its own subdirectories under `org-directory`, e.g. most tasks under `agenda/` (some things like timestamps and schedule/deadlines for agenda can still appear anywhere in the org-directory), notes under `notes/`, dailies under `journal/` and so on.

Tasks would be linked with each other through `task:id` links and would have its own `magit-section` displayed in `org-roam-mode`. There's also extra versions of `org-roam-node-{find,insert,capture}` for tasks and non-task based nodes and I probably try integrate other default `org-mode` for links with this system (like link storage and `org-insert-link` with completion menu).

Such workflow makes it really desirable to have a total control over automatic ID generation, because in some cases they can be useful, while in other they would simply add unwanted noise. Having some sort of facility to control per each template where I would need to have an automatic ID prescribed would be really nice, be it a file, a specific subset of headings in OLP, entry itself or a combination of such things. Maybe a special property akin to `:if-new` (`:id`?) to control that would be nice to have.

If you need a more detailed explanation on some things in the workflow, feel free to ping me again.

Wetlize:
@jethrokuan you could push [warning](https://www.gnu.org/software/emacs/manual/html_node/elisp/Warning-Basics.html)  about upcoming soon™ breaking changes with v2 to v1 users, so people who don't pin packages wouldn't be surprised that the package suddenly died. The warning could include migration instructions to v2 and link to v1's repo and MELPA recipe. To not annoy people too much with the warning you could flag warning behind some variable, that users would need set from `t` to `nil` in order to disable it.

org-roam.el
```diff
@@ -519,1159 +301,580 @@ Use external shell commands if defined in `org-roam-list-files-commands'."
...
+
+;;; Org-roam setup and teardown
+(defvar org-roam-find-file-hook nil
+  "Hook run when an Org-roam file is visited.")
+
+(defun org-roam-setup ()
+  "Setup Org-roam."
+  (interactive)
+  (unless (or (and (bound-and-true-p emacsql-sqlite3-executable)
+                   (file-executable-p emacsql-sqlite3-executable))
```
herbertjones, line 357: Looks like these may be intended to be `emacsql-sqlite-executable` as the ones here aren't defined in emacsql.

org-roam-db.el
```diff
@@ -78,6 +78,11 @@ value like `most-positive-fixnum'."
   :type 'int
   :group 'org-roam)
 
+(defcustom org-roam-db-node-exclude-function (lambda () t)
+  "A custom function to check if the headline at point as a node."
```
rodelrod, line 82: "*i*s a node"?

org-roam-db.el
```diff
@@ -276,7 +276,7 @@ If UPDATE-P is non-nil, first remove the file in the database."
   "Return t if headline at point is a node, else return nil."
   (and (org-id-get)
        (not (cdr (assoc "ROAM_EXCLUDE" (org-entry-properties))))
-       (funcall org-roam-db-node-exclude-function)))
+       (funcall org-roam-db-node-include-function)))
```
rodelrod: since this is inside an `and` with 2 other predicates, `org-roam-db-node-include-function` is not enough to guarantee the node is included. A clearer expression of the behaviour would be: 

```
(not (funcall org-roam-db-node-exclude-function))))
```

rodelrod: @jethrokuan  using the `org-roam-db-node-include-function` to exclude org-attach headings, I get a failure rebuilding the cache. There is a FK constraint violation in `links.source` → `nodes.id`. This occurs when it tries to insert the org-attach link from the heading to the attached file because the heading must be present in the `nodes` table.

rodelrod:
@jethrokuan  using the `org-roam-db-node-include-function` in the latest commits to exclude org-attach headings, I get a failure rebuilding the cache. There is a FK constraint violation in `links.source` → `nodes.id`. This occurs when it tries to insert the org-attach link from the heading to the attached file because the heading must be present in the `nodes` table.

rodelrod:
I had to revert to 0547dc723ddca5265d0e746c51cf2e682ab05866 to be able to rebuild the cache. Even without setting `org-roam-db-node-include-function` or setting it to `(lambda () t)` it would get stuck processing the first file.

rodelrod:
@jethrokuan Problem solved, both ROAM_EXCLUDE and `org-roam-db-include-function` work for me, thanks!

In the docs the variable is still called `org-roam-db-exclude-function` but I'm not sure if you are reconsidering the naming or the logic.

org-roam-db.el
jethrokuan: Here I wanted `ROAM_EXCLUDE` to always work, and also all nodes need an ID. so they take precedence over the include function.

rodelrod: That precedence would still hold with the logic I'm proposing:

```
  (and (org-id-get)
       (not (cdr (assoc "ROAM_EXCLUDE" (org-entry-properties))))
       (not (funcall org-roam-db-node-exclude-function)))
```

Which comes down to `(and has-id  (not is-excluded))` where `is-excluded` is `(or has-ROAM_EXCLUDE is-org-roam-db-node-excluded)`. It just feels closer to the intended semantics in the config file, and draws up a nice parallel with `ROAM_EXCLUDE`, making the final behavior easier to predict by the user:

```
(setq org-roam-db-node-exclude-function
      (lambda ()
        (member "ATTACH" (org-get-tags))))
```

myshevchuk:
Was this really necessary in https://github.com/org-roam/org-roam/pull/1401/commits/409a1476b1f3d64292ecff872411cdafd0f69c2e:

```
-  (org-roam-node-create :id id
-     :point (point))))))))))
+  (org-roam-node-create :id id))))))))
```

It has broken adding references in a new file, if the reference is not in the database yet and the node has no point associated with it. Although I've provided a workaround in ORB (see https://github.com/org-roam/org-roam-bibtex/issues/193), this may also affect Org-roam protocol (haven't checked if this this the case).

jethrokuan:
@myshevchuk right, I forgot why the `:point` was there. I'll revert the behaviour and add a note in the function doc.

TimQuelch:
Fantastic work. Thanks @jethrokuan.

d12frosted:
Happy day! Congratulations @jethrokuan! Huge thanks for your effort and kudos to everyone who helped this release to happen! Happy orging! 🎉 ❤️

PR #1404, opened 2021-01-25
(fix) inconsistency between props writing and reading (alias, tags)

### LINKED ISSUE #1403: org-roam-tag-add automatically deletes the double quotation marks of the tags that already exist.
I use semicolons to indicate the hierarchy of notes tag.
when tag has semicolon, it will cause an error. 
image

so, I add the double quotation marks.
image

but, When I try to add a tag using org-roam-tag-add, it automatically removes the double quotes.
image

d12frosted:
Fixes #1403

Right noow we have an inconsistency between how we write and read props like tags and aliases. I might be missing the rasons for using `org-roam--str-to-list`, but it seems that it's purpose was to just handle not balanced quotation marks `"`. If my understand is correct, I think we should use `split-string-and-unquote` (opposite to `combine-and-quote-strings` that we are using for writing props). It has simple implementation and there is no need to convert between other types to string.

But again, I might be missing something.

```
ELISP> (split-string-and-unquote "hello my friends")
("hello" "my" "friends")

ELISP> (split-string-and-unquote "hello \"my friends")
*** Eval error ***  End of file during parsing

ELISP> (split-string-and-unquote "hello \"my\" friends")
("hello" "my" "friends")

ELISP> (split-string-and-unquote "hello ;tag friends")
("hello" ";tag" "friends")

ELISP> (org-roam--str-to-list "hello my friends")
("hello" "my" "friends")

ELISP> (org-roam--str-to-list "hello \"my friends")
*** Eval error ***  End of file during parsing

ELISP> (org-roam--str-to-list "hello \"my\" friends")
("hello" "my" "friends")

ELISP> (org-roam--str-to-list "hello ;tag friends")
*** Eval error ***  End of file during parsing
```

Some benchmarks:

```
ELISP> (benchmark-run 100 (split-string-and-unquote "hello \"my\" friends"))
(0.000531 0 0.0)

ELISP> (benchmark-run 100 (org-roam--str-to-list "hello \"my\" friends"))
(0.003681 0 0.0)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - [#1352](https://github.com/org-roam/org-roam/pull/1352) fixed org-roam-{tag/alias}-{add/delete} altering the original case of the Org property
 - [#1374](https://github.com/org-roam/org-roam/pull/1374) fix headline completions erroring out
 - [#1375](https://github.com/org-roam/org-roam/pull/1375) fix org-roam-protocol to use existing ref file 
+- [#1403](https://github.com/org-roam/org-roam/issues/1403) fixed inconsistency between how we write and read props like alias and tags
 
 ## 1.2.3 (13-11-2020)
 
--- a/org-roam-doctor.el
+++ b/org-roam-doctor.el
@@ -53,7 +53,6 @@
 (declare-function org-roam--get-roam-buffers "org-roam")
 (declare-function org-roam--list-all-files "org-roam")
 (declare-function org-roam--org-roam-file-p "org-roam")
-(declare-function org-roam--str-to-list "org-roam")
 (declare-function org-roam-mode "org-roam")
 
 (defvar org-roam-verbose)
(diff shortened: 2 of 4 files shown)


jethrokuan:
LGTM, thanks! Appreciate the examples and benchmarks.

Imymirror:
this commit not fixes [#1403](https://github.com/org-roam/org-roam/issues/1403). 

I try org-roam-tag-add, it deletes the double quotation marks as before.

d12frosted:
> I try org-roam-tag-add, it deletes the double quotation marks as before.

Like I said in that ticket, removal of quotes is intentional. They are added only when needed. What this PR fixes - inconsistency between reading and writing tags and aliases. Previously you hit an issue when one logic was used to write them and the other to read. Now I can add tags like `;category` and `org-roam` is happy:

image

d12frosted:
@jethrokuan I see another issue with my code. 🤦 if the prop is missing, then I get an error:

```
ELISP> (split-string-and-unquote nil)
*** Eval error ***  Wrong type argument: stringp, nil
```

Will send a quick PR.

PR #1413, opened 2021-02-02
Fix & add to Link Completion doc

angrycandy:
- Fix missing '|' in the roam link.

- Add org-toggle-link-display to see literal link text instead of the
  description.

As a newbie to org-roam and a longtime user of org-mode, this was the only section of the doc that tripped me up so I'm fixing it for the next newbie.


d12frosted (approved):
Nice. I think it's good to mention link display toggle.

PR #1420, opened 2021-02-03
(feat) support passing tag/alias to add/delete functions

### LINKED ISSUE #1416: Add optional argument to org-roam-(add|delete)-(tag|alias) functions
### Brief Abstract

Now functions for add/delete tags can be used only by user, not from scripts. Adding optional argument can add possibility for creating e.g. user defined hydra for set/unset some tags.

d12frosted:

Fixes #1416

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1689,13 +1689,15 @@ command will offer you to create one."
         (org-roam-find-file "Index")))))
 
 ;;;###autoload
-(defun org-roam-alias-add ()
-  "Add an alias to Org-roam file.
+(defun org-roam-alias-add (&optional alias)
+  "Add an ALIAS to Org-roam file.
+
+When ALIAS is nil, it is read interactively.
 
 Return added alias."
   (interactive)
   (unless org-roam-mode (org-roam-mode))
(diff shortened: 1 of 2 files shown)


org-roam.el
```diff
@@ -1707,26 +1709,33 @@ Return added alias."
...
-      (let ((alias (completing-read "Alias: " aliases nil 'require-match)))
+      (let ((alias (or alias
+                       (completing-read "Alias: " aliases nil 'require-match))))
         (org-roam--set-global-prop
          "roam_alias"
          (combine-and-quote-strings (delete alias aliases)))
         (org-roam-db--update-file (buffer-file-name (buffer-base-buffer))))
-    (user-error "No aliases to delete")))
+    ;; error only during interactive usage
+    (unless alias
```
d12frosted, line 1727: I've added `user-error` here just to give feedback to the interactive users of this function when there is nothing to delete. I think for non-interactive usages it's better to simply noop. Same for tags.

d12frosted:
Now in V2 - https://github.com/org-roam/org-roam/pull/1515

PR #1445, opened 2021-03-04
(fix) Capture template initial text

### LINKED ISSUE #1433: Copy selection from browser when using it with org-protocol
What would be the org-roam-capture-templates / bookmark config when wanting to include selected text in the capture frame? Thanks.

seanfarley:

Mostly to fix #1433 but also discuss the format of the file name (slug).

### DIFF
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -72,11 +72,11 @@ It opens or creates a note with the given ref.
              (orglink
               (org-link-make-string ref (or (org-string-nw-p title) ref))))
         (when org-roam-protocol-store-links
-          (push (list ref title) org-stored-links)
-          (org-link-store-props :type type
-                                :link ref
-                                :annotation orglink
-                                :initial body))))
+          (push (list ref title) org-stored-links))
+        (org-link-store-props :type type
+                              :link ref
+                              :annotation orglink
+                              :initial body)))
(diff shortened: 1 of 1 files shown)


seanfarley:
Follow-up question: is there a technical reason that `%i` isn't included in the default `org-roam-capture-ref-templates`?

jethrokuan:
> Follow-up question: is there a technical reason that %i isn't included in the default org-roam-capture-ref-templates?

No, that's a mistake. Although `%i` could really be replaced with `{body}` in the ref template.

> Discuss the format of the file name (slug)

It's probably not a good idea to change the defaults at this time.

seanfarley:
> > Follow-up question: is there a technical reason that %i isn't included in the default org-roam-capture-ref-templates?
> 
> No, that's a mistake. Although `%i` could really be replaced with `{body}` in the ref template.

But wouldn't that mean `{body}` is _always_ required? e.g. I wouldn't be able to `org-roam-capture` a webpage without selecting text
 
> > Discuss the format of the file name (slug)
> 
> It's probably not a good idea to change the defaults at this time.

That's too bad.

I'll update the PR to drop that change.

jethrokuan (approved):

seanfarley:
Thanks!

PR #1457, opened 2021-03-25
Set ROAM_REFS during capture

myshevchuk:

The below commit allows for setting ROAM_REFS in the capture buffer from `org-roam-capture--info`. This is the first step toward re-enabling `org-roam-protocol` and `org-roam-bibtex`. Since an Org Roam file can have several refs, and it is in principle possible to capture into an existing file, the property should be appended rather than overwritten. As such, this is not ready yet for merging.


jethrokuan:
Added in 09b0d74

myshevchuk:
Great, thank you!

PR #1462, opened 2021-04-06
[V2] make org-roam-dailies-xxx work

alanz:
I found `org-roam-dailies-find-today` and `org-roam-dailies-capture-today` were not working.


jethrokuan:
Thanks, merged the correct fix into the v2 branch.

Issue #1467, opened 2021-04-13
[V2] `org-roam-node--completions` does not highlight matching string & it does not show alias


nobiot:
### Description

I am sending a PR to show what I mean more precisely.
Not an urgent issue; it would be great if you could look at the PR.

Two issues. 

1. Completion does not highlight matching strings
  This could be related to my set up. Screen shot and detail of my set up below
2. Aliases can be matched but not visible

I could only address matching string issue; I could not make aliases appear in the list of candidates. My PR actually makes aliases not even matching (I know why this is the case, but could not fix the problem).

![Snag_2b407f9](https://user-images.githubusercontent.com/12507865/114610636-6fd71880-9ca0-11eb-82b9-f63b0a040dd2.png)
**Figure 1.** Illustration of Issue 2. 
The org file in the main buffer has two aliases. The search in the completion matches the two aliases, but they are not visible -- `org-roam-node--format-entry` function for the display property does not appear to process aliases.

I could not make aliases appear and searchable at the same time with re-purposing `org-roam-node--format-entry`. 

![Snag_2b42610](https://user-images.githubusercontent.com/12507865/114610645-74033600-9ca0-11eb-8313-f5fd1908ccd3.png)
**Figure 2.** Illustration of issue 1. 
This is what completion looks like in my set up. Compare it with Figure 3 below after the PR is implemented. For completion, I use the standard Emacs completion (not even Icomplete) with [Vertico ](https://github.com/minad/vertico)and [Orderless](https://github.com/oantolin/orderless). Not sure if Selectrum you use has no issue with the display. The matched string in the display property does not get highlighted.

![Snag_2b43aff](https://user-images.githubusercontent.com/12507865/114610657-7796bd00-9ca0-11eb-995d-6098581906ab.png)
**Figure 3.** Illustration of issue 1, after PR.

#### Steps to Reproduce

Call `org-roam-node-find`, and type some string.

#### Backtrace

Not applicable

#### Expected Results

1. Completion minibuffer highlights matching strings
2. Aliases visible and searchable

#### Actual Results

See the images above.
1. Completion minibuffer does not highlight matching strings (at least in my set up with standard Emacs completion, Vertico and Orderless)
2. Aliases are searchable, but not visible

My PR at this stage does not address issue 2 properly -- aliases are not visible, and become unsearchable as it is not part of the cons.

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-w64-mingw32)
 of 2020-08-21
[...]

PR #1468, opened 2021-04-13
Fix/completing read, addressing #1467

nobiot:
See #1467

The PR is meant to clarify my explanation of the issue.
The PR is not complete -- it breaks the feature to use aliases for search in `org-roam-node-find`. I didn't test `org-roam-capture`.


nobiot:
Commit f52c3a47d42bbeab251a3b43ba41ca44d2841e7e fixes the problem. This PR is no longer needed.

PR #1469, opened 2021-04-15
fix: org-roam-node--completions #1467

nobiot:
@jethrokuan 

Thank you for f52c3a4.
`candidate-main` string also needs to pass node as `node` text-property.

I get the following error -- the error should not be vertico specific.

> post-command-hook (vertico--exhibit): (wrong-type-argument org-roam-node nil)

EDIT: The node prop needed for invisible text strings, too.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -613,7 +613,7 @@ is the `org-roam-node'."
                                   (candidate-main (org-roam-node--format-entry node (1- (frame-width))))
                                   (tag-str (org-roam--tags-to-str (org-roam-node-tags node))))
                        (cons (concat (propertize tag-str 'invisible t)
-                                     candidate-main
+                                     (propertize candidate-main 'node node)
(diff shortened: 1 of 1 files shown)


jethrokuan:
should be fixed in 624ac77

nobiot:
Thank you. I will pull the repo later :)

nobiot:
@jethrokuan 

Just as an idea for annotation.
I apprecaited the fact that I could access nodes in the annotation function.

I could do this, as noted in the [wiki](https://github.com/org-roam/org-roam/wiki/Hitchhiker's-Rough-Guide-to-Org-roam-V2#showing-the-number-of-backlinks-for-each-node-in-org-roam-node-find).  I wondered if you could reconsider adding some way to access node for annotation. This way, we can customize the annotation the way we like.

(I agree that tags in the annotation are redundant).

PR #1473, opened 2021-04-16
(feat) Make roam headline link completion robust #1472

### LINKED ISSUE #1472: Make `org-roam-link-complete-at-point` robust
### Brief Abstract
Use `org-link-escape` for its internal link insertion process in `org-roam-link-complete-at-point`.
### Long Description
While, with appreciation, enjoying your masterpiece, org roam, I found that it cannot link to roam file's headline that includes some other links. It should be fixed so that one can insert well-formed org link that happens to include other org link in it.  
Just wrap `str` in the `:exit-function` of `org-roam-link-complete-at-point`, just that.

HyunggyuJang:
Fixes  #1472

Headlines that includes links to others should be handled properly.

### DIFF
--- a/org-roam-link.el
+++ b/org-roam-link.el
@@ -318,7 +318,7 @@ DESC is the link description."
                                 (if headline-only-p 1 0)))
                 (insert (concat (unless (string= link-type "roam") "roam:")
                                 (when headline-only-p "*")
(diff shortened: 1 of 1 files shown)


jethrokuan:
Thanks!

Issue #1474, opened 2021-04-16
org-roam-refile is not required by org-roam.el


jmay:
`org-roam-refile` is very useful, but the file is not currently `require`d by `org-roam.el` (v2). Is there a reason for leaving this out?

When at the top level of file (outside of any heading), `refile` should transfer the entire contents of the file into the target node as a sub-node (including the top-level PROPERTIES drawer with the ID property).

jethrokuan:
Leaving it out first because it's not entirely ready... was just a quick experiment I was doing, but I intend to finish it.

PR #1481, opened 2021-04-20
Allow refile for top-level org-roam nodes.

jmay:
See https://github.com/org-roam/org-roam/issues/1474

org-roam-refile can now be invoked at the top level of any org-roam file, above the topmost heading line.

If the refile operation results in the source buffer being empty, then any matching file is deleted and buffer is killed.


jmay:
Cancelling this, will update & resubmit against latest master

PR #1495, opened 2021-04-29
speed up org-roam-db-sync by wrapping db activity in a transaction

zot:
[copying the commit comment here]
In my case of testing with 8253 zettels (available on request if
anyone wants to test with them), this improves cleanup speed by 6x
and update speed by 2x.

Improvements will be smaller for smaller batches as this really just
removes the transaction overhead.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -435,18 +435,19 @@ If FORCE, force a rebuild of the cache from scratch."
                          contents-hash)
           (push file modified-files)))
       (remhash file current-files))
-    (if (fboundp 'dolist-with-progress-reporter)
-        (dolist-with-progress-reporter (file (hash-table-keys current-files))
-            "Clearing removed files..."
-          (org-roam-db-clear-file file))
-      (dolist (file (hash-table-keys current-files))
-        (org-roam-db-clear-file file)))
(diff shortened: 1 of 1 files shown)


jethrokuan:
This is honestly kinda surprising, when I was working on v1 there was no difference when things were wrapped in a transaction and when they were not. I also came across this article on sqlite: https://sqlite.org/np1queryprob.html

That suggests that the current mutliple-query style is acceptable for sqlite. Curious where the speed gains are from, but shouldn't make a difference to me.

zot:
I checked the article but I don't see any mention of whether or not to use transactions in the article but rather a discussion about the number of SQL statements to use and the appropriate size of queries (like one insert vs multiple inserts for a lot of rows).

My code change doesn't reduce the number of SQL statements, it wraps them in a transaction.

There is an overhead for each transaction. If you don't wrap statements in a transaction, each one gets its own transaction, which is sometimes called "autocommit" mode. The overhead is small but it adds up, so you won't see much difference with a small number of files. You will see it with 8000+ files, however :).

Here's an article about that: https://medium.com/@JasonWyatt/squeezing-performance-from-sqlite-insertions-971aff98eef2

jethrokuan:
That makes a lot of sense, thanks!

PR #1500, opened 2021-05-01
fix no node at point during capture

myshevchuk:

Fix "No node at point" error when creating a node from a ref. See https://orgroam.slack.com/archives/C011XLMELUB/p1619564042037200 for a discussion.

In `org-roam-capture--get-point`, when creating a node from ref, i.e. following the first clause of `cond` because there is a ref and then the second clause of `if-let` because the node does not exist yet, the function `org-roam-capture--goto-location` is called.

First, `org-id-get-create` should be called before running any hooks, which may rely on an existing ID, as is indeed the case of `org-roam-capture--insert-ref`.

Second, just fixing this will lead to an `integer-or-marker nil` error because the node does not exist yet in the database. `org-roam-node-at-point` returns a node having only an ID with all other properties nil, including `point`. This causes `goto-char` to throw the error.

Maybe the second fix can be generalized? There are a few other places where `(goto-char (org-roam-node-point node))` assumes the point is non-nil, which may not necessarily be true.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -570,8 +570,8 @@ Return the ID of the location."
          (org-end-of-subtree t t))))
     (save-excursion
       (goto-char p)
-      (run-hooks 'org-roam-capture-new-node-hook)
-      (org-id-get-create))))
+      (org-id-get-create)
+      (run-hooks 'org-roam-capture-new-node-hook))))
 
 (defun org-roam-capture-find-or-create-olp (olp)
   "Return a marker pointing to the entry at OLP in the current buffer.
--- a/org-roam.el
+++ b/org-roam.el
@@ -472,36 +472,37 @@ OLD-FILE is cleared from the database, and NEW-FILE-OR-DIR is added."
 Uses the ID, and fetches remaining details from the database.
 This can be quite costly: avoid, unless dealing with very few
 nodes."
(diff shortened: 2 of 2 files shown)


alanz:
Works for me, thanks.

myshevchuk:
To address the second problem, I can suggest several possible solutions, from less to more general:

1. Provide a default value for `point` in `org-roam-node` structure. This could be `(point-min)` or simply 1.

2. Modify `org-roam-node-populate` so that it will set a slot's value to the corresponding value of the argument node if the value from the database is nil, e.g.:

```elisp
...
(setf (org-roam-node-file node) (or file (org-roam-node-file node))
         ...)
```

This way it would be possible to have statements like `(org-roam-node-populate (org-roam-node-create :id id :point (point-min)))` with nil values from the database replaced with non-nil values from the supplied node.

3. Implement a sort of union function `(org-roam-node-merge node1 node2)` that would return a new node with slot values merged from NODE1 and NODE2. This function might probably be useful elsewhere. With regard to the current issue, it could be used by say `(org-roam-node-merge-from-db (org-roam-node-create :id id :point (point-min)))`, which would essentially call `(org-roam-node-merge (org-roam-node-populate node) node)`, thus leaving the original `org-roam-node-populate` untouched.

myshevchuk:
Thanks!

PR #1501, opened 2021-05-01
35% faster completion startup (on ~8300 files)

zot:
Caching: make process-display-format cache its value
SQL: add index to DB, make SQL compute objects for query-nodes.
  Note that there is a dependency between the org-roam-node def and
  org-roam-query-nodes. This is indicated at both places in the code.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -621,6 +621,7 @@ Throw an error if multiple choices exist."
   "Return an alist for node completion.
 The car is the displayed title or alias for the node, and the cdr
 is the `org-roam-node'."
+  (org-roam--clear-cached-display-format)
   (let ((tags-table (org-roam--tags-table)))
     (cl-loop for row in (append
                          (org-roam-db-query [:select [file pos title title id properties olp]
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -435,18 +435,19 @@ If FORCE, force a rebuild of the cache from scratch."
                          contents-hash)
           (push file modified-files)))
       (remhash file current-files))
-    (if (fboundp 'dolist-with-progress-reporter)
-        (dolist-with-progress-reporter (file (hash-table-keys current-files))
-            "Clearing removed files..."
-          (org-roam-db-clear-file file))
-      (dolist (file (hash-table-keys current-files))
-        (org-roam-db-clear-file file)))
-    (if (fboundp 'dolist-with-progress-reporter)
-        (dolist-with-progress-reporter (file modified-files)
-            "Processing modified files..."
-          (org-roam-db-update-file file))
-      (dolist (file modified-files)
-        (org-roam-db-update-file file)))))
+    (emacsql-with-transaction (org-roam-db)
+      (if (fboundp 'dolist-with-progress-reporter)
+          (dolist-with-progress-reporter (file (hash-table-keys current-files))
+              "Clearing removed files..."
+            (org-roam-db-clear-file file))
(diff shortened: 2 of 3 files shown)


zot:
This adds two indexes to the database (on nodes.id and aliases.node_id) that speeds up the join quite a bit.

I did not update `org-roam-db--init` though -- it just does `create index if not exists` before the completion query. This is safe and takes negligible time but it's probably good to add it to the database initialization.

org-roam.el
```diff
@@ -617,28 +618,39 @@ Throw an error if multiple choices exist."
...
+        (org-roam--sql (org-roam-db) "
+    create index if not exists nodes_id on nodes (id);
+    create index if not exists aliases_id on aliases (node_id);"))))
+
+;; THIS MUST BE IN SYNC WITH ORG-ROAM-NODE DEF!!!
+(defun org-roam-query-nodes ()
+  "Get nodes from database"
+  (org-roam--db-prep)
+  (mapcar #'car (org-roam--sql (org-roam-db) "select '#s(org-roam-node ' || id || ' ' || file || ' nil ' || pos || ' nil nil nil nil ' || title || ' ' || properties || ' ' || ifnull(olp, 'nil') || ' nil nil nil' || ')' from nodes union select '#s(org-roam-node ' || node_id || ' ' || nodes.file || ' nil ' || pos || ' nil nil nil nil ' || alias || ' ' || properties || ' ' || ifnull(olp, 'nil') || ' nil nil nil' || ')' from aliases left join nodes on node_id = id;")))
+
```
jethrokuan, line 643: These aren't necessary anymore on latest commit.

org-roam.el
```diff
@@ -617,28 +618,39 @@ Throw an error if multiple choices exist."
...
+        (setq org-roam--db-prepped t)
+        (org-roam--sql (org-roam-db) "
+    create index if not exists nodes_id on nodes (id);
+    create index if not exists aliases_id on aliases (node_id);"))))
+
+;; THIS MUST BE IN SYNC WITH ORG-ROAM-NODE DEF!!!
+(defun org-roam-query-nodes ()
+  "Get nodes from database"
+  (org-roam--db-prep)
+  (mapcar #'car (org-roam--sql (org-roam-db) "select '#s(org-roam-node ' || id || ' ' || file || ' nil ' || pos || ' nil nil nil nil ' || title || ' ' || properties || ' ' || ifnull(olp, 'nil') || ' nil nil nil' || ')' from nodes union select '#s(org-roam-node ' || node_id || ' ' || nodes.file || ' nil ' || pos || ' nil nil nil nil ' || alias || ' ' || properties || ' ' || ifnull(olp, 'nil') || ' nil nil nil' || ')' from aliases left join nodes on node_id = id;")))
```
jethrokuan, line 642: Can we profile how much perf improvements come from manually constructing the query like this?

If the bulk of the perf improvements come from the indexes on the alias tables for the joins, then I'm leaning towards keeping query as is to keep the code clean.

The caching of the display format looks much more acceptable.

zot: Sure, I can check what difference it makes

zot: It's a good thing you asked that. You are correct that the index is what makes the difference. In fact, besides being harder to read and maintain, the inlined SQL version is actually around 7% slower than the original! I'm guessing that SQLite's string concatenation isn't as well implemented as Emacs'.

zot:
Hmm, I smashed my branch and there's a conflict now.

I'll close this and open another pull request.

PR #1502, opened 2021-05-01
Prefer mapconcat over string-join

progfolio:

Haven't benchmarked (the areas I changed did not seem to warrant it), but mapconcat should be slightly more performant than string-join, as it's implemented in C. Also a little cleaner than the `(string-join (mapcar ...` idiom.


jethrokuan (approved):
Thanks!

PR #1507, opened 2021-05-03
Correction in doc: Elisp code in directory local variables will not be executed

flakyhermit:
`expand-file-name` will not execute and the code itself will be stored as the values of the variables, since directory local variables is just a list and not elisp code to be executed.


jethrokuan:
Closing in favour of #2002

PR #1511, opened 2021-05-04
Global minor mode

publicimageltd:

Emacs strongly encourages the use of modes to 'switch' something on and off. And all users will look for that: How can I activate org roam? By using its mode, of course! Therefore, I think we should wrap setting up and tearing down the hooks for `org-roam` in an explicit global minor mode. That's all. If you accept the idea, I would add a second commit which documents that new mode.

Actually, that global mode should be called `org-roam-mode` proper. I am probably not the only one who would mistakenly think that `org-roam-mode` is the name of the mode belonging to org roam proper, and not to some specific feature (backlinks). But that change would mean to rename the mode which has now that name, e.g. to `org-roam-backlinks-mode`. I didn't want to propose that offhand, but I think it is a good idea.


publicimageltd:
Sorry, that was meant for v2. Merde!

PR #1512, opened 2021-05-04
Add gobal minor mode 'org-roam-active'

publicimageltd:
(I made the first PR against master; so here's against v2)

Emacs strongly encourages the use of modes to 'switch' something on and off. And all users will look for that: How can I activate org roam? By using its mode, of course! Therefore, I think we should wrap setting up and tearing down the hooks for org-roam in an explicit global minor mode. That's all. If you accept the idea, I would add a second commit which documents that new mode.

Actually, that global mode should be called org-roam-mode proper. I am probably not the only one who would mistakenly think that org-roam-mode is the name of the mode belonging to org roam proper, and not to some specific feature (backlinks). But that change would mean to rename the mode which has now that name, e.g. to org-roam-backlinks-mode. I didn't want to propose that offhand, but I think it is a good idea.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -793,17 +793,21 @@ window instead."
   (let* ((p (org-entry-get (point) prop))
          (lst (when p (split-string-and-unquote p)))
          (lst (if (memq s lst) lst (cons s lst))))
-    (org-set-property prop (combine-and-quote-strings lst))))
+    (org-set-property prop (combine-and-quote-strings lst))
+    s))
 
-(defun org-roam-remove-property (prop)
-  "Prompt to remove an item from PROP."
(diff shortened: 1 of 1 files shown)


publicimageltd:
*scratch head* Is that new, or did I miss something? I only had seen the functions `org-roam-setup` and `org-roam-teardown`, and over at discourse, we were told to use these function, well, to set up and tear down org roam. And we are talking about v2, right? 

But cool, that's exactly what I thought we need! :+1:

zaeph:
I haven't played around with v2 yet; but, regardless, the convention that you mention is something that we're aware of and that we'll implement if it makes sense.

Issue #1510, opened 2021-05-04
Make a defcustom for "svg", to allow for "pdf", "png", etc


gcoladon:
I would be happy to submit PRs for this, both for the v1 and v2 branches.

jethrokuan:
Closed with #1513

PR #1513, opened 2021-05-06
Custom graph filetype

gcoladon:
I wanted to print the graphs that are generated on multiple sheets of paper, since they were getting too large for a single sheet. I learned that one easy way to do this is to create a PDF, and then use a program like Acrobat Reader that can 'posterize' your document by printing it over n x m pages. 

Then I noticed that "svg" was hard-coded in this file, so I thought it made sense to be a customizable variable. Only 2 lines of code needed to be changed, and 6 added, to affect this change. 

The issue I'd previously posted about this:  https://github.com/org-roam/org-roam/issues/1510

The discussion on Discourse: https://org-roam.discourse.group/t/allow-graph-output-filetype-to-be-configurable-currently-svg-is-hardcoded/1552/5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Changed
 - [#1352](https://github.com/org-roam/org-roam/pull/1352) prefer lower-case for roam_tag and roam_alias in interactive commands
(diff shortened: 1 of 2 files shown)


jethrokuan (approved):
LGTM! One nicety would've been to enumerate the possibilities for `org-roam-graph-filetype`, but that's not that necessary I guess.

PR #1514, opened 2021-05-07
fix org-roam-capture--goto-location

myshevchuk:
Fix regression introduced in 7ed51329.
The function should return an ID.


jethrokuan:
good catch, thanks!

myshevchuk:
Well, I must actually say sorry, for suggesting that change without properly testing it. Anyway, thank you!

PR #1515, opened 2021-05-09
(feat) allow to specify props to remove and return them

d12frosted:

This is a follow up of #1420/ for v2.

Basically:

1. Allow to pass a tag/alias/ref to remove.
2. Return removed tag/alias/ref.
3. Return added tag/alias/ref to mirror (2).


jethrokuan:
Thanks!

PR #1517, opened 2021-05-10
Save before link updates when renaming notes

zaeph:
`org-roam-title-change-hook` conks out when links are added to an unsaved buffer.

Reproducible steps:
1. Go to an existing note `foo`
2. `org-roam-insert` → Create a new file `bar`
3. Validate the capture; notice that the link was added in `foo`, but the the buffer was not saved.
4. Go to `bar`
5. Rename the note to `baz` (via the title)
6. Link in `foo` is not updated

This is because we're retrieving `files-affected` from the db:
https://github.com/org-roam/org-roam/blob/89944230798e4a63b8ce5c345ac22c337414b734/org-roam.el#L1425

The db is updated prior to that, but it only includes what was saved:
https://github.com/org-roam/org-roam/blob/02e35e3b01914a0dceb39a116a978f523d5f4272/org-roam.el#L1436

I've elected to save the modified org-roam buffers prior to updating the db:
https://github.com/org-roam/org-roam/blob/89944230798e4a63b8ce5c345ac22c337414b734/org-roam.el#L1425

I've modularised the code to avoid repetitions elsewhere:
https://github.com/org-roam/org-roam/blob/89944230798e4a63b8ce5c345ac22c337414b734/org-roam.el#L1007-L1010

Also fixed a couple of docstrings which `checkdoc` didn't like.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -227,7 +227,8 @@ Template string   :\n%v")
      :head "#+title: ${title}\n#+roam_key: ${ref}"
      :unnarrowed t))
   "The Org-roam templates used during a capture from the roam-ref protocol.
-Details on how to specify for the template is given in `org-roam-capture-templates'."
+Details on how to specify for the template is given in
+`org-roam-capture-templates'."
   :group 'org-roam
   ;; Adapted from `org-capture-templates'
   :type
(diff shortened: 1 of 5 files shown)


zaeph:
I'm merging this into `master` for v1-legacy, and I'll merge `master` in `v2`.  The latter is of very little consequence (I did it locally), but I just wanted to gauge how well the surgery would go.

Issue #1480, opened 2021-04-20
[v2] duplicate node.id condition not handled


alanz:
### Description

If an org-roam file is accidentally duplicated (for me it was a syncthing conflict that did it), when org-roam starts up it reports a sql uniqueness constraint violated, and it stops the db update.

```
Error (use-package): org-roam/:config: EmacSQL had an unhandled condition: "near line 27: UNIQUE constraint failed: nodes.id"
```
I think it would help to have a handler that at least shows what the offending id is that triggered the problem, to help sorting it out.

#### Steps to Reproduce

Shut down emacs.

Duplicate any file in the org-roam directory

Start up emacs.

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)
 of 2021-03-27, modified by Debian
- Framework: N/A
- Org: Org mode version 9.4.5 (release_9.4.5-321-ge641d3 @ /home/alanz/.emacs.d/straight/build/org/)
- Org-roam: 2.0.0

- Org-roam commit: https://github.com/org-roam/org-roam/commit/be81a3a7383451043fe560c596e0ce8f925c6486

jmay:
Here is some work-in-progress for gracefully reporting duplicates. Updates code in `org-roam-db.el`. Does not handle duplicate IDs at the file level yet.

This has been an excuse to learn how to do exception handling in elisp. I'm bothered by my use of `t` in the handler clause; I couldn't figure out how to match the `emacsql-constraint` error condition.

```elisp
(defun org-roam-db-map-headlines (fns)
  "Run FNS over all headlines in the current buffer."
  (org-with-point-at 1
    (org-map-entries
     (lambda ()
       (catch 'loop
         (dolist (fn fns)
           (condition-case err
               (funcall fn)
             (t
              (message "%s at %s (%s) in %s"
                       (error-message-string err)
                       (org-link-display-format (nth 4 (org-heading-components)))
                       (org-id-get)
                       (buffer-file-name))
              (throw 'loop nil)))))))))
```

PR #1519, opened 2021-05-12
More graceful handling of ID conflicts #1480

jmay:
Behave more gracefully when ID conflicts are encountered when syncing org-roam files to the database.

Current behavior is to print "UNIQUE constraint failed: nodes.id" in the minibuffer and abort.

Proposed behavior: Do not abort. 
Log the offending entry details (ID, entry title, filename) to the *Warnings* buffer, and continue.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -237,15 +237,14 @@ If FILE is nil, clear the current buffer."
 ;;;;; Protection against unique-ID errors
 (define-error 'emacsql-constraint "SQL constraint violation")
 (defun org-roam--protected-node-insert (query vec)
-  (catch 'sqlerror
-    (condition-case err
-        (org-roam-db-query query vec)
-      (emacsql-constraint
-       (throw 'sqlerror
-              (message "%s for %s (%s) in %s"
-                       (error-message-string err)
-                       (aref vec 8) (aref vec 0) (aref vec 1)
-                       ))))))
+  (condition-case err
+      (org-roam-db-query query vec)
+    (emacsql-constraint
+     (lwarn 'org-roam :warning "%s for %s (%s) in %s"
+            (error-message-string err)
+            (aref vec 8) (aref vec 0) (aref vec 1)
+            )
+     nil)))
 
 ;;;;; Updating tables
 (defun org-roam-db-insert-file ()
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -234,6 +234,19 @@ If FILE is nil, clear the current buffer."
                       :where (= file $s1)]
                      file))
 
+;;;;; Protection against unique-ID errors
+(define-error 'emacsql-constraint "SQL constraint violation")
+(defun org-roam--protected-node-insert (query vec)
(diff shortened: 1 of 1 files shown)


org-roam-db.el
```diff
@@ -234,6 +234,19 @@ If FILE is nil, clear the current buffer."
...
                      file))
 
+;;;;; Protection against unique-ID errors
+(define-error 'emacsql-constraint "SQL constraint violation")
+(defun org-roam--protected-node-insert (query vec)
+  (catch 'sqlerror
+    (condition-case err
+        (org-roam-db-query query vec)
+      (emacsql-constraint
+       (throw 'sqlerror
```
jethrokuan, line 244: Doesn't seem useful to catch the error immediately in the same function. Could probably remove the catch and throw clauses, `condition-case` is enough

org-roam-db.el
```diff
@@ -234,6 +234,19 @@ If FILE is nil, clear the current buffer."
...
+              (message "%s for %s (%s) in %s"
```
jethrokuan, line 245: Instead of message, let's use `lwarn` so it appears in a warning buffer. You can see examples of its usage elsewhere in the repo.

jethrokuan (changes requested):

jethrokuan:
Also, It might be a good idea to just wrap all of `org-roam-db-query` to catch all sql errors and report them.

jmay:
I thought about that but I wanted to focus on the ID conflicts first. The other tables don't have unique primary keys. I haven't been able to trigger any other SQL errors besides the "unique constraint failed".

Also, I'm not sure if just writing these messages to `*Warnings*` is the best idea. Maybe there should be a special buffer for this, and a single summary message in the minibuffer at the completion of sync.

jethrokuan:
> I haven't been able to trigger any other SQL errors besides the "unique constraint failed".

I've seen some cases where the DB is the wrong version because of an upgrade and there's a schema mismatch (inserting wrong number of columns). 

I think *Warnings* is more appropriate than *Messages* to log errors, at least. We can still print a final message at completion of the command if there are any errors to check the warnings buffer.

org-roam-db.el
jmay: `(describe-function 'lwarn)` ... cool! Will do.

org-roam-db.el
jmay: I was having problems where `condition-case` was still aborting the control flow. I think I just figured out how this works.

zaeph:
LGTM; I'll need to try it out tomorrow to see if it behaves properly.

jethrokuan:
My preference still is to make the non-erroring insert the default for `org-roam-db-query`, but that's a small issue.

We need to also consider when the insert to the `files` table is successful, but the insert to the `nodes` table is not. Then cache state would be inconsistent: if the `files` table stores the content hash and if the file is never modified, the db will not contain the correct information.

The solution to this is to wrap each of the DB queries for a file in a single transaction.

PR #1522, opened 2021-05-15
V2 refile from top level 

jmay:
org-roam-refile should treat an entire file as a node if it is invoked at the top level (above any heading).

### DIFF
--- a/org-roam-refile.el
+++ b/org-roam-refile.el
@@ -30,8 +30,69 @@
 ;; Org-roam refile allows you to refile notes to your nodes.
 ;;
 ;;; Code:
+(eval-when-compile
+  (require 'org-roam-macs)
+  (require 'org-macs))
+
 (defvar org-auto-align-tags)
 (defvar org-loop-over-headlines-in-active-region)
+
+(defun org-roam--file-keyword-get (keyword)
+  "Pull a keyword setting from the top of the file."
+  (nth 1
+     (assoc keyword
+            (org-collect-keywords (list keyword))))
+)
+
+(defun org-roam--file-keyword-kill (keyword)
+  "Erase a keyword setting line from the top of the file."
+  (let ((case-fold-search t))
+    (org-with-point-at 1
+      (when (re-search-forward (concat "^#\\+" keyword ":") nil t)
+        (beginning-of-line)
+        (delete-region (point) (line-end-position))
+        (delete-char 1))
+      )))
+
+(defun org-roam-demote-entire-buffer ()
+    "Convert an org buffer with any top-level content to a single node.
+
+All headings are demoted one level.
+
+The #+TITLE: keyword is converted into a level-1 heading and deleted.
+Any tags declared on #+FILETAGS: are transferred to tags on the new top heading.
+
+Any top-level properties drawers are incorporated into the new heading.
+"
+    (interactive)
+    (org-with-point-at 1
+      (org-map-entries 'org-do-demote)
+      (insert (concat "* "
+                      (org-roam--file-keyword-get "TITLE"))
+              "\n")
+      (org-back-to-heading)
+      (org-set-tags (org-roam--file-keyword-get "FILETAGS"))
+      (org-roam--file-keyword-kill "TITLE")
+      (org-roam--file-keyword-kill "FILETAGS")
+      ))
+
+(defun org-roam--kill-empty-buffer ()
+    "If the source buffer has been emptied, kill it.
+
+If the buffer is associated with a file, delete the file.
+
+If the buffer is associated with an in-process capture operation, abort the operation.
+"
+    (if (eq (buffer-size) 0)
+        (progn
+          (if (buffer-file-name)
+              (delete-file (buffer-file-name)))
+          (set-buffer-modified-p nil)
+          (if (and org-capture-mode
+                   (buffer-base-buffer (current-buffer)))
+              (org-capture-kill)
+            (kill-this-buffer))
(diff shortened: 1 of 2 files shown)


org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+(eval-when-compile
+  (require 'org-roam-macs)
+  (require 'org-macs))
+
 (defvar org-auto-align-tags)
 (defvar org-loop-over-headlines-in-active-region)
 
+(defun org-roam--file-keyword-get (keyword)
+  "Pull a keyword setting from the top of the file."
+  (nth 1
```
jethrokuan, line 42: Use `car` instead of `nth 1`

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
 ;;; Code:
+(eval-when-compile
+  (require 'org-roam-macs)
+  (require 'org-macs))
+
 (defvar org-auto-align-tags)
 (defvar org-loop-over-headlines-in-active-region)
 
+(defun org-roam--file-keyword-get (keyword)
+  "Pull a keyword setting from the top of the file."
```
jethrokuan, line 41: Think it might be better to move the point to the top of file for this operation:

`(org-with-point-at 1 ...)`

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+)
+
+(defun org-roam--file-keyword-kill (keyword)
+  "Erase a keyword setting line from the top of the file."
+  (let ((case-fold-search t))
+    (org-with-point-at 1
+      (when (re-search-forward (concat "^#\\+" keyword ":") nil t)
+        (beginning-of-line)
+        (delete-region (point) (line-end-position))
+        (delete-char 1))
```
jethrokuan, line 54: Use `kill-line` instead. See https://gist.github.com/jethrokuan/02f41028fb4a6f81787dc420fb99b6e4#file-migrate-el-L24

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
 ;;
 ;;; Code:
+(eval-when-compile
+  (require 'org-roam-macs)
+  (require 'org-macs))
+
 (defvar org-auto-align-tags)
 (defvar org-loop-over-headlines-in-active-region)
 
+(defun org-roam--file-keyword-get (keyword)
```
jethrokuan, line 40: These should probably go into `org-roam-utils`.

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+
+The #+TITLE: keyword is converted into a level-1 heading and deleted.
+Any tags declared on #+FILETAGS: are transferred to tags on the new top heading.
+
+Any top-level properties drawers are incorporated into the new heading.
+"
+    (interactive)
+    (org-with-point-at 1
+      (org-map-entries 'org-do-demote)
+      (insert (concat "* "
```
jethrokuan, line 70: `insert` takes multiple string arguments: `concat` is not necessary here.

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+                      (org-roam--file-keyword-get "TITLE"))
+              "\n")
+      (org-back-to-heading)
+      (org-set-tags (org-roam--file-keyword-get "FILETAGS"))
+      (org-roam--file-keyword-kill "TITLE")
+      (org-roam--file-keyword-kill "FILETAGS")
+      ))
```
jethrokuan, line 77: it's generally bad style to have lines with only parens in lisp.

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+
+(defun org-roam--kill-empty-buffer ()
+    "If the source buffer has been emptied, kill it.
+
+If the buffer is associated with a file, delete the file.
+
+If the buffer is associated with an in-process capture operation, abort the operation.
+"
+    (if (eq (buffer-size) 0)
```
jethrokuan, line 86: Not sure about the logic here, I think you're looking for `when` instead of `if`.

org-roam-refile.el
```diff
@@ -30,9 +30,70 @@
...
+        (progn
+          (if (buffer-file-name)
+              (delete-file (buffer-file-name)))
+          (set-buffer-modified-p nil)
+          (if (and org-capture-mode
+                   (buffer-base-buffer (current-buffer)))
+              (org-capture-kill)
+            (kill-this-buffer))
```
jethrokuan, line 94: especially here...

also use `kill-buffer` rather than `kill-this-buffer`.

jethrokuan (changes requested):

org-roam-refile.el
jmay: You mean `cadr`. The `nth` function starts counting from zero.

org-roam-refile.el
jmay: Not necessary here; the `org-collect-keywords` internals does that already.

org-roam-refile.el
jmay: I wanted to avoid putting the removed text in the kill ring, and `kill-line` doesn't have an option to avoid that. This ugly delete sequence was the best alternative I could find.

org-roam-refile.el
jmay: Agreed. I was trying to keep all the refile-specific functions in `org-roam-refile.el` until you're ready to add that to the requires at the top of `org-roam.el`.

PR #1524, opened 2021-05-19
(feat) allow to provide custom finalize functions

d12frosted:

Unless `:immediate-finish` is non-nil, capture process is asynchronous, meaning that there is no way to do something AFTER the note is created. There is `org-roam-capture-new-node-hook`, but it's being called to early, during the capture process and the hook is not easy to configure for one specific capture.

So `org-roam` could provide a finalization hook in addition to `org-capture-after-finalize-hook`, but I still love the idea to be able to configure finalization per capture.


d12frosted:
P. S. forgot to say that this change is important becuase it does not enforce `org-roam` prefix for finalization functions, which is important namespacing 😸

jethrokuan (approved):

PR #1536, opened 2021-05-29
(fix) add :region to capture keywords

d12frosted:

Unless `:region` is part of this list, it's the value is not put into
the context using `org-roam-capture--put` making it inaccessible via
`org-roam-capture--get`, meaning that, for example,
`org-roam-node-insert` doesn't replace selected region, but instead
simply appends a link to selected region, which is not desired
intention.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -56,7 +56,7 @@ during the Org-roam capture process.")
 This variable is populated dynamically, and is only non-nil
 during the Org-roam capture process.")
 
-(defconst org-roam-capture--template-keywords '(:if-new :id :link-description :call-location)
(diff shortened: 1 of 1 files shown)


jethrokuan:
Thanks! Was just about to look into this.

d12frosted:
Wait, it seems that `org-roam-capture--template-keywords` can't be `nil`:

```emacs-lisp
(let* ((key (pop rest))
                (val (pop rest))
                (custom (member key org-roam-capture--template-keywords)))
           (when (and custom
                      (not val))
             (user-error "Invalid capture template format: %s\nkey %s cannot be nil" template key))
           (if custom
               (setq org-roam-plist (plist-put org-roam-plist key val))
             (setq options (plist-put options key val))))
```

@jethrokuan Is there a way to allow passing an optional keyword?

PR #1538, opened 2021-05-30
(fix) allow to insert with no active region

d12frosted:

Seems like I love to cause issues in this repository. I am sorry! 

Fix looks ugly, there might be a better solution (e.g. by having optional/nil-able values). This just to stop the bleeding :) 

See https://github.com/org-roam/org-roam/pull/1536#issuecomment-851036487 for explanation. Found in https://github.com/d12frosted/vulpea/runs/2705470297.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -826,11 +826,12 @@ which takes as its argument an alist of path-completions."
                          description)))
             (org-roam-capture-
              :node node
-             :props (list :region (when (and beg end)
-                                    (cons beg end))
-                          :insert-at (point-marker)
-                          :link-description description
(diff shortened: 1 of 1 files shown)


ErnestDong:
When I try to capture new node which doesn't exist before, it reads
 
`user-error: Invalid capture template format: (d default plain %? :if-new (file+head %<%Y%m%d%H%M%S>-${slug}.org #+title: Emacs the extensible, customizable self-documenting display editor) :unnarrowed t)`
`key :region cannot be nil`

I'm sorry to bother you. Could you please tell me how to fix it?

Issue #1543, opened 2021-06-08
Enable customization of character replacement when generating file names


salutis:
`org-roam` (v2) converts spaces to underscores (`_`) when forming file names:

```
(let* ((pairs `(("[^[:alnum:][:digit:]]" . "_")  ;; convert anything not alphanumeric
```

I would like to use the more standard (in Emacs) dash character (`-`). As an example, I would like "hello world" to become `hello-world` and not `hello_world`.

Would it be possible to add a customization point here?

P.S. What in the world are "slugs"?

aruscher:
> What in the world are "slugs"?
Basically, it's the short name of a given article or in other words the file name. (https://en.wikipedia.org/wiki/Slug_(publishing))

I will try to implement this. I will reference to this issue in the corresponding pull request.

aruscher:
@salutis See pull request

aruscher:
For a proper discussion why this feature has not been integrated, see: https://github.com/org-roam/org-roam/pull/1544#issuecomment-857605271 and the following two comment
This comment describes how the slug-generation has to be adapted for your own customization: https://github.com/org-roam/org-roam/pull/1544#issuecomment-860260777

@jethrokuan Can be closed.

salutis:
@aruscher Thank you!

PR #1544, opened 2021-06-08
Customization for character replacement generated slug

aruscher:
This change allows to modify the replacement rules for the slug names.

Addresses feature request #1543

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -163,6 +163,21 @@ Marks from the Latin alphabet."
   :type '(repeat character)
   :group 'org-roam)
 
+(defcustom org-roam-slug-name-replacement-rules
+  '(("[^[:alnum:][:digit:]]" . "_")  ;; convert anything not alphanumeric
+    ("__*" . "_")                    ;; remove sequential underscores
+    ("^_" . "")                      ;; remove starting underscore
+    ("_$" . ""))                     ;; remove ending underscore
+  "Replacement rules applied after Unicode normalization to get the slug.
+
+By default the following rules are applied:
+1. All not alphanumeric characters are replaced by underscores
+2. Sequential underscores are replaced by a single underscore
+3. Starting underscores are removed
+4. Ending underscores are removed"
+  :type '(alist :key-type regexp :value-type string)
+  :group 'org-roam)
+
 ;;;; ID Utilities
 (defun org-roam-id-at-point ()
   "Return the ID at point, if any.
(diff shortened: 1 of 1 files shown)


org-roam.el
```diff
@@ -163,6 +163,21 @@ Marks from the Latin alphabet."
...
+    ("__*" . "_")                    ;; remove sequential underscores
+    ("^_" . "")                      ;; remove starting underscore
+    ("_$" . ""))                     ;; remove ending underscore
+  "Replacement rules applied after Unicode normalization to get the slug.
+
+By default the following rules are applied:
+1. All not alphanumeric characters are replaced by underscores
+2. Sequential underscores are replaced by a single underscore
+3. Starting underscores are removed
+4. Ending underscores are removed"
```
salutis, line 177: This is great. So much passive voice, though: are applied, are replaced (2×), are removed (2×). I would go with apply, replace, and remove.

salutis (changes requested):

salutis:
> Addresses feature request #1543

This resolves the problem, so we can go with "Close #1543".

org-roam.el
aruscher: I used passive voice because something is done with something. I will check the other docs and revise this text if necessary.

jethrokuan:
In v2, these won't be customizable options. The way to change how the slug is computed is to override the function entirely:

```emacs-lisp
(cl-defmethod org-roam-node-slug ((node org-roam-node))
  "Return the slug of NODE."
  ;; My logic goes here
)
  ```

**Comment by aruscher** [comment]
@jethrokuan Where do you draw the line between customize options and method overriding? Is it intended that the user should by default override methods to implement the desired behavior and, thus, that every method should work without customize options?

**Comment by jethrokuan** [comment]
The default is customize options, if the options can be easily enumerated.

Slug computation is an open-ended configuration option: e.g. one may choose to include the file creation time, or tags in the slug in which case providing customize options would be far too messy.

In this case overriding the method goes in line with adding properties to nodes: you can define arbitrary properties on a node and use them in org-roam templates e.g.:

```
(cl-defmethod org-roam-namespace ((node org-roam-node))
  "Return the namespace of NODE."
  ;; Do something with folder structure maybe
)
```

and then in your templates use ${namespace}

**Comment by salutis** [comment]
The user here. :) So, what shall I do?

**Comment by aruscher** [comment]
> The user here. :) So, what shall I do?

In a nutshell: You are supposed to redefine the original function as part of your configuration instead of relying on a customization function. So you more or less should copy this code to your init.el to get your desired - instead of _. Please make sure that this code is evaluated after org-roam was loaded :

``` emacs-lisp
(cl-defmethod org-roam-node-slug ((node org-roam-node))
  (let ((title (org-roam-node-title node)))
    (cl-flet* ((nonspacing-mark-p (char)
				  (memq char org-roam-slug-trim-chars))
	       (strip-nonspacing-marks (s)
				       (ucs-normalize-NFC-string
					(apply #'string (seq-remove #'nonspacing-mark-p
								    (ucs-normalize-NFD-string s)))))
	       (cl-replace (title pair)
			   (replace-regexp-in-string (car pair) (cdr pair) title)))
      (let* ((pairs `(("[^[:alnum:][:digit:]]" . "-")  
		      ("--*" . "--")  
		      ("^-" . "")  
		      ("-$" . "")))
	     (slug (-reduce-from #'cl-replace (strip-nonspacing-marks title) pairs)))
	(downcase slug)))))
```

salutis:
@aruscher Thank you! I customize a large number of functions and variables in my Emacs initialization file, but I have yet to use (or see) `cl-defmethod`, `cl-flet*`, and the like. Hence, I got confused.

jimmerricks:
A small change needs to be made to the code snippet for it to work with Emacs 29. See here: https://github.com/org-roam/org-roam/pull/1982/commits/d184001e07d7f4e56c317c9e286eccb8f2f0b444.

rudolf-adamkovic:
> you more or less should copy this code

For future readers:

After a bit of research, I realized we can use CLOS, such as

```
(cl-defmethod org-roam-node-slug :around ((node org-roam-node))
  (string-replace "_" "-" (cl-call-next-method)))
```

PR #1545, opened 2021-06-10
Add options for browsing a published repo via the graphviz output

jasondunsmore:

Introduce two configuration variables to allow exporting of the graphviz output with URLs that link to a published repository.

See discussion here:
https://org-roam.discourse.group/t/script-to-convert-graph-for-offline-mobile-browsing/1674


mtbakerguy:
The following PR addressed this use case:  https://github.com/org-roam/org-roam/pull/1988

Issue #1553, opened 2021-06-18
Strings for node completions are too long


aruscher:
It seems that the strings for node completions are to long to be properly shown in the completion buffer.

Node:
``` Org
:PROPERTIES:
:ID:       7bc23411-8e60-46df-9405-a5acd18100e0
:END:
#+title: AAAA
#+filetags: A B C D E F G H I J K
```
Completion Buffer:
![image](https://user-images.githubusercontent.com/4226393/122608760-15ca4900-d07d-11eb-9cfe-11219852014f.png)

jethrokuan:
Please see https://github.com/org-roam/org-roam/issues/1530#issuecomment-849437672

Issue #1530, opened 2021-05-27
Minibuffer display of org-roam-node(ref)-find cut off #FILETAGS


herop:
### Description
Calling the above-mentioned functions does cut off the display (see screenshot)
#### Steps to Reproduce
`org-roam-node-find` opens the minibuffer

#### Expected Results
Showing the whole line and not huge whitespace instaed.

#### Actual Results
See screenshot
![Width-Minibuffer-roam-find](https://user-images.githubusercontent.com/658323/119782860-30fdca80-bebc-11eb-9efe-b6f01570896d.png)

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.29, cairo version 1.17.4)
 of 2021-05-19
- Framework: Doom
- Org: Org mode version 9.5 (nil @ /home/jburkhard/.emacs.d/.local/straight/build-28.0.50/org-mode/)
- Org-roam: 2.0.0

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

jethrokuan:
You'll need to manage your display template to make use of the limited screen real estate:

To assign a larger region to tags, you can do:

```
(setq org-roam-node-display-template "${title:*} ${tags:50}")
```

herop:
Ingenious. Thank you very much. Wasn't able to find this parameter.
Worked perfectly even with posframe!

PR #1554, opened 2021-06-18
Clarify 'org-roam-node-display-template' documentation

aruscher:
As already two issues addressing the "problem" of displaying nodes in the minibuffer (#1553 #1530), I found it necessary to clarify the documentation.


jethrokuan (approved):
Thanks!

PR #1555, opened 2021-06-21
Add a require for org-roam-protocol back in

gcoladon:
I am probably missing something, but I couldn't get roam-ref org-protocol links to work until I added this require back in.

org-roam-protocol is nice and seems to work as long as it's included?

### DIFF
--- a/org-roam-protocol.el
+++ b/org-roam-protocol.el
@@ -36,7 +36,7 @@
 ;;
 ;;; Code:
 (require 'org-protocol)
-(require 'org-roam)
+;; (require 'org-roam)
 (eval-when-compile
   (require 'org-roam-macs))
 (require 'ol) ;; for org-link-decode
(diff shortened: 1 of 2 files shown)


gcoladon:
OK I guess something is wrong with this PR, I didn't get the recursive require error on my side at first, but now I am getting  it. So please don't merge this until I get to the bottom of that.

gcoladon:
I still am somewhat baffled by how or why he require for org-roam-protocol has been left out, and I assume it was done on purpose, but can't find any explanation for why that would be.

jethrokuan:
Hasn't it always been left out? requiring it is supposed to be in your config file.

nobiot:
Yes, on V1, you needed to require it in config.

gcoladon:
OK, since I now know that `org-protocol` is itself not automatically loaded, I get why we wouldn't want to automatically load `org-roam-protocol`. Nevermind! :)

PR #1559, opened 2021-06-24
Encrypt specific capture templates

perfectayush:

Currently encryption behavior is globally controlled via org-roam-encrypt-files, and there isn't flexibility to control encryption for specific capture templates.

This change introduces a `:encrypt-file` keyword in org-roam-capture-templates, which can configure encryption for specific capture templates.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -618,7 +618,8 @@ the default template, reproduced below.
        "%?"
        :file-name "%<%Y%m%d%H%M%S>-${slug}"
        :head "#+title: ${title}\n"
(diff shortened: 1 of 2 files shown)


jethrokuan:
This is unnecessary in v2, which will be merged within the week.

As stated in the [encryption docs](https://github.com/org-roam/org-roam/blob/v2/doc/org-roam.org#encryption), just add the `.gpg` behind for the capture template you want to encrypt.

perfectayush:
@jethrokuan Oh cool. Looking forward.

PR #1560, opened 2021-06-24
Fix `org-roam-graph` command when used with `universal-argument`

Wetlize:

Previously, when the command would be called with `universal-argument` it would simply fall through, without running any code and hence, not generating a node-local graph. This is because the command didn't account correctly all possible value of ARG, where with `universal-argument` it becomes a list of one element, which is an integer.

I've also refactored the command to handle non-interactive usage and to better trace values when debugging it.

### DIFF
--- a/org-roam-graph.el
+++ b/org-roam-graph.el
@@ -264,15 +264,17 @@ ARG may be any of the following values:
   - nil       show the graph.
   - `\\[universal-argument]'     show the graph for NODE.
   - `\\[universal-argument]' N   show the graph for NODE limiting nodes to N steps."
-  (interactive "P")
-  (pcase arg
-    ('nil            (org-roam-graph--build (org-roam-graph--dot nil t)
-                                            #'org-roam-graph--open))
-    ((pred integerp) (org-roam-graph--build (org-roam-graph--dot
-                                             (org-roam-graph--connected-component
-                                              (org-roam-node-id (org-roam-node-at-point 'assert))
-                                              (abs arg)))
-                                            #'org-roam-graph--open))))
+  (interactive
+   (list current-prefix-arg
+         (and current-prefix-arg
+              (org-roam-node-at-point 'assert))))
+  (let ((graph (cl-typecase arg
+                 (null (org-roam-graph--dot nil 'all-nodes))
+                 (cons (org-roam-graph--dot (org-roam-graph--connected-component
+                                             (org-roam-node-id node) 0)))
+                 (integer (org-roam-graph--dot (org-roam-graph--connected-component
+                                                (org-roam-node-id node) (abs arg)))))))
+    (org-roam-graph--build graph #'org-roam-graph--open)))
 
 
 (provide 'org-roam-graph)


org-roam-graph.el
```diff
@@ -264,15 +264,17 @@ ARG may be any of the following values:
...
-                                              (abs arg)))
-                                            #'org-roam-graph--open))))
+  (interactive
+   (list current-prefix-arg
+         (and current-prefix-arg
+              (org-roam-node-at-point 'assert))))
+  (let ((graph (cl-typecase arg
+                 (null (org-roam-graph--dot nil 'all-nodes))
+                 (cons (org-roam-graph--dot (org-roam-graph--connected-component
+                                             (org-roam-node-id node) most-positive-fixnum)))
```
jethrokuan, line 274: In what scenario does this run? `org-roam-graph--connected-component` is quite expensive, i imagine `most-positive-fixnum` would just hang Emacs here.

Wetlize: Huh, since I haven't port yet my v1 notes to v2 nodes I'm running this against a generated `org-roam-directory`, which consist of only 100 nodes. The largest distance away from a root node doesn't reach there beyond of 5 edges in length and nor is it wide.

I do remember setting this value without checking-out what `org-roam-graph--connected-component` does in details, since `most-positive-fixnum` did the thing without breaking anything, and only now I've just realized it recursively queries to the DB. However, I cannot observe any performance differences between setting DISTANCE to 5 and using `most-positive-fixnum`, even after applying 55008e3707c4960afbc6e2c0fcd6e72a1e60fc0a. Both of them elapse under a single second.

I do run `native-comp` and `org-roam` is compiled by Doom, so it could eliminate some things away and apply tail call optimizations. Either that or my data set is too small to cause any hangs.

jethrokuan: I think it will eventually produce the same graph as with setting distance = 0 anyway, so there's no need to do all secondary json recursive stuff. I'm just wondering what arg would result in a `cons`.

Wetlize: It does look like `most-positive-fixnum` as `DISTANCE` produce the same graph as if you would set  it to 0.

[This](https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html) explains how Emacs maps `current-prefix-arg` (raw prefix argument value) in case of different prefix commands (`universal-argument` and `digit-argument` set 2 different values).

So, what do you think about handling of `ARG` here? Should `C-u` behave with `DISTANCE` set to 0? Or interpret it instead as integer based value? `C-u` = 4, `C-u C-u` = 16, `C-u C-u C-u 64`, etc.

jethrokuan: > So, what do you think now about handling of ARG? Should C-u set DISTANCE to 0? 

I think setting distance to 0 makes the most sense here

Wetlize:
This should be good to go now.

jethrokuan (approved):

PR #1567, opened 2021-07-02
(fix) SQL for querying notes with tags, aliases and refs

d12frosted:

See https://github.com/org-roam/org-roam/commit/a199886ef7ae208b0b10dc45e0df9b54d210cd4d#r52949692

P.S. I checked this query only outside of `org-roam`. And it worked :) But more real testing would be useful.


d12frosted:
Sorry, used tabs instead of spaces. Should be good now.

PR #1579, opened 2021-07-06
Inhibit org-capture from setting :exact-position by default

### LINKED ISSUE #1566: v2: All the variants of `:if-new (file+[headline] ...)` don't capture to the right places
### Description
If the target file doesn't exist or it doesn't contain any headings in it, then it will capture to the right place, as expected. However, it doesn't do so if the target file has some headings in it, independently of whether such headings match the defined in template OLP or these're some other non-matching headings, or whether they have an ID in them or not, none of this matter. 

When this happens, the capture will somehow randomly embed the new entry to the target file, for example, it can put entry as a level 1 heading, completely ignoring the defined OLP, or embed it as a child of some other random headings that were not in the OLP but were previously existing in the file.

I suspect this doesn't just happen with `(file+olp)`, but also with any other `(file+[heading])` targets, including `datetree`. I will update the title if I confirm that.

#### Steps to Reproduce
1. Use the next capture template:
```emacs-lisp
(setq-default
 org-roam-capture-templates
 `("d" "default" entry
   ,(mapconcat #'identity
               '("* ${title}%?"
                 ":PROPERTIES:"
                 ":ID: %(org-id-new)"
                 ":END:")
               "\n")
   :if-new (file+olp "test.org" ("Capture's OLP"))))
```
2. Run the capture for this template without previously creating the needed for it file.
3. Confirm that it captured the thing **as expected.**
3. Delete all the content inside the file, except file's title and properties.
3. Run the capture for already created `test.org` file, but with no headings or other content inside of it.
4. Confirm that it captured the things **as expected.**
4. Run the capture again for a new heading, keeping the previous capture inside the file.
5. Confirm that the captured heading **wasn't inserted at the right place.** 
5. Delete all the content inside the file, except file's title and properties.
6. Add some random headings inside this file that wouldn't match the defined OLP.
7. Run the capture for a new heading.
8. Confirm that the captured heading **was inserted at some random place.**

### Environment
- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0) of 2021-05-01
- Framework: Doom
- Org: Org mode version 9.5 (nil @ /home/user/.emacs.d/.local/straight/build-28.0.50/org-mode/)
- Org-roam: 2.0.0
- Org-roam commit: 31817f3483636300bd587b3f768ef31338427ad0

Wetlize:
`org-capture-set-target-location` will always try to set `:exact-position` for `function` based capture targets, which is what `org-roam-capture--get-point` is.

The problem with this is that `:exact-position` doesn't play well with various capture `type`s and it breaks `entry` based capture templates from proper insertion. The major reason for this is a conflict between `:exact-position` and `:target-entry-p` in org-capture, which conditionally compete with each other[0]. In `org-capture-set-target-location` only a very small subset of all the possible targets set `:exact-position`, and unfortunately `function` target is is one of them.

This won't prohibit org-roam's based capture targets from setting `:exact-position` for targets that would actually need it[1], but would inhibit org-capture from setting the `:exact-position` by default, when `org-roam-capture-` is running.

[0]: See `org-capture-place-{entry,item,table-line,plain-text}` and `org-capture-set-target-location` for how this is implemented.

[1]: This would be accomplished similarly to how `org-capture-set-target-location` does so.

This change introduces a hacky workaround for org-capture itself and hence, some tech debt. The thing is, this hack is a simplest way to tape everything to each other for a proper functionality of various combinations of `type`x`target` templates, while at the same time not requiring significant changes to the codebase. I see a 2 different paths towards a "clean" solution that can be taken in future:

- Request from `org-mode` maintainers to either, remove [this line](https://code.orgmode.org/bzg/org-mode/src/fdb98a436bbdb86834af08c168c3bcbcc73d349c/lisp/org-capture.el#L1072) with the motivation that the user should be in full-control of things whenever he uses `function` as capture target, or request them to introduce new target that would do the same thing without breaking existing target (and possibly deprecate the old one). In both of these cases `org-roam` would be able to remove the hack and keep the codebase intact.
- Rewrite a huge chunk of `org-roam-capture` to either, use an advice around `org-capture-set-target-location` or by `cl-letf`ing it from `org-roam-capture-`. The problem with both of these options is not just the cost of the rewrite, but also extra costs of maintaining this, where the newly introduced function would need to replicate what `org-capture-set-target-location` does, which can also introduce conflicts for people with different versions of org-mode.

Merging this should close #1566 and prevent similar new issues for other unreported `type`x`target` captures.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -727,7 +727,28 @@ This function is used solely in Org-roam's capture templates: see
                      (org-roam-capture--fill-template (org-capture-get :template)))
     (org-roam-capture--put :id id)
     (org-roam-capture--put :finalize (or (org-capture-get :finalize)
-                                         (org-roam-capture--get :finalize)))))
+                                         (org-roam-capture--get :finalize)))
+    ;; NOTE This should be always put at the end of the defun.
+    (advice-add 'org-capture-put :filter-args #'org-roam-capture--inhibit-exact-position-a)))
+
+;; HACK `org-roam-capture--get-point' designed as a (function ...) based target
+;; for `org-capture-set-target-location'. The problem with this is that the
+;; function based capture target will always try to set `:exact-position' there
+;; and this has a major conflict with org-roam's capture targets. Most (if not
+;; all) of its targets don't need `:exact-position' and those that do should set
+;; them explicitly in `org-roam-capture--goto-location', similarly to how
+;; `org-capture-set-target-location' does it for its own targets.
+;;
+;; This advice will be relevant as long as both, org-roam hooks itself to
+;; org-capture as a function based target and org-capture keep attempts to set
+;; `:exact-position' for function based capture targets.
+(defun org-roam-capture--inhibit-exact-position-a (plist)
+  "Filter out `:exact-position' out of PLIST.
+This function is designed as a self-removing advice for
+`org-captre-put' and is exclusively used inside
+`org-roam-capture--get-point'."
+  (prog1 (plist-put plist :exact-position nil)
+    (advice-remove 'org-capture-put #'org-roam-capture--inhibit-exact-position-a)))
 
 (defun org-roam-capture--convert-template (template &optional props)
   "Convert TEMPLATE from Org-roam syntax to `org-capture-templates' syntax.
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -617,7 +617,6 @@ Return the ID of the location."
            ;; Current date, possibly corrected for late night
            ;; workers.
            (org-today)))))
-       (org-end-of-subtree t t)
        (setq p (point)))
       (`(node ,title-or-id)
        ;; first try to get ID, then try to get title/alias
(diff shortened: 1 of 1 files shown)


progfolio:
> The problem with this is that :exact-position doesn't play well with various capture types and it breaks entry based capture templates from proper insertion. The major reason for this is a conflict between :exact-position and :target-entry-p in org-capture, which conditionally compete with each other

Is the problem specific to org-roam, or org-capture in general?
Can you provide a minimal example demonstrating the conflict you describe?

jethrokuan:
I don't think this is an issue with `org-capture` at all, with `function`, the caller is expected to move the point to the place where the template is to be inserted. That is, `org-roam-capture--get-point` has to move the point accordingly. That is what I had intended to achieve with `org-roam-capture--goto-location`.

So if the templates are not being inserted at the correct locations, it is because `org-roam-capture--goto-location` is not moving the point to the right location.

Wetlize:
@jethrokuan You might be right. I was looking specifically at [this](https://code.orgmode.org/bzg/org-mode/src/fdb98a436bbdb86834af08c168c3bcbcc73d349c/lisp/org-capture.el#L1162-L1165), where without `:local-target-p` it wouldn't adjust headline level for entries, but didn't notice that `org-capture-set-target-location` [sets both, :exact-position and :local-target-p](https://code.orgmode.org/bzg/org-mode/src/fdb98a436bbdb86834af08c168c3bcbcc73d349c/lisp/org-capture.el#L1067-L1069).

I'll try to readjust where point is moved to with `:exact-position` in mind and if necessary close this PR in favor of a new one.

Wetlize:
Closed in favor of #1581.

PR #1580, opened 2021-07-10
auto refresh for org-roam-buffer-toggle

yuchen-lea:

The guide says:

> To enable the org-roam to refresh based on the current node your point is at, use `org-roam-buffer-toggle`. You should see the backlink buffer refresh by just moving around in a file across different nodes.

However, if I call `org-roam-buffer-toggle` directly, it won't refresh automatically since `org-roam-buffer--post-command-h` is added to `post-command-hook` in some place not been touched.

So I add the `add-hook` in `org-roam-buffer-toggle` directly and make this global, because I think it's more convenient to refresh backlink buffer automatically when switching multiple org files.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -224,6 +224,7 @@ Valid states are 'visible, 'exists and 'none."
      (progn
        (setq org-roam-current-node (org-roam-node-at-point)
              org-roam-current-directory org-roam-directory)
+       (add-hook 'post-command-hook #'org-roam-buffer--post-command-h)
        (display-buffer (get-buffer-create org-roam-buffer))
        (org-roam-buffer-persistent-redisplay)))))
 


jethrokuan:
In https://github.com/org-roam/org-roam/blob/0547dc723ddca5265d0e746c51cf2e682ab05866/org-roam-mode.el#L250, we ensure that every Org-roam file gets this `post-command-hook`, so I don't think it's necessary to add to the global `post-command-hook`. It should still work when switching between multiple Org-roam files, but won't work with files not within Org-roam.

yuchen-lea:
<p>
Here is my config:
</p>

<div class="org-src-container">
<pre class="src src-emacs-lisp"><span style="color: #51afef;">(</span><span style="color: #26a6a6;">setq</span> <span style="color: #f8f8f0;">org-roam-directory</span> note-path<span style="color: #51afef;">)</span>
<span style="color: #51afef;">(</span><span style="color: #26a6a6;">use-package!</span> org-roam
  <span style="color: #ff5d38;">:after</span> org
  <span style="color: #ff5d38;">:commands</span>
  <span style="color: #c678dd;">(</span><span style="color: #f8f8f0;">org-roam-buffer</span>
   org-roam-buffer-toggle
   org-roam-setup
   org-roam-capture
   org-roam-node-find<span style="color: #c678dd;">)</span>
  <span style="color: #ff5d38;">:config</span>
  <span style="color: #c678dd;">(</span><span style="color: #bcd42a;">org-roam-setup</span><span style="color: #c678dd;">)</span><span style="color: #51afef;">)</span>
</pre>
</div>

<p>
Here are what I do:
</p>

<ol class="org-ol">
<li>Open emacs</li>
<li>Open some org-file</li>
<li>call <code>org-roam-buffer-toggle</code> on some heading in org-file</li>
</ol>

<p>
Here are the results under v2 branch:
</p>

<p>
The backlink buffer displays well, but cannot refresh automatically when pointer moving.
</p>



hook | function | hook contains function?
-- | -- | --
find-file-hook | org-roam–file-setup | yes
org-roam-find-file-hook | org-roam-buffer–redisplay | yes
post-command-hook | org-roam-buffer–post-command-h | no



<p>
It seems <code>org-roam-setup</code> works, <code>org-roam-buffer--redisplay</code> is added to <code>org-roam-find-file-hook</code>. However, when an Org-roam file is visited, <code>org-roam-find-file-hook</code> hasn’t been run, so that <code>post-command-hook</code> doesn’t contain <code>org-roam-buffer--post-command-h</code>. That’s why backlink buffer won’t refresh automatically.
</p>

Any suggestions?

jethrokuan:
> However, when an Org-roam file is visited, org-roam-find-file-hook hasn’t been run

Verify that your file is an Org-roam file with by evaluating `M-: (org-roam-file-p)`. If your above table is correct, then:

- On `find-file`, `find-file-hook` is run, running `org-roam--file-setup`
- If `(org-roam-file-p)`, `org-roam-find-file-hook` is run, including `org-roam-buffer--redisplay`
- `org-roam-buffer--redisplay` adds `org-roam-buffer–post-command-h` to the buffer-local `post-command-hook`

PR #1581, opened 2021-07-10
Adjust locations of where templates should be inserted for each target

### LINKED ISSUE #1566: v2: All the variants of `:if-new (file+[headline] ...)` don't capture to the right places

Wetlize:
This an alternative solution to #1579 to fix and close #1566. All it does is repositioning of the point, so each `org-capture-place-[type]` functions would do the the thing at the right position. From my internal semi-automated testing this works with all currently available `type`x`target` combinations in `org-roam-capture` and better than #1579, which still has some flows. For more details, please, read the commits message for this PR, the linked issue and the alternative PR.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -539,7 +539,7 @@ Return the ID of the location."
          (org-roam-capture--put :new-file path))
        (set-buffer (org-capture-target-buffer path))
        (widen)
-       (setq p (point)))
+       (setq p (goto-char (point-min))))
       (`(file+olp ,path ,olp)
        (setq path (expand-file-name
                    (string-trim (org-roam-capture--fill-template path t))
@@ -560,7 +560,7 @@ Return the ID of the location."
          (org-roam-capture--put :new-file path)
          (insert (org-roam-capture--fill-template head t)))
        (widen)
-       (setq p (point-min)))
+       (setq p (goto-char (point-min))))
       (`(file+head+olp ,path ,head ,olp)
        (setq path (expand-file-name
                    (string-trim (org-roam-capture--fill-template path t))
@@ -617,7 +617,6 @@ Return the ID of the location."
            ;; Current date, possibly corrected for late night
            ;; workers.
            (org-today)))))
-       (org-end-of-subtree t t)
        (setq p (point)))
       (`(node ,title-or-id)
        ;; first try to get ID, then try to get title/alias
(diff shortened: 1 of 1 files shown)


Wetlize:
Oh wow, the produced highlight for the diff really blows here on the GH. Dammit, it's not as bad as it looks :)

jethrokuan (changes requested):
:+1: The general direction of the fix looks good.

I think it would help to extract `move-point-based-on-capture-type-near` to an internal function within `org-roam-capture`, the `org-roam-capture--goto-location` function is already kinda unwieldy.

Wetlize:
@jethrokuan And done. I personally in favor of giant functions with some comments here and there, when it comes to things that heavily context dependent or only make sense to exist under a single common scope and never referenced in any other scope, but this can be a bit clumsy in Elisp. 

I've iterated a bit on this and only just now noticed that each capture target doesn't move point after setting it once, so the function doesn't need to be called under each target, but only once after the `pcase`. The mandatory argument was redundant too, so I removed that.

This solution was also written generalized, with assumption that some other capture types will require some level of adjustment, but it only looks like `plain` needs that. I think it's safe to keep it this way and if needed other types (e.g. some new ones) can easily extend this.

jethrokuan (approved):

PR #1583, opened 2021-07-12
Add `fd` as another backend to list files

Wetlize:
[fd](https://github.com/sharkdp/fd) is a popular alternative to `find` and is an optional dependency of Doom Emacs to index files, so there's plenty of users that use it. In some ways it's superior to `find`, while in others worse (mostly when it comes to power options). It doesn't by the default index hidden and ignored files, and I didn't set the flags to include them either. Let me know if you think this should be changed.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -109,23 +109,30 @@ ensure that."
 (defcustom org-roam-list-files-commands
   (if (member system-type '(windows-nt ms-dos cygwin))
       nil
-    '(find rg))
+    '(find fd fdfind rg))
   "Commands that will be used to find Org-roam files.
 
 It should be a list of symbols or cons cells representing any of the following
- supported file search methods.
+supported file search methods.
 
 The commands will be tried in order until an executable for a command is found.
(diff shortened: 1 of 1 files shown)


Wetlize:
While on this, is there any reason why you exclude by the default external file finders when it comes to Windows?
https://github.com/org-roam/org-roam/blob/0e5670f0b11259b7d74b726c924759b6512b8657/org-roam.el#L109-L112
For example, both, `fd` and `rg` are available as native binaries there and can be easily installed with something like `scoop`.

jethrokuan:
> While on this, is there any reason why you exclude by the default external file finders when it comes to Windows?
> https://github.com/org-roam/org-roam/blob/0e5670f0b11259b7d74b726c924759b6512b8657/org-roam.el#L109-L112
> 
> 
> For example, both, `fd` and `rg` are available as native binaries there and can be easily installed with something like `scoop`.

Some Windows users reported issues about the `rg` option not working, and I'm not able to test/fix them, so defaulting to the elisp option seemed like the safest option.

PR #1594, opened 2021-07-17
Refactor some docstring and parameters in `org-roam.el`

Wetlize:
General maintenance. There's a lot of things that preserved from v1/early v2 and wasn't updated. This PR doesn't address all the accumulated litter in the codebase, but only specific to `org-roam.el`, and even then, not all of it. A more serious refactoring is needed in some places. I've also updated some other docstrings that looked more as placeholders or wouldn't look clear enough.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -352,7 +352,9 @@ If BUFFER is not specified, use the current buffer."
 
 ;;;###autoload
 (defun org-roam-setup ()
-  "Setup Org-roam."
+  "Setup Org-roam and initialize its database.
+This will install the needed hooks and advices to keep everything
+in sync with the connected databases."
   (interactive)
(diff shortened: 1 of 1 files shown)


Wetlize:
BTW, is there some reason why you keep things that's only related to `org-roam-mode`/`magit-section` in `org-roam.el`? It looks like `org-roam-mode.el` would be a better place for all that.

jethrokuan:
> BTW, is there some reason why you keep things that's only related to `org-roam-mode`/`magit-section` in `org-roam.el`? It looks like `org-roam-mode.el` would be a better place for all that.

Something to do with circular imports, wasn't able to get things to cleanly load.

jethrokuan:
This was helpful, thanks!

PR #1595, opened 2021-07-18
Refactor backward compatibility related code

Wetlize:
Just tidying up the codebase to ease maintenance and improve its understanding by the other users. See the commit messages for more details about the changes.

### DIFF
--- a/org-roam-compat.el
+++ b/org-roam-compat.el
@@ -33,7 +33,90 @@
 ;;; Code:
 ;;;; Library Requires
 
+;;; Backports
+;; REVIEW Remove when 26.x support is dropped. This is exact the same as
+;; `directory-files-recursively' from Emacs 26, but with FOLLOW-SYMLINKS
+;; parameter from Emacs 27.
+(defun org-roam--directory-files-recursively (dir regexp
+                                                  &optional include-directories predicate
+                                                  follow-symlinks)
+  "Return list of all files under directory DIR whose names match REGEXP.
+This function works recursively.  Files are returned in \"depth
+first\" order, and files from each directory are sorted in
+alphabetical order.  Each file name appears in the returned list
+in its absolute form.
+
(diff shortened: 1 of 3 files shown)


org-roam-compat.el
```diff
@@ -33,7 +33,89 @@
 ;;; Code:
 ;;;; Library Requires
 
+;;; Backports
+(when (< emacs-major-version 27)
+  ;; Emacs 26 does not have FOLLOW-SYMLINKS in `directory-files-recursively'
+  (defun directory-files-recursively (dir regexp
```
jethrokuan, line 39: While harmless in this case, I don't think it's a good idea to shadow the function like this. i.e. we should leave the `org-roam-` prefix here.

Wetlize:
Should be good to go now.

jethrokuan (approved):

org-roam-dailies.el
```diff
@@ -327,25 +327,6 @@ negative, find note N days in the future."
 (define-key org-roam-dailies-map (kbd "v") #'org-roam-dailies-capture-date)
 (define-key org-roam-dailies-map (kbd ".") #'org-roam-dailies-find-directory)
 
-(define-obsolete-function-alias
```
d12frosted, line 330: While I agree with `org-roam--directory-files-recursively` which is exactly what `compat` file are about, IMO obsoletisation of a function should stay in the module where function is defined.

Wetlize: I kind of disagree, because it still promotes the usage of the obsolete functions and litters stuff. WDYT @jethrokuan?

jethrokuan: I'm ambivalent: I think having all obsolete declarations in a compat file helps with cleanup later on, but separates the function from its "alias". This is generally not an issue, it's a set-and-forget kind of thing. There are arguments for both sides, I guess it's just important to stick to one style.

PR #1610, opened 2021-07-18
Adjusted configuration examples for backlinks buffer.

carwin:
- Fixed the original example's extra parenthesis, which broke the alist.
 - Added a second example for configuring the buffer as a side window.

When migrating to v2 I found that the configuration example for the backlinks buffer (both in the wiki and official documentation) had an extra parenthesis, which would work if `setq` was being used instead of `add-to-list`.

This pull request fixes the original example and adds a second example for using the buffer as a side window.


jethrokuan:
Thanks!

PR #1611, opened 2021-07-18
migration: catch error and restore files

jethrokuan:
In case of migration error, restore original files, and print a message.

### DIFF
--- a/org-roam-migrate.el
+++ b/org-roam-migrate.el
@@ -78,23 +78,31 @@ To your init file.
   (when (yes-or-no-p "Org-roam will now convert all your notes from v1 to v2.
 This will take a while. Are you sure you want to do this?")
     ;; Back up notes
-    (let ((backup-dir (expand-file-name "org-roam.bak"
-                                        (file-name-directory (directory-file-name org-roam-directory)))))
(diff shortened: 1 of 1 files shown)


carwin:
Checking out this branching and running the `org-roam-migrate-wizard` on my system returned:

```
Symbol's value as variable is void: debug-dir
```

carwin:
Just to add a little more info/context, my v1 notes cause the migration script present in the current release to fail with "unexpected end of file" and nothing else.

I was able to get the script to run on this branch by manually setting `debug-dir`, and `backup-dir` (see previous comment for why that was necessary). The script runs, and dumps my v1 notes into the path set in `debug-dir` but no message is printed.

jethrokuan:
@carwin if it's okay with you could you send me your current org-roam-directory so I can figure out what's wrong? You can email me at jethrokuan95@gmail.com

carwin:
I've manually migrated my notes to v2, at this point - but I'll go try and find a backup after work.

carwin:
I completely forgot to do this - I don't have any of my old notes at this point, everything has been migrated manually.

PR #1622, opened 2021-07-19
(feat): formalize node dedicated buffers

Wetlize:
Replaces the `org-roam-buffer` command and formalizes the concept of node dedicated Org-roam buffers (compared to the persistent one). I'm not sure about the name, so if you have a better one let me know.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -120,16 +120,26 @@ and `:slant'."
   :group 'org-roam-faces)
 
 ;;; Variables
-(defvar org-roam-current-node nil
-  "The current node at point.")
+(defvar org-roam-buffer-current-node nil
+  "The node for which an `org-roam-mode' based buffer displays its contents.
+This set both, locally and globally. Normally the local value is
+only set in the `org-roam-mode' based buffers, while the global
+value shows the current node in the persistent `org-roam-buffer'.")
 
-(defvar org-roam-current-directory nil
-  "The `org-roam-directory' value for the current node.")
+(put 'org-roam-buffer-current-node 'permanent-local t)
+
+(defvar org-roam-buffer-current-directory nil
+  "The `org-roam-directory' value of `org-roam-buffer-current-node'.
+Set both, locally and globally in the same way as `org-roam-buffer-current-node'.")
+
+(put 'org-roam-buffer-current-directory 'permanent-local t)
 
 (defcustom org-roam-mode-section-functions (list #'org-roam-backlinks-section
                                                  #'org-roam-reflinks-section)
-  "Functions which insert sections of the `org-roam-buffer'.
-Each function is called with one argument, which is the current org-roam node at point."
+  "Functions that insert sections in the `org-roam-mode' based buffers.
+Each function is called with one argument, which is an
+`org-roam-node' for which the buffer will be constructed for.
+Normally this node is `org-roam-buffer-current-node'."
   :group 'org-roam
   :type 'hook)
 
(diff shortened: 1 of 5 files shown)


Wetlize:
Off-topic: Do you have any requirements for have commits should be titled? I've noticed that you try to tag some commits with `(feat):`, `(fix):` etc, but even then, you're not consistent with this yourself. There's also no list of recommended tags to use for a specific use case. Maybe you should add some short docs for the contributors about such thing? Assuming that's something important for you, otherwise w/e.

carwin:
Off-topic: Personally, I've always liked having commit / branch naming requirements. The Angular JS project has a pretty thorough outline on this kind of thing: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit-header

nobiot:
There is Contributing file that outlines what the project expects (see the bottom of readme). Is this too flexible?

jethrokuan:
Yeah I think commit naming requirements are nice and can help with search later on, but I don't want to raise the bar to contributing, so I'm relatively flexible here. Sometimes I just rename during merge.

I also find that they make the commits a bit long, so it's a delicate balance.

org-roam-mode.el
```diff
@@ -169,32 +169,48 @@ which visits the thing at point."
...
-                               (file-relative-name (buffer-file-name) org-roam-directory)))))
-          (with-current-buffer buffer
-            (org-roam-mode)
-            (setq-local org-roam-current-node node)
-            (setq-local org-roam-current-directory source-org-roam-directory)
-            (org-roam-buffer-render))
-          (switch-to-buffer-other-window buffer)))
-    (user-error "No node at point")))
+;;; Node dedicated buffer
+(defvar org-roam-dedicated-buffer-prefix "*org-roam: "
```
jethrokuan, line 173: It's a good rule to keep the functions and variables prefixed with the file name, in this case `org-roam-buffer`. Might be a bit strange sounding but I think `org-roam-buffer-dedicated-prefix` is more appropriate.

jethrokuan: Also, maybe use -format instead of prefix and use `format`? So people can choose whether anything should be added to the suffix (or at least close the * that is opened in the prefix)

org-roam-mode.el
```diff
@@ -169,32 +169,48 @@ which visits the thing at point."
...
+  "Prefix that's used by each node dedicated Org-roam buffer.
+See `org-roam-display-node-dedicated-buffer' for more details.
+Should start with \"*\" and end with a whitespace.")
+
+;;;###autoload
+(defun org-roam-display-node-dedicated-buffer (node)
```
jethrokuan, line 179: Same comment on buffer function name here.

jethrokuan (changes requested):

org-roam-mode.el
Wetlize: Wish it would be `org-roam-buffer.el`, but it's named as `org-roam-mode.el`. But I've got you 👍 
As for the prefix variable, I want to expose it to people for programmatic matching in Elisp, so putting the formatting specifiers would make it a bit difficult for people to use it for such purpose. I'll see what can be done about this.

Wetlize:
I'll turn this into draft for now, since there's quite large refactoring coming along for `org-roam-mode.el` and I'd like to avoid regression.

Wetlize:
OK, it didn't took me a while and this should be ready now for the review. There was some regression in the draft and I've fixed it. It should also cut more LoC than it was before the PR, since I've unified 2 different code paths into a single one. This also fixes small unreported bug with `org-roam-buffer-toggle`, where it wouldn't be able to reconstruct the buffer for the same node if you would manually kill the persistent buffer.

jethrokuan (approved):
LGTM, thanks!

Issue #1640, opened 2021-07-20
Helm Completion Buffer Displays "Extra" Tags Wrapping to Second Line


mchowning:
### Description

When searching for nodes, Helm is displaying [the `'invisible` tags](https://github.com/org-roam/org-roam/blob/master/org-roam.el#L710), causing the completion candidate to wrap to a new line. To be clear, I do not have this issue when using Ivy. For example, if I have

```
(setq org-roam-node-display-template "${title:*} ${tags:20}")
```

Then I end up getting duplicate tags (once for the `${tags:20}` and once from [the `'invisible` tags](https://github.com/org-roam/org-roam/blob/master/org-roam.el#L710).

![image](https://user-images.githubusercontent.com/4656348/126328918-15318dbd-af1d-446d-a121-848263f2b85a.png)

Although this issue has some similarities to [Helm completion too wide for side-by-side windows #1578](https://github.com/org-roam/org-roam/issues/1578), I think it is a different issue.

#### Occurs on Emacs 27.2, but not 27.1

This may be a helm or emacs issue. I did not observe this behavior on emacs 27.1, but I did observe it on 27.2. I actually had two computers sharing the same dotfile, and I had this problem on the one running 27.2, but not the one running 27.1. Then when I upgraded the computer running 27.1 to use 27.2, it immediately started showing this problem as well.

#### Steps to Reproduce

1. Load Emacs version 27.2
2. Run `org-roam-node-find` using Helm as your completion (I haven't tested other completion frameworks)
3. Observe that the tags for a node are listed once more than expected (based on the value of `org-roam-node-display-template`) and wrap to a second line (unless the tag is extremely short)

#### Expected Results

The completion candidates should only display tags if they are specified in `org-roam-node-display-template` and should not wrap the completion candidate to a second line.

#### Actual Results

The completion candidates should displays tags once more than indicated in `org-roam-node-display-template` and wrap the completion candidate to a second line so the "additional" tags are on the second line..

### Environment

I have this issue both on Linux/Ubuntu and Mac

#### Mac
- Emacs: GNU Emacs 27.2 (build 1, aarch64-apple-darwin20.5.0, NS appkit-2022.50 Version 11.4 (Build 20F71))
 of 2021-07-14
- Framework: Spacemacs
- Org: Org mode version 9.4.6 (9.4.6-11-g1ee52c-elpa @ /Users/matt/.emacs.d/elpa/27.2/develop/org-plus-contrib-20210712/)
[...]

PR #1641, opened 2021-07-20
(fix): remove invisible tag

mchowning:

Fixes https://github.com/org-roam/org-roam/issues/1640.

I'm not convinced this is the right fix for this issue because it is not solving the underlying problem of helm not respecting text properties, but it does fix the immediate problem of the default configuration of org-roam sometimes causing multiline completion candidates.

### Screenshots 
<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -739,12 +739,8 @@ GROUP BY id")))
 
 (defun org-roam-node--to-candidate (node)
   "Return a minibuffer completion candidate given NODE."
-  (let ((candidate-main (org-roam-node--format-entry node (1- (frame-width))))
-        (tag-str (org-roam--tags-to-str (org-roam-node-tags node))))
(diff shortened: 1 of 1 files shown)


PRESFIL:
Probably you've made too many changes, [see](https://github.com/org-roam/org-roam/issues/1640#issuecomment-887797750).

jethrokuan:
I'll merge this as a stop-gap, but it's unfortunate that the invisible strings aren't allowed in the completion... because I really wanted to work a sort of query language into the node completions interface.

mchowning:
>  I really wanted to work a sort of query language into the node completions interface.

That does sound like it would be very valuable, and I wouldn't want the issue with (some versions of) helm to block that. Like you said, this fix was very much just a stop-gap.

PR #1693, opened 2021-07-27
Adding parameters to three core functions

### LINKED ISSUE #1681: Add `filter-fn` and `templates` parameter to three of the main entry points to `org-roam` ecosystem.
### Brief Abstract

Adding a `filter-fn` parameter to `org-roam-capture` and a `templates` parameter to `org-roam-capture` , `org-roam-node-insert`, `org-roam-node-find` would improve their general utility, and reduce fiddling with the more inner functionality of `org-roam-capture-` and `org-roam-node-read`

_*Note:* This is a "nice to have" feature that has a viable work around with advising functions._

_*Update:* I started implementing the advising function work-around it's not as clean as I hoped (namely if you cancel a command it leaves attached advising functions)_

#### On `org-roam` contexts

In [Using Emacs While Running Online Games // Take on Rules](https://takeonrules.com/2020/11/09/using-emacs-while-running-online-games/), I wrote about a conceptual context.  To paraphrase, when I am running my New Vistas in the Thel-Sector game, I make regular use of `org-roam` features.  However, duing those sessions, I am only concerned about finding notes that are tagged `#thel-sector`, and when I capture a note, I always want to use a thel-sector specific template.

Another context that I've experienced: when I'm in a meeting and taking notes for that work meeting, I want my "context" to be scoped to my work notes.

#### Desired Functionality

```gherkin
Given that I have toggled on an org-roam context
When I create a node
Then I am only offered a choice from the given context’s list

Given that I have toggled on an org-roam context
When I search for a node
Then the results only include nodes from the given context

Given that I have toggled on an org-roam context
When I create a daily node
Then the template uses the org-daily template
```
    
#### The Four Main Functions

There are four functions to consider:

-   **`org-roam-capture`:** This function finds or creates a node then prompts for action (Save, Cancel, or Refile).
-   **`org-roam-dailies-capture-today`:** This function creates a new node in the daily template.
-   **`org-roam-node-insert`:** This function finds or creates a node then inserts a link to that node in the originating buffer.
-   **`org-roam-node-find`:** This function finds a node and opens it in a new buffer.

Each of these call two parameterized functions:

-   **`org-roam-node-read`:** relevant is the `filter-fn` parameter, which allows for narrowing of nodes presented for selection.
-   **`org-roam-capture-`:** relevant is the `templates` keyed parameter, which provides the list of applicable `org-roam` templates.

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">
<caption class="t-above"><span class="table-number">Table 1:</span> Function Exposes Parameter</caption>

<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Function</th>
<th scope="col" class="org-left"><code>filter-fn</code></th>
<th scope="col" class="org-left"><code>templates</code></th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left"><code>org-roam-capture</code></td>
<td class="org-left">No</td>
<td class="org-left">No</td>
</tr>

<tr>
<td class="org-left"><code>org-roam-dailies-capture-today</code></td>
<td class="org-left">No</td>
<td class="org-left">No</td>
</tr>

<tr>
<td class="org-left"><code>org-roam-node-insert</code></td>
<td class="org-left">Yes</td>
<td class="org-left">No</td>
</tr>

<tr>
<td class="org-left"><code>org-roam-node-find</code></td>
<td class="org-left">Yes</td>
<td class="org-left">No</td>
</tr>
</tbody>
</table>

#### Two Approaches

Ideally `org-roam-capture`, `org-roam-node-insert`, and `org-roam-node-find` would expose the `filter-fn` and `templates` parameter.  This would allow easier usage of those multi-purpose functions.  The solution would be to create one outer function for each of the capture, insert, and find.

However, without those parameterizations, I’m considering using the `advice-add` feature of Emacs.  I would add advice around both the `org-roam-node-read` and the `org-roam-capture-` functions.

In my experience, meddling with an inner function (e.g., `org-roam-node-read` and `org-roam-capture-`) creates more precarious situations.  That change impacts all callers of those inner functions.

#### Two Scenarios

There are two main scenarios to consider:

1.  I use my `hydra` menu to select a temporary context (e.g., I’m wanting to interact with a game note).
2.  A set `org-roam` context (e.g., I’m running a game and want all of my org-roam activity to default to that context).

In the case of the `hydra` menu, I could rely on the normal `template` selection mechanism and no applicable `filter-fn`.  However, when I’m about to find or create a node, I often have a narrowing concept in mind already (e.g., I already know that I want to write something for my game, let’s narrow it according to the relevant tag and only use the game related templates.)

In the case of the `org-roam` context, there’s a need to manage state.  Once I’ve toggled on that context, I could overwrite the `kbd` settings for the the capture, insert, and find functions (but not the daily).

#### Approaches Overlaying with Scenarios

Now that I’ve established the two scenarios and the two different approaches, I want to explore how I would implement that.  I also want to look at the challenges.

##### Outer Function Approach for Hydra Menu Scenario

For each applicable `org-roam` context, I would create three functions that “wrap” the capture, find, and insert behavior.  Then I would add those functions as menu items in the Hydra menu.

##### Advising Function Approach for Hydra Menu Scenario

For each applicable `org-roam` context, I would create five functions: three for the outer functions (capture, find, and insert) and two advising functions (one for `org-roam-node-read` and one for `org-roam-capture-`).  The three outer functions would follow three steps:

1.  Add the advising functions
2.  Call the inner function (e.g., my context capture would then call `org-roam-capture`)
3.  Remove the advising functions

##### Outer Function Approach for Org Roam Context

As the Outer Function Approach for Hydra Menu Scenario, except that when I’ve toggled on the `org-roam` context the `kbd` incantations for the original capture, find, and insert would map to the context specific functions.

##### Advising Function Approach for Org Roam Context

As the Advising Function Approach for Hydra Menu Scenario, except that when I’ve toggled on the `org-roam` context the `kbd` incantations for the original capture, find, and insert would map to the context specific functions.

#### Analysis of Approaches Overlaying with Scenarios

In the case of writing an outer function, I would write three functions per context.  In the case of advising function, I would need to write five functions.  However, for the outer function method to work, I would need the three functions (e.g., `org-roam-capture`, `org-roam-node-insert`, `org-roam-node-find`) parameterized to each accept a `filter-fn` and `templates` parameter.

I’m uncertain what the issues would be if I had an active `org-roam` context and performed nested `org-roam-capture` calls.  With advising functions, I may end up detaching the function then attempting to detach an advising function that’s no longer attached.

jeremyf:
The rationale for this change is to allow context-specific capture
consideration.  That is to say, in my experience, when I'm working on a
project, I often want to scope my note finding and creation to that
project.  By adding `filter-fn` and `templates` I can easily craft
context-aware functions for my note-taking.

I chose to switch these to `cl-defun` to expose `&key` parameters, which I
find more useful as the method signature grows.

Closes #1681

From the Github Issue:

> Adding a `filter-fn` parameter to `org-roam-capture` and a `templates`
> parameter to `org-roam-capture` , `org-roam-node-insert`,
> `org-roam-node-find` would improve their general utility, and reduce
> fiddling with the more inner functionality of `org-roam-capture-` and
> `org-roam-node-read`

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -807,14 +807,19 @@ TEMPLATES is a list of org-roam templates."
     (org-capture goto keys)))
 
 ;;;###autoload (autoload 'org-roam-capture "org-roam" nil t)
-(defun org-roam-capture (&optional goto keys)
+(cl-defun org-roam-capture (&optional goto keys &key filter-fn templates)
   "Launches an `org-capture' process for a new or existing note.
 This uses the templates defined at `org-roam-capture-templates'.
-Arguments GOTO and KEYS see `org-capture'."
+Arguments GOTO and KEYS see `org-capture'.
+FILTER-FN is a function to filter out nodes: it takes an `org-roam-node',
+and when nil is returned the node will be filtered out.
+The TEMPLATES, if provided, override the list of capture templates (see
+`org-roam-capture-'.)"
   (interactive "P")
-  (let ((node (org-roam-node-read)))
+  (let ((node (org-roam-node-read nil filter-fn)))
     (org-roam-capture- :goto goto
                        :keys keys
+                       :templates templates
                        :node node
                        :props '(:immediate-finish nil))))
 
--- a/org-roam.el
+++ b/org-roam.el
@@ -905,25 +905,30 @@ This section is made out of the next 2 `magit-section's:
       (insert ?\n))))
 
 ;;;###autoload
-(defun org-roam-node-find (&optional other-window initial-input filter-fn)
+(cl-defun org-roam-node-find (&optional other-window initial-input filter-fn &key templates)
   "Find and open an Org-roam node by its title or alias.
 INITIAL-INPUT is the initial input for the prompt.
 FILTER-FN is a function to filter out nodes: it takes an `org-roam-node',
 and when nil is returned the node will be filtered out.
-If OTHER-WINDOW, visit the NODE in another window."
+If OTHER-WINDOW, visit the NODE in another window.
+The TEMPLATES, if provided, override the list of capture templates (see
+`org-roam-capture-'.)"
   (interactive current-prefix-arg)
   (let ((node (org-roam-node-read initial-input filter-fn)))
     (if (org-roam-node-file node)
         (org-roam-node-visit node other-window)
       (org-roam-capture-
        :node node
+       :templates templates
        :props '(:finalize find-file)))))
 
 ;;;###autoload
-(defun org-roam-node-insert (&optional filter-fn)
+(cl-defun org-roam-node-insert (&optional filter-fn &key templates)
   "Find an Org-roam node and insert (where the point is) an \"id:\" link to it.
 FILTER-FN is a function to filter out nodes: it takes an `org-roam-node',
-and when nil is returned the node will be filtered out."
+and when nil is returned the node will be filtered out.
+The TEMPLATES, if provided, override the list of capture templates (see
+`org-roam-capture-'.)"
   (interactive)
   (unwind-protect
       ;; Group functions together to avoid inconsistent state on quit
@@ -948,6 +953,7 @@ and when nil is returned the node will be filtered out."
                          description)))
             (org-roam-capture-
              :node node
+             :templates templates
              :props (append
                      (when (and beg end)
                        (list :region (cons beg end)))


jethrokuan:
Thanks @jeremyf , the implementation here looks good, and the issue writeup was stellar.

PR #1707, opened 2021-07-29
(fix)capture: check for buffer existence before template insertion

jethrokuan:
Redefine a new file to Org-roam's capture system as one that does not
have a buffer visiting it yet. This means that if a buffer is already
visiting the file, then the HEAD template will not be inserted. This is
important because Org-roam capture does not enforce the save of files,
and this can lead to scenarios where the template is inserted multiple
times to a buffer for a file that has not yet been written to disk.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -528,15 +528,15 @@ also run Org-capture's template expansion."
 (defun org-roam-capture--goto-location ()
   "Initialize the buffer, and goto the location of the new capture.
 Return the ID of the location."
-  (let (p)
+  (let (p new-file-p)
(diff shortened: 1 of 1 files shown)


alanz:
This change breaks the `org-roam-dailies-capture-today` function.

Repro
- Use default `org-roam-dailies-capture-templates` value
- invoke `org-roam-dailies-capture-today`
- Save the file, and close it
- invoke `org-roam-dailies-capture-today`

At this point the template is re-inserted into the existing file, in the middle of it

jethrokuan:
@alanz I can't repro with your steps, actually I'd changed this precisely because I noticed the templates were inserted multiple times during some dailies operations...

PR #1710, opened 2021-07-30
(feat): add org-roam-extract-subtree

### LINKED ISSUE #1104: Promote headline to new file
### Brief Abstract
There should be a function equivalent to `org-brain`'s `org-brain-headline-to-file`. 
### Long Description
When expanding a note, if it becomes non-atomic, I wish to be able to select a region or a headline, create a new file, and move the contents there. Any links to the existing headline should then point to the new file.

NOTE: It is crucial to transform the headline's `:ID:` property to a file-level property to preserve existing `id:` links, as well as the workings of other packages.

- [x ] No similar feature requests

jethrokuan:
Adds org-roam-extract-subtree, which converts the subtree at point
into a top-level file node. This is useful for when a node you are
expanding on gets too big, and deserves its own file.

Closes #1104

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -1226,6 +1226,11 @@ Assumes that the cursor was put where the link is."
 (add-hook 'org-roam-find-file-hook #'org-roam-open-id-with-org-roam-db-h)
 
 ;;; Refiling
+(defcustom org-roam-extract-new-file-template "%<%Y%m%d%H%M%S>-${slug}.org"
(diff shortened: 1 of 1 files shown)


aragaer:
In case `org-roam-directory` has no trailing slash this code (using `read-file-name` function) creates files which have the last directory name glued to the file name. While I can fix this by changing `org-roam-directory` it would be nice to have a warning or something.

PR #1711, opened 2021-07-30
Correct type for INFO argument in org-roam-capture docstring

meliache:

According to the documentation of `org-roam-capture--info` and functions working with that variable, it is a property list and not an alist, so that should be stated correctly in the documentation for the `INFO` keyword argument.

Sorry for the minor PR, just found this while stepping through the code with edebug and looking at function signatures while trying to find the cause for another bug that I have.


jethrokuan:
> Sorry for the minor PR, just found this while stepping through the code with edebug and looking at function signatures while trying to find the cause for another bug that I have.

All PRs of this sort are welcome, I tend to miss out little things like these.

Issue #1700, opened 2021-07-28
org-roam capture template doesn't work correctly, capture buffer isn't being shown


RastiGiG:
### Description

I run into an error, whenever i try to run 'org-roam-node-find'. The command actually runs normally first, the error only occurs after i enter the name of the node (any node) into the minibuffer.
The following message shows up in the *warnings* buffer:

org-capture: Capture template ‘d’: Wrong type argument: listp, t

When i look up the variable 'org-roam-capture-templates' it has the value specified in the documentation:

(("d" "default" plain "%?" :if-new
  (file+head "%<%Y%m%d%H%M%S>-${slug}.org" "#+title: ${title}\n")
  :unnarrowed t))

I use the 'normal' org capture regularly and they work just fine for me, which is why i can't really determine, what causes the issue for me.
Also i use straight.el to install packages, but that should not make a difference, should it?
I couldn't find this issue being solved anywhere, so if it was, please point me there. 

#### Steps to Reproduce

1. Load Emacs
2. Run `org-roam-node-find`
3. Enter the name of the node

#### Backtrace

Debugger entered--Lisp error: (error "Capture template ‘d’: Wrong type argument: listp, ...")
  signal(error ("Capture template ‘d’: Wrong type argument: listp, ..."))
  error("Capture template `%s': %s" "d" "Wrong type argument: listp, t")
  org-capture(nil "d")
  org-roam-capture-(:node #s(org-roam-node :file nil :file-hash nil :file-atime nil :file-mtime nil :id nil :level nil :point nil :todo nil :priority nil :scheduled nil :deadline nil :title "Index" :properties nil :olp nil :tags nil :aliases nil :refs nil) :props (:finalize find-file :call-location #<marker at 11345 in *Messages*>))
  (if (progn (or (and (memq (type-of node) cl-struct-org-roam-node-tags) t) (signal 'wrong-type-argument (list 'org-roam-node node))) (aref node 1)) (org-roam-node-visit node other-window) (org-roam-capture- :node node :props '(:finalize find-file :call-location #<marker at 11345 in *Messages*>)))
  (let ((node (org-roam-node-read initial-input filter-fn))) (if (progn (or (and (memq (type-of node) cl-struct-org-roam-node-tags) t) (signal 'wrong-type-argument (list 'org-roam-node node))) (aref node 1)) (org-roam-node-visit node other-window) (org-roam-capture- :node node :props '(:finalize find-file :call-location #<marker at 11345 in *Messages*>))))
[...]

Issue #1688, opened 2021-07-27
Cannot find entry with ID, where the file ID exsit


lengyueyang:
### Description

Migrate from v1 to v2, after insert the node before, ENTER the link, say ""Cannot find entry with ID \"%s\"" "a1013ac4-b632-4c62-b58e-d87b24eb1a23")"

but newly insert node is OK.

#### Steps to Reproduce

1. Load Emacs, open the file aaa.org under org-roam directory
2. Run `org-roam-node-insert`
3. `click the link`

#### Expected Results

  error("Cannot find entry with ID \"%s\"" "a1013ac4-b632-4c62-b58e-d87b24eb1a23")
  org-id-open("a1013ac4-b632-4c62-b58e-d87b24eb1a23" nil)
  org-link-open((link (:type "id" :path "a1013ac4-b632-4c62-b58e-d87b24eb1a23" :format bracket :raw-link "id:a1013ac4-b632-4c62-b58e-d87b24eb1a23" :application nil :search-option nil :begin 217 :end 270 :contents-begin 260 :contents-end 268 :post-blank 0 :parent (paragraph (:begin 217 :end 272 :contents-begin 217 :contents-end 271 :post-blank 1 :post-affiliated 217 :parent nil)))) nil)
  org-open-at-point(nil)
  (if (or (equal (org-element-property :type lineage) "img") (and path (image-type-from-file-name path))) (+org--toggle-inline-images-in-subtree (org-element-property :begin lineage) (org-element-property :end lineage)) (org-open-at-point arg))
  (let* ((lineage (org-element-lineage context '(link) t)) (path (org-element-property :path lineage))) (if (or (equal (org-element-property :type lineage) "img") (and path (image-type-from-file-name path))) (+org--toggle-inline-images-in-subtree (org-element-property :begin lineage) (org-element-property :end lineage)) (org-open-at-point arg)))
  (let nil (let* ((lineage (org-element-lineage context '(link) t)) (path (org-element-property :path lineage))) (if (or (equal (org-element-property :type lineage) "img") (and path (image-type-from-file-name path))) (+org--toggle-inline-images-in-subtree (org-element-property :begin lineage) (org-element-property :end lineage)) (org-open-at-point arg))))
[...]

Issue #1354, opened 2020-12-20
org-roam-dailes-find-date deviates to "Select a capture template"


rakoenig:
### Description
Trying to open an exisiting dailies entry.

#### Steps to Reproduce

M-x org-roam-dailies-find date
The calendar appears, I see the dates that have entries and I selct one.

#### Expected Results
After selecting the date in the calendar the file should open directly without that step through "org capture". 

#### Actual Results

Instead of opening that entry, I end up in "Select a capture template" 
The strange thing is that this template just offers my one line:

[l] lab

above the usual [c] Customize and [q] abort.

 Pressing "l" then shows the entry for the selected date. 

This is kind of strange because org-roam-dailies-capture today shows

[l] lab
[j] journal

in the capture template.

### Environment

- Emacs: GNU Emacs 27.1 (build 1, x86_64-suse-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0)

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

jethrokuan:
Do you have anything in `org-roam-dailies-find-file-hook`?

Unfortunately I can't seem to reproduce. 

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin19.5.0, NS appkit-1894.50 Version 10.15.5 (Build 19F101))
 of 2020-11-13
- Framework: Doom
- Org: Org mode version 9.5 (nil @ /Users/jethrokuan/.emacs.d/.local/straight/build-28.0.50/org-mode/)
- Org-roam: 777f6d2

rakoenig:
`org-roam-dailies-find-file-hook` was nil. 

Now I updated form Org 9.3.x (which came with the emacs 27.1 package for openSUSE) to Org 9.4.2 and the problem is gone.

krummja:
I am also having this issue using emacs 27.1, using roam 1.2.3.

maxmahlke:
I can confirm that upgrading Org from 9.3 to the [latest version](https://code.orgmode.org/bzg/org-mode.git) got rid of the erroneous capture buffer.

meliache:
I have this issue again in the latest org-roam v2 [1] and the latest org [2] git revisions. My `org-roam-dailies-capture-templates` is quite complex, but I checked that I can reproduce the behaviour with 
``` elisp
  (setq org-roam-dailies-capture-templates
        '(("d" "default" entry "* %?" :if-new
[...]

PR #1724, opened 2021-08-01
(feat): globally restructure and refactor the codebase

### LINKED ISSUE #1690: Refactor Org-roam's modules
### Description

Org-roam's codebase is in need of some restructuring. For more context, see https://github.com/org-roam/org-roam/issues/1590#issuecomment-885817825.

### Migration plan

- Many modules have intricate interdependencies, so we migrate most of them into `org-roam.el`.
- Extensions that aren't core to org-roam's functionality are left in their own files, and perhaps in an extensions directory
  - [ ] graph
  - [ ] dailies
  - [ ] protocol
  - [ ] overlay

Wetlize:
This is an alternative PR to #1686 to address #1690. Merging this should close #1686 and fix #1690.

The next need to be addressed before this can be merged:
- [x] Figure out how to make Emacs recognize `extensions` subdirectory as part of the package.
- [x] Decouple `org-roam-capture` from `org-roam-protocol` and `org-roam-dailies` related code.
- [x] Rebase this branch to latest changes from `master` branch and merge conflicts.
- [x] Check and address byte compilation related warning/errors and update the related declarations.
- [x] Update the `make` file.


Wetlize:
> Figure out how to make Emacs recognize extensions subdirectory as part of the package.

@jethrokuan Do you have any ideas how to handle this without resolving to a complicated stuff? The only thing that I can think of so far is to have a sub-package with all the extensions in it and then add this sub-package as a hard dependency of the main package. The extensions from this sub-package wouldn't be automatically `require`d, but instead the sub-package itself would be declared in the `Package-Requires:` comment header of the `org-roam.el` file. This way the `load-path` should(?) work as expected (along with its autoloads) for the files from `extensions/` subdirectory.

jethrokuan:
> Do you happen to have any ideas how to handle this without resolving to a complicated stuff? 

Not sure tbh, I took a look at some of the larger packages (magit, org-mode), and they seem to have placed their extensions in the same directory ("lisp") as the main package. Moving them to "lisp" also seems to require some altercations to the MELPA recipe. I don't think it's a big deal having the extensions remain in the root directory here.

jethrokuan:
How does the load chain now work in this PR? It seems that requiring `org-roam` does not necessarily require `org-roam-db`, which seems problematic.

Also how do you intend to resolve `org-roam-capture` requiring functions from `org-roam-node` and vice-versa?

Wetlize:
> how do you intend to resolve org-roam-capture requiring functions from org-roam-node and vice-versa?

This is already solved.

> How does the load chain now work in this PR? 

https://github.com/org-roam/org-roam/blob/8f7ee477bd180d1aaf4f1dc66d7b6d424e9369fb/org-roam.el#L215-L224

And then each other file only needs to `(require 'org-roam)` and nothing else. It works quite simple (if you think about it), but not obvious: when `require` is called, it always starts by checking whether the corresponding `[feature]` was already previously loaded. It's accomplished by a simple lookup for a corresponding `[feature]` in the global `features` variable. 

If the `[feature]` wasn't loaded, then it will load the file and push the corresponding symbol (from the `(provide [feature])`) to the global `features` variable. Then, for each repeated evaluation of `(require '[feature])`, it will first lookup whether there's a corresponding `[feature]` symbol is in `features`, and if it finds it, it won't do any repeated loading. So if the `[feature]` was already previously loaded, it simply noops its loading.

[This](https://www.gnu.org/software/emacs/manual/html_node/elisp/Named-Features.html) entry from the manual should also help to understand the process. If you still don't understand, I can write up for you two examples: one would show how it loads starting from a different file (e.g. with an autoload) and the other how it loads from the main `org-roam.el` file.

Wetlize:
This also works as long as we don't do something weird *during* the loading. So, say, if you have a top-level form that calls something from another module that wasn't yet loaded, like this:
```emacs-lisp
;; This is `org-roam-capture.el`

;; Call something at the top level from a different (not yet loaded) module
;; during the loading process:
(org-roam-mode-go-bananas)
```
Things can get hairy. But since we don't do this kind of cross-reference and all the utils and third-party dependencies are preload before everything else, we should be good.

Wetlize:
> > Do you happen to have any ideas how to handle this without resolving to a complicated stuff?
> 
> Not sure tbh, I took a look at some of the larger packages (magit, org-mode), and they seem to have placed their extensions in the same directory ("lisp") as the main package. Moving them to "lisp" also seems to require some altercations to the MELPA recipe. I don't think it's a big deal having the extensions remain in the root directory here.

Hm, so I've investigated this for a bit and it certainly isn't something imposed by Emacs. By using the next recipe for Doom I was able to build the package as expected (with respect to the subdirectory):
```emacs-lisp
(package! org-roam
  :recipe (:local-repo "~/Projects/forks/org-roam"
           :files ("extensions/*.el"
                   ;; The default MELPA recipe
                   "*.el" "*.el.in" "dir"
                   "*.info" "*.texi" "*.texinfo"
                   "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
                   (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el"))
           :build (:not compile)))
```
I think if you could just reuse this `:files` property for the MELPA recipe it would just work™. However, if you're against that, I can get rid of the directory and keep it all mixed under the main root.

jethrokuan:
@Wetlize I think I get it now, the changes look good so far. I'll make a PR to MELPA to include the `extensions` directory.

Wetlize:
@jethrokuan I was wondering, does it really worth to keep all the utility based macros under "org-roam-macs.el"? It feels like another layer of mental overhead to keep the track of, and the benefits are not even notable. Same with various `eval-when-compile` and `eval-and-compile` calls all over the codebase.

I'm thinking of getting rid of them (or at least drastically minimize), while move all the macros to "org-roam-utils.el". WDYT?

jethrokuan:
> does it really worth to keep all the utility based macros under "org-roam-macs.el"?

I think it was worth it a long time ago, where the macros were few and far between, but the macros in there are quite commonly used now, especially `org-roam-with-file`, so the performance benefits aren't quite there anymore. I personally don't think it's that much mental overhead (do you expect the number of .el files to grow?), but if you think it matters then it might be another barrier to code contribution, so I'm for any simplification.

Wetlize:
>  you expect the number of .el files to grow?

Well, I do expect the code to change over time, and with each change you need to keep track of various things that can easily get "out of sync" from the code, e.g. comments and documentation. That's what I meant by the mental overhead, not the understanding of what it does, but keeping the track over time of such "directives", you know, it should worth it cost wise. I'll give it another shot tomorrow to think about.

Anyway, I'll do tomorrow the last iteration on the draft stage of this PR and it should be good then for the review, hopefully tomorrow too.

Wetlize:
This should be good for the review; feel free to change things yourself or nitpick.
There's still changes to make, but I don't think that any of them are blocking from the `master` branch. The bigger thing now would be the rename of symbols; I did that to some internal ones, but didn't touch anything public, except the backported change. There's some ambiguity in `org-roam-buffer-*` stuff, where in some places it means `org-mode` base buffers (e.g. `org-roam-buffer-p`), while in `org-roam-mode.el` it's used for `org-roam-mode` based buffers.

BTW, in dailies there's `org-roam-dailies-calendar--date-to-time` function that's not used anywhere. I think it can be removed? I didn't touch it though.

Wetlize:
https://github.com/org-roam/org-roam/blob/2b1238d7eb92bc703cc3ef6d74f6cff95d8e26e6/org-roam-utils.el#L37-L49

I see that you already use `truncate-string-to-width` in the codebase. Any reasons to keep this util?

Wetlize:
https://github.com/org-roam/org-roam/blob/2b1238d7eb92bc703cc3ef6d74f6cff95d8e26e6/org-roam-utils.el#L76-L82

This is also unused. Not sure if you that's needed.

jethrokuan (approved):
I've walked through all the changes, and they look good. Great work @Wetlize :+1: do you have any final changes?

Post this merge, I intend to update the changelog, list the breaking changes and cut a release.

Wetlize:
This is beyond of the scope of this PR, but since you intend to cut a new release the next minor things can be done on my side before doing this:
- Add a global minor mode in place of the setup and teardown. Having a mode for such thing is a canonical thing to have in Emacs, so the user can easily know whether the mode is enabled, hook to it and so on. Now, there's 2 things about this:
  1. `org-roam-mode` is now occupied and by the major mode and renaming it at this point will certainly break things on the user side. What I propose is just to name it as something abstract, e.g. `org-roam-watcher-mode`. In future you can also rename `org-roam-mode` with something more abstract too.
  2. Don't remove the setup and teardown functions. Instead, they can be simply called by the minor mode and we can modify their definition to automatically enable or disable the global minor mode. This way we won't break anything and automatically activate the mode for folks who have `(org-roam-setup)` form in their configs. We can also deprecate them with something like `org-roam-watcher-setup` and `org-roam-watcher-teardown`, or `org-roam-watcher-enable` and `org-roam-watcher-disable`.
- Gracefully handle missing `org-id-locations-file` by either, warning user about this or creating and polluting it ourselves. This can prevent issues like: #1734, https://github.com/org-roam/org-roam/issues/1700#issuecomment-889099832, #1688. Org-mode normally stores `org-id-locations-file` in `user-emacs-directory`, but some configurations can alter its location to a directory that wasn't previously created in the file system, which I suspect is the cause of these issues.
- It still in my todo list to address https://github.com/org-roam/org-roam/commit/829ee688606213da6aca11cbcecfbb512c159f0f#r54227932 and possibly subsequently address #1354.

Let me know what you think about these changes and if you think these changes should be part of the following release. If yes, then what the best way to handle all this, i.e. push all this to this branch, or work on this post merge as separate PRs and then cut a new release when done with this list?

Wetlize:
> do you have any final changes?

Yep, I will soon push them. They should be in scope of this PR.

jethrokuan:
@Wetlize I'm agreeable to all the changes you mentioned above, but let's make them in a separate PR post this merge, so master branch is unblocked and I can fix some of the reported bugs.

Wetlize:
@jethrokuan This should be the last batch of changes. Feel free to change or revert what you don't like.

Wetlize:
Ugh, I specifically renamed those symbols that are less likely to be on the user side of the configuration, even though most of them were public. Only the power users could be potentially damaged by this PR, since I didn't `define-obsolete-function-alias` such functions (to not pollute symbol space with notable chunk of deprecated symbols), but I think they will quickly figure that out without causing a fallout on the issue tracker. The only exception is `org-roam-visit-thing`, for which I handled deprecation as it should be. 

Anyway, no more commits from me on this branch unless you explicitly request some changes, so you can freely merge this without causing problems.

Wetlize:
@jethrokuan The current MELPA recipe doesn't pull extensions as expected. I think it's mandatory to include the wildcard for MELPA to correctly pull all the files for the proper load path and autoloads, i.e.
```diff
(org-roam :fetcher github
          :repo "org-roam/org-roam"
-          :files (:defaults "extensions"))
+          :files (:defaults "extensions/*")) ;; or *.el
```
In Doom, if you just use `:recipe (:files "extensions" ...) ` or `(:recipe (:files "extensions/" ...)` it won't do the right thing, but `:recipe (:files "extensions/*" ...)` will work as expected. Can you update the recipe?

SqrtMinusOne:
Same here, had to manually include the files in the straight.el recipe.
```elisp
(use-package org-roam
  :straight (:host github :repo "org-roam/org-roam"
             :files (:defaults "extensions/*.el"))
  ...)
```

Whil-:
> @jethrokuan I can be wrong, but I think the current MELPA recipe doesn't pull extensions as expected. It might be mandatory to include the wildcard for MELPA to correctly pull all the files for the proper load path and autoloads, i.e.
> 
> ```diff
> (org-roam :fetcher github
>           :repo "org-roam/org-roam"
> -          :files (:defaults "extensions"))
> +          :files (:defaults "extensions/*")) ;; or extensions/*.el
> ```
> 
> It's the same thing in Doom with `straight`. If you just use `:recipe (:files "extensions" ...) ` or `(:recipe (:files "extensions/" ...)` it won't do the right thing, but `:recipe (:files "extensions/*" ...)` will work as expected. ~Assuming I'm right, can you update then the recipe?~ Did it myself after the first reported issue.

Maybe good to patch the documentation for this also? At least until it's working as intended without any custom code..?

jethrokuan:
> Maybe good to patch the documentation for this also? At least until it's working as intended without any custom code..?

@Whil- it should be working now without anything extra from the user.

Whil-:
Hmm, okay. When I tried a moment ago it looked like I didn't have anything defined for the dailies extension. But I'll try again, maybe it was something I did wrong!

Whil-:
Oh, I should mention that I'm using straight, like @SqrtMinusOne. Not Melpa.

aragaer:
Got the update from melpa today. org-roam-dailies are not in "extensions" but still not in autoload. This means I have to add `(require 'org-roam-dailies)` but that isn't in the docs.

PR #1750, opened 2021-08-09
Update org-roam-mode.el

alaiyeshi025:
Modifiy preview related logic

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -1,6 +1,6 @@
 ;;; org-roam-mode.el --- Major mode for special Org-roam buffers -*- lexical-binding: t -*-
 
-;; Copyright © 2020-2021 Jethro Kuan <jethrokuan95@gmail.com>
+;; Copyright ? 2020-2021 Jethro Kuan <jethrokuan95@gmail.com>
 
 ;; Author: Jethro Kuan <jethrokuan95@gmail.com>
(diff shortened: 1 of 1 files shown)


jethrokuan:
It's not clear to me why this logic is an improvement over the current one. E.g. when I say parent element, it does not refer to the heading it is under. I would appreciate if you provided some screenshots on what changed and why.

alaiyeshi025:
> It's not clear to me why this logic is an improvement over the current one. E.g. when I say parent element, it does not refer to the heading it is under. I would appreciate if you provided some screenshots on what changed and why.

Thanks for replying.
I modified the logic because I've found that current code generate the following result:


Not sure the original result is intended or a "bug", while I thought the following would be more convenient for me (as of a newbie to org-roam):


Again, thanks for your time and wonderful work!

jethrokuan:
https://github.com/org-roam/org-roam/pull/1756 contains a change that pulls the outline into greyed preview, which should include your `sdfsadf` headline. Update to latest commit and rebuild your cache for these files and you should observe the change.

alaiyeshi025:
> #1756 contains a change that pulls the outline into greyed preview, which should include your `sdfsadf` headline. Update to latest commit and rebuild your cache for these files and you should observe the change.

Great! Thanks a lot!

PR #1754, opened 2021-08-09
(fix)completions: use full string

### LINKED ISSUE #1728: Note searching doesn't consider all tags
### Description
I just migrated to the org-roam v2. And I found that note searching can't find some notes by tags. For example:

#### Steps to Reproduce

Create two notes.

The first note
```
:PROPERTIES:
:ID:       E4F9AAA7-A49E-44DB-AD7D-0D08F4B28BF7
:END:
#+title: The Haunting of Bly Manor
#+filetags: :tv:tv_series:horror:drama:
```

The second note
```
:PROPERTIES:
:ID:       68E0EFA7-FB3A-420B-999F-DBB83E70E58B
:END:
#+title: Better call Soul
#+filetags: :tv:tv_series:dark_comedy:drama:
```

1. Search `drama`
2. Search `horror`

#### Expected Results

For drama search, both notes should be found.
For horror search, one note should be found

#### Actual Results

For drama search, only the second note is found
For horror search, nothing is found

### Environment

```
Copy info below this line into issue:
- Emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109))
 of 2019-09-02
- Framework: N/A
- Org: Org mode version 9.4.6 (9.4.6-gab9f2a @ /Users/ayrat/.emacs.d/elpa/org-9.4.6/)
- Org-roam: 2.0.0
```

I'm using the latest version of org-roam from melpa - `org-roam-20210801.626`

I migrated to v2 today, before migration note searching considered all tags

jethrokuan:
When strings in org-roam-node-display-template are truncated, they are
no longer searchable using completions, resulting in suprising
behaviour. This change allows truncated strings to still be searched
according to their original value. Closes #1728.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -481,12 +481,13 @@ Uses `org-roam-node-display-template' to format the entry."
          (if (not field-width)
              field-value
            (setq field-width (string-to-number field-width))
-           (truncate-string-to-width
-            field-value
-            (if (> field-width 0)
-                field-width
-              (- width (cdr fmt)))
-            0 ?\s)))))))
+           (propertize field-value
+                       'display (truncate-string-to-width
+                                 field-value
+                                 (if (> field-width 0)
+                                     field-width
+                                   (- width (cdr fmt)))
+                                 0 ?\s))))))))
 
 (defun org-roam-node-read--process-display-format (format)
   "Pre-calculate minimal widths needed by the FORMAT string."


Wetlize:
@jethrokuan I've noticed that starting from this PR (investigated after bisecting) completion frontends (tested with `vertico` and `ivy`) cannot anymore highlight matched completions. I don't think it's something critical and I didn't even notice that until yesterday, but just letting you know.

PR #1758, opened 2021-08-09
(feat): replace `org-roam-{setup,teardown}` with `org-roam-db-autosync-mode`

Wetlize:
Comply with the principle of least astonishment, where in Emacs having a global minor mode for this purpose would be the most expected thing to have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Here's a sample configuration with `use-package`:
              ;; Dailies
              ("C-c n j" . org-roam-dailies-capture-today))
       :config
-      (org-roam-setup)
+      (org-roam-db-autosync-mode)
       ;; If using org-roam-protocol
       (require 'org-roam-protocol))
 ```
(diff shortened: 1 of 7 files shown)


Wetlize:
@jethrokuan off-topic, but just to clarify, are you okay with me pushing PRs to the main branch without your reviews? So I can iterate faster without waiting on you; at least with things that I more or less confident about. 

Just to note: this PR is not one of such things that I'm confident about that you'll be okay with, so it waits for the review.

jethrokuan:
> @jethrokuan off-topic, but just to clarify, are you okay with me pushing/merging PRs to the main branch without your reviews? So I can iterate faster without waiting on you; at least with things that I more or less confident about.

@Wetlize Yeah feel free to merge most things (minor PRs, bug fixes). It's only PRs which add stuff that I'd like to look at, because I don't want to make the same mistake of letting the core grow too big.

PR #1759, opened 2021-08-10
Fix candidate display.

blester125:
Recent changes to how candidates are formatted (using a display property
to allow for searching in the full candidate while only showing enough
to fit in the frame), #1754, causes problems when the field value is already
preformatted (a user function, like the one provided in doom emacs,
already applies some faces with `propertize`). Generally the display
string get messed up and formatting gets applied to the whole display.
This PR updates the `propertize` call that adds the display property to the candidate
to remove all properties from the candidate before the styled (and
truncated) display string is added as a property. This allows the
display string to look correct according to the properties a user has
already assigned.

This PR also fixes another bug from this change, where if a
`field-value` is and empty string `""` it would not get padded out to
the `field-width` causing misalignment between candidates.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -481,13 +481,21 @@ Uses `org-roam-node-display-template' to format the entry."
          (if (not field-width)
              field-value
            (setq field-width (string-to-number field-width))
-           (propertize field-value
-                       'display (truncate-string-to-width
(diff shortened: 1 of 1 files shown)


jethrokuan (approved):
Thanks!

PR #1765, opened 2021-08-11
[Fix #1763] Added condition-case to org-roam-preview-get-contents

### LINKED ISSUE #1763: org-roam-preview-get-contents crashes when a contributing file has been removed
#### Steps to Reproduce
1. Create a kind of hub node in org-roam.
2. Make a link to this hub node in some other org-roam files.
3. Make sure the back-link structure is up to date, for example with `M-x org-roam-db-sync`
4. Remove one of the linked files outside of org-roam, for example with the shell command `rm <filename>`.
5. This causes the cache to be not updated.
5. Visit the hub node and construct the `*org-roam*` buffer with `M-x org-roam-buffer-toggle`.
6. The process will crash in the function `org-roam-preview-get-contents` because the file cannot be opened and either an incomplete buffer will be formed or a previous buffer will be left.

#### Expected Results

A properly formed `*org-roam*` buffer or at least a one with the existing files.

#### Actual Results

No buffer is formed, or one from another node is kept.

### Environment

    - Emacs: GNU Emacs 27.2 (build 1, x86_64-apple-darwin19.6.0, NS appkit-1894.60 Version 10.15.7 (Build 19H1030))
     of 2021-05-17
    - Framework: N/A
    - Org: Org mode version 9.4.6 (9.4.6-gab9f2a @ /Users/pieter/.emacs.d/elpa/org-9.4.6/)
    - Org-roam: 2.0.0
    
I have a patch, but for some reason GitHub doesn't allow me to upload my changes. That also means I can't make a Pull Request. So I patch it here.

    modified   org-roam-mode.el
    @@ -396,20 +396,22 @@ window instead."
     (defun org-roam-preview-get-contents (file point)
       "Get preview content for FILE at POINT."
       (save-excursion
    -    (org-roam-with-temp-buffer file
    -      (goto-char point)
    -      (let ((elem (org-element-at-point)))
    -        ;; We want the parent element always
    -        (while (org-element-property :parent elem)
    -          (setq elem (org-element-property :parent elem)))
    -        (pcase (car elem)
    -          ('headline                    ; show subtree
    -           (org-roam-preview-get-entry-text (point-marker) most-positive-fixnum))
    -          (_
    -           (let ((begin (org-element-property :begin elem))
    -                 (end (org-element-property :end elem)))
    -             (or (string-trim (buffer-substring-no-properties begin end))
    -                 (org-element-property :raw-value elem)))))))))
    +    (condition-case err
    +        (org-roam-with-temp-buffer file
    +          (goto-char point)
    +          (let ((elem (org-element-at-point)))
    +            ;; We want the parent element always
    +            (while (org-element-property :parent elem)
    +              (setq elem (org-element-property :parent elem)))
    +            (pcase (car elem)
    +              ('headline                    ; show subtree
    +               (org-roam-preview-get-entry-text (point-marker) most-positive-fixnum))
    +              (_
    +               (let ((begin (org-element-property :begin elem))
    +                     (end (org-element-property :end elem)))
    +                 (or (string-trim (buffer-substring-no-properties begin end))
    +                     (org-element-property :raw-value elem)))))))
    +      (error  (message "Error reading file: %s" err) (error-message-string err)))))

pietvo:
Take default action if FILE no longer exists

See issue #1763

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -396,20 +396,22 @@ window instead."
 (defun org-roam-preview-get-contents (file point)
   "Get preview content for FILE at POINT."
   (save-excursion
-    (org-roam-with-temp-buffer file
-      (goto-char point)
-      (let ((elem (org-element-at-point)))
-        ;; We want the parent element always
-        (while (org-element-property :parent elem)
-          (setq elem (org-element-property :parent elem)))
-        (pcase (car elem)
-          ('headline                    ; show subtree
-           (org-roam-preview-get-entry-text (point-marker) most-positive-fixnum))
-          (_
-           (let ((begin (org-element-property :begin elem))
-                 (end (org-element-property :end elem)))
-             (or (string-trim (buffer-substring-no-properties begin end))
-                 (org-element-property :raw-value elem)))))))))
+    (condition-case err
+        (org-roam-with-temp-buffer file
+          (goto-char point)
+          (let ((elem (org-element-at-point)))
+            ;; We want the parent element always
+            (while (org-element-property :parent elem)
+              (setq elem (org-element-property :parent elem)))
+            (pcase (car elem)
+              ('headline                    ; show subtree
+               (org-roam-preview-get-entry-text (point-marker) most-positive-fixnum))
+              (_
+               (let ((begin (org-element-property :begin elem))
+                     (end (org-element-property :end elem)))
+                 (or (string-trim (buffer-substring-no-properties begin end))
+                     (org-element-property :raw-value elem)))))))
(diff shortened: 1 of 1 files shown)


Wetlize:
This looks like a workaround, rather than a proper fix. The error IMO wouldn't be useful nor to the end user, nor to us. If the file doesn't exist, then it shouldn't be visited in the first place, so either we need to have a file watcher (e.g. with `file-notify-*`) to react to external changes related to the filesystem, or check whether the file exists before visiting it with something like `file-exists-p`.

pietvo:
A better solution will be appreciated.

PR #1767, opened 2021-08-13
(feat) scroll node to top

kimim:

When visit a headline node, it will display at the bottom. this PR scroll the headline to top of the buffer


Wetlize:
This might looks trivial, but I think different people will have different preferences where the display should start from. I'll add this to my todo list, but will need to spend a bit more time to think this through.

kimim:
Yes. I understand. This line works for me now:

```emacs-lisp
(advice-add 'org-roam-node-visit :after (lambda (&rest r) (recenter-top-bottom 'top)))
```

PR #1768, opened 2021-08-13
Remove duplicate edges from graphs

bram85:
When a node links to the another node twice, two edges appear in the
graph. To keep the graph clean, remove all duplicate edges.


jethrokuan (approved):

jethrokuan:
@bram85 can I trouble you to fix the linting error? Can't push to your branch. The lines are too long, I think using a newline right before `WHERE` would be sufficient.

bram85:
OK, fixed the linting issue.

Issue #1734, opened 2021-08-02
org-id-add-location: Wrong type argument: hash-table-p, nil


matheusfillipe:
### Description
When trying to create a new note there is that error and even though the new org buffer is created emacs doesn't switch to it automatically.
 
#### Steps to Reproduce
1. M-x org-roam-node-find
2. Type a new node name
3. Get error: `org-id-add-location:` Wrong type argument: hash-table-p, nil`

#### Backtrace
```lisp
Debugger entered--Lisp error: (wrong-type-argument hash-table-p nil)
  puthash("849d414d-ad2b-4640-90a5-0d70c5a86f12" "~/.org-roam/20210802165354-newfile.org" nil)
  org-id-add-location("849d414d-ad2b-4640-90a5-0d70c5a86f12" "/home/matheus/.org-roam/20210802165354-newfile.org")
  org-id-get(1 create)
  org-id-get-create()
  org-roam-capture--goto-location()
  org-roam-capture--get-point()
  org-capture-set-target-location(nil)
  #f(compiled-function (&optional goto keys) "Capture something.\n\\<org-capture-mode-map>\nThis will let you select a template from `org-capture-templates', and\nthen file the newly captured information.  The text is immediately\ninserted at the target location, and an indirect buffer is shown where\nyou can edit it.  Pressing `\\[org-capture-finalize]' brings you back to the previous\nstate of Emacs, so that you can continue your work.\n\nWhen called interactively with a `\\[universal-argument]' prefix argument GOTO, don't\ncapture anything, just go to the file/headline where the selected\ntemplate stores its notes.\n\nWith a `\\[universal-argument] \\[universal-argument]' prefix argument, go to the last note stored.\n\nWhen called with a `C-0' (zero) prefix, insert a template at point.\n\nWhen called with a `C-1' (one) prefix, force prompting for a date when\na datetree entry is made.\n\nELisp programs can set KEYS to a string associated with a template\nin `org-capture-templates'.  In this case, interactive selection\nwill be bypassed.\n\nIf `org-capture-use-agenda-date' is non-nil, capturing from the\nagenda will use the date at point as the default date.  Then, a\n`C-1' prefix will tell the capture process to use the HH:MM time\nof the day at point (if any) or the current HH:MM time." (interactive "P") #<bytecode 0x1591579f1b09>)(nil "d")
[...]

PR #1769, opened 2021-08-13
(fix)org-id: gracefully handle absence of `org-id-locations-file`

### LINKED ISSUE #1638: Recursively cache all org files in org-roam directory
### Brief Abstract

Function needed to recursively call `org-id-get-create` for all existing org files in the org directory without an `id` so that they can be cached by org-roam.

### Long Description

I am migrating from a regular org-mode notes directory into `org-roam`. It seem like only files and headings with an existing id are picked up by org-roam, but most of my org files have not been indexed this way. Is there an existing function that does this job? If not, is it possible to implement such function to help people migrating from regular org-mode notes?

Implement a function that creates an `ord-id` for all files in `org-roam` directory.

### LINKED ISSUE #1700: org-roam capture template doesn't work correctly, capture buffer isn't being shown

Wetlize:
This is the source of the errors reported in #1734, #1700, #1688. Ideally this problem should be handled by `org-id` itself, but for now this will be only patched in Org-roam.

Merging this fixes #1700 and fixes #1638.

### DIFF
--- a/org-roam-compat.el
+++ b/org-roam-compat.el
@@ -132,18 +132,24 @@ passed to it."
 WARNING: `org-id-locations-file' (%s) doesn't exist!
          Org-roam is unable to create it for you.
 --------
-This happens when Emacs doesn't have permissions to create the path to your `org-id-locations-file'.
-Org-roam will now fallback storing the file in your current `org-roam-directory', but the warning
-will keep popup with each new session.
 
-To stop this warning from popping up, set `org-id-locations-file' to the location you want and ensure
-that the path exists on your filesystem, then run M-x `org-roam-update-org-id-locations'.
+This happens when Emacs doesn't have permissions to create the
+path to your `org-id-locations-file'. Org-roam will now fallback
+storing the file in your current `org-roam-directory', but the
+warning will keep popup with each new session.
 
-Note: While Org-roam doesn't depend on `org-id-locations-file' to lookup IDs for the nodes that are stored in the database,
-it still tries to keep it updated so IDs work across other files in Org-mode, so the IDs used
-in your `org-roam-directory' would be able to cross-reference outside of `org-roam-directory'.
-It also allows to keep linking with \"id:\" links within the current `org-roam-directory' to headings
-and files that are excluded from identification (e.g. with \"ROAM_EXCLUDE\" property) as Org-roam
+To stop this warning from popping up, set `org-id-locations-file'
+to the location you want and ensure that the path exists on your
+filesystem, then run M-x `org-roam-update-org-id-locations'.
+
+Note: While Org-roam doesn't depend on `org-id-locations-file' to
+lookup IDs for the nodes that are stored in the database, it
+still tries to keep it updated so IDs work across other files in
+Org-mode, so the IDs used in your `org-roam-directory' would be
+able to cross-reference outside of `org-roam-directory'. It also
+allows to keep linking with \"id:\" links within the current
+`org-roam-directory' to headings and files that are excluded from
+identification (e.g. with \"ROAM_EXCLUDE\" property) as Org-roam
 nodes." org-id-locations-file)
                         (setq org-id-locations-file
                               (expand-file-name ".orgids" (file-truename org-roam-directory)))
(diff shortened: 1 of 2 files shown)


org-roam-compat.el
```diff
@@ -95,6 +95,61 @@ recursion."
             (push (concat dir "/" file) files)))))
     (nconc result (nreverse files))))
 
+;;; Compatibility hacks and patches
+(advice-add #'org-id-add-location :around #'org-roam--handle-absent-org-id-locations-file-a)
```
Wetlize, line 99: Going through the advice to `org-id-add-location`, because `org-id` related values can be altered through the runtime by the user (e.g. with dir-locals) and the related state can be configured in lazy loaded manner.

org-roam-compat.el
```diff
@@ -95,6 +95,61 @@ recursion."
...
+                     (make-directory (file-name-directory (file-truename org-id-locations-file)))
+                     (org-roam-update-org-id-locations)
+                     (apply fn args))
+            ;; In case of failure (lack of permissions), we'll patch it to at least handle the current session
+            ;; without errors.
+            (file-error (org-roam-message "Failed to regenerate `org-id-locations-file'")
+                        (lwarn 'org-roam :error "
+--------
+WARNING: `org-id-locations-file' (%s) doesn't exist!
+         Org-roam unable to recreate it for you.
```
daviwil, line 133:
```suggestion
         Org-roam is unable to recreate it for you.
```

PR #1776, opened 2021-08-15
Stop point moving when visiting node if point already at node

duncanburke:
When a buffer is buried and then brought up again, it is unexpected for
point to move. Currently, org-roam-node-find-noselect always sets point
to the pos column from the database, which corresponds to the very
beginning of the node in its file.

This patch extracts the logic in org-roam-node-at-point that finds the
node id associated with point into a new function,
org-roam-node--id-at-point. Then, org-roam-node-find-noselect is
modified to check whether point for the buffer associated with the node
is already at the target node. Only if it is not, is point moved to the
target node.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -355,12 +355,14 @@ GROUP BY id")))
 
 ;;;; Finders
 (defun org-roam-node-find-noselect (node)
-  "Navigate to the point for NODE, and return the buffer."
+  "Navigate to the point for NODE if point is not already at NODE, and return the buffer."
   (unless (org-roam-node-file node)
     (user-error "Node does not have corresponding file"))
   (let ((buf (find-file-noselect (org-roam-node-file node))))
     (with-current-buffer buf
-      (goto-char (org-roam-node-point node)))
+      (unless (equal (org-roam-node-id node)
+                     (org-roam-id-at-point))
+        (goto-char (org-roam-node-point node))))
     buf))
 
 (defun org-roam-node-visit (node &optional other-window)


Wetlize:
> This patch extracts the logic in org-roam-node-at-point that finds the
node id associated with point into a new function,
org-roam-node--id-at-point.

This looks exactly like `org-roam-id-at-point`
https://github.com/org-roam/org-roam/blob/8a21131d9f5f6a136bf355b2d3f3f1c765f0eddd/org-roam-node.el#L838-L848

> When a buffer is buried and then brought up again, it is unexpected for
point to move. Currently, org-roam-node-find-noselect always sets point
to the pos column from the database, which corresponds to the very
beginning of the node in its file.

> org-roam-node-find-noselect is
modified to check whether point for the buffer associated with the node
is already at the target node. Only if it is not, is point moved to the
target node.

What's your use case? You mean, brought up back if you visit it again with something like `org-roam-node-find`? If you just switch the buffer or change the window it should already preserve the current position, but if you visit it with again with `org-roam-node-find`, then it will change the position, in which case it should be expected behavior, just like in case of any other jump.

duncanburke:
> This looks exactly like org-roam-id-at-point

Oh, sorry. I didn't see that. I've updated the branch accordingly. I think that org-roam-node-at-point should be modified to use org-roam-id-at-point to remove the code duplication, but that doesn't belong in this pull request.

> What's your use case?

I'm writing a package org-roam-luhmann that enables a closer replication of the original Zettelkasten workflow used by Niklas Luhmann. In this some nodes have titles that are special, in that they imply additional structure. For example, the titles "foo1", "foo1a1", "foo1a2", "foo1b" are in order, and interactive functions are provided for moving between them in sequence.

If I'm at the node with the title "foo1a2" I want to be able to skip backwards to "foo1a1" and then back without the position in the buffer associated with "foo1a2" being changed.

> but if you visit it with again with org-roam-node-find, then it will change the position

That's not true. So long as the buffer wasn't actually closed, and point was left in the node that's being visited (i.e not under a heading with a different id) the point won't move.

Wetlize:
> For example, the titles "foo1", "foo1a1", "foo1a2", "foo1b" are in order, and interactive functions are provided for moving between them in sequence.

> If I'm at the node with the title "foo1a2" I want to be able to skip backwards to "foo1a1" and then back without the position in the buffer associated with "foo1a2" being changed.

So you need this behavior for `org-roam-node-find-noselect` just for your own custom commands and functions? And not to alter the behavior of all the current Org-roam functions that use it for their own purpose? Right now things that use `org-roam-node-find-noselect` in Org-roam make an assumption that the navigation will be always accomplished to where the node starts from. That's why this PR wasn't quickly merged.

I think what we can do is to introduce an `&optional` parameter to the function that would either, preserve the position of the point if it's already at the same ID, or it would force movement to the beginning of the node. A reply on the paragraph above would help to determine how to proceed with this.

> That's not true. So long as the buffer wasn't actually closed, and point was left in the node that's being visited (i.e not under a heading with a different id) the point won't move.

I hope there's no misunderstanding, because I'm talking specifically about the current implementation (i.e. without this patch), because that's not the case. The point will move if you e.g. bury the buffer and revisit the exact same node with `org-roam-node-find`. Maybe you have something in your config that doesn't allow it to do so?

duncanburke:
> I hope there's no misunderstanding, because I'm talking specifically about the current implementation (i.e. without this patch), because that's not the case. 

Okay, I think we had a misunderstanding.

I was talking about the behavior with my patch. With my patch, calling `org-roam-node-find` will not move the point when visiting a node with `org-roam-node-find`  if the point is already in the node. I see `org-roam-node-find` as morally equivalent to `find-file`. `find-file` doesn't move point to point-min if the buffer is already open, and I found it very unexpected and unpleasantly surprising that `org-roam-node-find` does.

To be completely clear, [here](https://gist.github.com/duncanburke/159dd5f8cf97c061c2cafd9910233bba) is how I tested it with `emacs -Q`. With master checked out the assert fails, with my patch it doesn't. And the same thing can be observed interactively with `org-roam-node-find`

> Right now things that use org-roam-node-find-noselect in Org-roam make an assumption that the navigation will be always accomplished to where the node starts from.

That makes sense. The problem with it is that that behavior is quite surprising when used with an interactive function.

> So you need this behavior for org-roam-node-find-noselect for your own custom commands and functions? 

I wouldn't say that. My own custom commands do much the same thing as `org-roam-node-find`: navigate to a particular node. I think this is a problem with the behavior of `org-roam-node-find`.

> Also, do your functions that use org-roam-node-find-noselect operate only in the context of current buffer or cross buffers/files too?

I'm not quite sure what you mean, but with the package I'm writing having multiple nodes in a single file is very much a niche use case. So the expected use is:

1. in a window visiting a node
2. a command is called that visits another node in the same window
3. another command is called that visits the first node in the same window

And I don't expect point to have moved.

carwin:
@duncanburke I tried reproducing the issue you're trying to solve and I actually didn't see the point moving.

Here's what I did:
1. Open a roam file in a buffer (the pointer was on a headline)
2. Switch to another buffer with a non-roam document opened in it
3. Jump back to the earlier roam buffer

In that little test, the pointer remained on the headline I left it on in step 1. Is this the behavior you're trying to get? It's possible I misunderstood what you're trying to say here.

duncanburke:
You didn't say what you used to go back to the buffer. `find-file` or `switch-buffer` will of course behave normally. The point is that anything that uses org-roam's mechanisms to grab the buffer for a particular node (e.g. `org-roam-node-find`) will always set the point to the very beginning of the node. In the usual case of a file with a single top-level node, that means it sets point to 1.

Wetlize:
Closing this in favor of #1778. This should now behave according to the description of this PR and the following discussion.

PR #1779, opened 2021-08-17
(feat)db: use filenotify to trigger db updates

jethrokuan:
Using filenotify to trigger db updates is the most reliable way to ensure the
database is always up to date, making it resilient to changes from outside
Emacs.

This will also address issues seen in #1765, although the root cause should also be patched.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -75,17 +75,17 @@ database."
 (defcustom org-roam-db-update-method (if file-notify--library
                                          'filenotify
                                        'on-save)
-  "How to keep Org-roam's database updated. This supports the
-following values:
+  "How to keep Org-roam's database updated.
+This supports the following values:
 
-  `on-save'
-    update the Org-roam database for the file that has changed.
+  'filenotify
+    Update Org-roam upon detecting changes from the filesystem.
 
-  `filenotify'
-   update Org-roam upon detecting changes from the filesystem.
+  'on-save
+    Update the Org-roam database for the file that has changed.
 
-  `nil'
-  Do not automatically update the Org-roam database."
+  nil
+    Do not automatically update the Org-roam database."
   :type '(choice (const :tag "On save" onsave)
                  (const :tag "Filenotify" filenotify)
                  (const :tag "Do not autoupdate" nil))
@@ -547,6 +547,9 @@ If FORCE, force a rebuild of the cache from scratch."
           (org-roam-db-update-file file))))))
 
 (defun org-roam-db-fn-callback (event)
+  "Handles filesystem EVENT.
+If Org-roam files are changed, we update the database
+accordingly."
   (pcase event
     (`(_ _ ,f)
      (when (org-roam-file-p f) (org-roam-db-update-file f)))
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -31,7 +31,6 @@
 ;;
 ;;; Code:
 (require 'org-roam)
-(require 'filenotify)
 (defvar org-outline-path-cache)
 
 ;;; Options
@@ -593,6 +592,7 @@ database, see `org-roam-db-sync' command."
    (org-roam-db-autosync-mode ; enabled
     (pcase org-roam-db-autosync-update-method
       ('filenotify
+       (require 'filenotify)
        (cl-pushnew
         (cons org-roam-directory
               (file-notify-add-watch org-roam-directory '(change)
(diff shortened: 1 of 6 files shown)


org-roam-db.el
```diff
@@ -544,33 +571,63 @@ database, see `org-roam-db-sync' command."
...
+  "Setup `org-roam-db-autosync-update-method' dependently on the mode's state."
+  (cond
+   (org-roam-db-autosync-mode ; enabled
+    (pcase org-roam-db-autosync-update-method
+      ('filenotify
+       (cl-pushnew
+        (cons org-roam-directory
+              (file-notify-add-watch org-roam-directory '(change)
+                                     #'org-roam-db-autosync--filenotify-update))
+        org-roam-db-autosync--filenotify-descriptors))
```
Wetlize, line 600: `file-notify-add-watch` isn't recursive, so we need to think through how to handle subdirectories. The current logic should handle multiple `org-roam-directory` just fine... I think (untested).

jethrokuan: > file-notify-add-watch isn't recursive, so we need to think through how to handle subdirectories.

ahh... we can watch every directory that's currently in `org-roam-directory`, and then on event, if a directory is created we can use the callback to create more file watchers. Not sure if it's efficient/worth it anymore though.

Wetlize: I mean, we kind of halfway there, so might as well try to handle this? Though if you think it's not worth the hassle to continue, then we can stop. 

If you wish to continue, then we can update [the current facility](https://github.com/org-roam/org-roam/blob/b8a66deae99e0feab3106d7b7ec36af1631f2dff/org-roam.el#L233-L307) to only list subdirectories given a special parameter, and then add something like `org-roam-list-(sub?)directories`. Though it shouldn't include patterns like `^\.`, so we wouldn't add file watchers to e.g. `git` related stuff. And to handle different `org-roam-directory` values we can use `add-variable-watcher` or `org-roam-find-file-hook`.

jethrokuan: I think i'll write a helper library for this and publish it to MELPA.

jethrokuan: https://github.com/jethrokuan/filenotify-recursive here it is... i think it works for now.

Wetlize: I'll take a look and might even try to integrate it there, before it will be added to Melpa. Do you think listing directories with Elisp won't be a bottleneck? I could add support for external finders (similarly to what we have in Org-roam) to the package, if you think it would worth it.

Wetlize: Huh, surprisingly `fnr--subdirectories-recursively` isn't slow. Don't think it will need external finders, at least until someone will show an actual demand for that, so I'll just integrate it to this PR as given.

jethrokuan: > I'll try to integrate it before it will be added to Melpa. But before doing so, do you think listing directories with Elisp won't be a problem? I could add support for external finders (similarly to what we have in Org-roam) to the package, if you think it would worth it.

I went ahead and did it already, and am noticing just one issue: It's not very smart in excluding hidden folders (like .git), so it does a lot of unnecessary watching.

Wetlize: Can you expose a `defvar` for the [predicate](https://github.com/jethrokuan/filenotify-recursive/blob/049de0b15d8478c2ffed008a580e7bc36c6b4a59/filenotify-recursive.el#L65)? Then we can use some dumb heuristic to determine whether each directory is hidden or not.

org-roam-db.el
```diff
@@ -597,7 +597,10 @@ database, see `org-roam-db-sync' command."
       ('filenotify
        (cl-pushnew
         (cons org-roam-directory
-              (fnr-add-watch org-roam-directory '(change) #'org-roam-db-autosync--filenotify-update))
+              (fnr-add-watch org-roam-directory
+                             '(change)
+                             #'org-roam-db-autosync--filenotify-update
+                             "\\`\\.")) ; Ignore directories that start with "."
```
Wetlize, line 603: This won't ignore hidden directories within subdirectories. Not sure you do this intentionally though.

Wetlize:
@jethrokuan Any arguments against https://github.com/org-roam/org-roam/pull/1779/commits/b5f9e40389551f8ea21f02c153040a1ccc914057?

jethrokuan:
@Wetlize No objections here. I was thinking of queuing changes to process, and then processing them only when required: this could be before the org-roam buffer display, or just on an idle-timer. This can reduce a lot of unnecessary re-computation, especially when the same file is frequently saved over.

PR #1787, opened 2021-08-19
(docs)readme: update installation section

Wetlize:

I've noticed that a lot of new people to Emacs just blindly copy the configuration sample from the "Installation" section, when they have easier or superior ways to do so in their environment (e.g. with Doom). This also now provides instructions how to properly install the package without using MELPA, so the `extensions` directory would be automatically included in the `load-path`.


Wetlize:
This doesn't include instructions for Spacemacs, with which I'm just not familiar with. If someone willing to include them, then it should follow similar instructions to Doom, i.e. include the default way to activate it there, then also specify how to alter the package recipe by downloading it from the source repository.

jethrokuan (approved):

PR #1798, opened 2021-08-21
fix: Do not skip invisible headings when getting node at point.

aragaer:
I'm using `org-roam-node-at-point` to fetch the title of the node when building the agenda. Not all my agenda-items are roam-nodes themself and I want to have access to the title of the parent node. However depending on whether the heading tree is visible or collapsed I'm getting different result. Turns out it's just a matter of `org-back-to-heading-or-point-min` which needed `invisible-ok` argument.

Example of the file that show different results depending on visibility
```
:PROPERTIES:
:ID:       dfe77ff7-7e27-4fdd-ad7c-ffd6aadc87bd
:END:
#+title: org-roam
#+startup: overview

* test
** inner
:PROPERTIES:
:ID:       4ced70aa-7b8c-41f0-b43e-3554dbbeeb10
:END:
*** item
SCHEDULED: <2021-08-21 Sat>
```

The result is either "inner" or "org-roam" depending on visibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#1795](https://github.com/org-roam/org-roam/pull/1795) buffer: optimized reflinks fetch
 
 ### Fixed
+- [#1798](https://github.com/org-roam/org-roam/pull/1798) org-roam-node-at-point: do not skip invisible headings
 
 ## 2.1.0
 ### Added
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -180,7 +180,7 @@ populated."
                                     (magit-section-up)
                                     (org-roam-node-at-point)))
         (t (org-with-wide-buffer
-            (org-back-to-heading-or-point-min)
+            (org-back-to-heading-or-point-min t)
             (while (and (not (org-roam-db-node-p))
                         (not (bobp)))
               (org-roam-up-heading-or-point-min))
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - [#1795](https://github.com/org-roam/org-roam/pull/1795) buffer: optimized reflinks fetch
 
 ### Fixed
+- org-roam-node-at-point: do not skip invisible headings
```
jethrokuan, line 10: Please include the PR link here as well.

org-roam-node.el
```diff
@@ -180,7 +180,7 @@ populated."
                                     (magit-section-up)
                                     (org-roam-node-at-point)))
         (t (org-with-wide-buffer
-            (org-back-to-heading-or-point-min)
+            (org-back-to-heading-or-point-min t)
```
jethrokuan, line 183: Good catch. There are other places where this needs to be fixed as well:

https://github.com/org-roam/org-roam/blob/82733e133ab03b0138201e28dcb06e073a4fb3cc/org-roam-node.el#L868
https://github.com/org-roam/org-roam/blob/82733e133ab03b0138201e28dcb06e073a4fb3cc/org-roam-node.el#L830

basically all calls to `org-back-to-heading-or-point-min`.

jethrokuan (changes requested):

jethrokuan (approved):
Thanks :)

PR #1799, opened 2021-08-22
Require org-macs when compiling org-roam-utils.el

vderyagin:
If org-roam-utils.el is byte-compiled when org-macs is not loaded (`org-with-point-at` macro not available), error can appear about missing `org-with-point-at`

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -32,6 +32,8 @@
 ;;
 ;;; Code:
 
+(require 'org-roam)
+
 ;;; String utilities
 ;; TODO Refactor this.
 (defun org-roam-replace-string (old new s)
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -32,6 +32,9 @@
 ;;
 ;;; Code:
 
+(eval-when-compile
+  (require 'org-macs))
+
 ;;; String utilities
 ;; TODO Refactor this.
 (defun org-roam-replace-string (old new s)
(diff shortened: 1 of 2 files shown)


Wetlize:
Since we don't use any of the utils in the main file, can you just move [(require 'org-roam-utils)](https://github.com/org-roam/org-roam/blob/04b7780ff9f916be0ed303e7614965ad7444a45e/org-roam.el#L97)  to the [bootstrap sequence](https://github.com/org-roam/org-roam/blob/04b7780ff9f916be0ed303e7614965ad7444a45e/org-roam.el#L312-L317) as the first entry there? And then use `(require 'org-roam)` at the top of the util file. This way it will pull all the needed dependencies during compilation and we wouldn't need to manage them in multiple places.

Wetlize (changes requested):
Since we don't use any of the utils in the main file, can you just move [(require 'org-roam-utils)](https://github.com/org-roam/org-roam/blob/04b7780ff9f916be0ed303e7614965ad7444a45e/org-roam.el#L97)  to the [bootstrap sequence](https://github.com/org-roam/org-roam/blob/04b7780ff9f916be0ed303e7614965ad7444a45e/org-roam.el#L312-L317) as the first entry there? And then add instead `(require 'org-roam)` at the top of the util file. This way it will pull all the needed dependencies during compilation (org already requires org-macs) and we wouldn't need to manage them in multiple places.

Wetlize (approved):
Thanks!

Issue #206, opened 2020-02-29
database not updated automatically


brabalan:
### Description
Since switching to the new sql database, it does not seem to update automatically

#### Steps to Reproduce
- create a new file `foo` using `org-roam-insert`
- navigate to another roam file
- try to do `org-room-insert` with `foo`
- `foo` is not in the list of files proposed
- running `org-roam-build-database` solves the issue

### Versions
- Emacs (`C-h v emacs-version`): GNU Emacs 26.3 
- Org-roam commit: I don't know, as of today

jethrokuan:
did you restart emacs after the switch?

brabalan:
I did restart emacs. It does not seem to happen to every file, however, but I don't know how to trigger it.

jethrokuan:
> foo is not in the list of files proposed

I can replicate this, it's because even though the file was created, the database wasn't updated for the created file.

brabalan:
Thanks!

zylinski:
I am experiencing an issue as described in the parent post. MELPA install. I do not have the command `org-roam-build-database`. I can create a file using `org-roam-find-file`. But, I need to restart my emacs to see it.

org-roam version 1.2.3

d12frosted:
@zylinski try `org-roam-db-build-cache`

d12frosted:
@zylinski btw, I advice to checkout [Getting Started][1] section of the documentation 😸 It describes how to build the cache and describes when the cache is automatically built (and how to configure that). 

[1]: https://www.orgroam.com/manual.html#Getting-Started

pajot:
I am experiencing this same problem. The cache updates only if I run org-roam-db-build-cache manually. I do not see any activity otherwise, not even an error. (I have tried running toggle-debug-on-error.)

albreche:
> I am experiencing this same problem. The cache updates only if I run org-roam-db-build-cache manually. I do not see any activity otherwise, not even an error. (I have tried running toggle-debug-on-error.)

same problem here,  emacs-29.1, org-roam 2.2.2, org-roam-db-on-save set to t and db updated at each emacs startup

kascesar:
Same here

PR #1804, opened 2021-08-24
(feat!)capture: unify template variables

Wetlize:
See the first commit message of this PR about the change along with [this](https://org-roam.discourse.group/t/org-roam-bibtex-for-org-roam-v2/1574/83?u=mshevchuk) Discourse discussion. 

Here's a very simplified example of how capture templates will look like, if and when this will be merged:
```diff
- (setq org-roam-dailies-capture-templates
-       '(("d" "default" entry "* %?"
-          :if-new (file+head "%<%Y-%m-%d>.org"
-                             "#+title: %<%Y-%m-%d>\n")))

-       org-roam-capture-ref-templates
-       '(("r" "ref" plain "%?"
-          :if-new (file+head "${slug}.org"
-                             "#+title: ${title}")
-          :unnarrowed t)))
+ (setq org-roam-capture-templates
+      '(...
+        ("d" "default" entry "* %?" :kind daily
+         :if-new (file+head "%<%Y-%m-%d>.org"
+                            "#+title: %<%Y-%m-%d>\n"))
+         ...
+        ("r" "ref" plain "%?" :kind protocol
+         :if-new (file+head "${slug}.org"
+                            "#+title: ${title}")
+         :unnarrowed t)))
```

There's no easy way ATM to dynamically process all the available capture templates to selectively extract from them the declared information, especially from the Org-roam itself, where you can't be aware about existence of all of the extensions.  With this PR doing so should be a piece of cake.

Additionally, the users and extensions now will be able to specify contexts for their own needs, to narrow down a subset of capture templates for a specific purpose, without inventing ways to hook such things to Org-roam. For example, with this PR, #1354 can be solved as such:

```emacs-lisp
        '(...
          ;; dailies
          ("d" "daily" entry "* %?" :kind daily :action any
           :if-new (file+head "%<%Y-%m-%d>.org" "#+title: %<%Y-%m-%d>\n"))
          ("x" "other daily" entry "* Something else %?" :kind daily
           :if-new (file+head "%<%Y-%m-%d>.org" "#+title: %<%Y-%m-%d>\nSomething else"))))
```
So if the user would run any of the `org-roam-dailies-goto-*` commands, the selection won't automatically popup, unless there's more than 1 `:kind daily :action goto` dedicated templates.

###### Things to get done before merging this
- [ ] Convert existing user's templates to the new format invisibly to the user.
  +  The change from this PR is breaking, but it can be soften. It won't handle all the edge cases, but will give the user time to readjust their own templates when they will have the time to do so.
- [ ] Reflect changes from this PR in the documentation and the changelog.
- [ ] Better think through the fallback behavior, e.g. what if there's no capture template for a given context? Like if there's no specified `:action goto` template, it should probably fallback to `:action capture` templates. 

Overall, it's fully working, but the code is still WIP and more like an MVP, so there will be some changes and refactors.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -38,6 +38,33 @@
 (defvar org-end-time-was-given)
 
 ;;; Options
+(defcustom org-roam-capture-load-default-templates t
+  "Whether to include default Org-roam capture templates during the loading.
+The value will also affect default templates provided by
+Org-roam's extensions. It can be:
+
+  t
+    Include all the default capture templates provided by
+    Org-roam and its extensions.
+
+  nil
+    Don't include the default capture templates provided by
+    Org-roam and its extensions.
+
+  a list of symbols
+    Each symbol in the list corresponds to a `provide'd FEATURE,
+    for which the default capture templates will be automatically
+    included (if any). The list can start with a special value
+    `:not', in which case the logic will be inverted to
+    exclusion, e.g. (:not org-roam-dailies) won't include the
+    default templates provided by `org-roam-dailies', but will
+    include for other features."
+  :type '(choice
+          (const :tag "Yes" t)
+          (const :tag "No" nil)
+          (repeat :tag "List features" symbol))
+  :group 'org-roam)
+
 (defcustom org-roam-capture-templates
   '(("d" "default" plain "%?"
      :if-new (file+head "%<%Y%m%d%H%M%S>-${slug}.org"
@@ -378,8 +405,17 @@ This variable is populated dynamically, and is only non-nil
 during the Org-roam capture process.")
 
 (defconst org-roam-capture--template-keywords (list :if-new :id :link-description :call-location
-                                                    :region)
+                                                    :region :kind :action)
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")
+
+(defun org-roam-capture--load-templates-p (feature)
+  "Return t if capture templates for FEATURE are allowed to be loaded.
+See `org-roam-capture-load-default-templates' for more details."
+  (let ((user-value org-roam-capture-load-default-templates))
+    (pcase user-value
+      ((pred consp) (org-roam--valid-option-p feature user-value))
+      ((pred null) nil)
+      ('t t))))
 
 ;;; Main entry point
 ;;;###autoload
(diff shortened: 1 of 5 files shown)


Wetlize:
@jethrokuan I need to know whether you will greenlight changes from this PR before I can continue further.

cc @myshevchuk, I think this sort of functionality you meant in #206?

org-roam-capture.el
```diff
@@ -390,11 +426,18 @@ INFO is a plist for filling up Org-roam's capture templates.
 NODE is an `org-roam-node' construct containing information about the node.
 PROPS is a plist containing additional Org-roam properties for each template.
 TEMPLATES is a list of org-roam templates."
-  (let* ((props (plist-put props :call-location (point-marker)))
+  (let* ((props (thread-first props
+                  (plist-put :call-location (point-marker))
+                  (plist-put :action  (or (plist-get props :action) (if goto 'goto 'capture)))
```
jethrokuan, line 431: what about templates that are for both goto and capture? I think there should be an `'any` keyword support here as well.

org-roam-capture.el
```diff
@@ -390,11 +426,18 @@ INFO is a plist for filling up Org-roam's capture templates.
...
+                  (plist-put :kind (or (plist-get props :kind) 'normal))))
```
jethrokuan, line 432: /s/kind/context
/s/normal/default

jethrokuan (commented):
I think the overall approach looks good, but once again it's not actually painpoint i'm personally facing, so maybe others can chime in as well (extension authors/users)

org-roam-capture.el
Wetlize: It used to be named "context" (before I've pushed it upstream), but there's couple of problems with this name:
1. If you think about it, both `:if-new`, `:id`, `:action` (probably even more) already sort of specifying a more global context for capture, so the word is too generic. So if you would ask someone in an issue "In which capture *context* this template doesn't work?" it would be a bit of ambiguous. 
2. "Context" already used in [org-capture](https://code.orgmode.org/bzg/org-mode/src/3664447eb2417b38f0053674f248d2b0f1592ff0/lisp/org-capture.el#L553) and implies a different thing.

I've also tried "type", but again, it already implies different thing in both, org-roam-capture and org-capture. Since "kind" is not occupied by the both, I've rolled with it. Do you think I should I still s/kind/context?

org-roam-capture.el
Wetlize: You mean declared in `org-roam-capture-templates` as `:action (capture goto)` or `:action any`? Or you mean the fallback behavior if it's not declared in the template? In the first case it should just work, while in the latter case I'm planning to fix it (was mentioned in the todo list).

org-roam-capture.el
jethrokuan: I do think `:kind` sounds a bit strange, and `:context` is much more appropriate (in the programming sense), but overloading terms with org-capture is definitely a bad idea. I was thinking something more descriptive would make sense like `:valid-if`, and then it could support more than just what `:kind` would (e.g. support arbitrary predicates).

PR #1806, opened 2021-08-25
(feat) core: support new org-mode citations

jethrokuan:

Support caching the new Org 9.5 citations.

Because citations now has first-class support, and are treated
differently from links, they are now cached in their own table.

Org-ref citations, instead of being stored in the links table, are now
stored in the citations table instead.

To use a citation as a ROAM_REF, use the `@citeKey` syntax.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -652,28 +652,9 @@ Since version 9.5, Org has first-class support for citations. Org-roam supports
 the caching of both these in-built citations (of form ~[cite:@key]~) and [[https://github.com/jkitchin/org-ref][org-ref]]
 citations (of form cite:key).
 
-** Setup
-
-*org-ref*
-
-To ensure that org-ref citations are cached, ~org-ref~ must be loaded prior to
-calls of ~org-roam-db-sync~. This can be achieved by placing ~(require
-'org-ref)~ before ~org-roam~ is loaded, for example in the ~:init~ block of the
-Org-roam use-package declaration:
-
-#+begin_src emacs-lisp
-(use-package org-roam
-  :init
-  (require 'org-ref)
-  :config
-  ;;...
-  )
-#+end_src
-
-*org-cite*
-
-Org-cite should be loaded by default, and no additional setup should be
-required.
+Org-roam attempts to load both the ~org-ref~ and ~org-cite~ package when
+indexing files, so no further setup from the user is required for citation
+support.
 
 ** Using the Cached Information
 
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -646,6 +646,68 @@ Org-roam also provides some functions to add or remove refs.
 
   Remove a ref from the node at point.
 
+* Citations
+
+Since version 9.5, Org has first-class support for citations. Org-roam supports
+the caching of both these in-built citations (of form ~[cite:@key]~) and [[https://github.com/jkitchin/org-ref][org-ref]]
+citations (of form cite:key).
+
+** Setup
+
+*org-ref*
+
+To ensure that org-ref citations are cached, ~org-ref~ must be loaded prior to
+calls of ~org-roam-db-sync~. This can be achieved by placing ~(require
+'org-ref)~ before ~org-roam~ is loaded, for example in the ~:init~ block of the
+Org-roam use-package declaration:
+
+#+begin_src emacs-lisp
+(use-package org-roam
+  :init
+  (require 'org-ref)
+  :config
+  ;;...
+  )
+#+end_src
+
+*org-cite*
+
+Org-cite should be loaded by default, and no additional setup should be
+required.
+
+** Using the Cached Information
+
+It is common to use take reference notes for academic papers. To designate the
+node to be the canonical node for the academic paper, we can use its unique
+citation key:
+
+#+begin_src org
+,* Probabilistic Robotics
+:PROPERTIES:
+:ID:       51b7b82c-bbb4-4822-875a-ed548cffda10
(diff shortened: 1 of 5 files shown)


doc/org-roam.org
```diff
@@ -646,6 +646,68 @@ Org-roam also provides some functions to add or remove refs.
...
+
+It is common to use take reference notes for academic papers. To designate the
+node to be the canonical node for the academic paper, we can use its unique
+citation key:
+
+#+begin_src org
+,* Probabilistic Robotics
+:PROPERTIES:
+:ID:       51b7b82c-bbb4-4822-875a-ed548cffda10
+:ROAM_REFS: @thrun2005probabilistic
```
Wetlize, line 688: I don't really have anything meaningful to add on this, but AFAIU this is not something that would be actionable (interactively) by org? Since this is a custom syntax. 

IIRC org-ref's "cite:" were actionable, just like URL links. I'm not sure whether this even important, but just wanted to highlight this, so someone else who's more knowledgeable on citations would evaluate this.

org-roam-db.el
```diff
@@ -79,7 +79,7 @@ slow."
   :group 'org-roam)
 
 ;;; Variables
-(defconst org-roam-db-version 16)
+(defconst org-roam-db-version 17)
```
Wetlize, line 82: Possibly unrelated to this PR, but I immediately noticed that building the database from scratch through automatic upgrade feels orders of magnitude more slower than building the database from scratch manually (i.e. through force sync). Not sure what's going on there.

Wetlize (commented):
I'm afraid that I'm not going to be a large help here, since citations are simply not present in my current workflow (though I might incorporate them later) and I'm completely unfamiliar with `org-cite` (except a very brief familiarization), and not so much with `org-ref`, which I've stopped using a while ago. In any case, I'll still try to help you.

However, I would try invite @bdarcus to review this. AFAIK he was actively participating through the design process of `org-cite` and also maintains bibtex/citation related packages, while also at the same time uses(?) org-roam. It also probably would make sense to invite @myshevchuk and @ThomasFKJorna, since both of them maintain packages that actively interface with citations.

Other than that, I've briefly tested this PR all over the place and it works just fine. So it LGTM.

doc/org-roam.org
```diff
@@ -646,6 +646,68 @@ Org-roam also provides some functions to add or remove refs.
...
+
+Since version 9.5, Org has first-class support for citations. Org-roam supports
+the caching of both these in-built citations (of form ~[cite:@key]~) and [[https://github.com/jkitchin/org-ref][org-ref]]
+citations (of form cite:key).
+
+** Setup
+
+*org-ref*
+
+To ensure that org-ref citations are cached, ~org-ref~ must be loaded prior to
```
bdarcus, line 659:
```suggestion
If you wish to use ~org-ref~ citations, to ensure they are cached, ~org-ref~ must be loaded prior to
```

doc/org-roam.org
bdarcus: I raised this question on the org-mode list recently, though have not yet had any responses.

https://lists.gnu.org/archive/html/emacs-orgmode/2021-08/msg00317.html

My tentative suggestion there was to allow citations in the property drawers, but I am still not sure.

As @Wetlize guesses, references of this form won't be treated as citations by org, and so they don't be actionable.

Thoughts?

doc/org-roam.texi
```diff
@@ -995,6 +1002,76 @@ ref to add.
...
+* Setup::
+* Using the Cached Information::
+@end menu
+
+@node Setup
+@section Setup
+
+@strong{org-ref}
+
+To ensure that org-ref citations are cached, @code{org-ref} must be loaded prior to
```
bdarcus, line 1022: If you accept the suggestion above, would want this consistent with it.

org-roam-db.el
jethrokuan: > Possibly unrelated to this PR, but I immediately noticed that building the database from scratch through automatic upgrade feels orders of magnitude more slower than building the database from scratch manually (i.e. through the force sync). Not sure what's going on there, but it's something reproducible.

Worth looking into separately, I guess.

doc/org-roam.org
```diff
@@ -652,28 +652,9 @@ Since version 9.5, Org has first-class support for citations. Org-roam supports
...
-  :config
-  ;;...
-  )
-#+end_src
-
-*org-cite*
-
-Org-cite should be loaded by default, and no additional setup should be
-required.
+Org-roam attempts to load both the ~org-ref~ and ~org-cite~ package when
```
bdarcus, line 655: What does this mean? How will you know "when needed"?

That if you have them installed, it will load them, but otherwise do nothing?

I was wondering about a user who has been using org-ref, but is migrating to org-ref-cite. For that user, both will be loaded until they uninstall org-ref?

Wetlize: @bdarcus It means that each time when org-roam will need to parse files it will try to automatically load both of the packages. If it detects that the package is loaded, it will parse cite links from the available packages. Otherwise, it will simply noop their parsing and do nothing. 

> I was wondering about a user who has been using org-ref, but is migrating to org-ref-cite. For that user, both will be loaded until they uninstall org-ref?

Yes. If org-roam detects that both, org-cite` and `org-ref are available, both of the citations types will be parsed and stored in the database.

doc/org-roam.org
bdarcus: Just to update, org-cite dev says it's a "terrible idea" to allow cites in property drawers :-)

https://lists.gnu.org/archive/html/emacs-orgmode/2021-09/msg00130.html

Issue #1814, opened 2021-08-27
(invalid-function (target (org-roam-capture--get :if-new)))


brabalan:
### Description
After updating org-roam, my dailies templates no longer work. I renamed `if-new` to `target` in the templates, to no avail.

#### Steps to Reproduce
- call `(org-roam-dailies-goto-today)`
- choose a template

#### Backtrace
```
Debugger entered--Lisp error: (invalid-function (target (org-roam-capture--get :if-new)))
  (target (org-roam-capture--get :if-new))()
  org-roam-capture--get-if-new-target-a((closure (org-end-time-was-given t) nil "Get the current capture :target for the capture te..." (or (org-roam-capture--get :target) (user-error "Template needs to specify `:target'"))))
  apply(org-roam-capture--get-if-new-target-a (closure (org-end-time-was-given t) nil "Get the current capture :target for the capture te..." (or (org-roam-capture--get :target) (user-error "Template needs to specify `:target'"))) nil)
  org-roam-capture--get-target()
  (let* ((val (org-roam-capture--get-target))) (if (consp val) (let* ((x103 (car val))) (cond ((eq x103 'file) (let* ((x104 ...)) (if (consp x104) (let* ... ...) nil))) ((eq x103 'file+olp) (let* ((x107 ...)) (if (consp x107) (let* ... ...) nil))) ((eq x103 'file+head) (let* ((x112 ...)) (if (consp x112) (let* ... ...) nil))) ((eq x103 'file+head+olp) (let* ((x117 ...)) (if (consp x117) (let* ... ...) nil))) ((eq x103 'file+datetree) (let* ((x124 ...)) (if (consp x124) (let* ... ...) nil))) ((eq x103 'node) (let* ((x129 ...)) (if (consp x129) (let* ... ...) nil))) (t nil))) nil))
[...]

PR #1818, opened 2021-08-28
(fix)capture: refactor `org-roam-capture--get-if-new-target-a`

Wetlize:
It looks like Emacs' byte compiler is very naive and cannot properly
sometimes compile closures, which causes error like in #1814, at least
on Emacsen <28. One of the possible solutions is to write the code using
less abstractions in between, so the byte compiler would do the right
thing when arranging instructions.

Possibly addresses #1814.

### DIFF
--- a/org-roam-compat.el
+++ b/org-roam-compat.el
@@ -153,17 +153,18 @@ nodes." org-id-locations-file)
                            (expand-file-name ".orgids" (file-truename org-roam-directory)))
                      (apply fn args)))))))
 
+;;;; Deprecated :if-new capture template keyword
 (with-eval-after-load 'org-roam-capture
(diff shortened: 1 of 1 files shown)


nobiot:
@Wetlize 
It took me some time to realize that `:if-new` keyword is now deprecated (it still works on my end).
Where am I supposed to get the warning? 

I don't seem to get the warning message and missed this change until now (I usually have the latest almost immediately after it is published to MELPA).

Wetlize:
@nobiot It's soft deprecated for now and probably won't be removed for a while. The warning is just not enabled for the time being. See #1809.

nobiot:
@Wetlize Thank you. Got it!

PR #1821, opened 2021-08-29
org-roam-node-visit: fix typo

listx:
This function ignored the `force' optional argument because it always
quoted it, resulting in a non-nil value every time.

This makes the suggested code snippet in
https://github.com/org-roam/org-roam/issues/597#issuecomment-907743125
work as expected (not resetting point to the beginning of the "Index"
buffer on every invocation).


Wetlize:
Good catch. Thanks!

PR #1849, opened 2021-09-11
(fix)db: Require 'oc to get org-cite functionality.

roshanshariff:
The org-cite library is called [oc.el](https://code.orgmode.org/bzg/org-mode/src/master/lisp/oc.el), not `org-cite.el`.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -512,7 +512,7 @@ If the file exists, update the cache with information."
             (org-roam-db-map-links
              info
              (list #'org-roam-db-insert-link))
-            (when (require 'org-cite nil 'noerror)
+            (when (require 'oc nil 'noerror)
               (org-roam-db-map-citations
                info
                (list #'org-roam-db-insert-citation)))))))))


chuxubank:
This should fix #1851.

carlosperezm:
I confirm this solves the issue. Thanks @roshanshariff!

jethrokuan:
My local copy of Org-mode (Org mode version 9.5 (9.5-b83ae59 @ /home/jethro/.emacs.d/.local/straight/build-28.0.50/org/)) has both `provides`:

image

but I just checked Org's git master, and it's now only `oc`:
https://git.savannah.gnu.org/cgit/emacs/org-mode.git/tree/lisp/oc.el#n1607

so I'm not sure if the fix is correct... Would be good to collect more data on Org version and which `provides` are available.

roshanshariff:
I looked at the git history and these are the relevent commits, and whether they `(provide 'oc)` or `'org-cite`:
| Commit | `'oc` | `'org-cite` | Notes |
| ------ | ----- | ----------- | ----- |
| bzg/org-mode@e844a6fd41ebede99b8b6544e6267afeebe02e8c | :heavy_check_mark: | :heavy_check_mark: | Add `oc.el` |
| bzg/org-mode@ad006e29305b4dd7af340631806c83d4d9ca2ab7 | :x: | :heavy_check_mark:  | Removes `(provide 'oc)` |
| bzg/org-mode@8df3e25db8a629bbffa883edadb13a55239ae4b7 | :heavy_check_mark: | :heavy_check_mark: | Revert previous change |
| bzg/org-mode@490eba3a423b4031cfc66915fe1e77f6b8391faf | :heavy_check_mark: | :x: | Current status |

The file name has always been `oc.el`, and it has always provided the `'oc` feature except between the second and third commits.  These two commits were made consecutively on 1 Aug, 2021 with a gap of 22 minutes. So the fix in this PR is correct except for that one brief instant.

_If_ the user happens to have that commit, then `oc.el` will be reloaded repeatedly since it doesn't `(provide 'oc)`. I don't think it's worth working around this very minor and transient bug. But if you really want, the code can be changed to
```elisp
(when (or (featurep 'org-cite) (require 'oc nil 'noerror)) ...
```

PR #1850, opened 2021-09-13
(refactor)completions: simplify org-roam-node-read--format-entry

real-or-random:
This commit simplifies the code in org-roam-node-read--format-entry,
saving the entire processing of display properties when no
field-length is specified and thus no truncation needs to happen.
Since in this case, we don't truncate, the bug fixed in #1759 won't
reappear.

This has the beneficial side-effect that existing display properties
in the field-value, e.g., display properties that replace text by
images/icon, are not overridden by us when no field-length is
specified. Note that images anyway don't need truncation of the text
they replace.

This commit also fixes a wrong word in the comment.


real-or-random:
force-pushed to fix the lint issue

jethrokuan (approved):

jethrokuan:
Thanks!

PR #1860, opened 2021-09-19
org-roam-db--insert-file: use file contents for hash

targit:
Use file contents for hash instead of buffer contents.
In case of encrypted files both ways lead to different results.
Otherwise encrypted files will need to be indexed each time
the cache will be updated/synced.


jethrokuan:
This was fixed in a commit earlier so this shouldn't be an issue anymore: https://github.com/org-roam/org-roam/pull/1725

https://github.com/org-roam/org-roam/blob/91d1793aea798d1fdc9da077ebaed67cb4cebc5f/org-roam-db.el#L476

targit:
Yes, that seems to work nicely.

Issue #1859, opened 2021-09-17
Links within properties drawers of a node are not indexed


rdrg109:
### Table of Contents

1.  [Description](#orgce6e532)
    1.  [Prologue](#org5ff381b)
    2.  [Steps to reproduce (mwe)](#org6aae6f5)
    3.  [Backtrace](#orga8be334)
    4.  [Expected results](#org31fe73c)
    5.  [Actual results](#org68d2883)
    6.  [Related issues](#org68d2884)
2.  [Environment](#org95186ea)

<a id="orgce6e532"></a>

### Description

<a id="org5ff381b"></a>

#### Prologue

I usually reference nodes in the properties drawer of another node.   To understand why I do this, let me explain one of my uses of Org Mode: I use Org Mode for storing bookmarks. Each bookmark is stored as a headline (see below)

```org
* Org Roam
:PROPERTIES:
:LINK: https://www.orgroam.com/
:END:
```

I can then use the properties of the headline to store metadata of that bookmark

```org
* Org Roam
:PROPERTIES:
:LINK: https://www.orgroam.com/
:TOPIC+: emacs
:INSTANCE_OF: official site
:LANGUAGE: english
:END:
```

Two days ago, I had this idea of treating each entity, that is used as metadata as an Org Roam node, as a node so that the graph feature can also show relevant information on my bookmarks. So, a node would look like this

```org
* Org Roam
:PROPERTIES:
:LINK: https://www.orgroam.com/
:TOPIC+: [[id:96db2d4a-51c9-48fc-9f45-ac8d3331f314][emacs]]
:INSTANCE_OF: [[id:3b9fa6fa-67ea-4f96-b35d-4971e3a5999c][official site]]
:LANGUAGE: [[id:dc42cf9e-d7b6-4b5c-9463-abecececc89a][english]]
:ID: e395b878-a574-4c74-b89d-7c14522b2e1f
:END:
```

However, I noticed that the links located in properties drawers were not being indexed.

<a id="org6aae6f5"></a>

#### Steps to reproduce (mwe)

Instead of explaining the steps to reproduce, I'm going to present a minimal working example (mwe).

Let's suppose we have 4 nodes. The files containing these nodes are shown below.

```sh
echo '++++++++++++++++++++++++++++++ foo1'
cat ~/org-roam/20210917175708-foo1.org
echo '++++++++++++++++++++++++++++++ foo2'
cat ~/org-roam/20210917175721-foo2.org
[...]

Issue #1464, opened 2021-04-09
Links from org properties are not indexed


lyndondrake:
### Description

While `org-roam` seems to be working, e.g. `org-roam-find-file` correctly shows all my notes including their tags, the `org-roam` buffer shows "No backlinks" even when backlinks do exist.

I've got rid of all my `.elc` files and tried deleting the `org-roam.db` file too. It gets rebuilt fine, but still no backlinks.

### Environment

- Emacs: GNU Emacs 27.2 (build 1, x86_64-apple-darwin20.3.0, Carbon Version 164 AppKit 2022.3)
 of 2021-03-28
- Framework: Doom
- Org: Org mode version 9.5 (nil @ /Users/lyndon/.emacs.d.doom/.local/straight/build-27.2/org-mode/)
- Org-roam: 1.2.3

lyndondrake:
Apologies - I got this wrong - it turns out that links from within an org properties bin are not indexed:
```
** Genesis 23
:PROPERTIES:
:Note: [[file:20210302093117.org][Genesis 23]]
:END:
```

doesn't generate a backlink in the Genesis 23 note. I'm assuming that's by design… I wondered if there's some way then to either change the behaviour, or to include a link in a form that's hidden from export (which is why I've dumped it in a properties bin).

jethrokuan:
This is because the Org parser itself doesn't detect items within property drawers as links. This could be changed to just searching forward for the link regex, but then it would be an inconsistency with the Org syntax parsing itself which can lead to its own surprises.

lyndondrake:
Ah, that makes sense then.

lyndondrake:
My use case is that in my thesis document, I want to add references to relevant notes. I don't want those links to show up in my formatted document though, hence using a property drawer item.

Can you suggest another option for having (effectively) a hidden link?

Or would there be a way for me to add a save-hook that looks for a link in a particular property — in my case, the `:Note:` property?

d12frosted:
@lyndondrake 

This is a bummer for me as well. While I was migrating some notes to
`org-roam` style of links I discovered that properties drawer does not
support links. They are displayed as links, but semantically they are
not! Meaning that open link may not work as expected (for example,
=id= links), elements API doesn't parse them as links (meaning that so
many things break, including org-roam backlinks). Though I can't find
a relevant link to documentation anymore.

In my case I needed to manipulate properties that link to other notes
(e.g. read and write) and using properties API would be convenient
[...]

PR #1868, opened 2021-09-22
fix(org-roam-db-map-links): Consider link in properties drawers

rdrg109:
###### Description

The current implementation uses `org-element-map` to iterate through all links in the current buffer. This function doesn't consider links within properties drawer. This implementation does consider such links.

###### Additional information

This pull request solves #1464 and #1859 .

A minimal working example that proves that `org-element-map` doesn't consider links in properties drawer can be found below.

```lisp
(let ((lines '("* John Doe"
               ":PROPERTIES:"
               ":FOO: [[1]]"
               ":END:"
               "[[2]]"
               "[[3]]")))
  (with-temp-buffer
    (org-mode)
    (insert (string-join lines "\n"))
    (org-element-map (org-element-parse-buffer)
        'link
      (lambda (link) (org-element-property :raw-link link)))))
```

```
("2" "3")
```

Another solution would be to make a patch for `org-element-map` so that this get fixed in the Org Mode codebase, but I think this would take much more time. For this reason, I created this pull request as a temporary workaround.

Some people, as me, store links in properties drawers. A use case of doing is shown below

```
* John Doe
:PROPERTIES:
:ID: 1daaebf6-f12b-449b-97d7-f56b57eee605
:KNOWS_OF+: [[id:303e7f25-21b1-4b40-81ab-8e4190fea512][GNU Emacs]]
:KNOWS_OF+: [[id:58893012-6137-4a9a-88c6-87098affe775][Guix]]
:LANGUAGE_NATIVE+: [[id:e4c3ad8c-fe23-4560-916c-52481b4ac4d7][Spanish]]
:LANGUAGE_FLUENT+: [[id:2333b842-0b0c-44f3-be19-eec8fbf4bdff][Russian]]
:END:
```

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -298,17 +298,18 @@ If UPDATE-P is non-nil, first remove the file in the database."
       ;; `re-search-forward' let the cursor one character after the link, we need to go backward one char to
       ;; make the point be on the link.
       (backward-char)
-      (let (element link bounds)
-        (setq element (org-element-context))
+      (let* ((element (org-element-context))
+             (type (org-element-type element))
+             link bounds)
         (cond
          ;; Links correctly recognized by Org Mode
-         ((eq (org-element-type element) 'link)
+         ((eq type 'link)
           (setq link element))
          ;; Links in property drawers. Recall that, as for Org Mode
          ;; v9.4.4, the org-element-type of links within properties
          ;; drawers is node-property.
-         ((and (or (eq (org-element-type element) 'node-property)
-                   (eq (org-element-type element) 'keyword))
+         ((and (or (eq type 'node-property)
+                   (eq type 'keyword))
                (setq bounds (org-in-regexp org-link-any-re))
                (setq link (buffer-substring-no-properties
                            (car bounds)
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -291,14 +291,35 @@ If UPDATE-P is non-nil, first remove the file in the database."
          (dolist (fn fns)
            (funcall fn)))))))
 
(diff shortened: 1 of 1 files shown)


org-roam-db.el
```diff
@@ -291,14 +291,35 @@ If UPDATE-P is non-nil, first remove the file in the database."
...
+         ;; Links in property drawers. Recall that, as for Org Mode
+         ;; v9.4.4, the org-element-type of links within properties
+         ;; drawers is node-property.
+         ((and (eq (org-element-type element) 'node-property)
+               (setq bounds (org-in-regexp org-link-any-re))
+               (setq link (buffer-substring-no-properties
+                           (car bounds)
+                           (cdr bounds))))
+          (with-temp-buffer
+            (org-mode)
```
jethrokuan, line 316:
```suggestion
            (delay-mode-hooks (org-mode))
```

Should be faster if people added hooks to `org-mode`.

rdrg109:
@jethrokuan I've also noticed that links in `#+TITLE` lines are not considered by neither `org-element-map` nor this approach. The latest commit solves this issue.

```lisp
(let ((lines '("#+TITLE: A title with [[https://link1.com][a]] link"
               "* My section"
               "[[https://link2.com]]")))
  (with-temp-buffer
    (org-mode)
    (insert (string-join lines "\n"))
    (org-element-map (org-element-parse-buffer)
        'link
      (lambda (link) (org-element-property :raw-link link)))))
```

```
("https://link2.com")
```

PRESFIL:
Backlinks of links inside PROPERTIES drawer do not refer to the current header but to the parent header within the backlinks buffer.

PR #1876, opened 2021-09-29
optimize(node): org-roam-node-read: reduce checks on filter-fn, gain O(n) -> O(1)

qzdl:
in org-roam-node-read, there is no need to check filter-fn for each node, so move the branch up and just check it once. 

the same logic could be used below, for `sort-fn`, instead of deviating with `setq`

found while debugging, easy fix

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -441,8 +441,10 @@ SORT-FN is a function to sort nodes. See `org-roam-node-read-sort-by-file-mtime'
 for an example sort function.
 If REQUIRE-MATCH, the minibuffer prompt will require a match."
   (let* ((nodes (org-roam-node-read--completions))
-         (nodes (cl-remove-if-not (lambda (n)
-                                    (if filter-fn (funcall filter-fn (cdr n)) t)) nodes))
+         (nodes (if filter-fn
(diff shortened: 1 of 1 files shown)


dlobraico:
I think this change breaks `org-roam-node-read` in the case where `filter-fn`is non-nil:

```
Debugger entered--Lisp error: (wrong-number-of-arguments #<subr cl-remove-if-not> 1)
  cl-remove-if-not(#f(compiled-function (n) #<bytecode -0x14277ba6c38e6572>))
  org-roam-node-read(nil (closure (t) (node) (let ((tags (progn (or (progn ...) (signal ... ...)) (aref node 15)))) (member "foo" tags))))
```

PRESFIL:
It seems like a couple of parentheses are mixed up in places...

It seems that it should be like this:

```diff
-         (nodes (if filter-fn
-                    (cl-remove-if-not
-                    (lambda (n) (funcall filter-fn (cdr n)) t))
-                  nodes))
+         (nodes (if filter-fn
+                    (cl-remove-if-not
+                    (lambda (n) (funcall filter-fn (cdr n)))
+                  nodes) t))
```

PR #1885, opened 2021-10-02
Put org-roam-db-autosync-mode into :init section

chuchana:
Put `(org-roam-db-autosync-mode)` into the `:init` section. 

I could not use `org-roam-node-visit` in my init file when it was in the `:config` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Here's a very basic sample for configuration of `org-roam` using `use-package`:
 ```emacs-lisp
 (use-package org-roam
   :ensure t
+  :init
+  (org-roam-db-autosync-mode)
   :custom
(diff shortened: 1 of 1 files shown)


jethrokuan:
`org-roam-node-visit` is not autoloaded as it's an internal function. Do you mean to use `org-roam-node-find`?

chuchana:
I want to automatically open some nodes from my init file.  At first I tried `(org-roam-node-find "(Node Title)")`, but that just got me a prompt asking for a node to open.  So I started digging through the code until I got it to work like this:
```
(dolist (node (list "(Node Id)"))
  (org-roam-node-visit (org-roam-node-from-id node)))
```

Can I achieve the same result more cleanly?

chuchana:
Does this usage clash roam's philosophy?

PR #1889, opened 2021-10-07
Remove f dependency

phikal:
The f library is being used, even though all the functionality you need is already built in to Emacs itself. This patch replaces f.el with the built-in analogues, removing an unnecessary dependency.

If interested, this could also serve as the basis to add org-roam to [NonGNU ELPA](https://elpa.nongnu.org).

### DIFF
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -37,7 +37,6 @@
 ;; scratch notes and whatever else you can came up with.
 ;;
 ;;; Code:
-(require 'f)
 (require 'dash)
 (require 'org-roam)
 
@@ -266,7 +265,7 @@ If FILE is not specified, use the current buffer's file-path."
     (save-match-data
       (and
        (org-roam-file-p path)
-       (f-descendant-of-p path directory)))))
+       (file-in-directory-p path directory)))))
 
 ;;;###autoload
 (defun org-roam-dailies-find-directory ()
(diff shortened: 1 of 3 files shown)


ThibautVerron:
I don't know if it introduces bugs in the case of org-roam, but the replacements are not equivalent to the `f` functions.

For the subdirectory testing, the `f-` version only compares the paths as strings, the `files.el` version does a lot more, including checking whether the directory exists.

Additionally, the `f` function considers that a directory is not a descendant of itself, the `files` function considers the opposite.

The equality testings are also not the same: `f` just compares the canonical paths and `files` compares all the file attributes. (I don't see how two files could have the same paths and different attributes, so it might be functionally equivalent.)

If the files do not exist, the return value of the `files.el` function is undefined (nil in my experiments), whereas `f` explicitly returns nil.

ThibautVerron:
`s-word-wrap` does not, as far as I know, have any convenient replacement outside of `s.el`. It's called only once, but I think that it's more readable like that than by expanding the `with-temp-buffer -> wrap -> buffer-substring` loop instead.

`s-join` is strictly equivalent to `string-join` and the packages already require `subr-x` so I guess there would be no harm in replacing that one.

skangas:
If anyone is interested, note that I've proposed another way to get rid of this dependency. See PR #2025.

Thanks!

jethrokuan:
Superceded by #2025

PR #1895, opened 2021-10-10
Fix highlighting of formatted/truncated strings

### LINKED ISSUE #1801: org-roam-read-node doesn't let completion highlight matches
### Description

I'd classify this as a regression due to [ad3e66](https://github.com/org-roam/org-roam/commit/aafc8606bb0802c57ebaa3b54f40d0db18ad3e66). Now it seems that all of a candidate string is replaced with an equivalent display-property for that same string even when its wide enough to not need truncating. The end result is that none of the candidates can be highlighted by a completion framework.

Screenshot_20210824_032454

You can see an example of how this should look in #1775.

#### Steps to Reproduce
1. Load Emacs
2. Run `org-roam--build-cache-async`
3. Run `org-roam-find-file`
4. Enter some search term

#### Expected Results

Candidates that match the search term will be highlighted as such by the completion-framework.

#### Actual Results
They aren't.

### Environment

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/e9ae19c
- Emacs: GNU Emacs 28.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.29, cairo version 1.17.4)
 of 2021-05-24
- Framework: N/A
- Org: Org mode version 9.4.6 (9.4.6-g604bfd @ /home/mohkale/.cache/emacs/straight/build/org/)
- Org-roam: 2.0.0

### Proposed Implementation

This fixes the issue for me and brings back the ability to omit the field-width of an entry in org-roam-display-format.

```diff
diff --git a/org-roam-node.el b/org-roam-node.el
index 07db362..aaa1a46 100644
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -511,18 +511,19 @@ Uses `org-roam-node-display-template' to format the entry."
                              (- width tmpl-width))
                             ((>= (string-to-number field-width) 0)
                              (string-to-number field-width))))
-         ;; Setting the display (which would be padded out to the field length) for an
-         ;; empty string results in an empty string and misalignment for candidates that
-         ;; don't have some field. This uses the actual display string, made of spaces
-         ;; when the field-value is "" so that we actually take up space.
-         (let ((display-string (if field-width
-                                   (truncate-string-to-width field-value field-width 0 ?\s)
-                                 field-value)))
-           (if (equal field-value "")
-               display-string
-             ;; Remove properties from the full candidate string, otherwise the display
-             ;; formatting with pre-prioritized field-values gets messed up.
-             (propertize (substring-no-properties field-value) 'display display-string))))))))
+         (if field-width
+             ;; Setting the display (which would be padded out to the field length) for an
+             ;; empty string results in an empty string and misalignment for candidates that
+             ;; don't have some field. This uses the actual display string, made of spaces
+             ;; when the field-value is "" so that we actually take up space.
+             (let ((display-string (truncate-string-to-width field-value field-width 0 ?\s)))
+               (if (or (<= (length field-value) field-width)
+                       (equal field-value ""))
+                   display-string
+                 ;; Remove properties from the full candidate string, otherwise the display
+                 ;; formatting with pre-prioritized field-values gets messed up.
+                 (propertize (substring-no-properties field-value) 'display display-string)))
+           field-value))))))
 
 (defun org-roam-node-read--process-display-format (format)
   "Pre-calculate minimal widths needed by the FORMAT string."
```

This isn't an ideal implementation because if only one character is larger than field-width then all of field-width is replaced with a display-property and cannot be highlighted by the completion framework. Although I believe this is the same behaviour as before so I'd say its an improvement compared to what we have now.

minad:
Fix #1801

The approach I used here is the simplest one which should fix the common cases. Unfortunately it won't work with preformatted strings which already use invisible/display properties. However for such strings the current implementation was already broken given that truncate-string-to-width doesn't take these properties into account.

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -127,7 +127,9 @@ value (possibly nil). Adapted from `s-format'."
                  (let ((v (progn
                             (set-match-data saved-match-data)
                             (funcall replacer var default-val))))
-                   (if v (format "%s" v) (signal 'org-roam-format-resolve md)))
+                   (if v
+                       (format (apply #'propertize "%s" (text-properties-at 0 var)) v)
+                     (signal 'org-roam-format-resolve md)))
                (set-match-data replacer-match-data))))
          (if (functionp template)
              (funcall template)


jethrokuan:
@minad thanks for the many contributions, helps a lot to have someone more experienced with elisp to fix these little things, learnt quite a bit here :)

PR #1896, opened 2021-10-10
Improve completion-at-point slightly

minad:
* Allow capfs to run after org-roam-complete-everywhere. This is necessary since
  org-roam-complete-everywhere gets triggered on every word, even if no match
  has been found.
* Simpler hook setup. org-roam-complete-at-point is unnecessary,
  better add the capfs directly.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -727,16 +727,17 @@ hence \"everywhere\"."
             :exit-function
             (lambda (str _status)
               (delete-char (- (length str)))
-              (insert "[[roam:" str "]]"))))))
-
-(defun org-roam-complete-at-point ()
-  "Try get completion candidates at point using `org-roam-completion-functions'."
-  (run-hook-with-args-until-success 'org-roam-completion-functions))
+              (insert "[[roam:" str "]]"))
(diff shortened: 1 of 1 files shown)


minad:
* If you want to avoid exclusive=no one could instead check the list of titles against the completion prefix or even better query the database directly with the prefix. I should probably rather implement this, since it would be more efficient!

* Do you want to deprecate org-roam-complete-at-point? I think a direct removal of the function is fine, the function shouldn't be used anywhere. It is an implementation detail.

jethrokuan:
> If you want to avoid exclusive=no one could instead check the list of titles against the completion prefix or even better query the database directly with the prefix. I should probably rather implement this, since it would be more efficient!

I think capf respects completion-styles, which (e.g. in the case of `orderless`) might not necessarily be a prefix match, so I think we can leave it as it is.

jethrokuan:
This otherwise LGTM, I was a bit worried about the setup/teardown (disabling org-roam should disable completions etc.) but that's not handled for completions anyway, so it's okay.

minad:
@jethrokuan Thanks for looking at this. I just added an edit above, I guess we wrote about the same time :)

PR #1912, opened 2021-10-19
Remove `org-roam-with-file` dir-local indent entry

conao3:
Remove `org-roam-with-file` dir-local indent entry

This `org-rorm-with-file` indent entry is not needed.
This macro is defined by this package and already has indent of 2.

    (defmacro org-roam-with-file (file keep-buf-p &rest body)
      (declare (indent 2) (debug t))

quoted from org-roam-utils.el


jethrokuan:
They're required for CI, as you can see.

PR #1931, opened 2021-10-30
(fix)utils: org-roam-set-keyword skip over all drawers

jethrokuan:
Previously org-roam-set-keyword was unable to handle the syntax of the
:LOGBOOK: drawer. We introduce `org-roam-skip-drawers` to move point
over all drawers, allowing us to set the keyword in the right place.

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -228,6 +228,41 @@ If BOUND, scan up to BOUND bytes of the buffer."
       (when (re-search-forward re bound t)
         (buffer-substring-no-properties (match-beginning 1) (match-end 1))))))
 
+(defun org-roam-end-of-meta-data (&optional full)
+  "Like `org-end-of-meta-data', but supports file-level metadata.
+
+When optional argument FULL is t, also skip planning information,
+clocking lines and any kind of drawer.
+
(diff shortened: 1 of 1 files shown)


Wetlize:
@jethrokuan I have the following util from one of my unpublished PRs that you can use instead of `org-roam-skip-drawers`:
```emacs-lisp
(defun org-roam-end-of-meta-data ()
  "Like `org-end-of-meta-data', but will also work from outline level zero.
Return back the newly adjusted character position of point."
  (org-back-to-heading-or-point-min t)
  (if (org-at-heading-p)
      (org-end-of-meta-data t)
    (let ((el (org-element-at-point)))
      (while (and (not (eobp))
                  (memq (org-element-type el)
                        '(drawer property-drawer keyword comment comment-block horizontal-rule)))
        (goto-char (org-element-property :end el))
        (setq el (org-element-at-point)))))
  (point))
```
I'm dumping it here just in case, since I'm not sure whether it's suitable to skip all the metadata, or just drawers.

jethrokuan:
@Wetlize thanks, is there some reason you went with the Org-element approach? I don't think it quite works for this use-case, but `org-end-of-meta-data` directly and that should work.

PR #1932, opened 2021-10-30
Insert links optimization: require org-ref once

hugoArregui:

A big file with a lot of links take a while to sync, even when its structure it's pretty simple (no headings), I discover one of the reasons for this is every time we insert a link we require org-ref. This is specially annoying if you sync in save.

By creating a special insert-links function, instead of mapping links and apply a generic set of functions, I'm able to only require org-ref once. In my local env the mapping/insertion was taking around 2.32 secs and now it takes 0.83 secs.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -360,6 +360,40 @@ If UPDATE-P is non-nil, first remove the file in the database."
           (dolist (fn fns)
             (funcall fn link)))))))
 
+(defun org-roam-db-insert-links ()
+  "Run FNS over all links in the current buffer."
+  (let ((org-ref-enabled (and (require 'org-ref nil 'noerror)
+                              (boundp 'org-ref-cite-types)
+                              (fboundp 'org-ref-split-and-strip-string))))
+    (org-with-point-at 1
+      (while (re-search-forward org-link-any-re nil :no-error)
+        ;; `re-search-forward' let the cursor one character after the link, we need to go backward one char to
+        ;; make the point be on the link.
+        (backward-char)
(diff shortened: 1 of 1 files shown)


jethrokuan:
If the issue is really with `(require 'org-ref)` being slow, it's possible to move the require out of the block to the top level update function, leaving the `boundp` check there (which should be relatively fast), which would leave most of the modularized code in place.

It might also make sense to have a macro `(org-roam-with-requires @body)` which does all the requiring so the body wouldn't need to do so.

hugoArregui:
@jethrokuan I'm not quite sure to understand your suggestion, do you mean doing something like 

```lisp
(let ((org-ref-enabled ...))
  (org-roam-db-map-links
             (list (lambda (link) (org-roam-db-insert-link org-ref-enabled link)))
```

in org-roam-db-update-file?

jethrokuan:
@hugoArregui I meant moving `(require 'org-ref nil 'noerror)` from `org-roam-db-insert-link` to `org-roam-db-update-file` and removing it from `org-roam-db-insert-link`. `boundp` checks should be much and take negligible time.

hugoArregui:
@jethrokuan sorry but I don't quite follow, we still want to run `org-roam-db-insert-link` if org ref is not available, so I don't quite know how we can remove it from `org-roam-db-insert-link`. Very sorry if I'm missing something obvious.

jethrokuan:
All the slowness has been addressed appropriately I think, with #1944 and #1946, closed in favour of these two PRs.

PR #1943, opened 2021-11-05
fix(backlinks in spacemacs): fix the weird backlink display in Spacemacs

### LINKED ISSUE #1772: Weird backlink buffer display by `org-roam-buffer-toggle` in Spacemacs
### Description

When using `org-roam-buffer-toggle` to display the backlink buffer in **Spacemacs**, the style/structure of the backlinks is messed up (e.g., title not at top of the buffer)

#### Steps to Reproduce

1. Run `org-roam-find-file` and open a node
2. Run `org-roam-buffer-toggle`

#### Expected Results

image

#### Actual Results

image

### Environment
- Emacs: GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0, Carbon Version 158 AppKit 1671.6) of 2021-04-17
- Framework: Spacemacs
- Org: Org mode version 9.4.6
- Org-roam: 2.0.0

### Possible Reason
- [People mentioned it related to page-break-line-mode](https://org-roam.discourse.group/t/backlinks-title-not-at-top-of-buffer/1209). However, turning all (Global-)page-break-lines-mode in Spacemacs doesn't seem to solve the problem. 

### A Workaround
- In the org-roam-mode.el. It was found such a weird behavior only happens when `org-roam-buffer` window is displayed before `org-roam-buffer-render-contents` is called. Making the rendering happens first helps solve the problem.
- Here is the [commit](https://github.com/swang251/org-roam/commit/3cc9ad89065dbf2f0e379edf83f70173e2ff72fe) that fixes the problem. However, since it's just a workaround instead of a "solution", I am not sure if a pull request is appropriate. Further tests in other environments would be appreciated.

swang251:
Backlinks will display weirdly in Spacemacs, which could be related to page-break-lines-mode. It was
found such a weird behavior only happens when org-roam-buffer-window is displayed before
`org-roam-buffer-render-contents` is called. Tricky fix was done by making sure the rendering
happens first.

Fixes #1772

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -267,7 +267,6 @@ To toggle its display use `org-roam-buffer-toggle' command.")
        (remove-hook 'post-command-hook #'org-roam-buffer--redisplay-h)))
     ((or 'exists 'none)
      (progn
-       (display-buffer (get-buffer-create org-roam-buffer))
        (org-roam-buffer-persistent-redisplay)))))
 
 (define-inline org-roam-buffer--visibility ()
@@ -287,9 +286,13 @@ Has no effect when there's no `org-roam-node-at-point'."
     (unless (equal node org-roam-buffer-current-node)
       (setq org-roam-buffer-current-node node
             org-roam-buffer-current-directory org-roam-directory)
-      (with-current-buffer (get-buffer-create org-roam-buffer)
-        (org-roam-buffer-render-contents)
-        (add-hook 'kill-buffer-hook #'org-roam-buffer--persistent-cleanup-h nil t)))))
+      (if (get-buffer-window org-roam-buffer)
+		  (delete-window (get-buffer-window org-roam-buffer)))
+      (let ((buffer (get-buffer-create org-roam-buffer)))
+		(with-current-buffer buffer
+		  (org-roam-buffer-render-contents)
+		  (add-hook 'kill-buffer-hook #'org-roam-buffer--persistent-cleanup-h nil t))
+		(display-buffer buffer)))))
 
 (defun org-roam-buffer--persistent-cleanup-h ()
   "Clean-up global state thats dedicated for the persistent `org-roam-buffer'."


nobiot:
@swang251 
Without Spacemacs (on vanilla Emacs), I can reproduce the issue by turing on `global-page-break-lines-mode`.
I can also fix the problem while it is still on by changing customizing variable `page-break-lines-modes`. Just remove `outline-mode` (this is for default values; if Spacemacs adds anything that is related to Org, remove them, too).

If this works on your end with Spacemacs, it's probably much less drastic solution -- no code change needed.

I have also asked if others can replicate the solution in the same way in [Discourse forum](https://org-roam.discourse.group/t/need-to-refresh-every-time-i-open-the-roam-buffer-spacemacs/2204/3?u=nobiot). I am hoping for a positive outcome.

swang251:
Hi @nobiot, thanks for your reply.
I've doubled checked that the `global-page-break-lines-mode` was off and there was noting under `page-break-lines-mode`.

lazzalazza:
I think that I should  leave a comment here (I have done so on the Discourse page also): disabling global-page-break-lines-mode didn't work. The pull request worked very well initally, but as soon as I quit the org roam buffer by pressing 'q', it was impossible to reactivate it without restarting emacs. I had to revert to the previous code...

nobiot:
@lazzalazza @swang251 for the solution I am suggesting here, the global-page-break-node is ON.

lazzalazza:
@nobiot Yep: it is ON, but I can confirm that the solution unfortunately doesn't work after quitting the org-roam-buffer the first time...

nobiot:
I installed Spacemacs. I can reproduce the issue but seem to be able to fix it by doing this in the `.spacemacs` file.
I'm not used to Spacemacs (very hard to use for me), so I may be missing something important.

Screenshot from 2021-11-10 17-54-40

nobiot:
@swang251 

Just to be sure:

> there was noting under page-break-lines-mode.

it’s not `page-break-lines-mode`; it’s `page-break-lines-modes` — the word “mode” is plural — it’s a list of mode that page-break-mode should be active. Org-mode and also Org-roam are derived mode from Outline-mode….

lazzalazza:
I put this in my .spacemacs configuration as you suggested (I’m transcribing it here so people might copy that directly): 

 (fset 'evil-redirect-digit-argument 'ignore)
  (add-to-list 'evil-digit-bound-motions 'evil-org-beginning-of-line)
  (evil-define-key 'motion 'evil-org-mode (kbd "0") 'evil-org-beginning-of-line)
  (global-page-break-lines-mode -1)

Unfortunately even **this didn’t work for me**. So I went back to the pull request and managed to get the whole thing working **by not deleting line 270**:

**(display-buffer (get-buffer-create org-roam-buffer))**

in **the org-roam-buffer-toggle function**.

Actually I don't see why it was necessary to delete that line in the first place, **am I missing something**?
Thank you so much.

nobiot:
@swang251 @lazzalazza 
Sorry, one last thing

Setting this has also worked on my end:

```
  (global-page-break-lines-mode +1)
  (setq page-break-lines-modes
        '(emacs-lisp-mode lisp-mode scheme-mode compilation-mode help-mode))
```

Screenshot from 2021-11-10 18-39-42

@swang251 since you report that the previous suggestion didn't work for you... You seem to have something more going on

nobiot:
I probably also should have mentioned this -- I added Org layer with Roam support like this:

Screenshot from 2021-11-10 18-44-36

No idea if this is relevant, though.

swang251:
Hi @nobiot , thanks for all your help. Here are some updates:
- The weird behavior in spacemacs is due to both the `global-page-break-lines-mode` and the `display-line-numbers-mode`.
- The `global-page-break-lines-mode` problem should be fixed by what is suggested above.
- The `display-line-numbers-mode` should be related to something that happens in *spacemacs* (not sure what it is yet), not about the `display-line-numbers-mode` itself. This means, only if one adds `display-line-numbers-mode` to `org-mode-hook` in spacemacs, the problem occurs. It won't be a problem in the plain emacs.

**Here is a workaround by modifying the `~/.spacemacs` without touching org-roam:**
1. If you toggled `dotspacemacs-line-number`, please modify it as 
```
   dotspacemacs-line-numbers '(:t
	                       :disabled-for-modes org-mode)
```
2. add the following to `dotspacemacs/user-config()`
```
(fset 'evil-redirect-digit-argument 'ignore)
(add-to-list 'evil-digit-bound-motions 'evil-org-beginning-of-line)
(evil-define-key 'motion 'evil-org-mode (kbd "0") 'evil-org-beginning-of-line)
(global-page-break-lines-mode -1)

(add-hook 'org-mode-hook 'display-line-numbers-mode nil t)
(defun diable-display-linum()
    (remove-hook 'org-mode-hook 'display-line-numbers-mode))
(add-hook 'org-roam-mode-hook 'diable-display-linum)
``` 

I will withdraw the PR if anyone could help confirm that the workaround above works. Thanks!

nobiot:
@swang251 Great... Thank you. There is a great detail in your investigation.
I already removed my Spacemacs installation. 
To get more feedback, perhaps you could ask the Spacemacs community or on Org-roam Discourse for feedback? 
I have a feeling that there are folks still struggling to get this fixed.

lazzalazza:
@swang251 @nobiot Swang: thank you so much for your insight! Unfortunately for those – such as myself – who are accustomed to use `display-line-numbers-mode` (especially in the `visual` mode) also in org-mode... If it doesn't break anything for other users, I suggest that **the PR should be kept and not removed**.

swang251:
Hi @lazzalazza , would you help try what I proposed above. It is just a workaround without turning off the `display-line-nubmers-mode`, even in `org-mode`.

lazzalazza:
Hi @swang251 !
Sorry, I had to mention that I have tried your solution: it works rather good, but it makes it impossible to use the “visual” relative line number mode for org-mode, which I love and would keep on using if possible: it solves all the problems related to folding.

Sorry for asking this so bluntly, but: would changing those 3 lines of code in org-roam-mode.el be such a big deal?

Again, thank you soooo much!

jethrokuan:
> Sorry for asking this so bluntly, but: would changing those 3 lines of code in org-roam-mode.el be such a big deal?

My concern is this certainly isn't the right fix: deleting and recreating the windows has other implications as well (e.g. a user has modified the window's properties). I think a better direction to look into is in `org-roam-fontify-like-in-org-mode`. We basically want to make sure that injecting the sections doesn't mess up the buffer. Perhaps something like this would work:

```
(defun org-roam-fontify-like-in-org-mode (s)
  "Fontify string S like in Org mode.
Like `org-fontify-like-in-org-mode', but supports `org-ref'."
  ;; NOTE: pretend that the temporary buffer created by `org-fontify-like-in-org-mode' to
  ;; fontify a `cite:' reference has been hacked by org-ref, whatever that means;
  ;;
  ;; `org-ref-cite-link-face-fn', which is used to supply a face for `cite:' links, calls
  ;; `hack-dir-local-variables' rationalizing that `bibtex-completion' would throw some warnings
  ;; otherwise.  This doesn't seem to be the case and calling this function just before
  ;; `org-font-lock-ensure' (alias of `font-lock-ensure') actually instead of fixing the alleged
  ;; warnings messes the things so badly that `font-lock-ensure' crashes with error and doesn't let
  ;; org-roam to proceed further. I don't know what's happening there exactly but disabling this hackery
  ;; fixes the crashing.  Fortunately, org-ref provides the `org-ref-buffer-hacked' switch, which we use
  ;; here to make it believe that the buffer was hacked.
  ;;
  ;; This is a workaround for `cite:' links and does not have any effect on other ref types.
  ;;
  ;; `org-ref-buffer-hacked' is a buffer-local variable, therefore we inline
  ;; `org-fontify-like-in-org-mode' here
  (with-temp-buffer
    (insert s)
    (let ((org-ref-buffer-hacked t))
      (org-mode)
      (page-break-lines-mode -1)
      (display-line-numbers-mode -1)
      (org-font-lock-ensure)
      (buffer-string))))
```

Either way I have never been able to reproduce this properly in my own setup, so I'd have to rely on external help here.

swang251:
@lazzalazza turning on the `visual` is simple, modify the step 2 as follows
```
(fset 'evil-redirect-digit-argument 'ignore)
(add-to-list 'evil-digit-bound-motions 'evil-org-beginning-of-line)
(evil-define-key 'motion 'evil-org-mode (kbd "0") 'evil-org-beginning-of-line)
(global-page-break-lines-mode -1)

(defun display-line-numbers-customize ()
	(setq display-line-numbers 'visual))
(add-hook 'org-mode-hook 'display-line-numbers-customize)
(advice-add 'org-roam-buffer-persistent-redisplay :before
			  (lambda () (remove-hook 'org-mode-hook 'display-line-numbers-customize)))
(advice-add 'org-roam-buffer-persistent-redisplay :after
			  (lambda () (add-hook 'org-mode-hook 'display-line-numbers-customize)))

```

swang251:
Hi @jethrokuan , thanks for the clarification and I agree that the way I dealt with the problem was not appropriate. It was initially taken as a workaround, instead of a solution. So I'd close the PR later.

I tried what you suggested but it didn't work. The format is ruined as soon as `display-line-numbers-mode` on, and toggle it back off won't help. I wouldn't imagine the `display-line-numbers-mode` itself is responsible for this but I have no idea what the real cause is.

lazzalazza:
@swang251 thank you so much for the solution for visual mode: it works perfectly and could be seen as a temporary patch to the problem. @jethrokuan It would be great if before closing the PR there could be some progress on how and why this problem occurs...

nobiot:
@lazzalazza 

> @jethrokuan It would be great if before closing the PR there could be some progress on how and why this problem occurs...

I don't know if Jethro can even give you any suggestion.

I added the two lines below. I can't reproduce the problem.
The visual display of line numbers seem to be working fine -- is it? I don't use it so I may be missing something.

As @swang251 analysed above, it seems to be an issue specific to Spacemacs.

```
(setq-default display-line-numbers 'visual)
(add-hook 'org-mode-hook 'display-line-numbers-mode)
```

Screenshot from 2021-11-13 16-19-29

swang251:
Hi @nobiot @lazzalazza @jethrokuan , I am closing this PR but let's keep the relative issues open until we figure out the causes of the weird format in the org-roam buffer. Thank you all.

PR #1944, opened 2021-11-06
(perf)db: require optional libraries only once per function

jethrokuan:
Require optional libraries 'org-ref and 'oc only once per function run:
org-roam-db-update-file and org-roam-db-sync will call these requires at
the top level instead of within the link mapper.

This is an alternative to #1932, but my own profiling on a file with a lot of
links show that this isn't actually the bottleneck.


jethrokuan:
cc @hugoArregui:

This is what I meant, but for me the bottleneck seems to be in `org-roam-id-at-point`, which calls `org-roam-db-node-p`, which is expensive because `org-entry-properties` is expensive. This slowness was introduced when we allowed people to exclude nodes.

image

PR #1947, opened 2021-11-06
(fix)db: don't consider links in ROAM_REFS as links

### LINKED ISSUE #1890: Multiple non-cite refs on a node cause duplicate self-references in Reflinks
### Description

If you have a note with several refs in its properties drawer, it will show up in its own roam buffer as reflinks.

A single `org-cite` ref in addition to (say) an `http` ref will not trigger this issue. Two `http` refs will.

#### Steps to Reproduce

1. Create a node
2. Add two `http` links to its `ROAM_REFS`. For example: `:ROAM_REFS: http://foo http://bar`
3. Open the roam buffer with `org-roam-buffer-toggle`
4. Observe the duplication in the roam buffer

#### Expected Results

A node should not appear in its own buffer, probably, but at a minimum the extra refs in the properties drawer should not show up.

#### Actual Results

```
Reflinks (2)
Test node (Top)
:PROPERTIES:
:ID:       1ab04e11-c761-4f3f-b3d0-76948d187c6f
:ROAM_REFS: http://foo http://bar
:END:

Test node (Top)
:PROPERTIES:
:ID:       1ab04e11-c761-4f3f-b3d0-76948d187c6f
:ROAM_REFS: http://foo http://bar
:END:
```

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.25, cairo version 1.16.0)
 of 2021-07-22
- Framework: N/A
- Org: Org mode version 9.5 (release_9.5-73-g57edfb @ /home/tracy/.emacs.d/hosts/archimedes.barbanon.org/straight/build/org/)
- Org-roam: v2.1.0-29-g54d17cc
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/54d17cc50f03f22ca44f4d88f7e589a00c59d05c

jethrokuan:
Links within ROAM_REFS are no longer added as links into the database.
This prevents self-referencing. Links within other properties are still
valid.

Closes #1890

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -342,6 +342,10 @@ If UPDATE-P is non-nil, first remove the file in the database."
          ;; Links correctly recognized by Org Mode
          ((eq type 'link)
           (setq link element))
+         ;; Prevent self-referencing links in ROAM_REFS
+         ((and (eq type 'node-property)
+               (string-equal (org-element-property :key element) "ROAM_REFS"))
+          nil)
          ;; Links in property drawers and lines starting with #+. Recall that, as for Org Mode v9.4.4, the
          ;; org-element-type of links within properties drawers is "node-property" and for lines starting with
          ;; #+ is "keyword".
@@ -354,7 +358,7 @@ If UPDATE-P is non-nil, first remove the file in the database."
           (with-temp-buffer
             (delay-mode-hooks (org-mode))
             (insert link)
-            (goto-char 1)
+            (point-min)
             (setq link (org-element-context)))))
         (when link
           (dolist (fn fns)


org-roam-db.el
```diff
@@ -342,6 +342,10 @@ If UPDATE-P is non-nil, first remove the file in the database."
          ;; Links correctly recognized by Org Mode
          ((eq type 'link)
           (setq link element))
+         ;; Prevent self-referencing links in ROAM_REFS
+         ((and (eq type 'node-property)
+               (string-equal (org-element-property :key element) "ROAM_REFS"))
```
nobiot, line 347: @jethrokuan `string-equal` is case sensitive (per its docstring). 
Are we letting properties to be specified by lower-case? 
I use lower-case and this condition returns `nil` for "roam_refs"...

jethrokuan: Sure, I'll submit another PR for this.

nobiot: Great. Thank you. There are some other places that use string-equal. I'm hoping that it won't cause cascade of changes...

PR #1949, opened 2021-11-07
(minor)db: make check for property drawers case-insensitive

jethrokuan:
As per title cc @nobiot 

I think other uses of `string-equal` are fine.


nobiot:
Thank you! Smart solution :)

PR #1955, opened 2021-11-09
(fix)buffer: change preview-content logic

### LINKED ISSUE #1934: roam-buffer shows full content of backlinks/reflinks
### Description
The roam backlinks buffer/window started showing the full content of nodes referencing the currently active node, and it's own content. Seems to be happening with all nodes I've tried.
 
#### Steps to Reproduce

1. Create two nodes a, b
2. Reference node a from node b
3. Activate backlinks buffer

See attachment.  The problem presents itself in all nodes currently. I've rebuild the db etc.

https://user-images.githubusercontent.com/68159077/139587767-f7601281-4db7-4a11-8833-7f31175fb953.mp4

#### Backtrace
n/a

#### Expected Results

Backlinks buffer should only show the paragaraph which includes a link to the currently shown/focused node.

#### Actual Results

The backlinks buffer show the complete content of the file referencing the active node, including headline and properies drawe.

### Environment

```
- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-redhat-linux-gnu, GTK+ Version 3.24.30, cairo version 1.17.4)
 of 2021-09-29
- Framework: Doom
- Org: Org mode version 9.5 (9.5-??-1b2d068 @ /home/[user]/.emacs.d/.local/straight/build-28.0.50/org/)
- Org-roam: 1b221a1
```

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/1b221a1d4ad0edae746a7eae435f4c3bca56c10e

jethrokuan:
Some commit in Org 9.5 changed the AST structure, so we our previous
preview content logic was borked and started showing the full contents
of the file. To be forward compatible we change the content preview
logic, and extract everything within the section.

Fixes #1934.


nobiot:
@jethrokuan 

So you ditched the use of org-element and entirely rely on a custom search function?

jethrokuan:
@nobiot yes, `org-element` was not giving the desired start and end points to extract.

nobiot:
@jethrokuan thank you 🙏 I also need to do something about Org-element for Org-transclusion—for a different reason and it’s sort of discussed in Org’s mailing list now. 

I was hoping I could “steal” your solution; will need to dig deeper. Thanks ☺️

yantar92:
In latest Org, `org-element-at-point' contains :parent elements even for headlines. Similar to what you get from `org-element-parse-buffer'.

PR #1961, opened 2021-11-10
Avoid surprises when STUFF contains the same key twice

pcrama:
Given that this code

    (let ((org-capture-plist (list :org-roam (list :z 26 :z "ignored"))))
      (org-roam-capture--get :z))

returns 26, I would expect

    (let ((org-capture-plist nil))
      (org-roam-capture--put :z 26 :z "ignored")
      (org-roam-capture--get :z))

to return 26, too, but it returns "ignored".

This patch fixes it by "reversing" the input list before using `plist-put` on its elements.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -445,13 +445,13 @@ the capture)."
   "Get the value for KEYWORD from the `org-roam-capture-template'."
   (plist-get (plist-get org-capture-plist :org-roam) keyword))
 
-(defun org-roam-capture--put (&rest stuff)
-  "Put properties from STUFF into the `org-roam-capture-template'."
+(defun org-roam-capture--put (prop value)
+  "Set property PROP to VALUE in the `org-roam-capture-template'."
   (let ((p (plist-get org-capture-plist :org-roam)))
-    (while stuff
-      (setq p (plist-put p (pop stuff) (pop stuff))))
     (setq org-capture-plist
-          (plist-put org-capture-plist :org-roam p))))
+          (plist-put org-capture-plist
+                     :org-roam
+                     (plist-put p prop value)))))
 
 ;;;; Capture target
 (defun org-roam-capture--prepare-buffer ()


jethrokuan:
Is it common to expect the "later" setting to be ignored? The only possible analogous scenario i can think of is alists where the items at the front get returned first, but that feels like a different mental model to me.

pcrama:
Using `plist-put` and `plist-get`, it is my expectation that only the first key/value pair is used in the presence of duplicate keys, see e.g. this experiment in M-x `ielm`:

    *** Welcome to IELM ***  Type (describe-mode) for help.
    ELISP> (plist-put (list :a 1 :a 2) :a 3)
    (:a 3 :a 2)

Only the first key/value pair is updated.

    ELISP> (plist-get (list :a 1 :a 2) :a)
    1 (#o1, #x1, ?\C-a)

Only the first key/value pair is returned.

My first lisp programming language was Common Lisp, not elisp and for what it's worth (elisp is not Common Lisp), the [CLHS glossary](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_p.htm) explicitly mentions this case in its definition of a property list
> property list n. 1. a list containing an even number of elements that are alternating names (sometimes called indicators or keys) and values (sometimes called properties). When there is more than one name and value pair with the identical name in a property list, the first such pair determines the property. 2. (of a symbol) the component of the symbol containing a property list.

OTOH, the [info manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Plists-and-Alists.html) seems to say that a property list with duplicate keys is an invalid plist, even though `plist-put`, `plist-get` and `plist-member` do not seem to validate them (probably to avoid the cost of walking the complete list?):
> ...In contrast to association lists, the order of the pairs in the property list is not significant, since the property names must be distinct...

The case that would cause a problem is where duplicate keys are present in `STUFF`: maybe this example where getting the same key from `STUFF` returns 2 different values is more convincing?

    ELISP> (let ((org-capture-plist nil)
                 (stuff (list :z 26 :z "ignored")))
             (apply 'org-roam-capture--put stuff)
             (list (plist-get stuff :z)
                   (org-roam-capture--get :z)))
    (26 "ignored")

If you are sure there will never be duplicate keys in `STUFF` then this branch (and extra runtime cost + garbage) is not worth it.  The `&rest` parameters of `org-roam-capture--put` is never really used: all usages of `org-roam-capture--put` are with a single key and value (i.e. you could have `(defun org-roam-capture--put (key value) ...)` that would not have the risk of duplicate keys in its input), even in [`org-roam-capture.el` lines 474-477](https://github.com/org-roam/org-roam/blob/50cc4d6990618e2510574159a9a5a52b1f928cd8/org-roam-capture.el#L475) where two such calls could have been merged into one function call to use the `&rest` parameter and pass both key/value pairs in one call.

My conclusion is a different proposal for `org-roam-capture--put`:

    (defun org-roam-capture--put (prop value)
      "Put PROP with VALUE into the `org-roam-capture-template'."
      (let ((p (plist-get org-capture-plist :org-roam)))
        (setq org-capture-plist
              (plist-put org-capture-plist
                         :org-roam
                         (plist-put p (pop stuff) (pop stuff))))))

It is a simplification because your code does not use the complicated case and then there is no discussion about the risk of duplicate keys in `org-roam-caputure--put`'s input.

jethrokuan:
@pcrama thanks for the detailed writeup, appreciate it. Perhaps it is best to simplify `org-roam-capture--put` to accept a single pair of values after all. I'd copied this off Org's `org-capture-put`, which also suffers from the same "issues" as `org-roam-capture--put`:

https://git.savannah.gnu.org/cgit/emacs/org-mode.git/tree/lisp/org-capture.el?h=cc2490a7061955395c4f5a1a23a088044554a2f7#n502

`org-roam-capture--put` is an internal function meant to only be used internally, so I do think a duplicate key issue is alright here. there's never a use case for setting the same key twice in the same function call.

pcrama:
I have rebased on the most recent develop and rewritten my branch according to the end of our discussion ([I said this](https://github.com/org-roam/org-roam/pull/1961#issuecomment-966482294), [you answered that](https://github.com/org-roam/org-roam/pull/1961#issuecomment-966868535)).  I hope that my commit message reflects the gist of our discussion (in case the issue is lost and only the commit remains).

Issue #1956, opened 2021-11-09
Add Title of File to the OLP


publicimageltd:
### Brief Abstract

I would like to know where a node "belongs to" if it is a heading. The data base field OLP seems to contain an outline path, but it begins at level 1 headings. I propose to include the title of the containing file as a level 0 heading, too. This way, an outline path can be easily displayed.

### Proposed Implementation (if any)

```elisp
(cl-defun org-roam-db-insert-node-data ()
....
(olp (append
      (list (org-link-display-format
              (or (cadr (assoc "TITLE" (org-collect-keywords '("title"))
                               #'string-equal))
                  (file-name-sans-extension
                   (file-relative-name file org-roam-directory)))))
      (org-get-outline-path nil 'use-cache)))
....
```

### Please check the following:

- [x] No similar feature requests

jethrokuan:
Please see this wiki entry: https://github.com/org-roam/org-roam/wiki/Hitchhiker's-Rough-Guide-to-Org-roam-V2#showing-node-hierarchy

publicimageltd:
Thanks for the link to the wiki. That's a fine approach, but it is too costly for longer queries since it requires to handle each node separately. `org-roam-filetitle` also requires the buffer to be loaded, if I understand it correctly, since `org-roam-get-keyword` presupposes to be in the active org mode buffer. So I'd need to loop through all nodes and even files to additionally add the filetitle to each node, right? My use case, however, is displaying nodes by querying them directly from the database (see [here](https://github.com/publicimageltd/delve)), and I would rather want to avoid looping over all nodes once more.

To put it differently: Isn't the method in the wiki relatively slow, so that it might be generally of benefit to add the file title to the OLP from begin with? I would do PR once you agree that it's worth it.

jethrokuan:
You're right that the wiki method is costly.

I'm disagreeable to modifying directly what the database stores as OLP: we should keep this as authentic to Org as possible. The file title is not part of the OLP.

I am agreeable to this approach instead, and open to a PR:

1. Add a column in the `files` table for `file-title`
2. Additionally caching the file-title in `org-roam-db-insert-file`
3. Adding an additional property `file-title` to the `nodes` structure
4. Ensure that the `file-title` property is properly populated in `org-roam-populate` and `org-roam-node-list`

Then in your own user config you can create your own custom `org-roam-node-olp-with-title` function that's just a string concatenation of the two and use it in your `org-roam-node-display-template`.

[...]

PR #1963, opened 2021-11-11
Add column "filetitle" to table "file"; adapt queries to that change

publicimageltd:
The column "filetitle" is populated with the document title or, if
there is none, the basic filename (sans extension) of the file. It is
copied into the org-roam-node when retrieving nodes.

See #1956

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -186,8 +186,7 @@ The query is expected to be able to fail, in this situation, run HANDLER."
 (defconst org-roam-db--table-schemata
   '((files
      [(file :unique :primary-key)
-      ;; "file-title" would be converted to "file_title"
-      filetitle
+      title
       (hash :not-null)
       (atime :not-null)
       (mtime :not-null)])
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -297,7 +297,7 @@ If FILE is nil, clear the current buffer."
 ;;;; Updating tables
 
 (defun org-roam-db--file-title ()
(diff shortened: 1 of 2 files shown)


org-roam-db.el
```diff
@@ -186,6 +186,8 @@ The query is expected to be able to fail, in this situation, run HANDLER."
 (defconst org-roam-db--table-schemata
   '((files
      [(file :unique :primary-key)
+      ;; "file-title" would be converted to "file_title"
```
jethrokuan, line 189: remove this comment. Also I think since it's the `files` table, `title` is sufficient, rather than `filetitle`.

org-roam-db.el
```diff
@@ -294,18 +296,30 @@ If FILE is nil, clear the current buffer."
                      file))
 
 ;;;; Updating tables
+
+(defun org-roam-db--file-title ()
+  "In current org document, get the title.
```
jethrokuan, line 301: /s/document/buffer
/s/org/Org

jethrokuan (commented):
Just some small nits, but looks good to me otherwise.

publicimageltd:
Sorry I am at a loss why "some checks where not successful." CI seems to complain about wrong indentation? If I run `elint-current-buffer` in `org-roam-node.el`, I find no problem. Don't know how to correct that.

jethrokuan:
@publicimageltd see https://github.com/skeeto/emacsql#emacs-lisp-indentation-annoyance

PR #1969, opened 2021-11-14
fix(org-roam-db-map-links): goto-char added to point-min

rdrg109:

Apparently, 4a2b44252ef4e71f183f627a49e6ec70c3cf8903 introduced a change that was not intended. The reason why I state this is because in that commit, `org-roam-db-map-links` evaluates `(point-min)` and it does nothing with the returned value. I think that `(goto-char (point-min))` was intended to be evaluated.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -375,7 +375,7 @@ If UPDATE-P is non-nil, first remove the file in the database."
           (with-temp-buffer
             (delay-mode-hooks (org-mode))
             (insert link)
-            (point-min)
+            (goto-char (point-min))
             (setq link (org-element-context)))))
         (when link
           (dolist (fn fns)
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -375,7 +375,6 @@ If UPDATE-P is non-nil, first remove the file in the database."
           (with-temp-buffer
             (delay-mode-hooks (org-mode))
             (insert link)
-            (goto-char (point-min))
             (setq link (org-element-context)))))
         (when link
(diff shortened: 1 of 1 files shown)


rdrg109:
I've made a new commit because I've noticed that there's no need to evaluate `(goto-char (point-min))` and consequently nor `(point-min)`. I can state this because `(org-element-context)` return information of the link at point even when the point is placed after the link. See minimal working example below.

```lisp
(with-temp-buffer
  (org-mode)
  (insert "[[a]]")
  (goto-char (point-min))
  (pp (org-element-context)))
```

```
(link
 (:type "fuzzy" :path "a" :format bracket :raw-link "a" :application nil :search-option nil :begin 1 :end 6 :contents-begin nil :contents-end nil :post-blank 0 :parent
        (paragraph
         (:begin 1 :end 6 :contents-begin 1 :contents-end 6 :post-blank 0 :post-affiliated 1 :parent nil))))
```

Now, let's try not to move the cursor.

```lisp
(with-temp-buffer
  (org-mode)
  (insert "[[a]]")
  (pp (org-element-context)))
```

```
(link
 (:type "fuzzy" :path "a" :format bracket :raw-link "a" :application nil :search-option nil :begin 1 :end 6 :contents-begin nil :contents-end nil :post-blank 0 :parent
        (paragraph
         (:begin 1 :end 6 :contents-begin 1 :contents-end 6 :post-blank 0 :post-affiliated 1 :parent nil))))
```

As we can see above, we get the same result.

Issue #1973, opened 2021-11-16
reflinks don't always show


zipy124:
When inserting references using org-ref and creating a org-roam notes file for that reference, the reflink does not always show using the org-roam buffer or org-roam-ui. Normal back-links work as expected.

nobiot:
Do you have reproducible examples?

zipy124:
Here we can see in window 1: the org-roam buffer for an org-roam notes file attached to an org-ref citation, the buffer open in windows 2 contains a link to the buffer open in window 3, but is not shown as a backlink:
![image](https://user-images.githubusercontent.com/6547350/142200049-1f9a808b-af5a-45b3-a7ae-786ee94b7cd4.png)

If i then cite the note in the note itself, it is however displayed:
![image](https://user-images.githubusercontent.com/6547350/142200428-f45d9e96-47a8-4625-86c0-7e20256c6305.png)

I then re-cite the note in the other note and we can see it is not however displayed:
![image](https://user-images.githubusercontent.com/6547350/142200637-e08d37ff-fd6c-4b79-85fc-10c9353f30c2.png)

In-fact if I even link to the roam ID directly using org-roam-node-insert it does not work, I can however do the reverse and link the original file:
![image](https://user-images.githubusercontent.com/6547350/142201311-7581453b-2975-4094-a385-43e3ca582ba7.png)
![image](https://user-images.githubusercontent.com/6547350/142201330-99c51fc9-ab7a-4b71-a8ee-3dd7f0d99607.png)

zipy124:
Upon reflection, perhaps it is because I updated to org-ref v3 and org-ref-bibtex only just got this commit:
https://github.com/org-roam/org-roam-bibtex/commit/f399b85f5c38bd52f6eb41da18fccfb971a04fe1

nobiot:
I don't seem to be able to reproduce the problem.

Does this correctly reflect what you have in the example?

- Window 1 (left) org-roam-buffer for "Literature note" (Window 3)
- Window 2 (top right) TODO with citing the literature note with `org-ref` syntax
- Window 3 (bottom right) "Literature note" with `roam_refs` property for `org-ref`

![Screenshot from 2021-11-17 17-30-50](https://user-images.githubusercontent.com/12507865/142241436-718426a9-153e-4db5-8f31-43d79158588a.png)

You seem to be using Spacemacs. Try without it.

nobiot:
[...]

PR #1975, opened 2021-11-18
fix:(org-roam-db)Support Org-ref V3 syntax #1973

nobiot:
Refer to issue #1973. This is meant to be a discussion starter with a concrete implementation example.
As the change will render `org-roam-db` incompatible with V2, we might not like to use it as is.

1. Function org-ref-split-and-strip-string does not exist in V3 Org-ref

2. org-ref-cite-types is now a different type in V3; `member` does not match; use `assoc` instead

3. match ROAM_REF without "&"

Because of Point 2 above, this change is not backward compatible with V2; the
change will make org-roam-db support only V3 Org-ref.

##### Test results:

Screenshot from 2021-11-18 17-36-06

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -507,10 +507,11 @@ INFO is the org-element parsed buffer."
       ;; For Org-ref links, we need to split the path into the cite keys
       (when (and source path)
         (if (and (boundp 'org-ref-cite-types)
-                 (fboundp 'org-ref-split-and-strip-string)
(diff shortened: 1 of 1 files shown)


jethrokuan:
Thanks for the preliminary investigation, I'll look into how we can maintain both v2/3 compatibility.

jethrokuan:
Superceded by #1977 , thanks @nobiot

PR #1977, opened 2021-11-20
(feat)db: support org-ref v3

jethrokuan:
Org-ref v3 introduced a new citation syntax, and removed some functions
that org-roam had previously relied upon (e.g. for extracting keys from
multi-citations).

This commit introduces support on two fronts:

1. Add `org-roam-org-ref-path-to-keys` which converts a link path to a
list of keys. This supports both Org-ref v2 and v3.

2. Adds support for parsing multi-citations in refs.

Addresses #1973.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -486,7 +486,13 @@ INFO is the org-element parsed buffer."
           (cond ((string-prefix-p "@" ref)
                  (push (vector node-id (substring ref 1) "cite") rows))
                 ((string-match org-link-plain-re ref)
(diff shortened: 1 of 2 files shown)


jethrokuan:
Let me know if this branch works for you @nobiot @jmkard @zipy124

jmkard:
Seems to work. Org-ref v3 citations are showing up in the roam buffer, no "&" hack required. Thanks for sorting this issue.

nobiot:
Thank you @jethrokuan. I won’t be able to test this weekend but I trust @jmkard. I don’t use Org-ref myself so i’m not a primary target user. Glad that it’s working for @jmkard.

PR #1988, opened 2021-11-24
Allow customized graph links and access to the temporary filenames.

mtbakerguy:

If a user wants to publish the graph with their notes, they need two things:

- a way to replace the org-protocol links in the generated graph
- the temporary files created so they can get access to them for publishing

PR 1727 can be closed as this is the same change done against the latest code.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -1439,6 +1439,82 @@ Essentially, to migrate notes from v1 to v2, one must:
    and the ~ROAM_TAGS~ property for headline nodes
 6. Replace existing file links with ID links.
 
+** How do I publish my notes with an Internet-friendly graph?
+
+The default graph builder creates a graph with an [[https://orgmode.org/worg/org-contrib/org-protocol.html][org-protocol]]
+handler which is convenient when you're working locally but
+inconvenient when you want to publish your notes for remote access.
+Likewise, it defaults to displaying the graph in Emacs which has the
+exact same caveats.  This problem is solvable in the following way
+using org-mode's native [[https://orgmode.org/manual/Publishing.html][publishing]] capability:
+
+1. configure org-mode to publish your org-roam notes as a project.
+2. create a function that overrides the default org-protocol link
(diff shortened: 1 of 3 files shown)


jethrokuan:
@mtbakerguy looks fine, just need to correct the lint errors.

mtbakerguy:
The texi changes surprised me.  I upgraded to 9.5 and saved the file again but that didn't account for the changes either.

jethrokuan:
@mtbakerguy The texi changes are due to your missing `ox-texinfo+`. If you give me push rights to your branch I can correct it for you, or you could install `ox-texinfo+` yourself.

mtbakerguy:
I added you as a contributor to my repository (thanks for suggesting that as I'd never done that before).  To make it easier to update the document, I'd suggest doing one of three things:

- use the texinfo exporter that comes with org-mode (this one's the easiest but I've no way to gauge ox-texinfo+'s additional value).
- update the `after-save-hook` to throw an error if someone's using the built-in exporter and give people a pointer to install it (this one is easy but contributor unfriendly which isn't what you want given people's reticence to writing documentation).
- create the .texi and .info files as part of some sort of packaging step in a Makefile so the .texi file isn't checked in (I prefer this one but it's the most work).

jethrokuan:
Usually I just update the texi manually myself (doc changes are rare, and most of the PRs are self-made anyway), so far it hasn't been worth the effort looking into.

nobiot:
@jethrokuan 

It's good to know about your use of `ox-texinfo+` (I believe it's [this](https://github.com/tarsius/ox-texinfo-plus)).
I think I broke the format of Info page when I wrote the Windows part of C compiler, but I just could not figure out how to make the format the way you do. 

It's true that most of the PRs are done by you.  Nevertheless, I believe you might get more contributions for the manual if your publishing workflow is a bit more explicit. 

Personally, I could not figure out how to fix the format breakage and didn't want to break it further so I left it as it was (sorry). I might try to come back to it once I install `ox-texinfo+` and figure out what it does.

Issue #1933, opened 2021-10-31
Multiple org-roam directories don't work according to manual instructions


josephmturner:
### Description

Following the [instructions in the manual](https://www.orgroam.com/manual.html#How-do-I-have-more-than-one-Org_002droam-directory_003f) for utilizing more than one org-roam directory leads to this warning: `org-roam-db--get-connection: Wrong type argument: stringp, (expand-file-name ".")`

The same warning appears if I try an absolute path as suggested in #1428.

I also tried the `eval` syntax as suggested in [this comment](https://github.com/org-roam/org-roam/issues/394#issuecomment-610123877) by @herbertjones (and use the same `eval` syntax for `org-roam-db-location`). This creates a new org-roam db but I am not able to reliably switch between org-roam directories.

#### Steps to Reproduce

1. Create a test directory
2. Inside the test dir, create `.dir-locals.el` containing
```
((nil . ((org-roam-directory . (expand-file-name "."))
         (org-roam-db-location . (expand-file-name "./org-roam.db")))))
``` 
2. Create a test config file containing only `(require 'org-roam)` (if necessary, set load-path as well)
3. Start emacs: `emacs -q -nw -l /path/to/test-config-file.el`
4. Open the test directory in dired (`C-x C-d` then type in `/path/to/test-dir`)
5. Run `M-x org-roam-db-sync` or `M-x org-roam-node-find`

#### Backtrace

```
Debugger entered--Lisp error: (wrong-type-argument stringp (expand-file-name "."))
  expand-file-name((expand-file-name "."))
  org-roam-db--get-connection()
  org-roam-db--close()
  org-roam-db-sync(nil)
  funcall-interactively(org-roam-db-sync nil)
  call-interactively(org-roam-db-sync record nil)
  command-execute(org-roam-db-sync record)
  execute-extended-command(nil "org-roam-db-sync" "org-roam-db-s")
  funcall-interactively(execute-extended-command nil "org-roam-db-sync" "org-roam-db-s")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

#### Expected Results

Depending on the run command, the org-roam db syncs or I can select a node.
#### Actual Results

Nothing happens, besides the error message popping up in the mode line.

### Environment
- `Emacs: GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)`

- Org-roam commit: v2.1.0

[...]

Issue #1527, opened 2021-05-26
Multiple org-roam directory - Ignoring unsafe file local variable for org-roam-directory 


Felixaverlant:
Hi there!
Thank you for this amazing product! I'm using it all the time and I feel the need to have more than one directory now.

### Description
I am trying to setup [21.1 How do I have more than one Org-roam directory?](https://www.orgroam.com/manual.html#How-do-I-have-more-than-one-Org_002droam-directory_003f) .

I've added the code provided in a .dir-locals.el.
Everything works fine on the primary org-roam. I can insert, search...

But for the second org-roam, it throws the error provided below. 
If I search or insert, it will do everything on the primary org-roam. 
If I describe org-roam-directory, it's set on the primary org-roam directory. 

Inside my config.el, I've tried to add: 
`(add-to-list 'safe-local-variable-values '(org-roam-directory . "."))`

But it has no effect whatsoever.

#### Steps to Reproduce

I just have to load an org file on the second org-roam to have this error thrown.

#### Backtrace
On load of an org-file of the second org-roam.
```
Ignoring unsafe file local variable: org-roam-db-location
Ignoring unsafe file local variable: org-roam-directory
```

#### Expected Results

I expect to be able to use the second one.

#### Actual Results

I can only use the primary org-roam.

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, aarch64-apple-darwin20.4.0, NS appkit-2022.44 Version 11.3.1 (Build 20E241))
 of 2021-05-23
- Framework: Doom
- Org: Org mode version 9.5 (nil @ /Users/me/.emacs.d/.local/straight/build-28.0.50/org-mode/)
- Org-roam: 1.2.3

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

Thank you!
Best.

Felixaverlant:
Hi, it is a doom-emacs related issue with how [elisp-mode package is configured](https://github.com/hlissner/doom-emacs/blob/26a0244f03a6beb5d9dfb25ccc27c8a69b33002a/modules/lang/emacs-lisp/config.el). 
Best

Just-Insane:
Were you able to find a resolution to this so that you are able to use a second org-roam with doom-emacs?

Felixaverlant:
> Were you able to find a resolution to this so that you are able to use a second org-roam with doom-emacs?
> 
[...]

Issue #1459, opened 2021-03-27
directory-local org-roam-directy value of "." becomes global


codygman:
### Description

If I use the recommended .dir-locals in a secondary org-roam folder:

```
((nil . ((org-roam-directory . ".")
         (org-roam-db-location . "./org-roam.db"))))
```

I somehow end up with `org-roam-directory` being globally set to `.` which causes all kinds of issues like running find calls on my entire $HOME directory.

#### Steps to Reproduce

1. Create new folder with `.dir-locals.el` above and add an org roam file
2. Visit that a file in that folder
3. Go back to your home directory and `C-h v org-roam-directory RET`

You'll likely see something like:

```
Its value is "."
Original value was 
"/home/cody/org-roam/"
Local in buffer 20210326144748-org_roam.org; global value is 
"/home/cody/org-roam/"

```

If not, wait until the update function runs. I think something in there is globally setting this value.

#### Expected Results

My org-roam-directory variable is *never* set to `.` and org-roam-db-update isn't run in every buffer I visit.

#### Actual Results

If I load that directory local variable it effectively means any directory I visit will get an org-roam.db and be updated on idle timer.

### Environment
Copy info below this line into issue:
- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)
- Framework: N/A
- Org: Org mode version 9.4.4 (9.4.4-27-gb712b9-elpa @ /nix/store/fslr7dhwzq7hi406292wsqb8h1kv4hk6-emacs-packages-deps/share/emacs/site-lisp/elpa/org-20210315/)
- Org-roam: 1.2.3

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/commithashhere

Not sure how to get the commit hash, but I'm using `org-roam-20210308.457`.

codygman:
I also have some notes on trying to work through the issue and debug it here: https://github.com/codygman/tech-roam/blob/master/20210326144748-org_roam.org#using-dir-locals-with-org-roam-somehow-results-in-org-roam-directory-being-set-globally-to-

codygman:
I think this could be specific to my setup and involve a weird interaction with with-editor, though that's just a theory.

codygman:
Reproduction steps are:

1. Go into org-roam directory, `org-roam-db-build-cache`.
2. create `org-roam-directory` and add .dir-locals.el with these contents
```emacs-lisp
[...]

PR #2002, opened 2021-12-11
(docs): fix examples for .dir-locals.el

nobiot:
Two items:

1. The current example does not work; in order to call a function wihtin
`.dir-locals.el`, you should use `eval`.

  - [StackExchange](https://emacs.stackexchange.com/questions/21955/calling-functions-in-dir-locals-in-emacs)

  - [Working example buried in an issue](https://github.com/org-roam/org-roam/issues/1527#issuecomment-933674233)

  - `(info "(emacs)Directory Variables")`

2. Add clear instruction to use an absolute path to define `org-roam-directory` in
this context (it's buried in this [issue](https://github.com/org-roam/org-roam/issues/1459#issuecomment-817259656)).

Questions relatd to mulitple Org-roam DB are occasionally posted on Discouse (a [recent example](https://org-roam.discourse.group/t/small-independent-org-roam-instances/2267))

There has been some issues around `.dir-locals.el` -- I hope this update will help close this one, for example:
- https://github.com/org-roam/org-roam/issues/1933

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -1391,8 +1391,12 @@ Note ~org-roam-directory~ and ~org-roam-db-location~ should be an absolute path,
 Alternatively, use ~eval~ if you wish to call functions:
 
 #+BEGIN_SRC emacs-lisp
-  ((nil . ((eval . (setq-local org-roam-directory (expand-file-name ".")))
-           (eval . (setq-local org-roam-db-location (expand-file-name "org-roam.db"))))))
+  ((nil . ((eval . (setq-local
+                    org-roam-directory (expand-file-name (locate-dominating-file
+                                                          default-directory ".dir-locals.el"))))
+           (eval . (setq-local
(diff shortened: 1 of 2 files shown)


Lord-Valen:
That solution breaks in the event that there are subdirectories.
```emacs-lisp
((nil . ((eval . (setq-local org-roam-directory (expand-file-name (locate-dominating-file default-directory ".dir-locals.el"))))
         (eval . (setq-local org-roam-db-location (expand-file-name (concat org-roam-directory "org-roam.db")))))))
```
Seems to work after some testing.

nobiot:
> That solution breaks in the event that there are subdirectories.
@Lord-Valen 
Thank you! Reproduced the breakage when sub-directories exist. Fixed the example with `eval`; lightly tested to work.
The other example with an absolute path should work as is (untested with subdirectory but it's absolute path...).

PR #2008, opened 2021-12-18
Support backlinks in notmuch email messages

Zetagon:
I've discussed this before on the discourse https://org-roam.discourse.group/t/how-easy-would-it-be-to-extend-org-roam-with-backlinks-to-other-types-of-items/1582 . I'm not sure if this is a feature that should be in org-roam so I'm leaving this as a draft for now (the commit messages aren't the best either).

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -657,12 +657,14 @@ database, see `org-roam-db-sync' command."
     (cond
      (enabled
       (add-hook 'find-file-hook  #'org-roam-db-autosync--setup-file-h)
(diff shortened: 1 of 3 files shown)


jethrokuan:
Support for external caching (non-Org) should be placed into the `contrib` section. Concretely that means `org-roam-db.el` needs to expose a way to consider and parse other types of files (beyond what `org-roam-list-files`). I'll think of a way to expose such an interface, but the current modifications are not ideal.

cooladam201212:
I think this is a very useful feature, there's even a stand-alone MacOS app called [Hookmark](https://hookproductivity.com/) to achieve this.

Issue #1927, opened 2021-10-27
sqlite3 option for org-roam-database-connector does not delete or update the database cache


nobiot:
### Description
Set `(setq org-roam-database-connector 'sqlite3)`. Org-roam caches a new node created.
It does not update or delete an existing node -- so C and R of CRUD work but U and D don't.

#### Steps to Reproduce

0. Delete configuration files (remove the directory `~/.config/emacs`)
1. Load Emacs; ensure it is loaded with no configuration
2. Set `package-archives` with MELPA and install `org-roam`, its dependecies, and `emacsql-sqlite3`
3. Ensure `sqlite3` is installed
4. Create a new file node in `org-roam-directory` -- e.g. first.org
5. Create a second file node -- e.g. second.org
6. Confirm both nodes appear with `org-roam-node-find`
7. Update second file node -- e.g. `org-roam-node-insert` to link to first.org
8. Save
9. See the screen shot

![Screenshot from 2021-10-27 17-41-57](https://user-images.githubusercontent.com/12507865/139133419-c3d18dfe-7ece-4346-ad1a-a963243b5fcd.png)

10. Use Dired and delete the second file
11. Run `org-roam-db-sync`
12. Run `org-roam-node-find`; you will see the second file still in the list of completion candidates

![Screenshot from 2021-10-27 17-43-25](https://user-images.githubusercontent.com/12507865/139134639-688f7b52-b983-423c-9b53-97bce384e0d3.png)

#### Backtrace
For update error:
```
Debugger entered--Lisp error: (emacsql-error "near line 331: UNIQUE constraint failed: nodes.id")
  signal(emacsql-error ("near line 331: UNIQUE c<!-- Example: File A is there -->onstraint failed: nodes.id"))
  #f(compiled-function (conn) #<bytecode -0x1a47a8250eaf33be>)(#<emacsql-sqlite3-connection emacsql-sqlite3-connection-155ef80413ce>)
  apply(#f(compiled-function (conn) #<bytecode -0x1a47a8250eaf33be>) #<emacsql-sqlite3-connection emacsql-sqlite3-connection-155ef80413ce> nil)
  emacsql-parse(#<emacsql-sqlite3-connection emacsql-sqlite3-connection-155ef80413ce>)
[...]

PR #2009, opened 2021-12-21
fix(db): fix support for sqlite3 connector

jethrokuan:
Moving the PRAGMA foreign_keys invocation out of the transaction seems
to work now.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -243,8 +243,8 @@ The query is expected to be able to fail, in this situation, run HANDLER."
 
 (defun org-roam-db--init (db)
   "Initialize database DB with the correct schema and user version."
+  (emacsql db [:pragma (= foreign_keys ON)])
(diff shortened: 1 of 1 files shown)


jethrokuan:
cc @nobiot if you want to confirm :)

nobiot:
@jethrokuan
Thank you. Will check on my end :)

nobiot:
See detail of  my tests in #1927.
I have confirmed that the fix works on Windows 10 :)

nobiot:
@jethrokuan 
Ubuntu struggles with `sqlite3`.

Ubuntu's Sqlite3 package gives me an older version (and has the same issue with the newest) so I compiled from source.
Creating new files seems to be OK (ish) but updating the same file gives me an error.

Screenshot from 2021-12-21 12-47-13

nobiot:
Maybe the Widows binary does something special?

PR #2013, opened 2021-12-22
docs: Correct how to install gcc on Windows

nobiot:

PATH needs to be manually set on Windows. Previously, it was explicitly
mentioned that it was unnecessary. More recent experience is that it's necessary.


nobiot:
Sorry, I have no idea how my previously merged commits are also listed. I only made one additional commit.

jethrokuan:
> Sorry, I have no idea how my previously merged commits are also listed. I only made one additional commit.

Because the branch commits have diverged (i've merged stuff after your last commit). What will fix this is rebasing your branch on top of master. I'll go ahead and do that for you (and regenerate the texi)

nobiot:
@jethrokuan Thank you!

Issue #1910, opened 2021-10-19
Error (use-package): org-roam/:config: EmacSQL had an unhandled condition: "near line 19: UNIQUE constraint failed: nodes.id"


aleneum:
### Description

Everytime I load `org-roam` I get a `UNIQUE constraint failed: nodes.id`. This also happens when there is only one ID in one file set. I checked for duplicates with `(org-id-update-id-locations (directory-files-recursively org-roam-directory ".org$\\|.org.gpg$"))` but found nothing (since it's just one file with one id). ~As far as I can tell, this issue only appear once when `org-roam-node-find` is called the first time.~ It also happens when calling `org-roam-db-sync` individually or when I save files. I also checked the SQLite `nodes` table where again only one ID is present. This problem only manifests when `emacsql-sqlite3` is used.

#### Steps to Reproduce

1. Load Emacs
2. Run `org-roam-node-find` (or  `org-roam-db-sync`)

#### Backtrace

```elisp
Debugger entered--Lisp error: (emacsql-error "near line 19: UNIQUE constraint failed: nodes.id")
  signal(emacsql-error ("near line 19: UNIQUE constraint failed: nodes.id"))
  #f(compiled-function (conn) #<bytecode 0x240a01d>)(#<emacsql-sqlite3-connection emacsql-sqlite3-connection-2f26548>)
  apply(#f(compiled-function (conn) #<bytecode 0x240a01d>) #<emacsql-sqlite3-connection emacsql-sqlite3-connection-2f26548> nil)
  emacsql-parse(#<emacsql-sqlite3-connection emacsql-sqlite3-connection-2f26548>)
  #f(compiled-function (connection sql &rest args) "Send SQL s-expression to CONNECTION and return the results." #<bytecode 0x23cbcf1>)(#<emacsql-sqlite3-connection emacsql-sqlite3-connection-2f26548> [:insert :into nodes :values $v1] (["9c488176-ad57-4d59-b944-be5adcfef474" "c:/Users/alneuman/sciebo/org/roam/backlog.org" 0 1 nil nil nil nil "📓 Backlog" (("CATEGORY" . "backlog") ("ID" . "9c488176-ad57-4d59-b944-be5adcfef474") ("BLOCKED" . "") ("FILE" . "c:/Users/alneuman/sciebo/org/roam/backlog.org") ("PRIORITY" . "B")) nil]))
[...]

PR #2018, opened 2021-12-27
(fix):db: pragma foreign keys to work with sqlite3

nobiot:
This commit is follow-up for PR #2009 should fix issues #1927 & #1910.

Currently, `[:pragma (= foreign_keys ON)]` is present only in function
`org-roam-db--init`. This means the `foreign_keys` is turned on only when the db
file is created for the first time. Subsequent Emacs sessions won't turn it on
as the db file is already present and does not evaluate `org-roam-db--init`.

This PRAGMA needs to be turned on each database connection at runtime according
to sqlite's documentation (https://sqlite.org/foreignkeys.html#fk_enable)

```
Foreign key constraints are disabled by default (for backwards compatibility),
so must be enabled separately for each database connection
```

I have observed that on Windows only the Emacs session that initially creates
the Org-roam db file works correctly with the `sqlite3` option. Subsequent Emacs
sessions have the same "Unique constraint failed" error messages.

I have tested this patch on Windows and Ubuntu; both seem to work in the first
and subsequent Emacs sessions.

I am not 100% sure if the location of the PRAGMA is the best; use it as a
proof-of-cocenpt for the fix. Thank you.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -103,9 +103,10 @@ slow."
 
 ;; TODO Rename this
 (defconst org-roam--sqlite-available-p
-  (with-demoted-errors "Org-roam initialization: %S"
-    (emacsql-sqlite-ensure-binary)
-    t))
+  (when (equal org-roam-database-connector 'sqlite)
+    (with-demoted-errors "Org-roam initialization: %S"
+      (emacsql-sqlite-ensure-binary)
+      t)))
 
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
   "Database connection to Org-roam database.")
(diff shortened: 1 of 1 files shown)


jethrokuan:
@nobiot thanks for investigating. I found it strange from a design perspective, but I found a forum post that describes a little on how this particular pragma came to be a *per connection* rather than *per db file* setting: https://sqlite.org/forum/info/c5dc50f61b88c587

Your fix is correct. We would then want to remove the other `pragma` call here: 
https://github.com/org-roam/org-roam/blob/ee9a8d423e24923ec60c3116a00c8a0e93095d86/org-roam-db.el#L246

Then this should be good to merge.

nobiot:
@jethrokuan 
Thank you for this. Removed, commit pushed. 
Tested OK with Windows and Ubuntu :)

nobiot:
@jethrokuan 

Started using `sqlite3` and noticed this error (see my third commit).

`org-roam--sqlite-available-p` does not seem to be used at all. Do you want to keep it for some reason?
In any case, the error message is not relevant and with "aborting" (it does not abort) it's misleading.

jethrokuan:
> org-roam--sqlite-available-p does not seem to be used at all. Do you want to keep it for some reason?

No, it can be removed

nobiot:
done :)

Issue #1843, opened 2021-09-10
Make node filenames and titles consistent (V2)


Hxppdv:
### Brief Abstract
I know that in org-roam V2, the links are based on the org-id so there is no need to force these to have the same name. But maybe this could be some kind of optional feature
### Long Description
In my personal practice, of course, it is useful to have different descriptions for the same link, but most of the time I need to make sure that the description of the link is consistent with the title of the content being linked to. Especially in some study notes, I may have created title A, but later, as I learn more, I may find that I should change title A to C. I can of course change all the links manually if there are few, but if there are many, this becomes a big problem.

Consistent filenames and titles seem to me to be good for use with other software. orgmode is text-based, which is good, and it would be nice to have consistent filenames and titles.

abhijeetsrivastava:
Meanwhile, hope this helps https://org-roam.discourse.group/t/does-renaming-title-no-longer-renames-the-filename/2018/3?u=abhiabhi

jpggithub:
The Emacs package `zk` (https://github.com/localauthor/zk) implements this functionality.

Complete demo of the functionality: https://www.youtube.com/watch?v=BixlUK4QTNk&t=495s

jpggithub:
It could be optional.
It could also be a function usable as a hook on saving the file. 
Currently, I rename manually  the buffer and file. It would be nice to to do it automatically based on the title of the node 😎

hpfr:
Removing descriptions from scope here since it’s more complicated and not necessarily desirable (as acknowledged). Feel free to open a separate issue

Inkbottle007:
I think automatic propagation (link description rewrites) is workflow dependent. (see for example https://org-roam.discourse.group/t/org-roam-v2-rename-file-or-title/1691/4)
In contrast I think a helper function to conveniently update the filename to reflect buffer's `#+title:` is very useful, because doing it by hand is tedious.
Here is my take on it, based on quolpr's code:
https://org-roam.discourse.group/t/does-renaming-title-no-longer-renames-the-filename/2018/11

hpfr:
Because capture templates set the filename and node title independently, this is complex to solve in the general case, especially if you want to keep the timestamping done in the original filename. My philosophy is to generally ignore filenames, node titles are what really matters. 

[...]

PR #2020, opened 2021-12-28
org-roam-node-rename

SeungheonOh:
Functionality to rename current node name minding the title, all links, and filename.

Previously, the links, filename, and title have to be renamed manually. It wound be nice to have clean function to rename a node with links and filename taken care of.

It is improved version of function mentioned in #1843. It uses org-roam-backlink to locate all occurrences of links and replaces. 

https://user-images.githubusercontent.com/24271442/147531641-6498dc78-2d25-4dda-ad3b-b08e60541db1.mov

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -496,11 +496,15 @@ title/filename/link titles."
     (save-buffer)
 
     ;; Rename current buffer
-    (let* ((fname (buffer-file-name))
-	   (newreg (format "\\1-%s.org" newtitle))
-	   (newfname (s-replace-regexp "\\([0-9]+\\)\\-.+\.org" newreg fname)))
+    (let* ((orif (buffer-file-name))
+	   (dir (file-name-directory orif))
+	   (file (file-name-nondirectory orif))
+	   (titlenoslash (s-replace-regexp "/" "_" (downcase newtitle)))
+	   (newreg (format "\\1-%s.org" titlenoslash))
+	   (nfile (s-replace-regexp "\\([0-9]+\\)\\-.+\.org" newreg file))
+	   (newfname (concat dir nfile)))
       (progn
-	(rename-file fname newfname)
+	(rename-file orif newfname)
 	(rename-buffer newfname)
 	(set-visited-file-name newfname)
 	(set-buffer-modified-p nil)))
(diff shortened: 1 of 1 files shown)


PRESFIL:
Sometimes it shouldn't to update the link name if it does not contain the node name,
but some other information associated with the context of the link.

For example, the link `[[id:...][look at the source]` does not need be updated, because
it does not contain the name of the node. How can such cases be ignored?

SeungheonOh:
@PRESFIL The regex match can include the node name; that will only match for the link with the current node name. 
Not sure if I want to do that though, because I think it should take care of corrupted links as well.

org-roam-node.el
```diff
@@ -476,6 +476,52 @@ The TEMPLATES, if provided, override the list of capture templates (see
...
+  (let* ((id (org-entry-get (point) "ID"))
+	 (node (org-roam-node-from-id id))
+	 (title (org-roam-get-keyword "title"))
+	 (newtitle (read-string "New title: " title)))
+    (when (null id) (error "Failed to fetch ID"))
+    (when (org-roam-node-from-title-or-alias newtitle)
+      (error (format "Node %s already exists" newtitle)))
+
+    ;; Set new title
+    (org-roam-set-keyword "title" newtitle)
```
jethrokuan, line 495: This function only works for file nodes: it needs to generalize to headline nodes as well.

jethrokuan (commented):
This function used to exist in v1, and was removed in v2 because there are a lot of edge cases here to consider here:

- `rename-file`, and `save-buffer` often both contain hooks that can cause this function to fail.
- the new file name is often dependent on the template used, and the slug function used. There is no way to determine this from the file alone.

SeungheonOh:
Hmm is this anyway to get the file name templates? That way I think I can force it to use certain file name template.

nigo81:
> Functionality to rename current node name minding the title, all links, and filename.
> 
> Previously, the links, filename, and title have to be renamed manually. It wound be nice to have clean function to rename a node with links and filename taken care of.
> 
> It is improved version of function mentioned in #1843. It uses org-roam-backlink to locate all occurrences of links and replaces.
> 
>  Screen.Recording.2021-12-27.at.11.34.24.PM.mov

I need this function too! your code works with my org-roam, i wish org-roam can use this function originally.Thanks a lot.

PR #2021, opened 2021-12-28
Make 'fd and 'fdfind backend actually work for listing files

nchachereau:

The `fd` backend for listing files, introduced in #1583, did not work for me, mainly because the command is actually:

    fd -L --type file -e .org -e .org.gpg . "/home/user/org-roam"

In other words, the directory comes last, and a pattern (a match-all pattern `.`) needs to be specified (otherwise `fd` assumes the last argument is the pattern, in this case `/home/user/org-roam`). There was a slight bug in the creation of the `extensions`variable, where only the second pattern (`.org.gpg`) would be prefixed by `-e`.

P.S. It might be good if @Wetlize would chime in, as they wrote the original PR.


nchachereau:
Great, thank you!

PR #2025, opened 2022-01-05
Drop f.el dependency

skangas:

It seems unnecessary to depend on an entire library just for one function that is trivially implemented ourselves. I suggest we drop that requirement.

Part of the motivation for this is that I am currently investigating if we can add this package to NonGNU ELPA, and we have not yet added f.el to that archive; and adding it will take more time.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 ### Changed
 - [#2040](https://github.com/org-roam/org-roam/pull/2040) completions: fix completions display-width for Helm users
+- [#2025](https://github.com/org-roam/org-roam/pull/2025) chore: removed the dependencies on f.el and s.el
 
 ## 2.2.0
 ### Added
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -269,7 +269,7 @@ If FILE is not specified, use the current buffer's file-path."
     (save-match-data
       (and
        (org-roam-file-p path)
-       (f-descendant-of-p path directory)))))
+       (org-roam-descendant-of-p path directory)))))
 
 ;;;###autoload
 (defun org-roam-dailies-find-directory ()
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -8,7 +8,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 2.2.0
-;; Package-Requires: ((emacs "26.1") (dash "2.13") (f "0.17.2") (org-roam "2.1"))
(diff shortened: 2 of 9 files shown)


skangas:
I see now that there is previous work on this in https://github.com/org-roam/org-roam/pull/1889. My pull request is probably less risky, as it more closely mimics the f.el behaviour.  For example, these rewritten tests from the f.el testsuite verify this:

```
(progn
  (require 'ert)
  (require 'f) ; for the f-expand etc. part

  (mkdir "foo/bar/baz/qux" t)
  (should (org-roam-descendant-of-p "foo/bar" "foo"))
  (should (org-roam-descendant-of-p "foo/bar/baz" "foo"))
  (should (org-roam-descendant-of-p "foo/bar/baz/qux" "foo"))
  (should (org-roam-descendant-of-p "foo/bar/baz" "foo/bar"))
  (should (org-roam-descendant-of-p "foo/bar/baz/qux" "foo/bar"))
  (should (org-roam-descendant-of-p "foo/bar/baz/qux" "foo/bar/baz"))
  (should (org-roam-descendant-of-p (f-expand (car (f-directories (f-root))) (f-root)) (f-root)))
  (should (org-roam-descendant-of-p "/foo/bar/baz/qux" "/foo/bar/baz"))

  (should-not (org-roam-descendant-of-p "foo" "foo/bar"))
  (should-not (org-roam-descendant-of-p "foo" "foo/bar/baz"))
  (should-not (org-roam-descendant-of-p "foo" "foo/bar/baz/qux"))
  (should-not (org-roam-descendant-of-p "foo/bar" "foo/bar/baz"))
  (should-not (org-roam-descendant-of-p "foo/bar" "foo/bar/baz/qux"))
  (should-not (org-roam-descendant-of-p "foo/bar/baz" "foo/bar/baz/qux"))
  (should-not (org-roam-descendant-of-p (f-root) (f-expand (car (f-directories (f-root))) (f-root))))
  (should-not (org-roam-descendant-of-p "/foo/bar/baz" "/foo/bar/baz/qux"))

  (should-not (org-roam-descendant-of-p "foo" "foo"))
  (should-not (org-roam-descendant-of-p "foo/bar" "foo/bar"))
  (should-not (org-roam-descendant-of-p "foo/bar/baz" "foo/bar/baz"))
  (should-not (org-roam-descendant-of-p "foo/bar/baz/qux" "foo/bar/baz/qux"))
  (should-not (org-roam-descendant-of-p (f-root) (f-root))))
```

jethrokuan:
@phikal happy to help w the process, i'm not sure if it's possible with emacsql being a non-negotiable dependency.

skangas:
Would org-roam work with emacsql version 3.0.0, released in 2018? Details here: https://stable.melpa.org/#/emacsql

If not, we would need this ticket merged first: https://github.com/skeeto/emacsql/pull/87

PR #2028, opened 2022-01-06
Add `keys` argument to `org-roam-dailies-capture-today` and `org-roam-dailies--capture` functions.

astery:
That gives convenience to create shorter keybindings that lead right into desired capture template buffer avoiding the need to select it each time in the capture selection screen. 

```elisp
    (leader-def
      "rt" '((lambda () (interactive) (org-roam-dailies-capture-today nil "d")) :wk "capture today daily")
      "ri" '((lambda () (interactive) (org-roam-dailies-capture-today nil "i")) :wk "capture interruption"))
      "ro" '(some-other-roam-bindings)
```

It makes `org-roam-dailies-capture-today` arguments to behave exactly as `org-capture` arguments do - (&optional goto keys).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## TBD
 ### Added
 - [#1806](https://github.com/org-roam/org-roam/pull/1806) db: support caching and usage of Org 9.5's in-built citations
+- [#1806](https://github.com/org-roam/org-roam/pull/2028) dailies: add `keys` argument to `org-roam-dailies-capture-today` and `org-roam-dailies--capture` functions to give the abilty to get into a capture buffer bypassing the selection screen
 
 ### Removed
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## TBD
 ### Added
 - [#1806](https://github.com/org-roam/org-roam/pull/1806) db: support caching and usage of Org 9.5's in-built citations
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## TBD
 ### Added
 - [#1806](https://github.com/org-roam/org-roam/pull/1806) db: support caching and usage of Org 9.5's in-built citations
+- [#1806](https://github.com/org-roam/org-roam/pull/2028) dailies: add `keys` argument to `org-roam-dailies-capture-today` and `org-roam-dailies--capture` functions to give the abilty to get into a capture buffer bypassing the selection screen
```
jethrokuan, line 6: Thanks for trying to update the changelog. I haven't been very diligent with this either, but the PR number is off here.

astery: Do you mean it shouldn't be there or it isn't working? 

I've copy-pasted it in the address bar and it works for me:
image

astery: Oh, now I see it - the label itself

astery: Fixed that

astery:
Great, thank you!

meliache:
Oh nice. This addresses #1354. I will use this in my hydra to replace my `org-roam-dailies-goto` commands with the `(org-roam-dailies-capture-... 'goto <key>)`. I'm thinking about asking to add this for convenience also to the `dailies-goto` commands, though I assume they are mainly for interactive use and for non-interactive use the capture commands can be used with the `goto` argument, so I think it's fine as it is.

Issue #1578, opened 2021-07-06
Helm completion too wide for side-by-side windows


bnordbo:
#### Steps to Reproduce

1. Set up two windows side-by-side using `C-x 3`.
2. Run `M-x org-roam-node-find`.

#### Expected Results

A list of completions in the inactive window.

#### Actual Results

As expected, except all the lines are way too long. This causes them to wrap, causing a very untidy display.

From what I can tell, the problem is caused by using `frame-width` as the basis for the width  of the completion list in [`org-roam-node--to-candidate`](https://github.com/org-roam/org-roam/blob/v2/org-roam.el#L738). Replacing `(1- (frame-width))` with  `(window-width)` works as long as the windows are of equal width, but I suspect this might affect completions using the minibuffer. I suppose the problem is that neither function return the width of the to-be completion buffer.

### Environment

- Emacs: GNU Emacs 28.0.50 (build 1, x86_64-apple-darwin20.4.0, NS appkit-2022.44 Version 11.3.1 (Build 20E241))
 of 2021-06-08
- Framework: N/A
- Org: Org mode version 9.4.4 (release_9.4.4 @ /usr/local/Cellar/emacs-plus@28/28.0.50/share/emacs/28.0.50/lisp/org/)
- Org-roam: 2.0.0
- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/90945e8c03b170731070af2009b89b9687ff2f58

jethrokuan:
Do you have a screenshot of what it actually looks like? Are your minibuffer completions never always the frame width? I've always assumed this to be the case.

bnordbo:
Sure, here's an example when the frame is split side-by-side:

![image](https://user-images.githubusercontent.com/696788/125159112-a6c0ab80-e175-11eb-8127-2be58271fee0.png)

When it's split below (or not split at all), it looks fine because the completions get the full width. Note that the lines are almost twice as long as the window width. I've set aside 20 characters for tags.

In the above example, `(frame-width)` returns 237, and `(window-width)` gives 117 for both windows. I have a stock helm setup, and it will show completions in the window to the right if split side-by-side (or the current window if invoked in the rightmost window).

Maybe there's a way to get the width of the window which will be used for completions given the current window setup?

jethrokuan:
@bdarcus do you have any insights here? I see that `bibtex-actions` also uses `frame-width` to compute the rich UI, I suppose it would see similar issues.

bdarcus:
[...]

PR #2040, opened 2022-01-15
(fix)Format completions to window-width, not frame-width

stuhlmueller:
Format minibuffer completion candidates using `window-width` instead of `frame-width`. This way completions render as a single line even if the frame is split.

Before: Empty line between completions
frame-width

After: No empty line between completions
window-width

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## TBD
+### Added
+### Removed
+### Fixed
+### Changed
+- [#2040](https://github.com/org-roam/org-roam/pull/2040) completions: fix completions display-width for Helm users
 
 ## 2.2.0
 ### Added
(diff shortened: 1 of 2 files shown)


jethrokuan:
@stuhlmueller see https://github.com/org-roam/org-roam/issues/1578 for some discussion about `frame-width` vs `window-width`.

The TLDR is that `window-width` seems to only be the optimal value for Helm because of the way completions are presented in Helm. `frame-width` works better for every other completion framework that uses the mini-buffer, which are usually frame-width wide.

stuhlmueller:
Ah, thanks! Any reason not to merge [the workaround](https://github.com/org-roam/org-roam/issues/1578#issuecomment-890486453) mentioned there?

jethrokuan:
> Ah, thanks! Any reason not to merge the workaround mentioned there?

Think it slipped me. Modify this PR to the workaround and I'll merge it?

stuhlmueller:
Done!

jethrokuan (approved):

Issue #2044, opened 2022-01-16
Newline at end of completion candidates for org-roam-node-find


spwhitton:
The completion candidates for `org-roam-node-find` seem to have newline characters on the end of them.  I use `fido-mode` with `icomplete-prospects-height` set to 1, and thus none of the completion candidates show up.  If I set `icomplete-prospects-height` to a higher value, I get something which looks like this: ![roam-candidates](https://user-images.githubusercontent.com/692648/149681416-c5b5c72e-f5e9-4012-8f74-134fd9ab5705.png) Note that I am *not* using `fido-vertical-mode`, so it should not be one completion candidate per line like this.  Thanks.

jethrokuan:
Can you try

```
(setq org-roam-node-display-template "${title}")
```

spwhitton:
This works.  Thanks.  Am I meant to set that in my init, such that there
isn't a bug here?

-- 
Sean Whitton

jethrokuan:
Yes, the default template is optimized towards vertical completion frameworks: we use spacing for aligning different components of the node. This seems to be the majority choice amongst users.

spwhitton:
Cool.  Happy to set it in my init.  But perhaps this could be documented
somewhere?  Thank you again.

-- 
Sean Whitton

spwhitton:
Thanks again!

On Tue 18 Jan 2022 at 03:56PM -08, Jethro Kuan wrote:

> Closed #2044 via #2047.

-- 
Sean Whitton

Issue #1941, opened 2021-11-03
completing-read selection has a trailing new line


memeplex:
### Description

![image](https://user-images.githubusercontent.com/2845433/140177687-d933cc40-1265-4de0-97a2-c857e6892986.png)

#### Steps to Reproduce

1. Create a node
2. Execute org-roam-node-find
3. Type the title of the new note
4. Press Tab

jethrokuan:
Org-roam's default display-template works best for the majority (ivy/helm/vertico users). If you wish to use the default completion framework you can change the node display template:

```emacs-lisp
(setq org-roam-node-display-template "${title}")
```

memeplex:
Ok, thanks!

Issue #1778, opened 2021-08-16
`org-roam-node-find` always creates a new node


tramzee:
### Out of the Box org-roam-node-find always creates new node

After installing org-roam on my emacs 27.1 (ubuntu 20.04) everything seems to work as advertised. The one exception is that any function that attempts to find or create a new org-roam node (e.g. `org-roam-node-find` and `org-roam-node-insert`) ends up *always* creating a new node even when the requested node already exists.

I can get things to work (apparently correctly) if I monkey-patch the `org-roam-node--to-candidate` like this:
````
(defun org-roam-node--to-candidate (node)
    "Return a minibuffer completion candidate given NODE."
    (let ((candidate-main (org-roam-node--format-entry node (1- (frame-width))))
          (tag-str (org-roam--tags-to-str (org-roam-node-tags node))))
      (cons (propertize (concat (string-trim candidate-main)
                                (propertize tag-str 'invisible t))
                        'node node)
            node)))
````
This is the same as the original `org-roam-node--to-candidate` except that I have put a `string-trim` around the `candidate-main` value. `candidate-main` comes back as a string that is padded with spaces out to the width of my emacs frame. So if I have a node called "MyNode", the `candidate-main` value is something like (note the extra spaces after 'MyNode')
````
"MyNode                                                            "
````
where the number of spaces after 'MyNode' is dependent on the width of my emacs frame
This does not match the `"MyNode"` string that is typed in in the minibuffer  'Node:' prompt and therefore creates the new node.

#### Steps to Reproduce

1. Install org-roam from melpa stable
2. Set the org-roam-directory to "~/org/roam"
3. `M-x org-roam-node-find`
4. Type the name of a new node ("MyNode")
5. Enter some text in the new node and save the node
6. `M-x org-roam-node-find`
7. Type (or complete) the name of the new node you just created ("MyNode")
8. Instead of going to the existing node it creates another "MyNode" node (different timestamp on the filename)

#### Expected Results
I expect `org-roam-node-find` to put me into the previously created node

#### Actual Results
I am always presented with a new node

### Environment
- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars)
 of 2020-08-18
- Framework: N/A
- Org: Org mode version 9.4.6 (9.4.6-gab9f2a @ /home/tramsey/.emacs.d/elpa/org-9.4.6/)
- Org-roam: 2.0.0

[...]

Issue #1735, opened 2021-08-03
org-roam-node-find adds whitespace to titles


vallon:
### Description

After upgrading to v2, I found my shortcut to org-roam-find-file was broken because the function disappeared.  I seem to find a replacement in org-roam-node-find.

However, the actual input string that it uses contains a lot of whitespace (seemingly an entire line).

I traced it down to the fact that it is using org-roam-node-display-template to generate its completion strings.  However, it is odd that the completion would be "title   #tag".  So, it seems to be done on purpose, but it is odd.

Or, maybe there is a better alternative to v1's org-roam-find-file than org-roam-node-find.

#### Steps to Reproduce

WIth a node with title "test":  M-x org-roam-node-find tes TAB.  Completion fills in "test{lots of spaces}".

Repl:

```
(org-roam-node-title (car (org-roam-node-list)))
"Channels"
(org-roam-node-title (cdr (car (org-roam-node--completions))))
"Channels"
(substring-no-properties (car (car (org-roam-node--completions))) nil nil
"Channels                                                                                  "
```

#### Backtrace

None

#### Expected Results

Completion (input) by title, not display-format.

#### Actual Results

Completion has lots of noise.

### Environment

Can't cut-paste.  GNU emacs 27.1, org 9.4.6 (9.4.6-g604bfd), org-roam 2.0.0

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/1d03f87 (master @ 8/2 21:37)

Wetlize:
Do you use the default Emacs completion interface? No package like vertico/selectrum/ivy/helm etc? Anyway, you can simply change `org-roam-node-display-template` to what you need (see its docstring). Here some examples:
```emacs-lisp
;; This would only display the title, without any whitespaces
(setq org-roam-node-display-template "${title}")

;; This would display title and tags without of extra whitespace
(setq org-roam-node-display-template "${title} ${tags}")
```

ebwb:
This got in my way when first learning org-roam (I'm new to it, even since v2). When I had two node names, one being a substring of another, I'd have to add a space after the smaller node name in order to get the tab-complete to work. Otherwise, if I only had the substring on an `org-roam-node-insert`, I'd be taken to a fresh node with the same name rather than insert the already existing node.

I use the default Emacs completion interface.

vallon:
My completion-styles is '(basic partial-completion emacs22), which is the default for my emacs.

[...]

PR #2054, opened 2022-01-20
(breaking!)node: simplify default display-template

jethrokuan:
Revert to a simplified `org-roam-node-display-template`, because on
non-vertical completion frameworks it looks and behaves strangely. To
restore the original default behaviour, set
`org-roam-node-display-template` in your Emacs configuration as such:

```
  (setq org-roam-node-display-template
        (concat "${title:*} "
                (propertize "${tags:10}" 'face 'org-tag)))
```

This should reduce confusion for newcomers, at the trouble of existing users,
eliminating issues of this sort:

#1778, #1735, #2044, #1941, #1627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 ## TBD
+### Breaking
+- [#2054](https://github.com/org-roam/org-roam/pull/2054) node: simplify default `org-roam-node-display-template`.
+  This was done so completions work fine by default on all completion systems. To restore the tabular vertical completion interface, set this in your configuration:
+  
+  ```emacs-lisp
+  (setq org-roam-node-display-template
+        (concat "${title:*} "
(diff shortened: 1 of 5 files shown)


hraban:
thanks! I'm eager to try this out.

minad:
@jethrokuan What about providing the alternative formatting templates in the `defcustom` type specification? This helps people who explore the configuration options via the customize interface.

~~~ elisp
:type '(choice (const :tag "Title only" "${title}")
               (const :tag "Title and tags" "${title} ${tags}")
               ...)
~~~

jethrokuan:
@minad makes sense! It's not something I use and it always slips my mind.

minad:
I also don't use it often, but it is a nice service. Btw, you can use `embark-export` from `describe-variable`. Then you will get a customize view of all the variables from completion. This is how I sometimes explore configuration options.

Issue #2039, opened 2022-01-15
org-roam-node-insert: captured headline-based nodes


hrdl-github:
### Brief Abstract
Unless creating a file-based node, `org-roam-node-insert` does not insert nodes, but creates headings in a file that is a node. This means that the inserted ID usually belongs to a heading's ancestor or the file (#1996).

### Long Description
Also discussed here: https://org-roam.discourse.group/t/how-to-set-up-org-roam-capture-templates-to-create-a-node-headline/2282
org-roam defines nodes as files or headlines with an ID property. When linking to a newly-created capture via `org-roam-node-insert` the link inserted is based on `(org-roam-capture--get :id)`, which is set by `org-roam-capture--prepare-buffer`
https://github.com/org-roam/org-roam/blob/6170cc99288e75b8e3f3fd3101b66acc12a9613a/org-roam-capture.el#L475 before the node is created, causing the id to be created for the file or first headline, not the heading to be created. This means that we are not capturing a node.

I've been using a second hook to use the correct id when finalising the capture. Does it make sense to implement this as the default behaviour? Do you see any issues with this approach?

Tested with capture template
```
("t" "thought" entry "* ${title}"
      :target (file "thoughts.org")
      :immediate-finish t)
```

### Proposed Implementation (if any)
```
(defun nom/org-roam-capture-create-id ()
  "Create id for captured note and add it to org-roam-capture-template." 
  (when (and (not org-note-abort)
             (org-roam-capture-p))
    (org-roam-capture--put :id (org-id-get-create))))
(add-hook 'org-capture-prepare-finalize-hook 'nom/org-roam-capture-create-id))
```

### Please check the following:

- [X] No similar feature requests

lennonhill:
This is still relevant even with https://github.com/org-roam/org-roam/commit/4fabea2757824c0571d3dfd0614208b7e69afd2d, as the ID is inserted at `(point-min)` for all target types but `file+datetree`. This makes perfect sense for target `file`, but for all the other types I'd disagree with this decision.

dustinfarris:
I'm not convinced this is fully resolved.

PR #2056, opened 2022-01-20
(feat!)capture: change id creation to headline on entry-type capture-templates

jethrokuan:
Addresses #2039.


jethrokuan:
cc @Wetlize

In my brief testing I found that `org-roam-capture--adjust-point-for-capture-type` wasn't really needed anymore, but preferably you can confirm this. A lot of the weird insertion locations was because `:target-entry-p` in org-capture was set based on the point the `org-roam-capture--prepare-buffer` ended up at, instead of the target type.

lennonhill:
At least some form of adjustment is probably desirable, as we'd end up at (point-min) otherwise. I will look into this as soon as I have time to see if there's a sensible default or whether something like Org's `%?` template element makes more sense.

lroolle:
> ... if there's a sensible default or whether something like Org's `%?` template element makes more sense.

Hi @lennonhill, I've checked out this branch for a couple of days, this feature works fine for me with the `entry` type for a heading node, and `plain` type for a file node, but the entry type for a file node actually not works fine.

I'm not sure if this is the `(point-min)` problem, to make more sense,  we may add another "entry type" with a file node?

lennonhill:
I've been running with this change, which seems to be working fine for `file+head`: https://github.com/lennonhill/org-roam/commit/8ba4c8e00afdace9a6f7cd393c8855585b5de3de. I haven't tested `file` yet.

PR #2060, opened 2022-01-21
[FIX] Added DOUBLE ACUTE ACCENT for unicode normalization

Hrothgar32:
Org-roam usually "normalizes" non-english characters in node titles, such as á becomes a, é becomes e, which is helpful with external programs which are finicky with this.
The double acute accent was left out, so ű, ő wouldn't normalize.


jethrokuan:
@Hrothgar32 thanks! could you add a small note under `changed` in the CHANGELOG?

Hrothgar32:
Yep, I updated it!

PR #2065, opened 2022-01-27
Allow specifying template keys in remaining org-roam-dailies capture/goto commands

meliache:
This is an extension of PR #2028 by @astery, where the `keys` argument was added to `org-roam-dailies--capture` `org-roam-dailies-capture-today`.

I extended this argument to also be available in other built-in dailies functions, as I didn't see any reason why not and it results in my opinion in more consistent behaviour. So far it seems to work for me. But I' wondering why @astery didn't implement that everywhere in the first place, so maybe there is a reason I don't know.

One thing that might be confusing for users is when they start adding the keys to all their dailies functions without reading their respective docstrings, since some have have already other optional arguments, so users might need to add some `nil`'s, e.g.

``` emacs-lisp
      (org-roam-dailies-goto-date nil "<key>")
```

Then, in this case, the key-string will be interpreted as the value for `prefer-future` and since it's non-nil, there won't be an error.

The rationale is the same as in #2028: Allow users to add keybindings (or in my case hydras) to calls of org-roam-dailies functions with specific template keys, thus circumventing the selection screen.

### DIFF
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -139,14 +139,14 @@ In this case, interactive selection will be bypassed."
   (org-roam-dailies--capture (current-time) goto keys))
 
 ;;;###autoload
-(defun org-roam-dailies-goto-today ()
+(defun org-roam-dailies-goto-today (keys)
   "Find the daily-note for today, creating it if necessary."
   (interactive)
-  (org-roam-dailies-capture-today t))
+  (org-roam-dailies-capture-today t keys))
 
 ;;;; Tomorrow
 ;;;###autoload
-(defun org-roam-dailies-capture-tomorrow (n &optional goto)
+(defun org-roam-dailies-capture-tomorrow (n &optional goto keys)
   "Create an entry in the daily-note for tomorrow.
 
 With numeric argument N, use the daily-note N days in the future.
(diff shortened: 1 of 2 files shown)


extensions/org-roam-dailies.el
```diff
@@ -139,55 +139,55 @@ In this case, interactive selection will be bypassed."
   (org-roam-dailies--capture (current-time) goto keys))
 
 ;;;###autoload
-(defun org-roam-dailies-goto-today ()
+(defun org-roam-dailies-goto-today (keys)
```
astery, line 142: I think this argument should be hidden as optional

extensions/org-roam-dailies.el
```diff
@@ -139,55 +139,55 @@ In this case, interactive selection will be bypassed."
...
 
 With a `C-u' prefix or when GOTO is non-nil, go the note without
 creating an entry."
   (interactive "p")
-  (org-roam-dailies--capture (time-add (* n 86400) (current-time)) goto))
+  (org-roam-dailies--capture (time-add (* n 86400) (current-time)) goto keys))
 
 ;;;###autoload
-(defun org-roam-dailies-goto-tomorrow (n)
+(defun org-roam-dailies-goto-tomorrow (n keys)
```
astery, line 160: The same is here

astery:
There was no special reason, I just missed them.

extensions/org-roam-dailies.el
meliache: Fixed in 6c0df349b92cf272216f4c4b4d637b19ce602693

jethrokuan:
Thanks!

PR #2068, opened 2022-01-28
docs: fix little typos

nashamri:
Found these typos while I was reading the manual.


jethrokuan:
@nashamri these aren't typos, I think your changes changed the meaning. the first two examples are for org-cite, and the last is for org-ref.

nashamri:
Oh right! :stuck_out_tongue_winking_eye: my bad, sorry about that.

Issue #2078, opened 2022-02-05
Capture of info's :ref only works if org-roam-protocol is required


jeremyf:
### Description

The hook that adds the `:ref` from `org-roam-capture--info` only does so when you've required `org-roam-protocol`.  I have a use-case for wanting the add ref behavior when I don't have `org-roam-protocol` enabled.

My proposal is to move the hook to `org-roam-capture` and rename accordingly.

#### Steps to Reproduce

Assuming you have one capture template:

```elisp
(org-roam-capture-
     :node (org-roam-node-create :title "Org Roam Homepage")
     :info '(:ref "https://orgroam.com")
     :props '(:immediate-finish nil))
```

#### Expected Results

```gherkin
Given I have not require `org-roam-protocol`
When I call `org-roam-capture-` with `:info '(:ref "https://orgroam.com")`
Then my capture has a `ROAM_REFS: https://orgroam.com` line in the PROPERTIES.
```

#### Actual Results

```gherkin
Given I have not require `org-roam-protocol`
When I call `org-roam-capture-` with `:info '(:ref "https://orgroam.com")`
Then my capture does not have a `ROAM_REFS: https://orgroam.com` line in the PROPERTIES.
```

### Environment

Copy info below this line into issue:
- Emacs: GNU Emacs 28.0.90 (build 1, aarch64-apple-darwin20.6.0, NS appkit-2022.60 Version 11.6 (Build 20G165))
 of 2021-12-07

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/eed1df90f571b19cae1581e7f5d938892b00c24d

jeremyf:
I'm preparing to write a pull request for this.

PR #2079, opened 2022-02-05
Ensuring that :ref info capture in all cases

jeremyf:
Let's assume we're evaluating the following region:

```elisp
(org-roam-capture-
     :node (org-roam-node-create :title "Org Roam Homepage")
     :info '(:ref "https://orgroam.com")
     :props '(:immediate-finish nil))
```

Prior to this commit, if you did not require "org-roam-protocol" then
`org-roam-capture-` would ignore the `:info '(:ref
"https://orgroam.com")` parameter.  Once you required
"org-roam-protocol" then the callback hooks that process `:info`'s
`:ref` will fire regardless of whether it comes from org-roam-protocol.

After this commit, you no longer need to require "org-roam-protocol" to
ensure that you capture an `:info`'s `:ref`.

Why is this useful?  When I'm reading through my elfeed RSS feed, I want
to capture an article.  I could use the org-roam-protocol to do this,
but that seems a bit unnecessary given that I'm already in Emacs.

Closes org-roam/org-roam#2078

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Breaking
 - [#2054](https://github.com/org-roam/org-roam/pull/2054) node: simplify default `org-roam-node-display-template`.
   This was done so completions work fine by default on all completion systems. To restore the tabular vertical completion interface, set this in your configuration:
-  
+
   ```emacs-lisp
   (setq org-roam-node-display-template
         (concat "${title:*} "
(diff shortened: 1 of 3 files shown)


jethrokuan:
Could you also add an entry into the changelog?

jeremyf:
Absolutely!

jeremyf:
@jethrokuan I added a CHANGELOG entry.

Thank you again for writing such a wonderful package.  I love org-roam!

PR #2080, opened 2022-02-06
(fix)dailies: prevent multiple "dailies/" subdirectory expansions

### LINKED ISSUE #2070: org-roam-dailies--daily-note-p fails inside org-roam-capture-new-node-hook
###  `org-roam-dailies--daily-note-p` fails inside `org-roam-capture-new-node-hook`

`org-roam-dailies--daily-note-p` function tests, if current file path is descendant of `(expand-file-name org-roam-dailies-directory org-roam-directory))`

But inside the `org-roam-capture-new-node-hook`, `org-roam-directory` already evaluates to the daily note's path. (e.g. /path/to/org/roam/daily). 

This makes detecting if new note is a daily note or a regular roam note trickier. We should have a sure way to test this.

#### Steps to Reproduce

1. Create a org-roam-capture-new-node-hook` as follows
```elisp
(defun my/capture-new-node-hook ()
"Define a test hook"
(when (org-roam-dailies--daily-note-p)
"Do Cool things"))
```
2. Register this hook
```elisp
 (add-hook! 'org-roam-capture-new-node-hook #'vk/add-auto-props-to-org-roam-dailies)
```
3. Reload emacs, call `org-roam-dailies-capture-today` or `org-roam-dailies-capture-date` to create new daily node.

#### Expected Results

The `when` block should get executed.

#### Actual Results

`when` block does not get executed.

### Environment

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/b6d59e22385dacc36108539ece3e9eb28c5e1042

juergenhoetzel:
When dynamic binding `org-roam-directory` to the "dailies/"
subdirectory also bind `org-roam-dailies-directory` to "./" to prevent
invalid expansions in user-defined hook functions.

Fixes #2070

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [#2065](https://github.com/org-roam/org-roam/pull/2065) dailies: add `keys` argument to the remaining dailies functions `org-roam-dailies-goto-yesterday`/`-today`/`-tomorrow`/`-date` and `org-roam-dailies-capture-yesterday`/`-tomorrow`/`-date` to give the abilty to get into a capture buffer bypassing the selection screen in all dailies commands. Extension of #2055
 ### Removed
 ### Fixed
+- [#2080](https://github.com/org-roam/org-roam/pull/2080) dailies: prevent multiple "dailies/" subdir expansions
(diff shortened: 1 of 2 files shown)


jethrokuan:
interesting xD I was thinking of trying to find a better fix but this should work

PR #2086, opened 2022-02-07
fix #2083 ID should be added to org-id-locations-file after capture

### LINKED ISSUE #2083: New note default ID generation doesn't update org-mode ID database
### Brief Abstract
It would be good if org-roam-capture- could use "org-id-get-create" instead of "org-id-new" so that org-mode's ID database is updated.

### Long Description
I was reviewing the way org-roam generates IDs for new notes and found that the org-roam-capture- function uses "org-id-new" instead of "org-id-get-create":

[https://github.com/org-roam/org-roam/blob/ccfa97ec3a70634bced494149eb85809a6e47cfb/org-roam-capture.el#L411](https://github.com/org-roam/org-roam/blob/ccfa97ec3a70634bced494149eb85809a6e47cfb/org-roam-capture.el#L411)

From what I understand, org-id-new simply returns a new ID:

[https://github.com/tkf/org-mode/blob/master/lisp/org-id.el#L323](https://github.com/tkf/org-mode/blob/master/lisp/org-id.el#L323)

However, org-id-get-create is the one that takes that ID (by first calling org-id-new) and updates org-mode's ID database:

[https://github.com/tkf/org-mode/blob/master/lisp/org-id.el#L227](https://github.com/tkf/org-mode/blob/master/lisp/org-id.el#L227)

It would be useful if org-id-get-create was used instead of just org-id-new so that the org-mode ID database is updated on the spot.

However, this appears to be a change from how part of org-roam used to work. Tracing recent history shows capture did at one time have org-id-get-create before its removal:

[https://github.com/org-roam/org-roam/pull/2049](https://github.com/org-roam/org-roam/pull/2049)

and later changed to its current form:

[https://github.com/org-roam/org-roam/pull/2052](https://github.com/org-roam/org-roam/pull/2052)

nobiot:
Currently, Condition 1 below is met. A fix will need to meet Condition 2 in addition.

1. The new Org ID must be available to other functions during the capture
   process

2. Adding the new Org ID to Org ID file/hash must be done after "Abort" check
   -- i.e., when the user aborts the capture process, the new ID must not be
   added to the Org ID file and hash database

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Removed
 ### Fixed
+- [#2086](https://github.com/org-roam/org-roam/pull/2086) capture: correctly update org-id-locations on capture
 - [#2082](https://github.com/org-roam/org-roam/pull/2082) buffer: don't destroy window if `org-roam-node-toggle` reuses window
 - [#2080](https://github.com/org-roam/org-roam/pull/2080) dailies: prevent multiple "dailies/" subdir expansions
 - [#2055](https://github.com/org-roam/org-roam/pull/2055) dailies: removed stray f require, which was causing require and compilation errors
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -714,7 +714,7 @@ the current value of `point'."
         (when (find-buffer-visiting new-file)
           (kill-buffer (find-buffer-visiting new-file)))
         (delete-file new-file))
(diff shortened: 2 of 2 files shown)


jethrokuan:
Thanks @nobiot , I think this fix looks good. Could you:

1. Add a space between the paranthesized expressions
2. Add an entry into the changelog describing the fix? (like "Correctly update `org-id-locations` on capture")

PR #2088, opened 2022-02-10
docs: Indicate how to start emacs server in manual page.

geomule:
As advised by @nobiot [here](https://org-roam.discourse.group/t/taking-notes-on-webpages-using-org-protocol/2446), the motivation is that the newbie approaching org-roam should not have to read the org-mode documentation to find out that in order to get org-protocol working, they have to enable Emacs server running in the background. I suggest adding this to the [Roam-Protocol](https://www.orgroam.com/manual.html#Org_002droam-Protocol) section.

Maybe reference that the best way to achieve this is to have Emacs run in daemon mode when logging into the user account. [This page](https://www.emacswiki.org/emacs/EmacsAsDaemon) describes how to implement this on the different distributions, including Debian although the method below is likely to work on most Free desktop based GNU/Linux distros.

I achieved this by creating the file `~/.config/autostart/emacs_daemon.desktop` and copying in the following content:

[Desktop Entry]
Exec=emacs --daemon
Icon=
Name=emacs_daemon
Path=
Terminal=False
Type=Application

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -288,29 +288,25 @@ in your Emacs environment as a prerequisite for Org-roam when you install it.
 ~emacsql-sqlite~ requires a C compiler (e.g. ~gcc~ or ~clang~) to be present in
 your computer. How to install a C compiler depends on the OS that you use.
 
-- For Windows:
+**** C Compiler for Windows
 
-There are various ways to install one, depending on how you have installed
-Emacs. If you use Emacs within a Cygwin or MinGW environment, then you should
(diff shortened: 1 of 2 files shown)


jethrokuan:
Don't think you have the right branch here?

jethrokuan:
Ah if you meant to just provide info, what you probably wanted to do was to [create an issue](https://github.com/org-roam/org-roam/issues) instead of a pull-request. A PR is for when you want code merged in.

jethrokuan:
Closed with #2116

PR #2089, opened 2022-02-12
Fix `org-roam-descendant-of-p` bug on Windows

etern:
On Windows, `file-truename` and `directory-file-name` downcase drive label: "C:/" => "c:/", while `expand-file-name` keep the case unchanged.

If `org-roam-directory` use upper case drive label, `org-roam-descendant-of-p` will alway return `nil`.

This make `org-roam` totally unusable!!

From [here](https://github.com/org-roam/org-roam/blob/b163c900b8ec2e3637bc251d9b90421efc771c02/org-roam-compat.el#L64), all the candidates file names have a lowercase drive label.

### DIFF
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -289,7 +289,7 @@ If FILE is not specified, use the current buffer's file-path."
     (save-match-data
       (and
        (org-roam-file-p path)
-       (org-roam-descendant-of-p path directory)))))
+       (file-in-directory-p path directory)))))
 
 ;;;###autoload
 (defun org-roam-dailies-find-directory ()
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -108,14 +108,6 @@ SPEC is a list, as per `dolist'."
       `(dolist-with-progress-reporter ,spec ,msg ,@body)
     `(dolist ,spec ,@body)))
 
-;;; File utilities
-(defun org-roam-descendant-of-p (a b)
-  "Return t if A is descendant of B."
-  (let ((a (file-truename a))
-        (b (file-truename b)))
-    (unless (equal a b)
-      (string-prefix-p b a))))
-
 (defmacro org-roam-with-file (file keep-buf-p &rest body)
   "Execute BODY within FILE.
 If FILE is nil, execute BODY in the current buffer.
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -111,9 +111,10 @@ SPEC is a list, as per `dolist'."
 ;;; File utilities
 (defun org-roam-descendant-of-p (a b)
   "Return t if A is descendant of B."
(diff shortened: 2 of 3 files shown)


jethrokuan:
I don't see the issue here, because both `file-truename` and `expand-file-name` are applied to both strings, so. the casing treatment should be exactly the same (both downcased). This function was lifted out a utilities library that is widely used across Emacs.

@nobiot can you confirm?

nobiot:
Sorry I am not entirely certain what exactly I should be looking at. 

> expand-file-name keep the case unchanged.

This does not seem true on my Windows machine with Emacs 27.2.
See the second example.

> If org-roam-directory use upper case drive label, org-roam-descendant-of-p will alway return nil.

This might be true but I don't know when this "If" condition is met. I use sub-directories work and personal for my main Windows machine for work everyday, and have never had any issue.

``` elisp
;; This returns nil
  (org-roam-descendant-of-p "C:/Users/nobiot/org-roam/personal/subdir/" "~/org-roam/personal/")
```

image

nobiot:
@etern Thank you. I can observe the same behaviour of `expand-file-name` in Emacs version 29.0.50 on my Ubuntu machine -- the first arg (`NAME`) is not downcased. 

When does it become an issue on Windows?
At the bottom of the image, you have an analysis of the code. I have not had this issue in my daily usage. I'm not saying it would never happen but do you experience this?

As far as I could see, the function is used only in `org-roam-file-p`, which is indeed fundamental to Org-roam.

etern:
I guess your `org-roam-directory` has a tilde: "~/your-roam", that will not trigger the bug.

Try `(setq org-roam-directory "C:/your-roam")`, this makes `org-roam-descendant-of-p` 's 1st arg having lower case label (c:/) and 2nd arg having upper case (C:/)

nobiot:
@etern thanks, yes I use tilde as you guess.

@jethrokuan It might be worth guarding against this case, then? Don’t know where is most appropriate; I guess it would be up to you.

jethrokuan:
can you check if `file-in-directory-p` works? maybe the helper function is not needed after all.

nobiot:
hmmm. It seems to work. 
One thing I noticed is that the function uses `file-truename` for both FILE and DIR, which does this:

> If FILENAME is not absolute, first expands it against ‘default-directory’.

So... Essentially:
1. DIR gets downcased
2. FILE does not have to exist -- see below. 
 
If the second point is not an issue (the utility function might well have the same quark, but I don't know), perhaps the built-in function can replace it.

```org
#+begin_src elisp
  (file-in-directory-p "non-existent.file"
                       "C:/Users/nobiot/org-roam")
#+end_src

#+RESULTS:
: t
```

etern:
Yes, tried in my environment, `file-in-directory-p` is good enough, I commited that change.

PRESFIL:
It seems this PR breaks the support of symlinks inside the `org-roam-directory`. ;)

Issue #1870, opened 2021-09-24
Executing `org-roam-extract-subtree` saves the file under the current directory instead of `org-roam-directory`


rdrg109:
#### Steps to Reproduce

The steps shown below were executed with a very minimal `init.el` file. The file is shown below.

```elisp
(setq org-roam-v2-ack t)
(require 'org-roam)
(org-roam-setup)
```

Also, `org-roam-directory` was empty (i.e. there were no files) in order to ensure that nothing is interfering with this issue.

1. Load Emacs

I opened it with

```sh
$ emacs -nw
```

2. Place your cursor in any headline. The location of the file containing the headline should be different than `org-roam-directory`.

I will use this file with name `foobar.org`. Suppose that my cursor is on the `* bar` headline.

```
* foo
* bar
```

3. Execute `org-roam-extract-subtree`.

#### Backtrace

Because this issue doesn't throw an error, there's no backtrace.

#### Expected results

The line containing `* bar` is moved to a node under `org-roam-directory`.

#### Actual results

The line containing `* bar` is not moved to a node under `org-roam-directory. Further details below.

After executing `org-roam-extract-subtree`, the following prompt is shown:

```
Extract node to: ~/org-roam/20210924120224-bar.org
```

Press RET to accept the default prompt for the desired location. Now, the following prompt is shown:

```
Non-existent agenda file ~/e/20210924120224-bar.org . [R]emove from list or [A]bort?
```

If I press `R`, `20210924120224-bar.org` is saved in the current directory instead of `org-roam-directory. If I press `A` instead, `20210924120224.org` is created in the current directory, but it is not saved. In either case (pressing `R` or `A`), the file is not created under `org-roam-directory`. This is definitively a bug because the node is ultimately created in the current directory instead of the path which I was prompt.

### Environment

```
- Emacs: GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4)
 of 2021-03-26
- Framework: N/A
- Org: Org mode version 9.4.4 (release_9.4.4 @ /usr/share/emacs/27.2/lisp/org/)
- Org-roam: 2.1.0
```

dustinfarris:
I am seeing this in a read-only buffer.  when I try to run `org-roam-extract-subtree` it interactively prompts to confirm the file locations, something like `roam/20220102_mything.org` which seems fine.  I hit enter to confirm and get:

```
Buffer is read-only
```

[...]

Issue #2058, opened 2022-01-20
org-roam-refile cannnot create new node


stshine:
### Description

When running `org-roam-refile`, you can only refile to an existing node instead of creating new one.

#### Steps to Reproduce

1. Load Emacs
2. Move cursor to an org heading
3. Run `org-roam-refile`
4. Enter a new node name

#### Expected Results

A new node with the name is created and contents refiled to it.

#### Actual Results

"No match" message displayed in minibuffer.

### Environment

- Emacs: GNU Emacs 27.2 (build 1, x86_64-w64-mingw32)
 of 2021-08-27
- Framework: 
- Org: Org mode version 9.4.4 (release_9.4.4 @ c:/msys64/mingw64/share/emacs/27.2/lisp/org/)
- Org-roam: 2.2.0

- Org-roam commit: https://github.com/jethrokuan/org-roam/commit/6f5d65abd9e707b3fdb10092a9fef3b739e143dd

Hrothgar32:
I'm also interested in this, since I also many times create manually a new node, just to be able to refile to it. As I see in the source code of the function however, this is intentional, why is that?

stshine:
> Is `org-roam-extract-subtree` what you want? That will create a new file-node.

It works nice! However there are still some bugs, e.g. title are not in the correct place.

Delapouite:
Hi.

The link merged PR by jmay appears to have fixed the problem.
If that's not the case don't hesitate to reopen this issue. Thanks.

PR #2091, opened 2022-02-14
Fixed org-roam-promote-entire-buffer structure errors

### LINKED ISSUE #1857: =org-roam-extract-subtree= prompts for a confirmation that is not related to the function itself
### Table of Contents

1.  [Description](#org1ce219c)
    1.  [Prologue](#org395cf56)
    2.  [The problem](#orgebc5c40)
2.  [Steps to reproduce](#orge6678e5)
3.  [Backtrace](#orgb5cb830)
4.  [Expected results](#org133a71b)
5.  [Actual results](#org87dc7a1)
6.  [Environment](#org1bf505f)

<a id="org1ce219c"></a>

### Description

<a id="org395cf56"></a>

#### Prologue

To understand this issue, it is first important to understand how the `org-map-entries` function behaves: Whenever `org-map-entries` is used in a buffer whose associated file doesn't exist yet, the following message is shown

```text
Non-existent agenda file ~/org/foo.org.  [R]emove from list or [A]bort?
```

This happens because `org-map-entries` calls `org-agenda-prepare-buffers` which calls `org-check-agenda-file` and the latter executes `file-exists-p` in the file of the current buffer.

<a id="orgebc5c40"></a>

#### The problem

The current implementation of `org-roam-extract-subtree` calls `org-roam-promote-entire-buffer`, which uses `org-map-entries`, before `save-buffer` is executed. Because of this, `file-exists-p` always yields `nil` causing the prompt shown above to be shown.

<a id="orge6678e5"></a>

### Steps to reproduce

1.  Create a subtree in any Org Mode file
2.  Execute `org-roam-extract-subtree`.

<a id="orgb5cb830"></a>

### Backtrace

The following backtrace was obtained by instrumenting `org-check-agenda-file` and calling `org-roam-extract-subtree`. When `org-check-agenda-file` started being instrumented, the backtrace was obtained by pressing `d` (`edebug-pop-to-backtrace`)

```text
org-check-agenda-file("/home/beep1560/org-roam/2021_09_15_22_31_36_885170999_-05.org")
org-agenda-prepare-buffers(("/home/beep1560/org-roam/2021_09_15_22_31_36_885170999_-05.org"))
org-map-entries((closure ((--mpom . 1) cl-struct-org-roam-node-tags t) nil (if (> (org-outline-level) 1) (progn (org-do-promote)))))
(save-restriction (widen) (goto-char (or --mpom (point))) (org-map-entries #'(lambda nil (if (> (org-outline-level) 1) (progn (org-do-promote))))) (let ((title (nth 4 (org-heading-components))) (tags (nth 5 (org-heading-components)))) (beginning-of-line) (kill-line 1) (org-roam-set-keyword "title" title) (if tags (progn (org-roam-set-keyword "filetags" tags)))))
(save-excursion (save-restriction (widen) (goto-char (or --mpom (point))) (org-map-entries #'(lambda nil (if (> (org-outline-level) 1) (progn (org-do-promote))))) (let ((title (nth 4 (org-heading-components))) (tags (nth 5 (org-heading-components)))) (beginning-of-line) (kill-line 1) (org-roam-set-keyword "title" title) (if tags (progn (org-roam-set-keyword "filetags" tags))))))
(save-excursion (if (markerp --mpom) (progn (set-buffer (marker-buffer --mpom)))) (save-excursion (save-restriction (widen) (goto-char (or --mpom (point))) (org-map-entries #'(lambda nil (if (> (org-outline-level) 1) (progn (org-do-promote))))) (let ((title (nth 4 (org-heading-components))) (tags (nth 5 (org-heading-components)))) (beginning-of-line) (kill-line 1) (org-roam-set-keyword "title" title) (if tags (progn (org-roam-set-keyword "filetags" tags)))))))
(let ((--mpom 1)) (save-excursion (if (markerp --mpom) (progn (set-buffer (marker-buffer --mpom)))) (save-excursion (save-restriction (widen) (goto-char (or --mpom (point))) (org-map-entries #'(lambda nil (if (> (org-outline-level) 1) (progn (org-do-promote))))) (let ((title (nth 4 (org-heading-components))) (tags (nth 5 (org-heading-components)))) (beginning-of-line) (kill-line 1) (org-roam-set-keyword "title" title) (if tags (progn (org-roam-set-keyword "filetags" tags))))))))
org-roam-promote-entire-buffer()
(save-current-buffer (set-buffer (find-file-noselect file-path)) (org-paste-subtree) (org-roam-promote-entire-buffer) (save-buffer))
(let* ((template-info nil) (node (org-roam-node-at-point)) (template (org-roam-format-template (string-trim (org-capture-fill-template org-roam-extract-new-file-path)) #'(lambda (key default-val) (let ((fn (intern key)) (node-fn (intern (concat "org-roam-node-" key))) (ksym (intern (concat ":" key)))) (cond ((fboundp fn) (funcall fn node)) ((fboundp node-fn) (funcall node-fn node)) (t (let ((r (completing-read (format "%s: " key) nil nil nil default-val))) (plist-put template-info ksym r) r))))))) (file-path (read-file-name "Extract node to: " org-roam-directory template nil template))) (if (file-exists-p file-path) (progn (user-error "%s exists. Aborting" file-path))) (org-cut-subtree) (save-buffer) (save-current-buffer (set-buffer (find-file-noselect file-path)) (org-paste-subtree) (org-roam-promote-entire-buffer) (save-buffer)))
(save-excursion (org-back-to-heading-or-point-min t) (if (bobp) (progn (user-error "Already a top-level node"))) (org-id-get-create) (save-buffer) (org-roam-db-update-file) (let* ((template-info nil) (node (org-roam-node-at-point)) (template (org-roam-format-template (string-trim (org-capture-fill-template org-roam-extract-new-file-path)) #'(lambda (key default-val) (let ((fn (intern key)) (node-fn (intern (concat "org-roam-node-" key))) (ksym (intern (concat ":" key)))) (cond ((fboundp fn) (funcall fn node)) ((fboundp node-fn) (funcall node-fn node)) (t (let ((r (completing-read (format "%s: " key) nil nil nil default-val))) (plist-put template-info ksym r) r))))))) (file-path (read-file-name "Extract node to: " org-roam-directory template nil template))) (if (file-exists-p file-path) (progn (user-error "%s exists. Aborting" file-path))) (org-cut-subtree) (save-buffer) (save-current-buffer (set-buffer (find-file-noselect file-path)) (org-paste-subtree) (org-roam-promote-entire-buffer) (save-buffer))))
org-roam-extract-subtree()
funcall-interactively(org-roam-extract-subtree)
call-interactively(org-roam-extract-subtree record nil)
command-execute(org-roam-extract-subtree record)
execute-extended-command(nil "org-roam-extract-subtree" nil)
funcall-interactively(execute-extended-command nil "org-roam-extract-subtree" nil)
call-interactively(execute-extended-command nil nil)
command-execute(execute-extended-command)
```

As you can see there, `org-check-agenda-file` is called in the newly created node. Because the node doesn't have an associated file, the message previously shown is shown.

```text
Non-existent agenda file ~/org-roam/2021_09_15_22_31_36_885170999_-05.org.  [R]emove from list or [A]bort?
```

<a id="org133a71b"></a>

### Expected results

The prompt (`Non-existent agenda file ...`) shouldn't be shown whenever `org-roam-extract-subtree` is called because the fact that the file doesn't exist is always true since a newly created node is inserted in a newly created file.

<a id="org87dc7a1"></a>

### Actual results

The prompt (`Non-existent agenda file ...`) is shown whenever `org-roam-extract-subtree` is called.

<a id="org1bf505f"></a>

### Environment

```text
- Emacs: GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4)
 of 2021-03-26
- Framework: N/A
- Org: Org mode version 9.4.4 (release_9.4.4 @ /usr/share/emacs/27.2/lisp/org/)
- Org-roam: 2.1.0
```

jmay:
Fixes #1857 and #2058 

###### Changes
Run safety checks before executing `org-roam-promote-entire-buffer`. These are unnecessary when it is called from `org-roam-extract-subtree`, but provide some extract protection if invoked directly.

Replaced an invocation of `(org-roam-set-keyword "title" title)` with explicit code, to avoid the situation where `org-roam-set-keyword` steps over a line and injects the `#+title:` line into the middle of body text instead of at the top.

Added a call to `save-buffer` to prevent `org-map-entries` from triggering a confusing & unwanted org-agenda prompt "Non-existent agenda file. [R]emove from list or [A]bort?".


jmay:
This might also fix #1870

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
     (org-roam-erase-keyword "title")
     (org-roam-erase-keyword "filetags")))
 
+(defun org-roam--heading-levels ()
+  "Extract the heading level structure from the current file."
+  (org-map-entries (lambda () (org-current-level))
```
jethrokuan, line 834: we don't want to be using `org-map-entries` here since it introduces dependencies on `org-agenda` related configuration. Use `org-map-region` istead as per line 848 to 854.

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
...
+                   nil 'file))
+
+(defun org-roam--can-promote-p ()
+  "Verify that this buffer is promoteable:
+There is a single level-1 heading
+and no extra content before the first heading."
+  (let ((level-1-headings-count (-count (lambda (level) (eq 1 level)) (org-roam--heading-levels)))
```
jethrokuan, line 841: let's not use anything from `dash` here, we have no explicit dependency on `dash` at the moment, although we do, I think, pull it in transitively.

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
...
+        (is-top-content-empty (org-with-point-at 1
+                                (org-at-heading-p))))
+    (and
+     (eq level-1-headings-count 1)
+     is-top-content-empty)))
+
```
jethrokuan, line 847: no `my/` prefix here.

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
...
-       (when (> (org-outline-level) 1)
-         (org-do-promote)))
-     (point-min) (point-max))
-    (let ((title (nth 4 (org-heading-components)))
-          (tags (nth 5 (org-heading-components))))
-      (beginning-of-line)
-      (kill-line 1)
-      (org-roam-set-keyword "title" title)
-      (when tags (org-roam-set-keyword "filetags" tags)))))
+  (save-buffer)
```
jethrokuan, line 862: as far as possible I want to avoid calling `save-buffer` here, is it necessary?

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
...
+              (tags (nth 5 (org-heading-components))))
+          (my/org-roam--outdent-headings)
+          (kill-whole-line)
+          (org-roam-end-of-meta-data)
+          ;;(org-roam-set-keyword "title" title)
+          ;;(when tags (org-roam-set-keyword "filetags" tags))))
+          (insert "#+title: " title "\n")
+          (when tags (insert "#+filetags: " tags "\n"))
+          (org-roam-db-update-file)))
+    (user-error "Cannot promote: Multiple root headings or there is extra file-level text")))
```
jethrokuan, line 875: Let's error at the beginning in a `when` block instead.

org-roam-node.el
```diff
@@ -829,23 +829,50 @@ Any top level properties drawers are incorporated into the new heading."
...
+      (org-with-point-at 1
+        (let ((title (nth 4 (org-heading-components)))
+              (tags (nth 5 (org-heading-components))))
+          (my/org-roam--outdent-headings)
+          (kill-whole-line)
+          (org-roam-end-of-meta-data)
+          ;;(org-roam-set-keyword "title" title)
+          ;;(when tags (org-roam-set-keyword "filetags" tags))))
+          (insert "#+title: " title "\n")
+          (when tags (insert "#+filetags: " tags "\n"))
```
jethrokuan, line 873: use `org-roam-tags-add` here.

jethrokuan (commented):
I think this PR definitely fixes a bunch of issues existing in `org-roam-extract-subtree` and will make it strictly better, but fundamentally I think the approach of cut-paste-promote is brittle and needs to change. Open to merging after the comments have been addressed, but I'll need some time to rethink the right way to go about extraction.

jmay:
I agree that this warrants a proper rethink. Maybe this should really be treated as a special case of refile.

org-roam-node.el
jmay: I've removed it, but I'm worried that I ran into issues earlier where an unsaved buffer had node ID values in some state that broke things when the buffer was promoted. I can't reproduce the situation now, but I'll keep watching in case I can cause errors again. The `org-roam-db-update-file` is definitely needed to ensure that node state is persisted properly to the database.

PR #2092, opened 2022-02-14
(fix)capture: don't force :immediate-finish

vherrmann:
Currently, `org-roam-capture` forces :immediate-finish to be false, this
pr changes this behavior, so that the value of the template is used.

-------

I personally see no reason for `org-roam-capture` changing the templates's specified behavior and I also need it to work as customized.


nickanderson (approved):
Ah, yes. I was looking to see if there was a bug filed for this. I couldn't figure out what I was doing wrong and why my immediate-finish dailies worked fine, but not my immediate-finish org-roam-capture templates.

I tried this patch and it fixed my issue. Hopefully @jethrokuan will merge it.

nickanderson:
Nudge. Would be nice if this got merged. I keep forgetting I need to re-apply the patch after updating.

nickanderson:
Bump.

hpfr (approved):
`:immediate-finish` defaults to nil, so this seems straightforward. There’s also an approval from a second user daily driving it.

hpfr:
Thanks for your patience!

PR #2098, opened 2022-02-20
(chore): pin eldev install to 10.3

jethrokuan:


jethrokuan:
cc @doublep, this is needed because the the web-install pulls from MELPA stable, so a version bump in Eldev should resolve things.

https://github.com/org-roam/org-roam/runs/5261592640?check_suite_focus=true

doublep:
@jethrokuan: I've just released 0.11, should become available from MELPA in a few hours. Hopefully this bug will be gone, though I don't quite understand it. Ping me again if it is not resolved.

Issue #2102, opened 2022-02-21
org-roam-capture--finalize fails for non-file targets


lennonhill:
### Description
https://github.com/org-roam/org-roam/commit/b163c900b8ec2e3637bc251d9b90421efc771c02 breaks when the captured nodes introduces a new ID, but does not require creating a new file. This can be experienced with some uncommon targets (e.g. file+olp targets) and entries when using the unmerged https://github.com/org-roam/org-roam/pull/2056/commits/b5acacac20d1fd1554d5a6b858015e6686a39a7e .

#### Backtrace
```
Debugger entered--Lisp error: (error "‘org-id-get’ expects a file-visiting buffer")
  signal(error ("‘org-id-get’ expects a file-visiting buffer"))
  (if (file-exists-p (file-truename org-id-locations-file)) (signal (car err) (cdr err)) (or org-id-locations (setq org-id-locations (make-hash-table :test 'equal))) (condition-case _err (progn (org-roam-message (concat "`org-id-locations-file' (%s) doesn't exist. " "Trying to regenerate it (this may take a while)...") org-id-locations-file) (make-directory (file-name-directory (file-truename org-id-locations-file))) (org-roam-update-org-id-locations) (apply fn args)) (file-error (org-roam-message "Failed to regenerate `org-id-locations-file'") (lwarn 'org-roam :error "\n--------\nWARNING: `org-id-locations-file' (%s) do..." org-id-locations-file) (setq org-id-locations-file (expand-file-name ".orgids" (file-truename org-roam-directory))) (apply fn args))))
[...]

PR #2103, opened 2022-02-21
(fix)capture: don't update org-id-locations if file is unknown

lennonhill:
My attempt to address https://github.com/org-roam/org-roam/issues/2102 .
Feedback is welcome.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -716,7 +716,7 @@ the current value of `point'."
           (kill-buffer (find-buffer-visiting new-file)))
         (delete-file new-file))
     (when-let* ((buffer (plist-get org-capture-plist :buffer))
-		(file (buffer-file-name buffer)))
+                (file (buffer-file-name buffer)))
       (org-id-add-location (org-roam-capture--get :id) file))
     (when-let* ((finalize (org-roam-capture--get :finalize))
                 (org-roam-finalize-fn (intern (concat "org-roam-capture--finalize-"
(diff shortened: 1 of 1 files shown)


org-roam-capture.el
```diff
@@ -715,7 +715,10 @@ the current value of `point'."
         (when (find-buffer-visiting new-file)
           (kill-buffer (find-buffer-visiting new-file)))
         (delete-file new-file))
-    (org-id-add-location (org-roam-capture--get :id) (org-roam-capture--get :new-file))
+    (when-let (file (or (org-roam-capture--get :new-file)
```
jethrokuan, line 718: you're missing at least a bracket here, so file is always `nil`, and same below.

lennonhill: `when-let ` and `if-let` use the same SPEC, which "As a special case, interprets a SPEC of the form `(SYMBOL SOMETHING)` like `((SYMBOL SOMETHING))`", according to the documentation if `if-let`.

lennonhill:
I've switched to double parentheses. The other syntax should be functionally equivalent, but it mostly exists for backward compatibility, so I'm happy to adopt the modern one.

org-roam-capture.el
```diff
@@ -715,7 +715,10 @@ the current value of `point'."
         (when (find-buffer-visiting new-file)
           (kill-buffer (find-buffer-visiting new-file)))
         (delete-file new-file))
-    (org-id-add-location (org-roam-capture--get :id) (org-roam-capture--get :new-file))
+    (when-let ((file (or (org-roam-capture--get :new-file)
```
jethrokuan, line 718: It looks like for even for the `file` options, `:buffer` populated correctly, so it may be best to add remove this line.

lennonhill: This is also what I've been experiencing during my brief experiments.

PR #2121, opened 2022-03-12
Adding unique option fro org-roam-backlinks-section

### LINKED ISSUE #2119: Allow for `org-roam-buffer`'s Backlinks section to be unique per source
### Brief Abstract

Allow customization so that Backlinks portion of `org-roam-buffer` can be unique per source.

### Long Description

Part of my writing process is to include the same reference multiple times in an `org-roam-node`.  (For example, when I reference a person in my dailies they might have multiple references within the same day).  When I look at the backlinks I see duplicate source titles.  This can lead to a huge backlink buffer.

The following query appears to produce a desired effect for what I'm after:

```elisp
(org-roam-db-query
		    [:select :distinct [source dest (funcall min pos) (funcall min properties)]
		     :from links
		     :where (= dest $s1)
		     :and (= type "id")
		     :group :by source]
		    (org-roam-node-id node))
```

A concession of the query is to use the "first" `pos`.  I'm uncertain the best approach for `properties` but went with something similar.

I see two routes to proceed:

1. Create a new `org-roam-backlinks-unique-source-section` function2. 
2. Create a variable (`(defvar org-roam-backlinks-section-limit-to-unique-sources)`? then add a switch in `org-roam-backlinks-get`

My preference would be to have the unique customizable variable.

Also, I'm prepared to create the solution as I think I have all the necessary context to proceed once I have guidance on an approach.

- [X] No similar feature requests; though it's almost close to https://github.com/org-roam/org-roam/issues/2029

jeremyf:
Prior to this commit, when we would render backlinks in the
`org-buffer`, we would render each reference.  This meant that if a
source file referenced the node 5 times, there would be 5 backlink
references in the `org-buffer` (one for each position of the reference).

With this change, we add a parameter that specifies that backlink
sources should be unique.  In the above example, that would mean instead
of 5 backlinks we'd see 1.  And, as a concession, we'd use the lowest
position (e.g. the first reference in the source).

Closes #2119

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -459,14 +459,23 @@ headline, up to the next headline."
         (org-roam-populate (org-roam-backlink-target-node backlink)))
   backlink)
 
-(defun org-roam-backlinks-get (node)
-  "Return the backlinks for NODE."
-  (let ((backlinks (org-roam-db-query
-                    [:select [source dest pos properties]
-                     :from links
-                     :where (= dest $s1)
-                     :and (= type "id")]
-                    (org-roam-node-id node))))
+(cl-defun org-roam-backlinks-get (node &key unique)
+  "Return the backlinks for NODE.
+
+ When UNIQUE is nil, show all positions where references are found.
+ When UNIQUE is t, limit to unique sources."
+  (let* ((sql (if unique
+                  [:select :distinct [source dest pos properties]
+                   :from links
+                   :where (= dest $s1)
+                   :and (= type "id")
+                   :group :by source
+                   :having (funcall min pos)]
+                [:select [source dest pos properties]
+                 :from links
+                 :where (= dest $s1)
+                 :and (= type "id")]))
+         (backlinks (org-roam-db-query sql (org-roam-node-id node))))
     (cl-loop for backlink in backlinks
              collect (pcase-let ((`(,source-id ,dest-id ,pos ,properties) backlink))
                        (org-roam-populate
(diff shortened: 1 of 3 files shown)


org-roam-mode.el
```diff
@@ -459,14 +459,22 @@ headline, up to the next headline."
...
-                     :where (= dest $s1)
-                     :and (= type "id")]
-                    (org-roam-node-id node))))
+(cl-defun org-roam-backlinks-get (node &key unique)
+  "Return the backlinks for NODE.
+
+ When UNIQUE is nil, show all positions where references are found.
+ When UNIQUE is t, limit to unique sources."
+  (let* ((sql (if unique
+                  [:select :distinct [source dest (funcall min pos) (funcall min properties)]
```
jethrokuan, line 468: I'm a bit worried `min` will cause the pos and properties to not match. Maybe `first` would be a better choice.

jeremyf: Good call, amending!

jeremyf:
Hold on merging, I think I need "first_value" (testing locally again)

jeremyf:
This is in "ready to merge" status

org-roam-mode.el
```diff
@@ -459,14 +459,23 @@ headline, up to the next headline."
...
+ When UNIQUE is nil, show all positions where references are found.
+ When UNIQUE is t, limit to unique sources."
+  (let* ((sql (if unique
+                  [:select :distinct [source dest pos properties]
+                   :from links
+                   :where (= dest $s1)
+                   :and (= type "id")
+                   :group :by source
+                   :having (funcall min pos)]
+                [:select [source dest pos  properties]
```
jethrokuan, line 474: nit: could you rid of the additional space here between pos and properties

org-roam-mode.el
```diff
@@ -482,9 +491,12 @@ Sorts by title."
...
            (org-roam-node-title (org-roam-backlink-source-node b))))
 
-(defun org-roam-backlinks-section (node)
-  "The backlinks section for NODE."
-  (when-let ((backlinks (seq-sort #'org-roam-backlinks-sort (org-roam-backlinks-get node))))
+(cl-defun org-roam-backlinks-section (node &key (unique nil))
+  "The backlinks section for NODE.
+
+When UNIQUE is nil, show all positions where references are found.
+  When UNIQUE is t, limit to unique sources."
```
jethrokuan, line 498: and the space here at start of docstring.

org-roam-mode.el
jeremyf: Absolutely!

PR #2122, opened 2022-03-12
Exposing defcustom for uniqueness of backlinks

jeremyf:
This builds on #2121 by providing a `defcustom` variable.  I include it
separately as I'm uncertain if this is a direction @jethrokuan wishes to
go with this feature.  (If it is not a direction to go, then please
close the PR).

This provides, perhaps, a cleaner customization experience.  For myself,
I prefer the uniformity of the list of functions (as below) comapred to
a list that is a mix of functions and lambdas.

```elisp
(setq org-roam-mode-section-functions
	(list
	 #'org-roam-backlinks-section
	 #'org-roam-reflinks-section))
```

In addition, I haven't delved into writing `defcustom` declarations so
this narrow scope commit helps me further my understanding of Emacs.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -657,6 +657,16 @@ Or, if you want to render unique sources for backlinks (and also keep rendering 
          #'org-roam-reflinks-section))
 #+end_src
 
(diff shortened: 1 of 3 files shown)


jethrokuan:
I'll write the pcase version, which should be cleaner and more generalizable.

PR #2123, opened 2022-03-13
(feat)buffer: pass args to org-mode-section-functions

jethrokuan:

Update `org-mode-section-functions` to accept a list, that passes the arguments into the section rendering function. cc @jeremyf


jeremyf:
Thank you for pinging me.  This is very helpful (`pcase` paired with <kbd>`</kbd> and <kbd>,</kbd> continue to be floating just at the limit of my comprehension of emacs-lisp.  I'm reading this and learning.

PR #2125, opened 2022-03-14
(feat): add org-roam-post-node-insert-hook

lafirest:
some times the node title and the tag are the same, so I added this callback to automatically add the label every time when I insert a node

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -123,9 +123,8 @@ All Org files, at any level of nesting, are considered part of the Org-roam."
   :type 'hook)
 
 (defcustom org-roam-post-node-insert-hook nil
-  "Hook run when an Org-roam node is insert.
-The first parameter is the id of the node's link,
-The second parameter is the link's description"
+  "Hook run when an Org-roam node is inserted as an Org link.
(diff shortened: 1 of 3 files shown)


org-roam.el
```diff
@@ -122,6 +122,13 @@ All Org files, at any level of nesting, are considered part of the Org-roam."
   :group 'org-roam
   :type 'hook)
 
+(defcustom org-roam-post-node-insert-hook '()
```
jethrokuan, line 125:
```suggestion
(defcustom org-roam-post-node-insert-hook nil
```

org-roam.el
```diff
@@ -122,6 +122,13 @@ All Org files, at any level of nesting, are considered part of the Org-roam."
...
+  "Hook run when an Org-roam node is insert.
+The first parameter is the id of the node's link,
+The second parameter is the link's description"
```
jethrokuan, line 128:
```suggestion
  "Hook run when an Org-roam node is inserted as an Org link.
Each function takes two arguments: the id of the node, and the link description."
```

PR #2130, opened 2022-03-18
(fix) unlinked-references: search symlinked directories

rdparker:
Org-roam finds nodes across symlinked directories.  The `org-roam-unlinked-references-section` function does not.

This addresses that by adding the `-L` option to tell ripgrep to follow directory symlinks while looking for unlinked references. which matches what `org-roam--list-files-rg` does.


jethrokuan:
Thanks!

Issue #2131, opened 2022-03-20
Make Org Export/Publish work out of the box


salutis:
I decided to try Org Roam again and found that everything works "out of the box" except exporting/publishing.  I found [1] as a suggested workaround and it solved the problem.  So, both Org Export and Org Publish works now.  Then, I do not understand the logic.  Do we expect everyone to find this (rather large) procedure and add it to their configuration file?  If so, then why not bundle it with the Org Roam package and call it `org-roam-export-insinuate` or something similar?

What do I miss?

[1]:

```emacs-lisp
(defun org-html--reference (datum info &optional named-only)
  (let* ((type (org-element-type datum))
	 (user-label
	  (org-element-property
	   (pcase type
	     ((or `headline `inlinetask) :CUSTOM_ID)
	     ((or `radio-target `target) :value)
	     (_ :name))
	   datum))
         (user-label
          (or user-label
              (when-let ((path (org-element-property :ID datum)))
                (concat "ID-" path)))))
    (cond
     ((and user-label
	   (or (plist-get info :html-prefer-user-labels)
	       (memq type '(headline inlinetask))))
      user-label)
     ((and named-only
	   (not (memq type '(headline inlinetask radio-target target)))
	   (not user-label))
      nil)
     (t
      (org-export-get-reference datum info)))))
```

_edit: wording and syntax highlighting_

jethrokuan:
Addressed with #2138.

salutis:
@jethrokuan Thank you!

PR #2138, opened 2022-03-27
(feat)export: init

jethrokuan:

Initialize a new module `org-roam-publish` that can be `require`d to patch `org-export` to support publishing Org-roam files.

Addresses #2131.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 ## TBD
 ### Breaking
 ### Added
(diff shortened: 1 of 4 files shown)


igorcafe:
it would be very nice if you could provide an example... im seeking for this for a long time but i have no idea on what to do after requiring it.. i can just ruim org-publish-all and it will work properly with roam?

igorcafe:
@jethrokuan

Issue #2134, opened 2022-03-26
org-roam-dailies-goto-today can not go to today's note directly


ZucchiniY:
### Description

if `org-roam-dailies-capture-template` has `%^G`, I must input the tags first and then goto today's note when I use this command `org-roam-dailies-goto-today`.

#### Steps to Reproduce

1. the template as the following

``` emacs-lisp
  (setq  org-roam-dailies-capture-templates
        '(("d" "Daily" entry
           "* %? %^G"
           :target (file+head "%<%Y-%m-%d>.org"
                              "#+title: %<%Y-%m-%d>\n")
           :empty-lines 1
           :unnarrowed t)
           ))
```

2. I use the command `org-roam-dailies-goto-today`, it will go to the *CAPTURE* page first
3. after input the tags, it goes to today's note

#### Expected Results

it will go to today's note immeditely after I use the command `org-roam-dailies-goto-today`.

#### Actual Results

it will go to the *CAPTURE* page first. and then i have to input tags, then it goes to today's note.

### Environment

- Emacs: GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95))
 of 2021-11-19
- Framework: N/A
- Org: Org mode version 9.5.2 (9.5.2-gfbff08 @ /Users/dylan/.emacs.d/elpa/org-9.5.2/)
- Org-roam: 2.2.1

swflint:
This appears to hold even for a template of `"* %?"` as well.

ZucchiniY:
It can go to ensure file if I use same template in org-capture.

I guess it may happened when I use `org-roam-capture--prepare-buffer` method.

hpfr:
This appears to still be an issue and has an open PR with numerous requests for merge.

Wheest:
Also experiencing this issue, org-roam version f4ba41c.

Frestein:
> Also experiencing this issue, org-roam version [f4ba41c](https://github.com/org-roam/org-roam/commit/f4ba41cf3d59084e182a5186d432afc9aa3fc423).

Try this https://github.com/Frestein/dotfiles/commit/6ffc39aa82f8d2e3f37bd6c072d30ed0082b788a

hpfr:
Removing milestone since this is more nuanced than I thought as described in https://github.com/org-roam/org-roam/pull/2141#issuecomment-3021291947.

[...]

PR #2141, opened 2022-03-29
(fix) goto-today: can not goto today note directly(#2134)

ZucchiniY:

Fixed: #2134 

###### Change

There is no need to call `org-roam-capture--fill-template` method when I use this command `org-roam-dailies-goto-today`. So comment `org-roam-capture--prepare-buffer` out as the following:

``` emacs-lisp
(let ((template (org-capture-get :template)))
      (when (stringp template)
        (org-capture-put
         :template
         (org-roam-capture--fill-template template))))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Removed
 ### Fixed
 - [#2130](https://github.com/org-roam/org-roam/pull/2130) buffer: unlinked-references section now also searches within symlinked directories
+- [#2134](https://github.com/org-roam/org-roam/issues/2134) goto: fix `org-roam-dailies-goto-today` can not goto today's note.
 
 ### Changed
 
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -469,11 +469,11 @@ capture target."
   (let ((id (cond ((run-hook-with-args-until-success 'org-roam-capture-preface-hook))
                   (t (org-roam-capture--setup-target-location)))))
     (org-roam-capture--adjust-point-for-capture-type)
-    (let ((template (org-capture-get :template)))
-      (when (stringp template)
-        (org-capture-put
-         :template
-         (org-roam-capture--fill-template template))))
+    ;; (let ((template (org-capture-get :template)))
+    ;;   (when (stringp template)
+    ;;     (org-capture-put
+    ;;      :template
+    ;;      (org-roam-capture--fill-template template))))
     (org-roam-capture--put :id id)
     (org-roam-capture--put :finalize (or (org-capture-get :finalize)
                                          (org-roam-capture--get :finalize)))))


dangom:
Any plans to have this merged?

dangom:
You can work around it by doing the following:

```elisp
(setq org-roam-dailies-capture-templates
        '(("a" "direct" plain "%?" :if-new (file+datetree "%<%Y>.org" day)))

And then on your use-package declaration you bind keys as in
("C-c t" . (lambda () (interactive) (org-roam-dailies-goto-today "a")))
```

ZucchiniY:
> How can we get a maintainer to review and merge this?

you can clone https://github.com/ZucchiniY/org-roam.git to use this feature.

cyb3rm4x1:
I'm not sure the best way to get the maintainer's attention to merge, will @jethrokuan help?

cyb3rm4x1:
Thank you Daniel for the workaround. I'm still wondering the process of
getting fixes merged - is there anything special about org-roam that we
don't want to merge fixes like this?

natecox:
I just came across this looking for why the capture dialog was coming up when I tried to goto today.

I've implemented the lambda workaround above, but adding another voice to the requests to get this merged.

hpfr:
Sorry for the extended wait, folks. Those of you who are asking here, have you tried this PR locally? It looks to me like this will disable org-roam-capture template preprocessing for any org-roam-based capturing, not just `org-roam-dailies-goto-today`.

It seems like two issues are being discussed here.

1. PR author doesn’t want to be prompted by capture template for input when using goto functions.

   I think this is our bug to fix since `org-roam-capture--fill-template` calls the internal `org-capture-fill-template` without considering goto beforehand. As mentioned, I think this PR will break all capture template prompting, though. I think the ideal fix for this would be to only process the org-roam specific parts of the template before handing off to org-capture.

1. Commenters don’t want to be prompted to choose from multiple templates when using goto functions.

   This is expected behavior defined by `org-capture`. From the docstring, emphasis mine:
   > When called interactively with a C-u prefix argument GOTO, don't capture anything, just go to the file/headline where the *selected template* stores its notes.

   Given the context of daily notes, presumably those of you with multiple templates have the exact same `:target` defined for all of them, so you don’t find the template prompt useful. My main concern is that there are other users out there who have multiple `org-roam-dailies-capture-templates` with different target files (e.g. a template targeting `%<%Y-%m-%d>-food.org` for meal tracking and another targeting `%<%Y-%m-%d>.org` for general notes)). I don’t think multi-file workflows were intended but they are possible. We should probably add a warning if the daily templates have the wrong target configured and see if users show up to complain. In the meantime, this advice might be more convenient than individual wrappers like the one in https://github.com/org-roam/org-roam/pull/2141#issuecomment-1209367970:
   ```elisp
   (defun my/org-roam-dailies--capture-goto-quick (fn time &optional goto keys)
     "Skip prompt for capture template with dailies goto functions."
     (apply fn time goto (if goto (caar org-roam-dailies-capture-templates) keys)))

   (advice-add #'org-roam-dailies--capture :around #'my/org-roam-dailies--capture-goto-quick)
   ```

Frestein:
> ```emacs-lisp
> (defun my/org-roam-dailies--capture-goto-quick (fn time &optional goto keys)
>   "Skip prompt for capture template with dailies goto functions."
>   (apply fn time goto (if goto (caar org-roam-dailies-capture-templates) keys)))
> 
> (advice-add #'org-roam-dailies--capture :around #'my/org-roam-dailies--capture-goto-quick)
> ```

Thank you! 

This worked for me:

```elisp
(defun my/org-roam-dailies--capture-goto-quick (fn time &optional goto keys)
  "Skip prompt for capture template with dailies goto functions."
  (apply fn time goto
    (if goto
      (list (car (car org-roam-dailies-capture-templates)))
    (list keys))))

(advice-add #'org-roam-dailies--capture :around #'my/org-roam-dailies--capture-goto-quick)
```

PR #2152, opened 2022-04-09
(fix) org-roam-preview-default-function doesn't copy next node

### LINKED ISSUE #2151: `org-roam-preview-default-function` copies next heading if no content under current heading
### Description
When there is no content after the end of meta-data under a heading the `org-roam-preview-default-function` function gets the next heading

#### Steps to Reproduce
1. Create two consecutive headline nodes with no content under it
2. Link to the node first node created above
3. run `org-roam-buffer` on the node which links to the first node created in step 1.

#### Expected Results
The roam-buffer should only show the first created node with no content

#### Actual Results
The roam-buffer shows the first created node with the second created node as it's content

### Environment

- Emacs: GNU Emacs 27.2 (build 1, x86_64-w64-mingw32)
 of 2021-03-26
- Framework: N/A
- Org: Org mode version 9.5.2 (9.5.2-ge82229 @ c:/Users/amsha/.emacs.d/straight/build/org/)
- Org-roam: v2.2.1-4-gfee008c

ahmed-shariff:
Fix #2151 
Using save-excursion in `org-roam-preview-default-function` to avoid copying next heading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Removed
 ### Fixed
 - [#2130](https://github.com/org-roam/org-roam/pull/2130) buffer: unlinked-references section now also searches within symlinked directories
+- [#2152](https://github.com/org-roam/org-roam/pull/2152) org-roam-preview-default-function: doesn't copy copy content of next heading node when current node's content is empty
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


jethrokuan:
Fantastic, thanks!

Issue #2163, opened 2022-04-18
`org-roam-file-p` with no `file` argument, local variable `path` to be `nil`


tommyjiang:
Thanks for the great mode.

After recent upgrade https://github.com/org-roam/org-roam/commit/61a544cebd45c9f4823440ffe2c32c0686b19f63, when I open epub file with `nov.el`, the following backtrace would be triggered:

```
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  file-truename(nil)
  org-roam-descendant-of-p(nil "~/org/org-roam")
  org-roam-file-p()
  org-roam-db-autosync--setup-file-h()
  run-hooks(find-file-hook)
  after-find-file(nil t)
```

I suppose this is caused by the local variable `path` in `org-roam-file-p` function to be `nil`. I revise this line https://github.com/org-roam/org-roam/commit/61a544cebd45c9f4823440ffe2c32c0686b19f63#diff-2f1e3beb58c04fd0b840c2761afd33f1fddc1b6ef25860d52af87341c3392b51R200 by my self with this:

```emacs-lisp
(org-roam-dir-p (if path (org-roam-descendant-of-p path org-roam-directory)))
```

Maybe it is related to `nov.el` when opening epub files, just FYI.

tommyjiang:
Thanks for the quick fix.

PR #2169, opened 2022-04-23
(fix) `org-roam-file-p` handle opening a buffer with no path

c1-g:

Errors when opening a .epub file with nov.el (a package
that renders EPUB file) because nov.el doesn't store the filename
in `buffer-file-name`, opening it caused errors with `org-roam-file-p`.

The error is documented in #2163

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -192,29 +192,30 @@ FILE is an Org-roam file if:
 - It's located somewhere under `org-roam-directory'
 - It has a matching file extension (`org-roam-file-extensions')
 - It doesn't match excluded regexp (`org-roam-file-exclude-regexp')"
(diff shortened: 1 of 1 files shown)


jethrokuan:
Good catch, but I think it's better to just return `nil` when `path` is nil. Like
```
if not path:
  nil
else:
  rest of logic assuming path is not nil
```

c1-g:
With the new commit, I've done exactly as you said. Now, it will just return `nil` if the `buffer-file-name` is `nil.`

jethrokuan:
Thanks!

Gleek:
@jethrokuan This seems to be reverted in  #2171. Was it intended? I'm facing the same issue in the latest version of org-roam

jethrokuan:
urgh my bad

PR #2177, opened 2022-04-28
support custom minibuffer matching function

cuttlefisch:

Added support for custom `completion-read` predicate to the
`org-rom-node-find` function.

The `org-roam-node-find` function relies on `org-roam-node-read`,
which already provides the `require-match` argument to support this
behavior.

This PR adds an optional argument to `org-roam-node-find` which is
passed to the embedded `org-roam-node-read` call in order to support
custom `completion-read` predicates when searching for nodes in the
minibuffer.

# Example Usage

Case-insensitive search in the minibuffer:

```elisp
(defun case-insensitive-matchp (nodes string)
  "Check if the interactive prompt contains a case insensitive
substring of an Org Roam node title."
  (interactive)
  (lambda (needle haystack) (cl-search (downcase needle) (downcase haystack))) nodes)

(defun my-org-roam-node-find ()
  "Wrapper to run case-insensitive search for Org Roam node
titles."
  (interactive)
  (org-roam-node-find nil nil nil #'case-insensitive-matchp))
```

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -477,7 +477,7 @@ interactive calls FORCE always set to t."
                       force))
 
 ;;;###autoload
-(cl-defun org-roam-node-find (&optional other-window initial-input filter-fn &key templates)
+(cl-defun org-roam-node-find (&optional other-window initial-input filter-fn pred &key templates)
   "Find and open an Org-roam node by its title or alias.
 INITIAL-INPUT is the initial input for the prompt.
 FILTER-FN is a function to filter out nodes: it takes an `org-roam-node',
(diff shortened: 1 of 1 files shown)


russmatney:
I tried to make this work today, but I don't understand how this could ever work - it seems to pass the `pred` into `org-roam-node-read` as the `sort-fn`? Maybe a merge changed those args up? I can't figure it out. Just wanted to call it out, as it was confusing and might be worth cleaning up.

I tried to make `pred` work as an additional `org-roam-node-find` arg (maybe those should switch to `&key` args at some point?), but I don't understand `completing-read` well enough... ultimately I found [this vertico documentation](https://github.com/minad/vertico#completion-styles-and-tab-completion). The very last code block in that section offers:

``` elisp
(setq read-file-name-completion-ignore-case t
      read-buffer-completion-ignore-case t
      completion-ignore-case t)
```

I was able to get case-insensitive completion by wrapping `org-roam-node-find` and `org-roam-node-insert` with a let, like:

``` elisp
;;;###autoload
(defun russ/org-roam-find-file ()
  ""
  (interactive)
  (let ((completion-ignore-case t))
    (org-roam-node-find
     nil nil (lambda (node)
               (= 0 (org-roam-node-level node))))))

;;;###autoload
(defun russ/org-roam-insert-file ()
  ""
  (interactive)
  (let ((completion-ignore-case t))
    (org-roam-node-insert
     (lambda (node)
       (= 0 (org-roam-node-level node))))))
```

Not sure if that applies to all completion frameworks though - it seems worth supporting a passed completion-predicate like this PR intended.

PR #2178, opened 2022-04-29
(fix)org-roam-file-p: don't exclude org-roam-directory

### LINKED ISSUE #2165: org-roam-node-find or org-roam-node-insert  not showing nodes
After running `org-roam-db-sync` on emacs (29.0.50), I don't see nodes anymore with "org-roam-node-find" command, trying to find or insert nodes just shows the history, when I select history node it prompts me to select the roam template - as if I am creating a new node. I installed emacs 28.1, re-installed all packages, that did not solve the problem. Using org-roam version 2.2.1.

The output of `org-roam-db-diagnose-node`

```
#s(org-roam-node nil nil nil nil nil "9e8a966b-4220-4b54-b29d-ba5317c8ac7e" nil 129 nil nil nil nil nil nil nil nil nil nil) [2 times]

```
I tried deleting the 'org-roam.db' database and ran the `org-roam-db-sync` command, but this did not work either.

toregilhk:
Don't exlude the org-roam-directory even if it matches
org-roam-file-exclude-regexp.

Should fix #2165.

This makes `org-roam-file-p` compare `org-roam-file-exclude-regexp` only with files and subdirectories within  `org-roam-directory`, not the entire path.

When someone has an `org-roam-directory` that contains a (sub)string that is a part of `org-roam-file-exclude-regexp`, the entire org-roam directory is excluded, and all files removed from the database.

For example, when `org-roam-directory` is "~/data/org-roam/" and `org-roam-file-exclude-regexp` is "data/", the entire roam directory is exluded, and no files are found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Added
 ### Removed
 ### Fixed
+- [#2165](https://github.com/org-roam/org-roam/pull/2165) (fix)org-roam-file-p: don't exclude org-roam-directory
 - [#2168](https://github.com/org-roam/org-roam/pull/2168) (perf)node-read: filter nodes before mapping --to-candidate
 ### Changed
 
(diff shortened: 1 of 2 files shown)


toregilhk:
When I'm looking at it again, I think it would probably be better to have `(file-relative-name path org-roam-directory)` defined only one place, instead of having it being recalculated each time the dolist loops through.

jethrokuan:
Thanks!

Issue #2192, opened 2022-05-12
Respect blank lines in capture templates


salutis:
Given `template.org`

```
#+title: ${title}

%?

* References
:PROPERTIES:
:UNNUMBERED: t
:END:
#+PRINT_BIBLIOGRAPHY:
```

and

```elisp
(setq org-roam-capture-templates
        '(("d" "default" plain (file "~/org/template.org")
           :target (file "${slug}.org")
           :empty-lines-before 1
           :unnarrowed t)))
```

I expect

```
:PROPERTIES:
:ID: 163005CB-0AA5-4AD4-AFBD-7E10272DF748
:END:

#+title: Example

|

* References
:PROPERTIES:
:UNNUMBERED: t
:END:
#+PRINT_BIBLIOGRAPHY:
```

but I get

```
:PROPERTIES:
:ID: 163005CB-0AA5-4AD4-AFBD-7E10272DF748
:END:

#+title: Example
|
* References
:PROPERTIES:
:UNNUMBERED: t
:END:
#+PRINT_BIBLIOGRAPHY:
```

Any tips how I can make Org Roam respect the template?  I want to avoid typing `RET RET C-p` every time I create a new node.

salutis:
I tracked down the problem down to the following line in `org-roam-capture.el`:

```elisp
(setq template (replace-regexp-in-string "\n$" "" (org-capture-fill-template template)))
```

It seems to remove _all_ empty lines from the template the user defined.

salutis:
@alopezrivera 

I have just noticed that I do _not_ need to adjust new notes anymore.  Thank you!

PR #2203, opened 2022-05-24
Solved issue #2192 (Respect blank lines in capture templates)

alopezrivera:
Issue: https://github.com/org-roam/org-roam/issues/2192

Unwanted removal of `\n` characters in Org Roam template strings.

-----

`org-capture-fill-template` fills the template string, but it removes whitespace and newlines at the beginning and end of the template.*

`org-roam-capture--fill-template` removes all blank lines and restores them after calling `org-capture-fill-template`.

The regex formula used for this however (`\n$`) eliminates all blank lines in the template, whether or not they are at the beginning and end of the template.

By replacing it with `[\n]*\\'` only the blank lines at the end of the template will be removed -only the ones which will be later restored.

###### Example

With a default template as follows:

```
      '(("d" "default" plain "%?"
	    :target (file+head "%<%Y%m%d%H%M%S>-${slug}.org"
			       "#+STARTUP: overview\n\n\n\n#+title:${title}\n\n\n")
	    :unnarrowed t)))
```

`org-roam-capture--fill-template` would previously output

```
"#+STARTUP: overview#+title:${title}\n\n\n"
```

while with the new formula it outputs (the user's input as would be expected)

```
"#+STARTUP: overview\n\n\n#+title:${title}\n\n\n"
```

-----
*: paraphrased from the comment in `org-roam-capture--fill-template`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   build:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ on:
   pull_request:
   push:
     branches:
-      - main
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


jethrokuan:
if you don't mind could you add a test case for your template so it doesn't regress

alopezrivera:
Just added the test case. Changed the CI workflow so it'll run on pushes to "main", as that's the default branch in my fork.

alopezrivera:
Just reverted the CI trigger change (so runs are again triggered by pushes to "master")

PR #2215, opened 2022-06-08
(db)fix: FOREIGN KEY error in narrowed buffer

psii:
When trying to save from a narrowed buffer, only the Roam nodes visible
in the narrowed buffer are written to the database.

This commit ensures that org-roam behaves the same whether the current
buffer is narrowed or not.


jethrokuan (approved):

jethrokuan:
great catch!

PR #2219, opened 2022-06-09
[Fix #2188] Insert Org ID links in org-roam-completion

### LINKED ISSUE #2188: Insert Org ID links via CAPF
When I request a link via `org-roam-node-insert`, Org Roam inserts a standard Org ID link, as expected.  However, when I request a link via CAPF (with Corfu), Org Roam inserts a `roam:` link instead of a standard ID link, and I have to save, which takes 10 seconds with a relatively small file (249k), to resolve the new link(s).  Why does Org Roam not insert a fully baked Org ID link when using CAPF, like `org-roam-node-insert` does?

leinfink:
Updates `org-roam-complete-link-at-point` and
`org-roam-complete-everywhere` to create Org ID links instead of the
outdated "roam:" links. The functions now complete for node title or alias and then link with the respective ID and the returned completion string as the link description.

For example: `[[Hou|]]` can complete into `[[id:01234][Houses]]`.

If there is already an "id:" link at point, completion works inside the description of it and then updates the ID accordingly.

-------

As noted in #2188, this seemed to be a remnant of org-roam v1. Org ID links are now the default, so the completion options should reflect that.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -761,26 +761,34 @@ Assumes that the cursor was put where the link is."
 
 ;;;;;; Completion-at-point interface
 (defconst org-roam-bracket-completion-re
-  "\\[\\[\\(\\(?:roam:\\)?\\)\\([^z-a]*?\\)]]"
+  "\\[\\(?:\\[id:\\([^z-a]*?\\)]\\)?\\[\\([^z-a]+?\\)]]"
   "Regex for completion within link brackets.
-We use this as a substitute for `org-link-bracket-re', because
-`org-link-bracket-re' requires content within the brackets for a match.")
+Matches both empty links (i.e. \"[[|]]\") and existing \"id:\"
+links (e.g. \"[[id:12345][|]]\").")
 
 (defun org-roam-complete-link-at-point ()
-  "Complete \"roam:\" link at point to an existing Org-roam node."
+  "Complete \"id:\" link at point to an existing Org-roam node.
+Places the completed string into the link description. If there
+was already an \"id:\" link at point, overwrites it to fit the
+new description."
   (let (roam-p start end)
     (when (org-in-regexp org-roam-bracket-completion-re 1)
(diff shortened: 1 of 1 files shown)


leinfink:
There is the small potential issue that, with font-lock mode active, it can be hard to tell whether the ID has been updated.

So I can imagine cases where 1) you want to change the description of a link without changing its target, and then 2) use a completion where it's not clear if it was just a "in-buffer" completion etc. or an org-roam completion, and therefore you maybe wouldn't be sure whether the ID was changed against your will or not.

If this is a concern, I guess one could inhibit "updating" completions and only allow for org-roam completion when creating new links? That alone is already a very practical feature in my opinion. In any case, the current "roam:" link completions are not useful anymore.

dustinfarris:
Is this PR going to be merged?

leinfink:
> There is the small potential issue that, with `font-lock-mode` and `org-link-descriptive` active, it can be hard to tell whether the ID has been updated.

I haven't tested it, but this might have been solved with the new Org version 9.6?

> org-catch-invisible-edits now works for hidden parts of the links and for emphasis markers

So I'd assume with `org-catch-invisible-edits` set to `show`, you would see if there was an ID update. However, if set to `error` it might not work at all? Or does `org-catch-invisible-edits` only apply to direct user input?

I will check later.

publicimageltd:
This merge seems to be overdue, any reasons it didn't happen yet?

hpfr:
For the in-link CAPF, I’m hesitant to provide completion when modifying existing id: links, since users might want to change a link description without changing the target node. Also, not all ID links necessarily link to org-roam nodes (`org-roam-db-node-include-function`).

Changing both CAPFs to insert ID links definitely makes sense though. I opened https://github.com/org-roam/org-roam/pull/2595 which includes some other cleanup and avoids breaking muscle memory of users expecting completion on `[[roam:]]` links.

If you are watching this PR, help reviewing and testing that one is appreciated!

PR #2221, opened 2022-06-13
(db)fix: file modification detection failing in some edge cases

kisaragi-hiu:

`org-roam-db--file-hash` potentially returns different values depending on whether it is run with the file path passed to it. That is,

```elisp
(equal (org-roam-db--file-hash)
       (org-roam-db--file-hash (buffer-file-name))
```

can be nil.

One example of an edge case that triggers this:

- If a file has LF line endings,
- and is situated in a repository where .editorconfig asks for CRLF,
  (Such repositories exist, for example <https://github.com/obsidianmd/obsidian-docs/>)
- and editorconfig-mode is enabled,

org-roam-db--file-hash would then return different results depending on whether the file path is passed to it or not.

The result is that these files will always be updated regardless of whether they are modified or not.

While this is a contrived edge case, there may be other cases like this where the data from `insert-file-contents-literally` and the data from `find-file-noselect` makes `secure-hash` return different values.

This PR removes the branching path in `org-roam-db--file-hash` to compute the hash based on buffer content, instead making it always require a file path.

(This commit also allows `org-roam-db-insert-file` to take an optional HASH argument, allowing it to avoid recalculating the hash.)

Since now we always compute the hash with the file on disk, the special case for encrypted files is no longer needed.

This should have no impact on whether db is synchronized. The only case when db--file-hash uses the "calculate in current buffer" path is in db-insert-file; the only use of db-insert-file is in db-update-file, which already calls db--file-hash with the file path, and does not rely on the value from the current buffer.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -349,15 +349,16 @@ If there is no title, return the file name relative to
                                   (buffer-file-name (buffer-base-buffer))
                                   org-roam-directory)))))
 
-(defun org-roam-db-insert-file ()
+(defun org-roam-db-insert-file (&optional hash)
   "Update the files table for the current buffer.
-If UPDATE-P is non-nil, first remove the file in the database."
+If UPDATE-P is non-nil, first remove the file in the database.
+If HASH is non-nil, use that as the file's hash without recalculating it."
   (let* ((file (buffer-file-name))
          (file-title (org-roam-db--file-title))
          (attr (file-attributes file))
          (atime (file-attribute-access-time attr))
          (mtime (file-attribute-modification-time attr))
-         (hash (org-roam-db--file-hash)))
+         (hash (or hash (org-roam-db--file-hash file))))
     (org-roam-db-query
      [:insert :into files
       :values $v1]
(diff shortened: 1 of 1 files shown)


org-roam-db.el
```diff
@@ -606,19 +607,12 @@ INFO is the org-element parsed buffer."
...
-        (insert-file-contents-literally file-path)
-        (secure-hash 'sha1 (current-buffer)))
-    (org-with-wide-buffer
-     (secure-hash 'sha1 (current-buffer)))))
+(defun org-roam-db--file-hash (file-path)
+  "Compute the hash of FILE-PATH."
+  (with-temp-buffer
+    (set-buffer-multibyte nil)
+    (insert-file-contents-literally file-path)
+    (secure-hash 'sha1 (current-buffer))))
```
jethrokuan, line 615: hmm i then wonder whether we really need to compute the content hash, or just get the hash of the file directly then. I forget why we needed to use the file contents

kisaragi-hiu: There doesn't seem to be a way to compute the hash from the file directly, as `secure-hash` only works on a string or a buffer; I'm not sure if shelling out would be worth it either.

jethrokuan: yeah I think this is fine too.

PR #2227, opened 2022-06-21
Add org-roam-link-face

notevenaperson:
This registers an additional font-face distinct from org-link. This pull request doesn't change how anything appears on screen, it only makes it easier to customize the element. The (My) use case is to color "internal" links different than external World Wide Web links. I like to tell them apart at a glance, colored in pink:
image

Roam too, has a cue to visually distinguish [[roam]] links from external links. (image i found on google) The user can easily customize org-roam-link-face to be visually distinct, if he or she wants.
image

I've given thought whether my proposed change may be org-roam overreaching its territory; since id links are a native feature of org, should the only face to style them be named after org-roam? Yes. Org-roam v2 itself was redesigned in part, to be more synonymous with the native id: links. When the user activates org-roam her or she has chosen to treat id: links in the org-roam context. He or she has chosen to apply org-roam-link-face.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -107,6 +107,11 @@
   :group 'org-roam
   :group 'faces)
 
+(defface org-roam-link-face
+  `((t :inherit org-link)))
+  "Customizable face to make id: links stand out from other org links")
+(org-link-set-parameters "id" :face 'org-roam-link-face)
+
 (defcustom org-roam-verbose t
   "Echo messages that are not errors."
   :type 'boolean


Sachin-12:
> This registers an additional font-face distinct from org-link. This pull request doesn't change how anything appears on screen by default, it only makes it easier to customize the element, it's optional. The (My) use case is to color "internal" links different than external World Wide Web links. I like to tell them apart at a glance, colored in pink: image
> 
> Roam too, has a cue to visually distinguish [[roam]] links from external links. (image i found on google) The user can easily customize org-roam-link-face to be visually distinct, if he or she wants. image
> 
> I've given thought whether my proposed change may be org-roam overreaching its territory; since id links are a native feature of org, should the only face to style them be named after org-roam? Yes. Org-roam v2 itself was redesigned in part, to be more synonymous with the native id: links. When the user activates org-roam her or she has chosen to treat id: links in the org-roam context. He or she has chosen to apply org-roam-link-face.

This is helpful. In Logseq it is shown differently like you said. Pls merge this soon

Sachin-12:
This is helpful. It should be able to differentiate a website link and org node link. Merge this soon

JohnJohnstone:
(defun +org-link--id-face-fn (link)
  (let ((m (org-roam-id-find link)))
    (if m
        `(:inherit org-link
          :foreground ,(doom-color 'base1)
          :background ,(doom-color 'green)
          :weight bold
          :underline nil
          )
      '(:inherit org-link
          :weight bold))))

(org-link-set-parameters "id"
                          :follow 'org-roam-id-open
                          :face '+org-link--id-face-fn)

I am using the above hack to get same functionailty does sem bringing back the face would be a better route.

ntevenhere:
I haven't taken a look at this since I proposed it 3 years ago, though it's a simple addition that shouldn't take any work to integrate into org-roam's current codebase.

hpfr:
> I've given thought whether my proposed change may be org-roam overreaching its territory; since id links are a native feature of org, should the only face to style them be named after org-roam? Yes. Org-roam v2 itself was redesigned in part, to be more synonymous with the native id: links. When the user activates org-roam her or she has chosen to treat id: links in the org-roam context. He or she has chosen to apply org-roam-link-face.

Thanks for noting this. The problem here is that even among org-roam users, not all `id:` links are necessarily org-roam links. `org-roam-db-node-include-function` can be configured to exclude IDs, and ID links can exist outside org-roam-directory. So calling the face `org-roam-id` is a bit misleading. Since it’s just a few lines that don’t depend on org-roam, I think it’s best left to users’ Emacs configs (or a tiny separate package).

```elisp
(with-eval-after-load 'org-id
  (defface org-id-link '((t :inherit org-link))
    "Face for Org id: links."
    :group 'org-faces)
  (org-link-set-parameters "id" :face 'org-id-link))
```

PR #2230, opened 2022-06-30
fix outline of link in properties

bhuztez:
I added a link in the properties drawer, however it does not show up in the backlinks. so I checked org-roam.db. It turns out that the outline of the link is empty.


VitruvianVice:
Without having tested it yet. This should also fix #2252

jethrokuan (approved):

PR #2240, opened 2022-07-23
Force capture with prefix arg in org-roam-node-insert

Hugo-Heagren:
Sometimes I'm writing a note and want to insert a link to something else, and write something about the connection between the two in *both notes*. `org-roam-node-insert` *nearly* does this, but by default it doesn't capture to the target node if it already exists. It also doesn't currently use a prefix arg, so that seemed a good way to trigger this. The previous behaviour remains the default, so users who prefer that won't notice any difference.

Original commit message:
``` text
(feat)capture: prefix arg forces capture in org-roam-node-insert

Add an argument FORCE-CAPTURE to `org-roam-node-insert', which is a
prefix arg when called interactively. If FORCE-CAPTURE is non-nil,
then capture to the selected node before inserting the link regardless
of whether it already exists (previous behaviour was to only capture
to new nodes, this remains the default without FORCE-CAPTURE).

Add a line to docstring explaining default behaviour.

Update docstring for new arg.
```

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -662,14 +662,16 @@ COMPLETION-A and COMPLETION-B are items in the form of
 ;;;; Linkage
 ;;;;; [id:] link
 ;;;###autoload
-(cl-defun org-roam-node-insert (&optional filter-fn &key templates info)
+(cl-defun org-roam-node-insert (force-capture &optional filter-fn &key templates info)
   "Find an Org-roam node and insert (where the point is) an \"id:\" link to it.
+If the node does not exist capture to it. If prefix arg FORCE-CAPTURE is
+non-nil, capture even the node already exists.
 FILTER-FN is a function to filter out nodes: it takes an `org-roam-node',
 and when nil is returned the node will be filtered out.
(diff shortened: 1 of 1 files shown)


Hugo-Heagren:
Bump. This has no conflicts and I still think it's a useful feature.

hpfr:
The implementation adds a required positional argument, which will break existing calls from users’ Emacs Lisp. It would have to be a new keyword argument (which `interactive` could still assign from the prefix argument by using a custom list instead of `"P"`).

Aside from that, my hesitation with introducing prefix arguments is that they really have to be an intuitive extension of the command’s base functionality, and I’m not sure capturing to the link target while inserting the link in the current buffer meets that bar. One alternative prefix argument use off the top of my head would be prompting for a link description instead of using the node title.

You’re right that current users can just not use it, but the issue is it has to be maintained and documented indefinitely, and every additional branch in user-facing interactive functions adds complexity that newcomers have to wrap their heads around when learning how to use the package. So for now, I’m inclined to let power users define their own insert functions in their personal configs. What do you think?

Issue #1043, opened 2020-08-16
Backlinks filtering


wedens:
### Brief Abstract

It'd be really useful to be able to filter the backlinks buffer by selecting other links that are shown in the backlinks buffer. In a regular tag-based system it's basically the same as filtering by multiple tags.

Randomly found screenshot of the filter in Roam:
![image](https://user-images.githubusercontent.com/697162/90327455-004c4480-dfbe-11ea-86ca-ec1ff82826d2.png)

### Please check the following:

- [x] No similar feature requests

dangom:
This would be extremely useful to avoid cluttering. Filtering out backlinks to dailies or showing them with a toggle would be ideal.

Here's a workaround hack that illustrates how this could be done (note I use a single file for dailies).

```
(setq dg/dailies-filename "/Users/dg/org/roam/dailies.org")

(defun dg/org-roam-backlinks-section (node)
    "The backlinks section for NODE."
    (when-let ((backlinks (seq-sort #'org-roam-backlinks-sort (org-roam-backlinks-get node))))
      (magit-insert-section (org-roam-backlinks)
        (magit-insert-heading "Backlinks:")
        (dolist (backlink backlinks)
        ;; This is a filtering condition that only lets backlinks through that are not sourced from my dailies file.
          (unless
              (string= dg/dailies-filename (org-roam-node-file (org-roam-backlink-source-node backlink)))
            (org-roam-node-insert-section
             :source-node (org-roam-backlink-source-node backlink)
             :point (org-roam-backlink-point backlink)
             :properties (org-roam-backlink-properties backlink))))
        (insert ?\n))))

(setq org-roam-mode-section-functions (list #'dg/org-roam-backlinks-section #'org-roam-reflinks-section))
```

A more sustainable solution would be to have `org-roam-backlinks-section` accept an optional filter function, and the documentation to give examples of filter functions users could read to implement their own filters.

ahmed-shariff:
I am wondering if org-ql-view like interface might be of interest here. I already have a very rough workaround to view a list of roam nodes with view, which I find is far better for my purposes than the org roam buffer.

wuqui:
That sounds great!

ahmed-shariff:
A simple hack for using org-ql-view with org-roam-nodes:
This is modified from [org-ql-search](https://github.com/alphapapa/org-ql/blob/e41cdb45912f1731b5b060f9e05d011096663655/org-ql-search.el#L119)
```emacs-lisp
[...]

PR #2247, opened 2022-08-10
Add support for filtering backlinks.

swflint:
See #1043.

Being able to filter things from the backlinks buffer (for instance, dailies) can be helpful to reduce on-screen clutter.

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -655,6 +655,19 @@ rendering reference links), set ~org-roam-mode-sections~ as follows:
           org-roam-reflinks-section))
 #+end_src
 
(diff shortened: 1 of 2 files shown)


swflint:
I can add the same functionality to org-roam-reflinks-section as well if that would be desired.

swflint:
@jethrokuan, could I get this reviewed?

swflint:
I'd like to see this usable without having to maintain a separate copy of the function in my .emacs.d, and it seems that this has been discussed elsewhere.  Would it be possible to get this reviewed?

jethrokuan (approved):

PR #2258, opened 2022-09-11
(Docs): orgmode.org/elpa has been shut down

aviad:
See https://list.orgmode.org/87blb3epey.fsf@gnu.org/ for details. Removed it from the documentation in the manual to (try and) assist newcomers.

Trying to install `org-mode` according to the manual fails, and on my setup (Emacs 28.1) is unnecessary.

If it is still required for the manual to help users install a recent version of `org-mode`, its [installation docs](https://www.gnu.org/software/emacs/manual/html_mono/org.html#Installation) recommend:
1. Using Emacs packaging system (no mention of adding the elpa repo)
2. Using Org’s git repository

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -178,18 +178,7 @@ archives to =package-archives=:
 #+END_SRC
 
 Org-roam also depends on a recent version of Org, which can be obtained in Org's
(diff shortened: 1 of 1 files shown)


aviad:
@jethrokuan - randomly pinging you since I saw you reviewed the last docs PR. 🙏🏿

PR #2260, opened 2022-09-13
save source & target buffers after refile #2257

### LINKED ISSUE #2257: org-roam-refile should save both source and target files
### Brief Abstract

`org-roam-refile` should save the source file and the target file, *in that order*, after the node has been moved.

### Long Description

The org-roam database records the mapping between IDs and files, updating when a file is saved. When a node is refiled, the mapping should be updated in the database so that `org-roam-node-find` can locate it correctly.

If we save the target file first, we'll get a duplicate node error.

There's a risk: If we save the source file first, there is a small time window where the node doesn't appear anywhere in the database; it could be lost if that target file save doesn't take place for some reason.

When `org-roam-refile` completes, trigger a save on the source file buffer and then the target file buffer.

jmay:

Avoid potential loss of changes following a refile. Ensure that org-roam database is updated immediately, so that `org-roam-find` locates the refiled node in the right place.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -916,6 +916,8 @@ If region is active, then use it instead of the node at point."
          (delete-region
           (and (org-back-to-heading t) (point))
           (min (1+ (buffer-size)) (org-end-of-subtree t t) (point)))))
+      (save-buffer)
+      (with-current-buffer nbuf (save-buffer))
       ;; If the buffer end-up empty after the refile, kill it and delete its
       ;; associated file.
       (when (eq (buffer-size) 0)


hpfr:
In general, saving buffers is something that should be left to the user, since users might prefer different auto save mechanisms and might run arbitrary code in hooks before and after save. Upstream `org-refile` doesn’t save at all, for example. I would maybe suggest advice here?

meedstrom:
> Upstream org-refile doesn’t save at all, for example. 

This is a pretty strong argument not to save, I agree.

At the same, upstream Org does not have an equivalent to `org-roam-node-find`. It'd be nice if that command could find the nodes at their new locations immediately. And the simplest way to do that is to save...

I've tried writing code that can find the nodes despite them being unsaved, and it's not worth the complexity IMO.

It's theoretically possible to have `org-roam-db-sync` always prefer any currently open buffer if one exists, over reading a file...

hpfr:
Let’s try consolidating discussion in the issue for now and keep comments here to the implementation when we’re ready

PR #2263, opened 2022-09-20
Minor typo in the commented text of the definition of `org-roam-refile`

apc:


apc (commented):
I was just reviewing the code and noticed a simple typo. Nothing should be affected by this.

PR #2292, opened 2022-12-10
don't complete org-roam nodes in source blocks

ParetoOptimalDev:

This fixes the following situation where
`org-roam-completion-everywhere = t` and you are working with org-babel
blocks in one of your org-roam nodes.

```
#+begin_src emac|
#+end_src
```

Before my nodes would be suggested, but after this change only valid
babel headers are suggested as desired.

Previously completions from org-roam nodes would be suggested


ParetoOptimalDev:
@jethrokuan Thanks for fixing the lints.

I missed the failure somehow, sorry!

Issue #2207, opened 2022-05-26
Allow more than one node associated with a ref?


bdarcus:
I'm trying to get this little package working https://github.com/bdarcus/citar/pull/609.

From my understanding, org-roam v2 should allow two use cases, neither of which was supported in v1, or in citar now.

1. multiple refs/citekeys per bibliographic note
2. multiple bibliographic notes per ref/citekey

From experimenting with `org-roam-ref-find`, it appears the second is not currently supported.

Example:

![image](https://user-images.githubusercontent.com/1134/170560276-daf8f67d-7619-4481-affb-377f4fcb49ba.png)

See those first two candidates? The first is on a headline node, and the second on a file.

If I select the first, it opens the second. This is because there are duplicate keys, so `completing-read` can’t distinguish them. 

Or note this discrepancy:

```elisp
ELISP> (org-roam-node-from-ref "@ahrens2017")
#s(org-roam-node :file "/home/bruce/org/roam/biblio/ahrens2017.org" :file-title "Notes on: Ahrens, S. (2017): How to take smart notes" :file-hash nil :file-atime
                  (25057 29042 708718 37000)
                  :file-mtime
                  (24920 53783 24747 92000)
                  :id "6661bba3-4b23-4934-aec4-389676bd5b31" :level 0 :point 1 :todo nil :priority nil :scheduled nil :deadline nil :title "Notes on: Ahrens, S. (2017): How to take smart notes" :properties
                  (("CATEGORY" . "ahrens2017")
                   ("ROAM_REFS" . "@ahrens2017")
                   ("ID" . "6661bba3-4b23-4934-aec4-389676bd5b31")
                   ("BLOCKED" . "")
                   ("FILE" . "/home/bruce/org/roam/biblio/ahrens2017.org")
                   ("PRIORITY" . "B"))
                  :olp nil :tags nil :aliases nil :refs
                  ("ahrens2017"))

ELISP> (org-roam-db-query [:select * :from refs :where (= ref "ahrens2017")])
(("6661bba3-4b23-4934-aec4-389676bd5b31" "ahrens2017" "cite")
 ("eeafba7b-d1e1-4349-8ab2-2a7e62e714c2" "ahrens2017" "cite"))
```

## The issue

You have an inconsistency:

1. org-roam will populate the database with multiple nodes per ref key
2. but most of the rest of the code enforces a single node per ref key
3. except `org-roam-ref-read--completion` does not; hence this bug

[...]

PR #2305, opened 2023-01-13
(node): add function to return all nodes with a ref

Hugo-Heagren:
* org-roam-node.el (org-roam-nodes-from-ref): return all nodes with a
given ref. Return nil if there are no such nodes. Always returns a
list.

-------

Sometimes, I want to get a list of all the node with a given ref. At the moment, there is no programmatic facility for this besides writing a manual query.. I think others might find this useful as well, so I've made a PR.

A concrete example specific to me: I use org-roam as a backend for notes on literature in [citar](https://github.com/emacs-citar/citar). This requires a function to get a hash-table of relevant note nodes. At the moment my function looks like this:

``` emacs-lisp
(defun my/citar-org-roam-titles-hash-table (keys)
    "Return hash table of KEYS and lists of notes.
Each key in KEYS is a key in the hash table. Each value is a list
of strings: the titles of every org roam node which has a
ROAM_REFS property with that KEY."
    (let ((table (make-hash-table)))
      (mapcar
       (lambda (key)
     ;; NOTE Only one node returned
	 (when-let ((node (org-roam-node-from-ref (concat "@" key)))
		    (formatted (org-roam-node-title node)))
	   (puthash key `(,formatted) table)))
       keys)
      table))
```

This only ever returns one node. For large, complex works (especially those where I am working on the same ideas myself), I find it useful to dedicate smaller nodes to individual ideas, but still use the `ROAM_REFS` property to link the node to the work. This means that I often have more than one node with the relevant ref, and it would be good if a function existed for getting them all.

###### Implementation note
At the moment, the new function is implemented in the same way as the previously existing `org-roam-node-from-ref`. It would be easy to reimplement `org-roam-node-from-ref` using this new function (just as `(car (org-roam-nodes-from-ref REF))`). I haven't done this, as I wasn't sure if there was a reason (some sort of optimisation?) the earlier function was limited to one node. Would be happy to add if necessary.


Hugo-Heagren:
Gentle bump. I'm happy to alter this in line with feedback if need be, but it would be great if something like it could be merged -- I need it for my related-files workflow and having it as a local monkey-patch is a bit annoying.

hpfr:
Seems related to https://github.com/org-roam/org-roam/issues/2207

Hugo-Heagren:
Bump. This has no conflicts and I still think it's a useful feature.

hpfr:
I don’t have a full understanding, but this is discussed at length in https://github.com/org-roam/org-roam/issues/2207. It seems like people previously thought that properly supporting multiple nodes per ref needed to be addressed with care.

@bdarcus previously attempted this in https://github.com/org-roam/org-roam/commit/ee79bba3fd2ccd55a8680c61ad49fb9ff873cdaf, but dropped the commit after Jethro suggested in https://github.com/org-roam/org-roam/issues/2207#issuecomment-1146877644 that he would improve it to account for factors mentioned in https://github.com/org-roam/org-roam/issues/2207#issuecomment-1140322208. So https://github.com/org-roam/org-roam/pull/2208 only addressed one facet.

It also ties into a design question (https://github.com/org-roam/org-roam/issues/2207#issuecomment-1140228892):
> If we go down this path, I do start to wonder what the difference would be between a note with a ref, and a note with a citation to that same citekey. E.g. is a note whose body is `Some text with two citations: [cite:@doe] and [cite:@smith]` just a bibliographic note for two refs?

I would suggest reading through that entire issue, particularly https://github.com/org-roam/org-roam/issues/2207#issuecomment-1140322208, and hopefully @bdarcus can offer advice or further context?

In my own org-roam priorities, addressing this would come after 1-4 in https://github.com/org-roam/org-roam/issues/2474#issuecomment-3701549146. Hopefully that won’t take too long now that there’s a few moderately active maintainers.

PR #2334, opened 2023-03-26
[Fix #1732] Fix contents in org roam buffer being displayed in the wrong order

### LINKED ISSUE #1732: clicking on any link within *org-roam* buffer fails with an error message
After upgrading to org-roam version 2, none of the backlinks shown in my `*org-roam*` buffer work.  If I click on any link  with my mouse or hit `RETURN` key with cursor on a link, nothing happens other than the following error being displayed in the mini-buffer:
```
There is no thing at point that could be visited
```
When I click on any link within `*org-roam*` buffer, I expected the the content of the target link to pop up and be displayed in a buffer.  However nothing happens others than the error message being displayed in the mini-buffer.

I have many emacs startup files setup using chemacs2 package on ubuntu 20.04 so that I can choose one of many setups.  The reported problem occurs with my recent builds of emacs 27 or 28, with package.el or straight.el, with spacemacs or no spacemacs.

Sergey-Makarov:
This pull request is aimed to fix the issue when the preview sections of the backlinks inside an org roam buffer are not displayed where they should.

Steps to reproduce:
1. tested only on Spacemacs distro
2. page-break-lines-mode, global-page-break-lines-mode, display-line-numbers-mode and global-display-line-numbers-mode are enabled in org mode buffer
3. invoke org-roam-buffer-toggle to show up an org roam buffer

Result:

Also hitting RET on any section inside such buffer would result in the following error message:
```
There is no thing at point that could be visited
```

Result if the fix is applied:

What is also interesting is that even without this fix, if we navigated to org roam buffer and called org-roam-buffer-refresh, the buffer would redraw just fine. But if we then navigated back to the original org mode buffer and triggered org roam buffer refresh (by, say, pointing to another node that has it's own backlinks), the problem would occur once again.

I have no clear idea of why this happens, but here is a line that actually caused this issue:
```
  (defun org-roam-fontify-like-in-org-mode (s)
    "..."
    (with-temp-buffer
      (insert s)
      (let ((org-ref-buffer-hacked t))
        (org-mode) ;; after this call, the point inside the org roam buffer is set to 1 for some reason. this is why all the previews are displayed at the top of the buffer
        (setq-local org-fold-core-style 'overlays)
        (font-lock-ensure)
        (buffer-string))))
```

This fix is expected to close these issues:
https://github.com/org-roam/org-roam/issues/1732
https://github.com/syl20bnr/spacemacs/issues/14969

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -225,13 +225,19 @@ Like `org-fontify-like-in-org-mode', but supports `org-ref'."
   ;;
   ;; `org-ref-buffer-hacked' is a buffer-local variable, therefore we inline
   ;; `org-fontify-like-in-org-mode' here
-  (with-temp-buffer
-    (insert s)
-    (let ((org-ref-buffer-hacked t))
-      (org-mode)
-      (setq-local org-fold-core-style 'overlays)
-      (font-lock-ensure)
-      (buffer-string))))
+  ;;
+  ;; NOTE: we have to use `save-excursion', because after calling `org-mode' in temp buffer, the point in the "main" buffer
+  ;; might change to 1 for unknown reason, which we want to avoid. It's somehow related to using `page-break-lines-mode',
+  ;; `display-line-numbers-mode' and the corresponding global versions of those modes.
+  ;; See <https://github.com/org-roam/org-roam/issues/1732>.
+  (save-excursion
+    (with-temp-buffer
+      (insert s)
+      (let ((org-ref-buffer-hacked t))
+        (org-mode)
+        (setq-local org-fold-core-style 'overlays)
+        (font-lock-ensure)
+        (buffer-string)))))
 
 ;;;; Shielding regions
 (defface org-roam-shielded


real-or-random (approved):
I've tested the branch and can confirm that it resolves the issue.

jpstone:
Curious when this will get merged?

NahsiN:
I had a similar issue in spacemacs. However OP correctly mentions the related issue on spacemacs repo https://github.com/syl20bnr/spacemacs/issues/14969 which has now been closed. Pulling the latest changes resolved this for me. To be specific it was this commit https://github.com/syl20bnr/spacemacs/commit/dd99bf45ee5fe0206db4817ba0668720923c1dbe

Sergey-Makarov:
Hey @akashpal-21, thanks for your inputs. Currently, I don't have time to apply the suggested fix and test it properly, but maybe someone else does 🙂
I don't mind if my PR gets closed in favor of a new, more proper one. Edits by maintainers is enabled for this PR, too.

dustinfarris:
I am having trouble reproducing this with a minimal config.  Is it possible that this is no longer an issue?

```emacs-lisp
;; minimal-init.el

;; Package setup
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
(package-initialize)

;; Bootstrap use-package
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))
(require 'use-package)

;; Ensure packages are installed by default
(setq use-package-always-ensure t)

;; Install org-roam from MELPA
(use-package org-roam
  :custom
  (org-roam-directory "~/org-roam/")
  :config
  (org-roam-db-autosync-mode)
  :bind (("C-c n f" . org-roam-node-find)
         ("C-c n i" . org-roam-node-insert)
         ("C-c n l" . org-roam-buffer-toggle)))

;; Install page-break-lines from MELPA
(use-package page-break-lines
  :config
  (global-page-break-lines-mode))
```

Screenshot 2025-06-06 at 11 28 39 PM

akashpal-21:
Could be. The problem was never from the side of org roam but from the other package. Do you think it would be a general good idea to make sure that such other packages do not cause conflict anyways in the future ? 

When org-mode is called; many a packages that hook onto it will be called. We can prevent it by wrapping it in (delay-mode-hooks.

Do you think it would be a good preventive step nevertheless?

akashpal-21:
Counter argument to my previous point:

The user may "expect" fortification to be exactly like in org-mode, which means all modifications that user makes manually. Including all hooks that user explicitly adds. This could very well be - but the user then would be responsible for throwing the wrench indirectly.

But the user may also not be able to pin point what is causing breakage because the user cannot be expected to remember all packages they have installed.

I remember another problem that is related to this issue: 

https://org-roam.discourse.group/t/resolved-styling-links-in-the-roam-buffer/3720/20

Same problem.

Sergey-Makarov:
For the record, I tried the mentioned steps using my own Emacs config (I am not using Spacemacs anymore), and the problem does not reproduce anymore for me either.

hpfr:
Based on the recent feedback I think we can close this and revisit it if anyone pipes up with steps to reproduce, thanks all

PR #2336, opened 2023-03-28
(fix) Replace olp with heading list

### LINKED ISSUE #2335: `org-roam-dailies-capture` make OLP incorrectly  when captured again
### Description
When I make capture task by code, org-roam generates OLP heading again sometimes.

#### Steps to Reproduce

Run this code and check `org/roam/daily/test.org` Then run the code again.
```elisp
(let ((org-roam-directory (expand-file-name org-roam-dailies-directory org-roam-directory))
      (org-roam-dailies-directory "./")
      (org-roam-dailies-capture-templates
       `(("t" "Tasks" entry "* %?"
          :if-new (file+head+olp "test.org" ;; "%<%Y-%m-%d>.org"
                                 "* Notes\n\n* Tasks\n\n* Journal"
                                 ("Tasks" "[2023-03-28 Tue]"))))))
  (save-window-excursion
    (org-roam-dailies--capture (current-time) t)))
```

#### Backtrace
There is no error for this.

#### Expected Results
```org
* Notes

* Tasks

** [2023-03-28 Tue]

* Journal

```
#### Actual Results
`'("Tasks" "[2023-03-28 Tue]")` OLP is generated again.
```org
* Notes

* Tasks

** [2023-03-28 Tue]

* Journal

* Tasks

** [2023-03-28 Tue]
```

### Environment
```
- Emacs: GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0)
 of 2023-03-20
- Framework: Doom
- Org: Org mode version 9.6.1 (9.6.1-??-fe92a3c @ /home/gglee/.config/emacs/.local/straight/build-30.0.50/org/)
- Org-roam: 5c06471-dirty
- sqlite-connector: sqlite-builtin
```

hwiorn:
Closes #2335

After `org-roam-capture--fill-template` which actually calls `(kill-buffer)` inside it, strangely `kill-buffer` reverts buffer position from `(point-min)` to original. So `re-search-forward` loop can't find headings correct and `org-roam-capture-find-or-create-olp` keep creating headings if the buffer is already visited and buffer position is under below the OLP.

So I change `olp` to `(setq headings (mapcar #'org-roam-capture--fill-template olp))` then use it in the search loop.

After this patch, below test code doesn't make new "Tasks" heading again when it exists already.
```
(let ((org-roam-directory (expand-file-name org-roam-dailies-directory org-roam-directory))
      (org-roam-dailies-directory "./")
      (org-roam-dailies-capture-templates
       `(("t" "Tasks" entry "* %?"
          :if-new (file+head+olp "test.org" ;; "%<%Y-%m-%d>.org"
                                 "* Notes\n\n* Tasks\n\n* Journal"
                                 ("Tasks" "[2023-03-28 Tue]"))))))
  (save-window-excursion
    (org-roam-dailies--capture (current-time) t)))
```

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -612,13 +612,14 @@ you can catch it with `condition-case'."
          (lmax 1)
          (start (point-min))
          (end (point-max))
+         headings
          found flevel)
     (unless (derived-mode-p 'org-mode)
       (error "Buffer %s needs to be in Org mode" (current-buffer)))
     (org-with-wide-buffer
+     (setq headings (mapcar #'org-roam-capture--fill-template olp))
      (goto-char start)
-     (dolist (heading olp)
-       (setq heading (org-roam-capture--fill-template heading))
+     (dolist (heading headings)
        (let ((re (format org-complex-heading-regexp-format
                          (regexp-quote heading)))
              (cnt 0))
--- a/tests/test-org-roam-dailies.el
+++ b/tests/test-org-roam-dailies.el
@@ -1,0 +1,55 @@
+;;; test-org-roam-dailies.el --- NONDESCRIPT FILE  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2026 Martin Edström
+;;
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; Code:
+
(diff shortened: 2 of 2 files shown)


hpfr:
The test snippet in the description should be converted to a true test.

meedstrom:
You mention that point moves when it shouldn't:

> After org-roam-capture--fill-template which actually calls (kill-buffer) inside it, strangely kill-buffer reverts buffer position from (point-min) to original. 

That indicates a bug!

I'm not familiar with this code at all, but I found some potential smells.

Let's recap:  
1. We're sitting in `org-roam-capture-find-or-create-olp`, which uses `org-with-wide-buffer` (after apparently let-binding some variables `start` and `end` based on a NOT wide buffer, I don't know if that's intentional),
2. then in the wide buffer, calls `org-roam-capture--fill-template` a number of times, 
3. which calls `org-capture-fill-template`, which... does something I find suspicious given the aforementioned context: it generates a temp buffer and uses `save-window-excursion` and `switch-to-buffer-other-window`...

Maybe that's sane, maybe not.

I'm guessing the `kill-buffer` you saw is the one at the end of `org-capture-fill-template`.  And that something weird happens then on cleanup.

meedstrom:
Huh, the test suite doesn't have load path to org-roam-dailies?

    [00:01.702]  Cannot open load file: No such file or directory, org-roam-dailies

Issue #2241, opened 2022-07-24
Lists are slow; why not using hash tables?


jcguu95:
### Brief Abstract
I have many nodes and things start getting slower and slower in org-roam. I suspect one core issue is that it uses lists rather than hash table (e.g. `#'org-roam-node-list`). **I want to help.**

### Long Description
I can count myself as a "long-term" org-roam user. Thanks for the wonderful package! However, that means I have thousands of nodes, and each time the db update itself, it takes more than 20 seconds. 

I have a tiny hack uses memoization. However, it takes a lot of memory and sometimes it still gets slow.
``` emacs-lisp
  (setq org-roam-node-default-sort nil)
  (require 'memoize)
  (memoize 'org-roam-node-read--completions "10 minute")
  (defun my/force-update-org-roam-node-read (&optional timeout)
    (interactive)
    (flet ((memoize-force-update (fn-symbol &optional timeout)
                                 (when (get fn-symbol :memoize-original-function)
                                   (progn (memoize-restore fn-symbol)
                                          (memoize fn-symbol timeout)))))
      (memoize-force-update 'org-roam-node-read--completions
                            (or timeout memoize-default-timeout))))
  (run-with-idle-timer 60 t #'my/force-update-org-roam-node-read)
```

**QUESTION** Why do we use lists instead of hash tables? With the latter, it's easier to update and make changes to a specific node without traversing over the whole thing.

### Proposed Implementation (if any)
I'd like to know the principle of design behind, and how much it would need to change. And then I will come up with a concrete implementation.

jethrokuan:
That's primarily because the emacs APIs (e.g. minibuffer completion) take in a list and not a hash table. Traditionally hash tables have not been very popular and association lists are also used in place of hash tables.

jcguu95:
Indeed, the traditional `#'completing-read` only takes lists. However, nowadays many people have hooked it so it calls some modern completion framework instead. For example, `#'ivy-read` and `#'consult--read` both take hash tables.

A proposal would be to let the user choose if they want to enable hash table support. In that case, **(1)** do you think this will make org-roam (with thousands of nodes) significantly faster? For example, we can update the very specific entry of the database only when we need to, without touching (e.g. traversing through the list) other nodes at all. **(2)** What are some starting points and caveats that you'd advice me to pay attention to? We don't need to make this into the official branch, but I think it will be a nice thing to have, especially for those who have thousands of nodes like me :)

```
[...]

PR #2341, opened 2023-04-22
(feat): Increase speed of org-roam-node-find

vherrmann:
Increase speed of `org-roam-node-find` by caching last results. The validity of the cache is ensured by comparing the modification attribute of the database file with the last modification time of the cache.

This commit only increases the speed if you didn't modify any node since the last time you used `org-roam-node-find`.


vherrmann:
Please elaborate what the problem is.

vherrmann:
The cache from https://github.com/Konubinix/Devel/blob/30d0184db0a61f46ca35e102302d707fef964a8c/elfiles/config/after-loads/KONIX_AL-org-roam.el#L770-L787 works on a different level. It's caching the whole evaluation of org-roam-node-read--completions, while this pr doesn't cache the sorting of the entries. This pr also caches `org-roam-node-list`, so that even if you use a different filter-function at least the db doesn't have to be queried again.

Also this pr works at least for me with filter functions.

Maybe I'll change

vherrmann:
I added caching for sort-fn. This makes it viable to add sth like
```emacs-lisp
(setq iko/org-roam-node-find-precache-timer
        (run-with-idle-timer 6 t (defun iko/org-roam-node-find-precache ()
                                   (org-roam-node-read--completions org-roam-node--candidate-list-prev-filter-fn
                                                                    (or org-roam-node--candidate-list-prev-sort-fn
                                                                        (when org-roam-node-default-sort
                                                                          (intern (concat "org-roam-node-read-sort-by-"
                                                                                          (symbol-name org-roam-node-default-sort)))))))))
```
to ones config. This code rebuilds the cache if needed after 6 seconds of inactivity.

vherrmann:
It will only rebuild the cache if a change in the db has occured and emacs has been idle for 6 seconds.

vherrmann:
Another trick i use is the following: I memoize `org-roam-node-read--to-candidate`:

```
(defvar org-roam-node-read--to-candidate--memoization-map (make-hash-table :test #'equal :size 1000))
;; FIXME: cleanup org-roam-node-read--to-candidate--memoization-map after long time
(defun org-roam-node-read--to-candidate (node template)
  "Return a minibuffer completion candidate given NODE.
TEMPLATE is the processed template used to format the entry."
  (let ((val (gethash (cons node template) org-roam-node-read--to-candidate--memoization-map)))
    (if val
        val
      (let* ((candidate-main (org-roam-node--format-entry
                              template
                              node
                              (1- (if (bufferp (current-buffer))
                                      (window-width) (frame-width)))))
             (res (cons (propertize candidate-main 'node node) node)))
        (puthash (cons node template) res org-roam-node-read--to-candidate--memoization-map)
        res))))

```

vherrmann:
> I dont really understand this sort fn thing -- can I update with the new patch and if I dont set the timer -- will it work equivalently as before?

Yes

> Is the sort fn put so that the cache could be reset during idle time rather than on every save? 

It's a needed optimization.

> Is it plausible to do delta update to the cache rather than a full reset on change ?

No, this doesn't seem possible to me

dmgerman:
The database query only takes a fraction of the total time. The actual culprit is org-roam-node-read--to-candidate.
Also, the formatting of the data coming out of the database takes also much more time than the actual query.

vherrmann:
I profiled multiple rebuilds of the cache with the memoized version of `org-roam-node-read--to-candidate`.
I got the the following results: `org-roam-node-read--to-candidate` uses 30-60% of the samples, while `org-roam-node-list` uses 30-50% of them. Therefore the caching of org-roam-node-list seems to be a relevant improvement on performance, at least on my system.

dmgerman:
I think whether the user uses the advice or not should be optional.

All of these solutions are intrusive, some more than others. Thus, the user should be confronted with the choice of enabling the advice or not (specially the ones caching the database).

In many ways, these optimizations can live in a separate file (or even a module) and only be enabled (one at a time) if so desired. Also, org-roam uses dynamic scoping, so it is possible to create a file/module that contains redefinitions of functions in org-roam functions (and get fully tested) before they replace the originals. 

by the way, I run the emacs profiler very little time is actually spent in the query. Almost all the time is spent building the data structures (including the reformatting of the query--the cl-loop in org-roam-node-list).

vherrmann:
I'm using sort functions. It is relevant for me that the cache works with sort functions. I don't think that code that silently uses the previous cache (so that the sort function is not applied) can be called a final version. Also due to the (when (>= org-roam-node-cache-level …)) blocks the code only works on a switch of the value of `org-roam-node-cache-level`, if you specify `(setq org-roam-node-cache-level 3)` before loading the file.

dmgerman:
Mmm, see this: 

https://org-roam.discourse.group/t/rewriting-org-roam-node-list-for-speed-it-is-not-sqlite/3475/3?u=dmg

I get a huge difference in garbage collection.

Can you run the following test. This would show how fast the DB is:

https://gist.github.com/dmgerman/f8679f7a11bbdd748f648dd3f30151c5

This is how long it takes on my computer, with 1000 nodes:

```
dmg@cobalt:~/.emacs.d|master⚡ ⇒  /tmp/rip.sh
      66     173    1364
-- Loading resources from /Users/dmg/.sqliterc

real	0m0.009s
user	0m0.004s
sys	0m0.005s
```

it is basically instantaneous

honza:
I just tested this and it works amazingly well.  I have about 3k nodes.  It goes from about 2 seconds to instantaneous. 

@vherrmann Please rebase the PR for easier merging.  I don't think we want the merge commit in there.

vherrmann:
@honza  Like this?

honza:
That works!

dmgerman:
I think it is better to memoeize org-roam-node-read--to-candidate. It is slightly slower, but it keeps track of new nodes and changes to current nodes.

Memoizing org-roam-node-read--completions, it would be needed to reset it to keep track of modifications/new nodes.

vherrmann:
On slower machines the cache is in my experience quite relevant. It is the difference between the function being instantaneous and having a noticeable delay. Even with memoization of read--to-candidate the list still needs to be filtered and/or sorted again.

dmgerman:
But thinking more broadly. The org-roam is a persistent cache. We are trying to build a cache of the cache.

The first time emacs finds a roam file, this cache is read from disk. But after that, unless the user do es a org-roam-db-sync, then this cache can be maintained by modifying org-roam-db-update-file. When a file is created/saved, the cache and db is updated.

This way, the cache will _always_ be up to date, and we will need to read the database very rarely.

vherrmann:
> But thinking more broadly. The org-roam database is a persistent cache. We are trying to build a cache of the cache.
> 
> The first time emacs finds a roam file, this cache is read from disk. But after that, unless the user do es a org-roam-db-sync, then this cache can be maintained by modifying org-roam-db-update-file. When a file is created/saved, the cache and db is updated.
> 
> This way, the cache will _always_ be up to date, and we will need to read the database very rarely.

Yes, this is indeed a better approach that would be even more efficient.

vherrmann:
But it will require a much more intrusive refactor of the db functions

vherrmann:
> @vherrmann doing such a cache we can only save the GC that comes from the db query. We would still need to format entries and this is where the bottleneck is mostly. How is it more efficient.

Having an persistent cache in memory has the simple advantage that we don't have to read in the database all the time. Also if a node/file is changed we only need to format it's entry, which is very fast.
Such a system therefore would be more efficient (since org-roam-node-list does not have to be called if a file changes) and it would keep the calls to format entries to separate, unnoticeable moments.

vherrmann:
For now this pr is good enough for me and I also think that is fares very well in it's maintenance cost to feature ratio.

vherrmann:
> But we do break existing functionalities -- hiding inefficiencies in the code base behind caches is a very bad practice fundamentally.

I'm not sure what functionalities are broken by which code. Also I don't see why caches are bad practice. Org-roam itself is a cache. All the data in the database is already in the file system…. For every cache there should be a good reason. With the current implementation of org-roam I think that there is legitimate reason to implement a cache instead of doing sth different. This cache certainly is not be the best solution, but It's a working one with a rather small code footprint :)

vherrmann:
- A library shouldn't advice it's own functions.
- I don't see why one wouldn't want one of the caches. Therefore I also think that such a variable would place a unnecessary burden on the maintainer of this package. Though I think that for the memoization would should in fact consider such a variable (That's also the reason why I haven't yet added it to this pr).

> At least one variable related to timekeeping is redundant here

Which?

vherrmann:
I'm not convinced that is the case. If the sort-fn or the filter-fn changes the candidate-list is changed but the node-list isn't. Also, even if they were redundant I still might leave them, since they allow easier reasoning about the code (one time stamp for one cache).

dmgerman:
I have created a minor mode that enables the improvements of processing I described above. Disabling the minor mode goes back to the original processing. In my org-roam the saves are 20 times faster.

https://github.com/dmgerman/org-roam-gt/blob/main/org-roam-gt.el

jmay:
> It will only rebuild the cache if a change in the db has occured and emacs has been idle for 6 seconds.

@vherrmann Where is this 6-second idle timer implemented? I don't see anything for this in the code.

vherrmann:
Yes. It's in a [previous message](https://github.com/org-roam/org-roam/pull/2341#issuecomment-2068863385). I did not add it my pr, since it freezes emacs for short moments and this should not be default behavior. Also the memoization of #'org-roam-node-read--to-candidate makes it quick enough for me.

hpfr:
With https://github.com/org-roam/org-roam/pull/2511 merged to improve node display performance, how much benefit does the additional caching provide? What about when a function is used for display instead of a template string?

dmgerman:
> With #2511 merged to improve node display performance, how much benefit does the additional caching provide? What about when a function is used for display instead of a template string?

Note that this does not do any caching. the fix replaced inefficient code (and code that generated a lot garbage).

We would need a standardized method to measure it. @akashpal-21 did a lot of benchmarking  but compared to my org-roam files we saw a lot of difference in results.

For example, I reckon the benefit of a function vs a string for the template will depend on the complexity of the template.

hpfr:
Right, I’m wondering whether the caching introduced in this PR still has measurable benefits now that #2511 made the node display more efficient.

akashpal-21:
> Right, I’m wondering whether the caching introduced in this PR still has measurable benefits now that #2511 made the node display more efficient.

Try it, 
https://gist.github.com/akashpal-21/5991e99949af5feb0f75729374ce3d66

The same cache, implemented using advice rather than changing core fn defs.

hpfr:
> Try it

I don’t have many nodes, so `org-roam-node-find` is already instantaneous for me.

Presumably `org-roam-node-read--to-candidate` is faster with https://github.com/org-roam/org-roam/pull/2511, which you said consumed 98% of the time in https://github.com/org-roam/org-roam/pull/2341#issuecomment-2121820974.

> We would need a standardized method to measure it.

Test code to generate a directory with a few thousand placeholder node files seems like it would be useful for evaluating PRs like this one.

In general though, I tend to agree with the sentiments expressed previously that we should first look to improve inefficiencies in database reads and node display before adding another caching layer.

Also, it seems like having `org-roam-node-list` make a hash table would be worth exploring before caching the list: https://github.com/org-roam/org-roam/issues/2241

edit: oh yeah, and doing the joins on the db side https://github.com/org-roam/org-roam/issues/2474#issuecomment-2369959995

vherrmann:
While search is fast with #2511, it still is not instant for me. Thus I think this PR is still relevant.

vherrmann:
Because

vherrmann:
Closing this in favor of #2589

PR #2343, opened 2023-04-28
Allow `org-roam-ref-add` to take a list of refs

Hugo-Heagren:
Allow `org-roam-ref-prompt-function` to return a list of strings. In this case, each string is added in order, instead of just a single string.

It is occasionally useful to add multiple refs in one go, such as when writing about an academic work and a translation or commentary on it. Adding these one after the other is annoying, and `org-roam-ref-prompt-function` can easily be customised to provide a list of references.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -300,6 +300,34 @@ Return nil if there's no node with such REF."
                               type path))))
           (org-roam-populate (org-roam-node-create :id id)))))))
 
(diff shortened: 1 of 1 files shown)


Hugo-Heagren:
Side note -- I haven't added corresponding behaviour for `org-roam-ref-remove`, but I could if people think this is a good idea? Even if not used very much, it might be useful for some people, and maybe nice to have the same interface for both functions.

PR #2351, opened 2023-06-05
(feat)capture: add customizable defaults for deleting aborted files

### LINKED ISSUE #2277: Aborted capture always prompts for file deletion
I just did a big upgrade of my setup include Emacs and Spacemacs, upgrading org-roam from v1 -> v2 and similar upgrades of org-ref, org-roam-bibtex, etc. So it's possible this issue isn't a direct org-roam issue, but any guidance is appreciated.

I've managed to get everything more or less working and similar to what I had before with v1. The issue I'm having now is that if I make new note using `org-roam-node-insert` or `org-roam-node-find`, and then abort the capture with `C-c C-k`, I get the following message at the bottom that I'm required to respond to by typing `y`:
```
Delete file for aborted capture? (y or n)
```
Previously this was the default before, and it's pretty annoying to have to type this every time. How can I restore the previous behavior where it would automatically delete the file?

Hugo-Heagren:
* org-roam-capture.el (org-roam-capture-delete-aborted-files): new
custom var
(org-roam-capture--check-aborted-delete): util function to check
whether we should delete the file after an aborted capture.
(org-roam-capture--finalize): use
`org-roam-capture--check-aborted-delete' rather than `yes-or-no-p'.

-------

I abort captures pretty regularly but I always answer the same to the prompt which asks whether I want to delete the aborted file -- I always want to delete it. This is a (tiny) friction which could be easily customised away. This PR makes that possible.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -343,6 +343,24 @@ The hooks must not move the point."
   :group 'org-roam
   :type 'hook)
 
+(defcustom org-roam-capture-delete-aborted-files 'ask
+  "What to do when a new capture is aborted.
+When running `org-capture-kill':
+- if `ask' (the default) ask whether to delete the new file.
+- if t always delete the file
+- if nil never delete
+- any other value is called as a function and should return one of the
+  above values.
+
+Internally, this functionality is controlled by the function
+`org-roam-capture--check-aborted-delete'."
+  :group 'org-roam
+  :type '(choice
+          (const :tag "Ask whether to delete the new file" ask)
+          (const :tag "Always delete the file" t)
+          (const :tag "Never delete the file" nil)
+          (function :tag "Call with no arguments to get one of the above values")))
+
 (defvar org-roam-capture-preface-hook nil
   "Hook run when Org-roam tries to determine capture location of the node.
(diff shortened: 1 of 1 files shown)


m-e-leypold:
Just stumbled over the very same problem. What's keeping this PR from progressing?

hpfr (changes requested):
Custom variables that can invoke arbitrary functions need to be marked risky. An idiomatic approach would be something like `org-roam-capture-delete-after-abort-predicate` (automatically marked as risky due to `-predicate`) where the default would be a function with the current prompt (`org-roam-capture-delete-after-abort-p`?), and offering `ignore` and `always` as additional choices.

An upstream example:
```elisp
(defcustom copy-region-blink-predicate #'region-indistinguishable-p
  "Whether the cursor must be blinked after a copy.
When this condition holds, and the copied region fits in the
current window, `kill-ring-save' will blink the cursor between
point and mark for `copy-region-blink-delay' seconds."
  :type '(radio (function-item region-indistinguishable-p)
                (function-item :doc "Always blink point and mark." always)
                (function-item :doc "Never blink point and mark." ignore)
                (function :tag "Other predicate function"))
  :group 'killing
  :version "29.1")
```

PR #2359, opened 2023-06-27
Allow word boundary regex to be customized for unlinked reference

### LINKED ISSUE #1290: org-roam-unlinked-references doesn't work for Chinese sentence
#### Steps to Reproduce

Imagine I have two files like this
```
#+ TITLE: 二三四

This is 二三四
```

```
#+ TITLE: 四五六

一二三四五六
```

1. Load doom-emacs
2. Go to the roam file  二三四
3. M-x org-roam-unlinked-references
4. Nothing happens
#### Expected Results

The 二三四 of 一二三四五六 linked the 二三四

#### Actual Results

The unlinked reference buffer appears with none found references.

### Environment
```
- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30)
 of 2020-09-19
- Framework: Doom
- Org: Org mode version 9.4 (nil @ /home/ualnomis/.emacs.d/.local/straight/build-27.1/org-mode/)
- Org-roam: 1.2.2
```

okomestudio:

The unlinked references are searched for by ripgrep using a simple regex pattern `(\b%s\b`). This works fine with English, but in a language for which the concept of regex word boundary doesn't apply (e.g., CJK), the pattern doesn't work. The issue https://github.com/org-roam/org-roam/issues/1290 for Chinese is an example.

This PR introduces `org-roam-unlinked-references-word-boundary-re` custom variable so that users can modify the regex pattern used by ripgrep.

For example, adding a pattern like this

``` emacs-lisp
(set-q org-roam-unlinked-references-word-boundary-re
       "|(\\b%1$s\\b|(?<=[^\x20-\x7e\xff61-\xff9f])%1$s(?=[^\x20-\x7e\xff61-\xff9f]))")
``` 

the regex would work sufficiently well in both English and Japanese.

This PR should fix #1290, though improving regex pattern is delegated to users.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -677,7 +677,7 @@ References from FILE are excluded."
                                (mapconcat (lambda (glob) (concat "-g " glob))
                                           (org-roam--list-files-search-globs org-roam-file-extensions)
                                           " ")
-                               (format " '\\[([^[]]++|(?R))*\\]%s' "
+                               (format " \"\\[([^[]]++|(?R))*\\]%s\" "
                                        (mapconcat (lambda (title)
                                                     (format org-roam-unlinked-references-word-boundary-re
                                                             (shell-quote-argument title)))
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -86,6 +86,15 @@ applied in order of appearance in the list."
   :group 'org-roam
   :type 'hook)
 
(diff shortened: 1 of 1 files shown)


okomestudio:
To avoid clutter, closing this PR in favor of https://github.com/org-roam/org-roam/issues/2416.

Issue #2365, opened 2023-07-14
Update org-roam-extract-subtree to insert a link to the newly created node


Haxxflaxx:
### Brief Abstract
It would be useful if org-roam-extract-subtree left behind a link to the newly created node.

### Long Description
I've been looking at adding some advice around the org-roam-extract-subtree function, as I in most cases want to reference the note from where I extracted it from.
I think this would be fairly simple to do, though maybe a bit fragile the way I'm envisioning it (my elisp skills are meager at best). I believe it would be easier and more robust to update the function itself, as well as I could see how other people could benefit from it too.

### Proposed Implementation (if any)
At the end of the function we could return to the starting buffer and insert a roam link at point using the stored node name.
If we don't want to change the default behavior, then maybe take an argument, or look at some org-roam-extract-subtree-insert-link variable, that we can set.

### Please check the following:

- [X] No similar feature requests

jmay:
I've been thinking about the same thing. Perhaps enable this behavior when the function is invoked with a prefix argument? That would leave the current behavior unchanged when no argument is provided.

Should the created link be inserted alone on a line where the previous entry was extracted? Or create a header at the same level as the extracted entry?

Haxxflaxx:
Good point with the heading. From an outlining standpoint you probably would want the heading to remain. Also when using org-transclusion you would want the structure to be the same.

jmay:
Maybe there should be a hook function here. After a node has been extracted, pass that information to the hook which can inject a text link, or a heading, or a `#+transclude:` line.

jmay:
@Haxxflaxx Does this do what you need?

With the PR above, you can inject references with
```
(add-hook 'org-roam-post-extraction-functions #'my/link-extracted-node)

(defun my/link-extracted-node (node)
  "Paste link to NODE at current point."
  (let* (
         (id (org-roam-node-id node))
         (description (org-roam-node-formatted node))
         )
    (org-insert-heading)
    (org-do-demote)
    (insert (org-link-make-string
             (concat "id:" id)
             description))
    ))

(defun my/transclude-node (node)
  "Transclude NODE at current point."
  (let* (
         (id (org-roam-node-id node))
         (description (org-roam-node-formatted node))
         )
    (insert (format "#+transclude: %s :level %d"
                    (org-link-make-string
                     (concat "id:" id)
                     description)
                    (+ 1 (org-outline-level))))
    ))
```

Haxxflaxx:
[...]

PR #2366, opened 2023-07-18
feature: add an abnormal hook for org-roam-extract-subtree #182

jmay:
https://github.com/org-roam/org-roam/issues/2365

Refactoring: extracted pieces of `org-roam-extract-subtree` to utility functions.
Added a `defcustom` for the hook (abnormal hook, takes a parameter). Default is an empty hook, so no change in standard behavior.

For consideration: include 2 sample hook functions, unused by default: one to insert a link to the extracted node, one to insert a `#+transclude:` line.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -118,6 +118,11 @@ Should take no arguments, prompt the user, and return a string."
   :group 'org-roam
   :type 'function)
 
+(defcustom org-roam-post-extraction-functions ()
+  "List of functions to be run when a node is extracted to a separate file."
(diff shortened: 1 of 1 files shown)


jmay:
Hey @jethrokuan can you see why this test case fails under emacs 27? I haven't use buttercup before, I'm probably relying on some behavior that doesn't exist in older emacsen.

Also for the new test file `test-org-roam-extract.el` I just replicated the file header from the other tests. Not sure if that's the correct practice, I don't want to invoke your name in there without an OK from you.

PR #2380, opened 2023-08-14
Allow customization of the org-roam visibility function

### LINKED ISSUE #2379: Support frame-based workflow for the org-roam buffer
### Brief Abstract

The org roam persistent buffer does not really support a workflow that makes use of frames.

### Long Description
Currently the org-roam buffers visibility is defined as follows:
```
(define-inline org-roam-buffer--visibility ()
  "Return the current visibility state of the persistent `org-roam-buffer'.
Valid states are 'visible, 'exists and 'none."
  (declare (side-effect-free t))
  (inline-quote
   (cond
    ((get-buffer-window org-roam-buffer) 'visible)
    ((get-buffer org-roam-buffer) 'exists)
    (t 'none))))
```

This makes the assumption that if the org-roam buffer is not part of the active frame that it is not visible and therefore should not be updated. It should be possible to do better than this, but it may require some platform specific things and external tools.

In my package frame-mode, I have a function that relies on xwinfo to determine if a frame is viewable:

https://github.com/IvanMalison/frame-mode/blob/ae2366969927c9f89ea07c999bef382b0b47cac1/frame-mode.el#L83

This is obvious not only platform, but display server (XOrg vs Wayland etc.) specific, which is not ideal. Still, I think that maybe the move here is to allow the user to specify a visibility function and default to the current behavior.

colonelpanic8:
Fixes #2379

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -307,7 +307,7 @@ To toggle its display use `org-roam-buffer-toggle' command.")
 (defun org-roam-buffer-toggle ()
   "Toggle display of the persistent `org-roam-buffer'."
   (interactive)
-  (pcase (org-roam-buffer--visibility)
+  (pcase (funcall org-roam-buffer-visibility-fn)
     ('visible
      (progn
        (quit-window nil (get-buffer-window org-roam-buffer))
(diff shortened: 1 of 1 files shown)


hpfr:
Sorry this was missed @colonelpanic8, what do you think of the proposed approach in https://github.com/org-roam/org-roam/pull/2487?

colonelpanic8:
not opinionated.

hpfr:
Ok, well if you feel like applying that (https://github.com/org-roam/org-roam/pull/2487#issuecomment-3021073829) here I can proceed here instead, looks like it’s been open longer anyway. I’ll get one or the other merged before too long

hpfr:
Superseded by https://github.com/org-roam/org-roam/pull/2594, reviews appreciated

PR #2388, opened 2023-09-02
Take org-roam-node as argument to #'org-roam-refile

pestctrl:
With this change, one can programmatically refile headings to org-roam-nodes.


pestctrl:
ping

PR #2389, opened 2023-09-02
fix(node): allow all diacritical marks in slugs

hpfr:
2d586516999d limited the removal of nonspacing characters to combining
diacritical marks targeting the Latin alphabet because removing marks
for other alphabets results in undesirable changes to the semantics of
the filename.

Even for the Latin alphabet, though, removing these accents is not
necessarily desirable and can change meaning. Filesystem support for
Unicode is widespread. If users can type diacritical marks for their
language into their node titles, they can likely type the same marks
when managing files at a command line shell. Furthermore, multilingual
users will still end up with non-ASCII-compatible filenames if their
node titles happen to include letters from other languages.

Therefore, be consistent across alphabets by default, and preserve all
alphanumeric Unicode characters in filenames, regardless of whether
those characters happen to be Latin-based. `string-glyph-compose` is
retained, so NFC normalization still happens. While we’re at it, tidy
the function as a whole.

Ref: https://github.com/org-roam/org-roam/pull/1460

Cc @khinsen (#230): Any thoughts on this? The above PR has already introduced
NFC normalization and diacritics, but not for the Latin alphabet. I’m curious
how your dual-boot setup looks where you’re running into normalization issues
between macOS and Linux. Are you mounting a filesystem other than HFS or APFS in
macOS that doesn’t have an `nfc` mount option?


meedstrom:
For Latin-based alphabets, there are some pragmatic benefits in coercing to pure ASCII, so that the string is typeable by any QWERTY/AZERTY keyboard in the world with no need to configure "regional settings" on the device. 

Even more when you boot up a Linux recovery disk and carry out some data repair in a TTY or busybox environment... 

I bet many users like the current behavior, but perhaps this PR would garner more interest if you made it configurable ;-)

------

Aside: That just gave me an idea.

Perhaps the cleanest algorithm is not to delete specific diacritics chosen from the range U+0300 to U+0331 because they happen to commonly occur with Latin characters, but to check whether the *preceding char* belongs to ASCII, and then delete the diacritic no matter what it is.

hpfr:
> For Latin-based alphabets, there are some pragmatic benefits in coercing to pure ASCII, so that the string is typeable by any QWERTY/AZERTY keyboard in the world with no need to configure "regional settings" on the device.
> 
> Even more when you boot up a Linux recovery disk and carry out some data repair in a TTY or busybox environment...

This seems a little contrived. The user has already demonstrated they can type (or copy) the string by creating the title in the first place. In the default capture templates, the ASCII-based date precedes the title in the filename, so it’s still pretty likely that the filename can be typed with shell completion or globs. I have a hard time imagining people fiddling with org-roam files in a Busybox-based environment. If they are dealing with individual org-roam files (i.e. not recursively copying the entire org-roam directory), they realistically have Emacs available, and Emacs has plenty of support for multilingual input. So I think we should prioritize accuracy to titles in filenames over being able to manually type out individual filenames in a minimal initramfs.

> I bet many users like the current behavior, but perhaps this PR would garner more interest if you made it configurable ;-)

It’s configurable by redefining the method, so this is just about simplifying the default for consistency across scripts.

meedstrom:
I understand, it’s surely a matter of who you are and what you tend to do with tech. In my own experience, I definitely wouldn’t want to rely on this assumption:

> The user has already demonstrated they can type (or copy) the string by creating the title in the first place.

The user has actually only demonstrated that they could at some point in the past, on one of their devices. I frequently set up new devices and rarely bother to configure regional settings.

Anyway, just saying it probably should be easy to configure, and my unqualified guess is that the majority would prefer to stick with a ”filename-safe” charset for filenames and if I’m right, a sensible default is opt-in to allow diacritics, not opt-out, but I don’t have a horse in this race.

Your point about tab completion is a good one.

> so this is just about simplifying the default for consistency across scripts.

I’m curious: what does this mean? What kind of scripts would be inconsistent?

hpfr:
> The user has actually only demonstrated that they could at some point in the past, on one of their devices. I frequently set up new devices and rarely bother to configure regional settings.

I mean, if you have Emacs, you can definitely manage non-ASCII filenames. Any other platform default file manager GUI should work too. So this is really only an issue in the Busybox scenario which I think is not very realistic.

> I’m curious: what does this mean? What kind of scripts would be inconsistent?

Script is just a Unicode and Emacs term for a collection of characters composing a writing system. Some scripts aren’t alphabetic so it’s technically more accurate than alphabet, I guess. What I meant was currently Japanese, Korean, etc are faithfully reproduced in filenames, but various Latin languages are not. Since we are already using Unicode in filenames with non-Latin characters, there is no point trying to coerce Latin-based titles to ASCII.

Nobody else seems to care though so I’ll close this for now.

PR #2394, opened 2023-09-26
Link Martin Edström's knowledge base

meedstrom:

One more example.  Just a suggestion, but it's fine if you don't want to.


Delapouite:
Thanks for the link.
What's nice about it, is that on top of using Org-Roam, the content is about Emacs / Org, which makes it quite "meta" :)

PR #2400, opened 2023-11-26
Add exclude patterns to org-roam-graph

braun-steven:
Define `org-roam-graph-exclude-patterns` which allows filtering the graph nodes by title/filenames.

The org-roam directory may include files that should not be visualized in the org-roam-graph. This change allows defining string patterns which are used to filter nodes in the node collection process of `org-roam-graph--dot`.


hpfr:
I think we can implement this in a more general way, something like a hook variable where arbitrary filtering functions can be called with id/file/title. That way people aren’t limited to regexps matched against both file and title

hpfr:
Closing this due to lack of response, if someone else wants to pick this up just open a new PR referencing https://github.com/org-roam/org-roam/issues/2434 and keeping in mind https://github.com/org-roam/org-roam/pull/2400#issuecomment-3021637001

Issue #2407, opened 2024-01-25
Node title with single quote(s) breaks ripgrep for unlinked references


okomestudio:
### Description

When constructing the ripgrep shell command for grepping unlinked references, each title (and aliases) of that node gets sanitized by ``shell-quote-argument``, which (properly) escapes quotes. However, escaped single quote “breaks” the constructed ripgrep shell command, since the regex gets expanded within single-quoted argument. For example, for title/alias of “Don’t Repeat Yourself” and “DRY”, the generated ripgrep command looks like this: 

    rg -L -o --vimgrep -P -i -g "*.org" -g "*.org.gpg" -g "*.org.age" \
        '\[([^[]]++|(?R))*\]|(\bDon\'t\ Repeat\ Yourself\b)|(\bDRY\b)' /org/roam/

Since an escaped single quote cannot appear within single quoted token (in Bash, at least), this command appears to silently fail, and the unlinked references section produces empty result because of this.

A solution may be to use double quotes to wrap the constructed regex. There, escaped characters, including single quote, should be interpreted as intended. A PR implementing a fix may follow.

#### Steps to Reproduce

1. Create a node with title text including single quote character (``'``)
2. Visit that node with org-roam
3. Run ``org-roam-buffer-toggle``
4. See that no entries appear under the Unlinked References section

#### Backtrace

No error occurs that produces a backtrace. The issue appears to be silently ignored.

#### Expected Results

A list of node filename and the paragraph including the node title text/aliases will appear.

#### Actual Results

No entries appear under the Unlinked References section of the Org Roam buffer.

### Environment

- Emacs: GNU Emacs 29.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.16.0)
 of 2024-01-18
- Framework: N/A
- Org: Org mode version 9.7-pre (release_9.6.15-1044-gaab2c9 @ /home/taro/.config/emacs/straight/build/org/)
- Org-roam: v2.2.2-44-g8667e44-dirty
- sqlite-connector: sqlite-builtin

okomestudio:
I don't believe this bug has been fixed. Given that the fix seems relatively simple (see the PR), I'd argue that this is worth taking care of, despite being marked stale.

dustinfarris:
> I don't believe this bug has been fixed. Given that the fix seems relatively simple (see the PR), I'd argue that this is worth taking care of, despite being marked stale.

Agreed

PR #2408, opened 2024-01-25
(fix): Use double quotes for proper single quote escape

okomestudio:

This PR provides a fix for https://github.com/org-roam/org-roam/issues/2407.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -664,9 +664,12 @@ This is the ROW within FILE."
           (mapconcat (lambda (glob) (concat "--glob " glob))
                      (org-roam--list-files-search-globs org-roam-file-extensions)
                      " ")
-          (format " \"\\[([^[]]++|(?R))*\\]%s\" "
+          (format " '\\[([^[]]++|(?R))*\\]%s' "
                   (mapconcat (lambda (title)
-                               (format "|(\\b%s\\b)" (shell-quote-argument title)))
+                               (format "|(\\b%s\\b)"
+                                       (mapconcat #'shell-quote-argument
+                                                  (split-string title "'")
+                                                  "'\"'\"'")))
                              titles ""))
           (shell-quote-argument org-roam-directory)))
 
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -668,7 +668,7 @@ References from FILE are excluded."
                                (mapconcat (lambda (glob) (concat "-g " glob))
                                           (org-roam--list-files-search-globs org-roam-file-extensions)
                                           " ")
-                               (format " '\\[([^[]]++|(?R))*\\]%s' "
+                               (format " \"\\[([^[]]++|(?R))*\\]%s\" "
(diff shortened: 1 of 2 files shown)


Delapouite:
Hello.

After merging a related PR about this area of the code (https://github.com/org-roam/org-roam/pull/2411) and extracted the construction of the rg command in order to have a way to test it (https://github.com/org-roam/org-roam/pull/2449), I inadvertently introduced a merge-conflict in your branch.

I'm sorry for the trouble. 

This lead me to a few points:

- you mention that your fix is now capable to support titles with a single quote which is great, thanks.
- but what happen the other way around, if the node title now contains a double quote? Is it supported?
- other corner case, what if the node title contains a dollar sign resulting by accident to an environment variable (i.e. `$SHELL`). Will the fact that it's now a double-quote pattern will perform the expansion of this env var?

If you have time to give me some clues, it will be great. Otherwise I'll make my own investigation about this soon.

In any case, now that the resulting rg command can be tested in the newly introduced test file, it will be a nice addition to also cover these situations.

Thanks.

Delapouite:
I now see that you already did a tremendous job on the matter in another PR. https://github.com/org-roam/org-roam/pull/2417

(sorry I've started to visit and merge PRs in reverse chronological order)

So I guess most of my questions are probably covered, so I'll have a look to it.

okomestudio:
Thanks for getting the ball rolling with all the actions, @Delapouite ! It's very nice to see the "revival" of the tool that's become essential to my Emacs work flow.

Please feel free to let me know how we should proceed with work that I have submitted, this one and another you mentioned already. I also have some time to work on them myself if that helps take off a bit of work load off of you.

Again, thanks for your time!

okomestudio:
I merged the current main branch and resolved the conflict. I've added a couple test cases to check the regex construction for both single and double quotes.

> * you mention that your fix is now capable to support titles with a single quote which is great, thanks.
> * but what happen the other way around, if the node title now contains a double quote? Is it supported?

These should be fine. The double quote gets properly escaped, as seen in the newly added test case.

> * other corner case, what if the node title contains a dollar sign resulting by accident to an environment variable (i.e. $SHELL). Will the fact that it's now a double-quote pattern will perform the expansion of this env var?

I'll see if the potential environment variable expansion causes any issue.

okomestudio:
On the point about special characters like "$" used as part of input... the regex special characters can be special characters in the shell, so indeed it may be safer to stick to single quotes for enclosure (so that they lose their special meaning when used in the shell).

In practice I haven't been bitten by this, but a safer approach may be to escape single quote characters within single quoted enclosure. I'll see if this improves the PR.

okomestudio:
The latest commit "escapes" explicitly single quotes using double quotes in order to keep the regex pattern shell safe. All the points of three raised in https://github.com/org-roam/org-roam/pull/2408#issuecomment-2206900792 get proper treatment.

Thanks in advance for code review.

Issue #2410, opened 2024-02-01
Unlinked References doesnt work if the `org-roam-directory` has a space in it.


vikram-mandyam:
### Description
In the function `org-roam-unlinked-references-section` the `org-roam-directory` variable is concated to form the `rg_command`. This doesn't cover any spaces in the `org-roam-directory` path. 
So if the path is `/Users/user123/MyDrive/My Files/org-mode/org-roam`, due to the space in `My Files` rg returns an error: 
```
rg: /Users/user123/MyDrive/My: No such file or directory (os error 2)
rg: Files/org-mode/org-roam: No such file or directory (os error 2)
```
This is due to the [lines here](https://github.com/org-roam/org-roam/blob/8667e441876cd2583fbf7282a65796ea149f0e5f/org-roam-mode.el#L667-L675):
```elisp
           (rg-command (concat "rg -L -o --vimgrep -P -i "
                               (mapconcat (lambda (glob) (concat "-g " glob))
                                          (org-roam--list-files-search-globs org-roam-file-extensions)
                                          " ")
                               (format " '\\[([^[]]++|(?R))*\\]%s' "
                                       (mapconcat (lambda (title)
                                                    (format "|(\\b%s\\b)" (shell-quote-argument title)))
                                                  titles ""))
                               org-roam-directory))
```

#### Steps to Reproduce
Set the `org-roam-directory` to a directory containing spaces like shown above.

### org-roam-diagnostics

Copy info below this line into issue:
- Emacs: GNU Emacs 28.2 (build 2, aarch64-apple-darwin23.2.0, NS appkit-2487.30 Version 14.2.1 (Build 23C71))
 of 2024-01-09
- Framework: N/A
- Org: Org mode version 9.4.6 (9.4.6-796-ge68375.dirty-elpa @ ....)
- Org-roam: 2.2.2- sqlite-connector: sqlite

vikram-mandyam:
I have created a pull request for fixing this - please have a look https://github.com/org-roam/org-roam/pull/2411

PR #2411, opened 2024-02-01
Fix Issue #2410 - Unlinked References doesnt work if the org-roam-directory has a space in it. 

vikram-mandyam:
Fix Issue #2410 
In the function `org-roam-unlinked-references-section` the `org-roam-directory` variable is concated to form the `rg_command`. This doesn't cover any spaces in the `org-roam-directory` path.
So if the path is `/Users/user123/MyDrive/My Files/org-mode/org-roam`, due to the space in My Files rg returns an error:
```
rg: /Users/user123/MyDrive/My: No such file or directory (os error 2)
rg: Files/org-mode/org-roam: No such file or directory (os error 2)
```

This is due to the [lines here](https://github.com/org-roam/org-roam/blob/8667e441876cd2583fbf7282a65796ea149f0e5f/org-roam-mode.el#L667-L675):
```elisp
(rg-command (concat "rg -L -o --vimgrep -P -i "
                               (mapconcat (lambda (glob) (concat "-g " glob))
                                          (org-roam--list-files-search-globs org-roam-file-extensions)
                                          " ")
                               (format " '\\[([^[]]++|(?R))*\\]%s' "
                                       (mapconcat (lambda (title)
                                                    (format "|(\\b%s\\b)" (shell-quote-argument title)))
                                                  titles ""))
                               org-roam-directory))
```

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -672,7 +672,7 @@ References from FILE are excluded."
                                        (mapconcat (lambda (title)
                                                     (format "|(\\b%s\\b)" (shell-quote-argument title)))
                                                   titles ""))
-                               org-roam-directory))
+                               (shell-quote-argument org-roam-directory)))
            (results (split-string (shell-command-to-string rg-command) "\n"))
            f row col match)
       (magit-insert-section (unlinked-references)


Delapouite:
Thanks!

I added follow-up commits in https://github.com/org-roam/org-roam/pull/2449 so that the output of the problematic `rg` command can be tested and tweaked by the end user.

PR #2413, opened 2024-02-03
Added error checking org-roam-capture--setup-target-location

dmgerman:
If a template is invalid org-roam-capture--setup-target-location does not set any target. Then (goto) fails due to lack of buffer.

For example, this template:

        '(
          ("m"
           "movie" plain
           "*** ${title}\n :PROPERTIES:\n :RATING:\n:END:\n%t\n"
           :target (file+olp+datetree "Weekly.org" "abc")
           ))))

generates the error:

    save-excursion: Wrong type argument: integer-or-marker-p, nil

which is not very helpful

This patch adds an error message in such a case. It will report instead:

   Invalid org-roam capture specification (file+olp+datetree "Weekly.org" "abc")

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -548,10 +548,10 @@ capture target."
        (setq position! (goto-char (point-min))))
       
       (`(file+head+olp ,path ,head ,olp)
-       (widen)
+       (setq position! (point-min))
+       ;(widen)
        (when new-file-p
          (insert (org-roam-capture--fill-template head 'ensure-newline)))
-       (setq position! (point-min))
        (let ((m (org-roam-capture-find-or-create-olp olp)))
          (goto-char m)))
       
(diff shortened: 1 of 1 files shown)


dmgerman:
One extra comment. I am not sure if this is an error or on purpose. The documentation of org-mode says that the destionation is file+olp+datetree, but that destination is not handled by org-roam. Instead, file+datetree is used. It might be worthwhile to document this discrepancy

dmgerman:
I am going to close this PR. Since I submitted I have kept working on the main branch of my repo revamping/improving the templating system of org-roam and to make more compatible with org's template system.

One of the most important changes I have made is not to ask for a node title before a org-roam template. Sometimes I already know which node I want to go jump to, so don't ask me! ;)

Anyways, if there is interested on revamping the template system, see my main branch.

Issue #2416, opened 2024-02-04
Refactor unlinked references section code for more customizability


okomestudio:
### Brief Abstract

I propose refactoring the `org-roam-unlinked-references-section` function to give users more options to customize the section behavior through custom variables and advising functions used in key steps within the function without replacing it in its entirety.

### Long Description

The unlinked references section of `org-roam` buffer is constructed through a basic algorithm using ripgrep and filtering, much of it hard-coded which prevents users to tune any part of the behavior. I am fond of unlinked references for serendipitous discovery, but the current implementation limits the use cases to the simplest word matching with word-level tokenization.

For example, I have the following issues with the current implementation.

-   The hard-coded word boundary regex in `org-roam` currently doesn’t accommodate regex parsing documents written in one of the CJK languages, which needs more than `\b` for identifying word tokens

-   The regex matching isn’t flexible enough to handle variant characters, e.g., “Denny's” and “Denny’s” are considered different, when a regex pattern matching them is straightforward to write

Currently, implementing these would require replacing the
`org-roam-unlinked-references-section` function in its entirety. After the proposed refactoring, these can be solved though customizing variables and advising functions without intrusive code change to the source code.

The `org-roam` philosophy when it comes to customization seems to be to let users replace the whole function(s) after feature loading (see, e.g., [this discussion](https://org-roam.discourse.group/t/solved-exclude-properties-drawer-from-unlinked-references-rg/2025/12). While this might be sufficient and incurs lower maintenance cost, by refactoring relevant code judiciously, users can tailor the code behavior more to their needs through custom variables and function advising while still benefiting from the overall logic, which is be mostly sound. Pure refactoring maintains the current logic, leaving the existing users unaffected by it.

### Proposed Implementation (if any)

A proposed implementation is given in https://github.com/org-roam/org-roam/pull/2417.

### Please check the following:

- [X] No similar feature requests

okomestudio:
The merge conflicts for https://github.com/org-roam/org-roam/pull/2417 have been resolved. I'd like to make some comments on the PR.

While the refactoring performed in the PR is not very aggressive in my opinion, the primary intention is to give the users more opportunities to advise any function they want within the top-level magit section function for rendering unlinked references section. Advising, however, is rather intrusive and aggressive in its nature.

[...]

PR #2417, opened 2024-02-04
(refactor)org-roam-unlinked-references-section: make steps advisable

okomestudio:

This PR implements a proposed refactoring of the `org-roam-unlinked-references-section` function, so that

-   The word boundary regex is supplied via the custom variable `org-roam-unlinked-references-word-boundary-re` to allow more complex logic
-   The functions `org-roam-unlinked-references-title-regex`, `org-roam-unlinked-reference-apply-word-boundary-re`, `org-roam-unlinked-references-file-glob-args`, and `org-roam-unlinked-references-result-filter-p` are refactored out to allow targeted advising for custom behavior
-   The `org-roam-unlinked-references-max-results-count` custom variable limits result to avoid “freezing” when too many results are matched
-   The inner loop rendering the preview line is passed the values for file, row number, and col number from the previous iteration to allow more controlled rendering, in case when the matched term appears multiple times within the same line

The overall behavior of the refactored function does not change. The rationale for this change is described in https://github.com/org-roam/org-roam/issues/2416.
](https://github.com/org-roam/org-roam/issues/2416).

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -86,6 +86,25 @@ applied in order of appearance in the list."
   :group 'org-roam
   :type 'hook)
 
+(defcustom org-roam-unlinked-references-word-boundary-re "|(\\b%1$s\\b)"
+  "The word bounday regex used by ripgrep for unlinked references.
+
+In such languages as CJK, the regex's word boundary (\b) does not
+correctly determine how words and phrases should be tokenized.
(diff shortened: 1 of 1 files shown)


okomestudio:
Instead of this proposed enhancement, I will take advantage of the plugin architecture and provide the feature as a library. Please take a look at https://github.com/okomestudio/org-roam-cjk if interested.

Issue #1937, opened 2021-11-02
ROAM_REFS is being inserted into file top PROPERTIES when capturing ref as an entry


AloisJanicek:
### Description
When I attempt to capture some ref from browser as entry (as level 1 heading inside org-roam file node), the ROAM_REFS gets inserted into its file-top :PROPERTIES: drawer in addition to correct expansion from the capture template itself.

I tracked down this behavior to `org-roam-protocol--insert-captured-ref-h` and its invocation in `org-roam-capture--prepare-buffer` through `(run-hooks 'org-roam-capture-new-node-hook)` 

#### Steps to Reproduce
1. Have some org-roam file node of title "Some title"
2. Set following ref capture template:
```lisp
(setq
 org-roam-capture-ref-templates
 `(("r" "ref" entry (function
                     (lambda ()
                       (format
                        "* ${title} :ref:\n:PROPERTIES:\n:ID: %s\n:ROAM_REFS: ${ref}\n:END:"
                        (org-id-uuid))))
    :target (node "Some title")
    :unnarrowed t
    :immediate-finish t)))
```
3. Capture url/title from browser via usual bookmarklet using template "r"

#### Expected Results

ROAM_REFS gets added only to the newly created level 1 org entry in accordance with capture template "r"

#### Actual Results

ROAM_REFS gets added to the file top PROPERTIES drawer in addition to expansion of template "r"

### Environment

latest org-roam commit, up-to-date doom-emacs, emacs 28.60 native-comp

sg-qwt:
I hit the exact issue today. All my roam_refs are essentially duplicated by this. Do you find a way to not automatically insert ROAM_REFS at file level?

sg-qwt:
`(add-hook 'org-roam-capture-new-node-hook #'org-roam-capture--insert-captured-ref-h)`

This hook is automatically added for capture which caused the problem. I worked around by setting org-roam-capture-new-node-hook to nil.

Some relevant discussion [here](https://org-roam.discourse.group/t/how-to-prevent-org-roam-capture-from-adding-id-property/2669/5)

martyone:
For me this change fixed the behavior for `file+olp` and `file+head+olp` targets:

```diff
diff --git a/org-roam-capture.el b/org-roam-capture.el
index 8ec9170..4c9b328 100644
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -495,9 +495,8 @@ Return the ID of the location."
[...]

PR #2420, opened 2024-02-12
(feat)capture: Add file+ref+title target

martyone:
Suitable for maintaining a list of references in a single specific file, possibly in cooperation with org-bibtex.

Related to issue #1937.

### DIFF
--- a/extensions/org-roam-protocol.el
+++ b/extensions/org-roam-protocol.el
@@ -96,7 +96,12 @@ See `org-roam-capture-templates' for the template documentation."
                                                         (string :tag "  File")
                                                         (string :tag "  Head Content")
                                                         (list :tag "Outline path"
-                                                              (repeat (string :tag "Headline"))))))
+                                                              (repeat (string :tag "Headline"))))
+                                                  (list :tag "File & Reference & Title"
+                                                        (const :format "" file+ref+title)
+                                                        (string :tag "  File")
+                                                        (string :tag "  Reference")
+                                                        (string :tag "  Title"))))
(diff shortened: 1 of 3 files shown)


dmgerman:
I did a similar patch but for nodes. See https://www.reddit.com/r/orgmode/comments/1b1msff/improved_orgroam_templates/

I wonder if we might need to start maintaining a "org-roam-NG" branch in a different repo. So we can have some testing of these features before if/when they get merged into the main repo.

Perhaps we should think about what the main issues of current templates are rather than patch the current system

martyone:
Rebased

martyone:
Rebased, fixed conflicts, but not tested yet against the updated upstream.

hpfr:
Sorry about the wait, thanks for your patience.

I’m not a heavy capture user personally and we’ve been burned by capture recently (https://github.com/org-roam/org-roam/issues/2591), so I don’t have this PR planned for [2.4.0](https://github.com/org-roam/org-roam/milestone/13) as of yet, but we’ll get to it eventually. Most of what we have been working through in the past year is straightforward bug fixes/performance improvements, build system/testing/CI, or docs.

I’m a little concerned about adding more custom target support, especially ones rigidly tied to particular note types (in this case, reference notes). I see in https://github.com/org-roam/org-roam/issues/1937#issuecomment-1937347850 you implemented a fix for the existing `olp` targets to improve single-file ref capture. Is that still viable?

martyone:
Use of the existing `olp` targets was my first attempt, but soon I hit the drawback that they do not allow capturing additional notes based on the captured URL. So I created the dedicated target instead.

PR #2432, opened 2024-05-02
Compatibility with latest org-id version: advise org-id-find rather than overwriting id link

ricklupton:

There have been recent changes in org-mode to improve `org-id-open` to be able to apply search strings in id: links.  But, org-roam overrides the `:follow` parameter for id: links to its own `org-roam-id-open` function and therefore misses this.

The only change that `org-roam-id-open` makes is to try calling `org-roam-id-find` before falling back on `org-id-find`.  This commit uses advice to do this directly, thereby integrating better with `org-id-open`.

There was some discussion on the org-mode list about adding a custom variable to avoid using advice, but this accomplishes the result without requiring changes in org:

https://list.orgmode.org/87jzlxjiuf.fsf@localhost/

### DIFF
--- a/org-roam-id.el
+++ b/org-roam-id.el
@@ -63,32 +63,10 @@ With optional argument MARKERP, return the position as a new marker."
         (cons (org-roam-node-file node)
               (org-roam-node-point node))))))
 
-(defun org-roam-id-open (id _)
-  "Go to the entry with id ID.
-Like `org-id-open', but additionally uses the Org-roam database."
-  (org-mark-ring-push)
-  (let ((m (or (org-roam-id-find id 'marker)
-               (org-id-find id 'marker)))
-        cmd)
-    (unless m
-      (error "Cannot find entry with ID \"%s\"" id))
-    ;; Use a buffer-switching command in analogy to finding files
-    (setq cmd
-          (or
-           (cdr
-            (assq
-             (cdr (assq 'file org-link-frame-setup))
-             '((find-file . switch-to-buffer)
-               (find-file-other-window . switch-to-buffer-other-window)
-               (find-file-other-frame . switch-to-buffer-other-frame))))
-           'switch-to-buffer-other-window))
-    (if (not (equal (current-buffer) (marker-buffer m)))
-        (funcall cmd (marker-buffer m)))
-    (goto-char m)
-    (move-marker m nil)
-    (org-show-context)))
+(defalias 'org-roam-id-open 'org-id-open
+  "Obsolete alias - use `org-id-open' directly.")
 
-(org-link-set-parameters "id" :follow #'org-roam-id-open)
+(advice-add 'org-id-find :before-until #'org-roam-id-find)
 
 ;;;###autoload
 (defun org-roam-update-org-id-locations (&rest directories)


bpanthi977 (approved):

aragaer:
Current implementation fails to find links like this as backlinks. Either the backlink code has to be updated to search for all links that start with a given ID or the db code has to be updated to strip the "option" part of the link. It looks like the DB approach is easier.

nobiot:
Happy New Year to all!

I have also started using this PR as a patch on my local (just today; I never needed to fix it until now...). It is working as far as I am using it.

@Delapouite @lewisbrown @dustinfarris  @jethrokuan A gentle ping. Hoping that one of you could review and merge the PR in the new year? It's probably also a good time to bump up the version to clear up the DB issue... (in reference to https://github.com/org-roam/org-roam/issues/2485#issuecomment-2495398128).

nobiot:
@bpanthi977 has approved the PR; so, perhaps it's a matter of simple merge operation (and count up the version in the source of `org-roam.el`?).

nobiot:
> Current implementation fails to find links like this as backlinks. Either the backlink code has to be updated to search for all links that start with a given ID or the db code has to be updated to strip the "option" part of the link. It looks like the DB approach is easier.

What do you refer to as "current implementation"? PR #2432 or the current main branch of Org-roam? If this PR has an issue, perhaps #2494 (my one) can be an option. I am not sure, though. This PR seems to work on my very casual use...

aragaer:
> > Current implementation fails to find links like this as backlinks. Either the backlink code has to be updated to search for all links that start with a given ID or the db code has to be updated to strip the "option" part of the link. It looks like the DB approach is easier.
> 
> What do you refer to as "current implementation"? PR #2432 or the current main branch of Org-roam? If this PR has an issue, perhaps #2494 (my one) can be an option. I am not sure, though. This PR seems to work on my very casual use...

I refer to current implementation of `org-roam-db-insert-link` and `org-roam-backlinks-get` functions. The value stored in database is "id::option" and when searching for links where the target is exactly "id" these do not match.

dustinfarris:
I’ll take a look tomorrowDustinSent from my iPhoneOn Jan 3, 2025, at 17:19, Ilya Konovalov ***@***.***> wrote:﻿


Current implementation fails to find links like this as backlinks. Either the backlink code has to be updated to search for all links that start with a given ID or the db code has to be updated to strip the "option" part of the link. It looks like the DB approach is easier.

What do you refer to as "current implementation"? PR #2432 or the current main branch of Org-roam? If this PR has an issue, perhaps #2494 (my one) can be an option. I am not sure, though. This PR seems to work on my very casual use...

I refer to current implementation of org-roam-db-insert-link and org-roam-backlinks-get functions. The value stored in database is "id::option" and when searching for links where the target is exactly "id" these do not match.

nobiot:
> I refer to current implementation of `org-roam-db-insert-link` and `org-roam-backlinks-get` functions. The value stored in database is "id::option" and when searching for links where the target is exactly "id" these do not match.

Thank you. Understood. Opened issue #2945 and sent PR #2946 for this cache db and backlink issue. Closed #2494; I am in favour of this PR because it uses the upstream `org-id-open` directly. I find the advice appropriate. This will reduce maintenance effort in long term as `org-id` upstream may be further refined.

dustinfarris:
Thanks @ricklupton for your contribution and for the discussion on the org mailing list. 

I'm a Doom user, and It took me a while to understand why this is needed.  Turns out doom has been advising both `org-id-open` _and_ `org-roam-id-open` with a custom search implementation.

https://github.com/doomemacs/doomemacs/commit/aeef05bfd75b1693eb6813107166edbaca67ca26

What a trip.

Thanks again.

PR #2441, opened 2024-06-18
Update org-roam.org

marcosbodio:
Add additional information for setting up org-roam-protocol on Mac OS when using emacs installed from Homebrew (see also https://org-roam.discourse.group/t/problem-with-org-roam-protocol/3473)

see also https://org-roam.discourse.group/t/problem-with-org-roam-protocol/3473


Delapouite:
Thanks for providing this helpful procedure after this long investigation on the forum.

PR #2443, opened 2024-06-24
Add org-roam-tag-list-all

forkrul:
List all tags and file-count per tag in a new buffer. 

This shows the weight of topics covered by a given org-roam directory, and is useful to find mistagging, explore the tag-set, and clean up.


hpfr:
Going to close this as I’m not sure there’s a widespread need and it introduces a new kind of buffer which is substantial. Feel free to open an issue for discussion. In general I think hacking on presentation of statistics via database queries is an interesting thread to explore, but we can start out-of-tree for now.

PR #2444, opened 2024-06-25
fix #2425 Prevent data loss when user has called org-roam-extract-subtree on folded org headline

### LINKED ISSUE #2425: org-roam-extract-subtree results in data loss when extracting folded headings with properties under them
#### Steps to Reproduce

1. Navigate to a heading
2. Run org-roam-extract-subtree

#### Backtrace
No error occurs

#### Expected Results

A file is created with the contents of the heading extracted.

#### Actual Results

The file is subtly pseudo-deleted. The contents are deleted from the Org buffer you're in, but they are not taken to the proper Roam file. This can result in a scare later where you believe the text has been utterly deleted when, in the extracted roam file, all you see is:
```
#+title: The failed-extract's headline text here
```

However, there is a saving grace: in the Emacs autosave file (e.g. `20240314-foo.org~`), the full file contents exist.

### Reproducibility
 - About 20% of the time? 
   - It must have happened a half-dozen to a dozen times when I've extracted subtrees, and I only have 167 nodes total, many of which were created without extraction
    -  I'll try and gain more knowledge here, as I go.

I'm not sure what's interacting here, but even the same snippet of text fails to repeat the same behaviour if I undo in my Org buffer and then run the extract again. Maybe it's extremely sensitive to the headline text (which I need to add `foo` to in order to get a new file to auto populate a distinct file name), but I doubt it.

### Environment

- Emacs: GNU Emacs 30.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0)
 of 2024-03-03
- Framework: N/A
- Org: Org mode version 9.6.15 (release_9.6.15 @ /opt/emacs-30.0.50/share/emacs/30.0.50/lisp/org/)
- Org-roam: 2.2.2- sqlite-connector: sqlite-builtin

#### Further Environment Details
This happens on every Emacs from 27 forward. I've been updating various things to see if that helps

### Open to beta-test
I'm happy to begin digging in to any targets to test. I see that my 2.2.2 has the exact same org-roam-extract-subtree as `main`, should I try a different branch?

akashpal-21:
(kill-whole-line) kills folded text if called from column 1 Make sure to unfold text before calling it --
Prevents data loss if user has decided to extract subtree on a folded org headline.

As described in issue #2425 - org-roam-extract-subtree (by calling org-roam-promote-entire-buffer) deletes folded text because (kill-whole-line) when called from column 1 kills folded text -

Proposed change

```
-- org-roam-node-old.el	2024-06-26 00:04:34.713642244 +0530
+++ org-roam-node.el	2024-06-26 00:03:41.285582526 +0530
@@ -878,6 +878,7 @@
   (org-with-point-at 1
     (let ((title (nth 4 (org-heading-components)))
           (tags (org-get-tags)))
+      (org-fold-show-all)
       (kill-whole-line)
       (org-roam-end-of-meta-data t)
       (insert "#+title: " title "\n")
```

Best,

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -878,6 +878,7 @@ node."
   (org-with-point-at 1
     (let ((title (nth 4 (org-heading-components)))
           (tags (org-get-tags)))
+      (org-fold-show-all)
       (kill-whole-line)
       (org-roam-end-of-meta-data t)
       (insert "#+title: " title "\n")


n-hebert:
Definitely worth merging.

I ran into this a lot and nearly lost quite a few notes.

dustinfarris:
Thank you @akashpal-21! ❤️

PR #2463, opened 2024-07-16
org-roam-db-insert-aliases clear link with org-link-display-format 

Cletip:
(Like the title)

Not have this type of things during the search when having the alias in org-roam-node-display-template :
A nice alias with a [[id:54819299][description of a link]]

Instead, we want this :
A nice alias with a description of a link


hpfr:
Aliases are for alternative names for a node; they shouldn’t contain links themselves. Can you describe your use case more concretely?

hpfr:
I’ll close this but feel free to reply with more info.

PR #2466, opened 2024-08-13
Depend on emacsql 4.0.0

bcc32:
See https://github.com/magit/emacsql/issues/113.  Emacsql 4.0.0 has
been released and should now be used in the package-requires header.

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -6,7 +6,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 2.2.2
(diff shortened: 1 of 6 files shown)


dustinfarris:
Hi, have you tested this locally?  And what are the implications for users not on Emacs 29?

bcc32:
Thanks for your comment.  Yes, this package is in active use with the specified emacsql dependency version, by users at my site (with this patch applied).  I'm not aware of any implications for users on older Emacs versions---[emacsql-4.0.0](https://github.com/magit/emacsql/blob/3d9622aabbf3ab925beadd12ce928370d7704573/emacsql.el#L9-L10) only requires Emacs 25.1.

Issue #2478, opened 2024-10-15
org-roam-unlinked-references--rg-command constructs path with a leading '\' causing rg to fail with directory not found


nvimtor:
### Description
Unlinked references don't seem to work for me. Trying to debug it, I ran the command `org-roam-unlinked-references--rg-command '("Site Controller")`, which returned 

```
rg --follow --only-matching --vimgrep --pcre2 --ignore-case --glob "*.org" --glob "*.org.gpg" --glob "*.org.age" '\[([^[]]++|(?R))*\]|(\bSite\ Controller\b)' \~/org/roam
```

#### Steps to Reproduce
N/A, no particular steps

#### Backtrace
N/A

#### Expected Results

The rg command should execute correctly with a proper path, e.g., `~/org/roam`

#### Actual Results

When executed, the rg command fails with an IO error due to the incorrect path `\~/org/roam`:
```
rg: ~/org/roam: IO error for operation on ~/org/roam: No such file or directory (os error 2)
```

### Environment

```
- Emacs: GNU Emacs 29.1 (build 1, aarch64-apple-darwin24.0.0, Carbon Version 170 AppKit 2566)
 of 2024-09-29
- Framework: N/A
- Org: Org mode version 9.6.6 (release_9.6.6 @ /opt/homebrew/Cellar/emacs-mac/emacs-29.1-mac-10.0/share/emacs/29.1/lisp/org/)
```

hpfr:
Will need an `expand-file-name` somewhere, plus a test, ideally. Or updating the docstring to mention that `expand-file-name` should be used, but probably better to follow Postel’s law

hpfr:
This was (inadvertently?) fixed in https://github.com/org-roam/org-roam/commit/0786f73669baa468596b29bd0f0d26956770676d

PR #2479, opened 2024-10-15
fix: expand file name for org-roam-directory

nvimtor:
I am new to both Emacs & Emacs Lisp. I do not fully understand the implications of using `expand-file-name`.

Issue #2478.

I had set my `org-roam-directory` to be `~/org/roam` and that made `rg` silently fail. Running 

```
(org-roam-unlinked-references--rg-command '("Site Controller"))
```

then returned

```
rg --follow --only-matching --vimgrep --pcre2 --ignore-case --glob "*.org" --glob "*.org.gpg" --glob "*.org.age" '\[([^[]]++|(?R))*\]|(\bSite\ Controller\b)' \~/org/roam
```
which caused `rg` to fail with:

```
rg: ~/org/roam: IO error for operation on ~/org/roam: No such file or directory (os error 2)
```

It was solved by expanding the `~/org/roam` path.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -668,7 +668,7 @@ This is the ROW within FILE."
                   (mapconcat (lambda (title)
                                (format "|(\\b%s\\b)" (shell-quote-argument title)))
                              titles ""))
-          (shell-quote-argument org-roam-directory)))
+          (shell-quote-argument (expand-file-name org-roam-directory))))
 
 (defun org-roam-unlinked-references-section (node)
   "The unlinked references section for NODE.


meedstrom:
Afaik, expanding it transforms ~/ into /home/USER/.

But I find it interesting that rg could not handle your ~/org/roam path. Something's weird there.

nvimtor:
@meedstrom Ah, thanks. 

I have resorted to just expanding it in my `init.el`.

I don't think this is something the library should do as it implicitly changes the user's given directory so I will close this. I will leave the issue open, though

Issue #2485, opened 2024-11-16
emacsql package updated. No emacsql-sqlite.el available. 


lnguyen4:
There is a new update for emacsql package. I believe the emacsql-sqlite.el is no longer existing. On my emacs setup, it will throw error at initialization with 

`Cannot open load file: No such file or directory, emacsql-sqlite`.

EDIT: `emacsql` 's `melpa` version `20241117.1614` should be working with `org-roam`.

benthamite:
I came here just to report this. `emacsql-sqlite` [was removed two hours ago](https://github.com/magit/emacsql/commit/8ebf559f58fdeba53d23e80562181fb67917942e). This removal prevents `org-roam` from loading because the package is required by `org-roam.el`, regardless of which database connector is set in `org-roam-database-connector`. For those who do not have that user option set to `sqlite`, one can solve the problem by just commenting out `(require 'emacsql-sqlite)` until this problem is patched.

benthamite:
@jethrokuan, I don’t quite understand if `sqlite` is actually needed. Folks who use Emacs +29 can use `sqlite-builtin`, while the rest can use `sqlite-module`, right? In which case, it’s just a matter of removing the `emacsql-sqlite` requirement and updating the `org-roam-database-connector` docstring.

lnguyen4:
> I came here just to report this. `emacsql-sqlite` [was removed two hours ago](https://github.com/magit/emacsql/commit/8ebf559f58fdeba53d23e80562181fb67917942e). This removal prevents `org-roam` from loading because the package is required by `org-roam.el`, regardless of which database connector is set in `org-roam-database-connector`. For those who do not have that user option set to `sqlite`, the problem can be temporarily solved by commenting out `(require 'emacsql-sqlite)` in `org-roam.el` until the code is patched properly. If you use `sqlite`, the problem will persist because `sqlite` itself requires `emacsql-sqlite`, so consider setting `org-roam-database-connector` to one of the other admissible values.

Good that some other is also reporting the same problem like me. For now, I just installed the melpa-stable version of `emacsql` and using it until new fix comes in for the `org-roam`.

Proven0366:
> Good that some other is also reporting the same problem like me. For now, I just installed the melpa-stable version of `emacsql` and using it until new fix comes in for the `org-roam`.

[...]

PR #2486, opened 2024-11-17
Rely on emacsql-sqlite-open to pick the best available back-end

tarsius:
That function was added two years ago and the first release that
provided it was v4.0.0.  It automatically picks the best available
back-end, `emacsql-sqlite-builtin` or `emacsql-sqlite-module`.
In v4.0.0 it could also fall back to the legacy `emacsql-sqlite`.
The inferior third-party back-ends are no longer supported.

Emacsql v4.1.0, which will be releases in early December, removes
the legacy `emacsql-sqlite` back-end (which used a custom binary).

This means that we now require an Emacs that was built with SQLite
and/or module support.

-------

EmacSQL now takes care of picking the best back-end itself.  Let's
use that.

By removing an option that offers no longer valid choices, we make
sure users cannot pick an invalid choice.

### DIFF
--- a/extensions/org-roam-export.el
+++ b/extensions/org-roam-export.el
@@ -6,7 +6,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 2.2.2
-;; Package-Requires: ((emacs "26.1") (org "9.4") (org-roam "2.1"))
+;; Package-Requires: ((emacs "26.1") (org "9.6") (org-roam "2.1"))
 
 ;; This file is NOT part of GNU Emacs.
 
--- a/extensions/org-roam-graph.el
+++ b/extensions/org-roam-graph.el
@@ -6,7 +6,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 2.2.2
-;; Package-Requires: ((emacs "26.1") (org "9.4") (org-roam "2.1"))
+;; Package-Requires: ((emacs "26.1") (org "9.6") (org-roam "2.1"))
 
 ;; This file is NOT part of GNU Emacs.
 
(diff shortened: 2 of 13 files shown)


tarsius:
Re #2485.

tarsius:
`emacsql-sqlite-open` calls its SETUP function with the database as its only argument and expects it to only access via that binding. When I wrote 118be8a225cb588d171a147ac8a9bf2723f46026 I did not notice, that while none of the functions called in the old `org-roam-db` do that directly, they do call functions that use `org-roam-db`, leading to this infinite loop.

I.e., we have to give up on using the SETUP argument, which means that Org-Roam doesn't follow the (refined) conventions. Alternatively the problematic functions could be changed to take the database as an optional argument, but I don't understand enough about what Org-Roam is doing here (and why), so it would be too risky for *me* to do it.

tarsius:
I've pushed an additional commit to address the compiler warning (which are unrelated to the changes in the other commits). Normally that should be done in a separate pull-request, but due to the radio silence, I have chosen to append that here, so I don't have to juggle with two branches when/if I decide to play with Org-Roam.

aragaer:
I think that `org-roam-db--setup` is for setting up database when it already exists and `org-roam-db--init` for its first creation, so if we pass the correct one, that should work. Unfortunately the `org-roam-db--setup` checks if upgrade is needed and that uses `(org-roam-db-sync t)` which actually closes and reopens the connection. While emacql approach is fine we might want to rewrite the setup function so that it detects whether an update is needed and postpones it until after return from setup/open.

dustinfarris:
@tarsius I just merged a couple of related changes in #2503 that do just enough to get CI passing.  Would you be willing to update this branch based off of latest main?

tarsius:
I've rebased, but note that I have not tested, since I don't use this package.

dustinfarris:
It is working well for me locally.  Thank you so much for this thorough PR!  ❤️

PR #2487, opened 2024-12-01
org-roam-mode: Customize which frames to check for org-roam-buffer

aragaer:
Current implementation of `org-roam-buffer--redisplay-h` doesn't allow having backlink buffer opened in a different frame. Since `get-buffer-window` allows different behavior using the additional parameter it makes sense to add a new customization option to control that behavior.

### DIFF
--- a/org-roam-mode.el
+++ b/org-roam-mode.el
@@ -86,6 +86,19 @@ applied in order of appearance in the list."
   :group 'org-roam
   :type 'hook)
 
+(defcustom org-roam-buffer-frames nil
+  "Where to search for org-roam-buffer when updating list of
+backlinks.
+
+See the description of `all-frames' argument of
+`get-buffer-window' function."
+  :group 'org-roam
+  :type '(choice (const :tag "All frames" t)
+                 (const :tag "Visible frames" visible)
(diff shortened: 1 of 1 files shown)


hpfr:
Rather than introducing an additional option can we just change to use `visible`? Having a hard time imagining the other options being useful.

aragaer:
I assumed that the original behavior was intended for some reason.

hpfr:
Since it’s not explicitly set to `nil`, I think it’s likely that the original developers just don’t have a multi-frame workflow (fairly common) and weren’t thinking of it. The blame seems to corroborate this based on my cursory look. It shouldn’t be costly performance-wise.

So if you update to set it to `visible`, that sounds good to me. (edit: don’t forget to rebase)

Aside: I also noticed there are two other `get-buffer-window` uses, but they are for toggling the buffer, and I think it’s reasonable to need to toggle the buffer independently per frame to handle cases like workspaces and Wayland where Emacs can’t detect a frame is invisible. This contrasts with the redisplay you are fixing, where you are probably finding nodes in a different frame from the backlinks buffer.

hpfr:
Superseded by https://github.com/org-roam/org-roam/pull/2594, reviews appreciated

PR #2492, opened 2024-12-31
Add advice to org-id-find instead of full org-id-open rewrite

aragaer:

Original `org-id-open` handles links like `id:XXX::opts`, but `org-roam-id-open` doesn't copy that functionality. Since `org-roam-id-open` is actually an incomplete copy of `org-id-open` it makes sense to use the original function and only add advice to `org-id-find`.


aragaer:
Seems to be duplicate of #2432

PR #2494, opened 2025-01-03
fix: org-roam-id-open (alternative to #2432)

nobiot:
This is an alternative to PR #2432: (fix): advise org-id-find rather than overwriting id link

The motivation is the same (pasting the phrase from #2432 below)

There have been recent changes in org-mode to improve `org-id-open' to be able to apply search strings in id: links.  But, org-roam overrides the `:follow` parameter for id: links to its own `org-roam-id-open' function and therefore misses this.

The difference is this patch "merges" the change in the upstream Org-ID change with Org-roam, thus avoiding an approach with advice. I guess this may add to a maintenance burden a little, but there seems to be an issue with #2432: https://github.com/org-roam/org-roam/pull/2432#issuecomment-2567226639

### DIFF
--- a/org-roam-id.el
+++ b/org-roam-id.el
@@ -67,9 +67,19 @@ With optional argument MARKERP, return the position as a new marker."
   "Go to the entry with id ID.
 Like `org-id-open', but additionally uses the Org-roam database."
   (org-mark-ring-push)
-  (let ((m (or (org-roam-id-find id 'marker)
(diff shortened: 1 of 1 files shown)


nobiot:
Closing in favor of #2432. 
Refer to
https://github.com/org-roam/org-roam/pull/2432#issuecomment-2569984761

Issue #2495, opened 2025-01-04
Org-roam cache database does not store Org-ID with search option correctly


nobiot:
### Description

Related to issue #2489.

Org has introduced a new feature in version 9.7 ([release notes](https://orgmode.org/Changes.html#orga1dc5df)), where the user can add a search option to be added to the Org-ID link like this example (from #2489): 

```
[[id:f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtree heading][test link]]
```

PR #2432 fixes the issue where `org-roam-id-open` does not resolve these ID links with an search option; however, it does not address the issue where Org-roam's cache database does not properly store the ID (`dest` field) for such links. Refer to this comment https://github.com/org-roam/org-roam/pull/2432#issuecomment-2569984761 by @aragaer .

#### Steps to Reproduce

A quick description of steps to reproduce is as follows:

1. Create a destination and source nodes (eg both file nodes)
   Include a headline in the destination file-node like this below and save the buffer.

   ```org
   :properties:
   :ID:       1234
   :end:
   #+title: Destination node
   
   * Example
   ```

2. In the source node, add a link to the destination with a search option like this:
   
   ```org
   [[id:1234::*Example]]
   ```

3. Save the source buffer; observe that the backlink does not appear in the destination.

#### Backtrace

Not applicable

#### Expected Results

Backlink appears in the destination.

#### Actual Results

Backlink does not appear.

### Environment

- Emacs: GNU Emacs 29.4 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.43, cairo version 1.18.2)
 of 2025-01-02
- Framework: N/A
- Org: Org mode version 9.7.18 (9.7.18-1463ab @ /home/nobiot/.config/emacs/elpa/org-9.7.18/)
- Org-roam: v2.2.2-69-g4fcc0b3-dirty
- sqlite-connector: sqlite-builtin

Issue #2489, opened 2024-12-14
org-roam doesn't resolve links created with `org-id-link-consider-parent-id` enabled


raphaelventura:
### Description

I'm trying to make org links leveraging the `org-id-link-to-org-use-id` and `org-id-link-consider-parent-id` variables, as described in section [4.5](https://orgmode.org/manual/Handling-Links.html) of the org manual (Org mode buffers paragraph's example). But org-roam doesn't resolve these links.

#### Steps to Reproduce

1) create this example file:
```org
* example heading
:PROPERTIES:
:ID:       f3805217-ff72-4dbe-b8aa-72a266eb5498
:END:
** first subtree heading

** second subtree heading

[[id:f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtree heading][test link]]
```

2) Try to use `org-open-at-point` on this link.

#### Backtrace

```lisp
Debugger entered--Lisp error: (error "Cannot find entry with ID \"f3805217-ff72-4dbe-b8aa...")
  error("Cannot find entry with ID \"%s\"" "f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtr...")
  org-roam-id-open("f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtr..." nil)
  org-link-open((link (:standard-properties [139 nil 206 215 217 0 nil nil nil nil nil nil nil nil #<buffer test.org> nil nil (paragraph (:standard-properties [139 139 139 218 218 0 nil planning element t nil nil nil nil #<buffer test.org> nil nil (section ...)]))] :type "id" :type-explicit-p t :path "f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtr..." :format bracket :raw-link "id:f3805217-ff72-4dbe-b8aa-72a266eb5498::*first su..." :application nil :search-option nil)) nil)
  org-open-at-point(nil)
  funcall-interactively(org-open-at-point nil)
  command-execute(org-open-at-point)
```

#### Expected Results

Point should go to to the first subtree heading, but I get an error instead.

#### Actual Results

`org-roam-id-open: Cannot find entry with ID "f3805217-ff72-4dbe-b8aa-72a266eb5498::*first subtree heading"`

### Environment

- Emacs: GNU Emacs 29.4 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.43, cairo version 1.18.2)
- Framework: N/A
[...]

PR #2496, opened 2025-01-04
fix(db): Org-roam does not store Org-ID w/ search option  #2495

nobiot:

Fix issue #2495
Related to issue #2489 and related discussion in PR #2432 around https://github.com/org-roam/org-roam/pull/2432#issuecomment-2569984761.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,11 @@ jobs:
     strategy:
       matrix:
         emacs_version:
-          - 27.1
+          # REVIEW: we do not yet have the bootstrapping in place to test
+          #         versions of emacs without built-in sqlite (pre 29.1)
+          # - 27.2
+          # - 28.2
+          - 29.4
           - snapshot
     steps:
       - uses: purcell/setup-emacs@master
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -268,40 +268,6 @@ file:
 install-info /path/to/my/info/files/org-roam.info /path/to/my/info/files/dir
 #+end_src
 
-** Installation Troubleshooting
-*** C Compiler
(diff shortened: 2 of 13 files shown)


aragaer:
This implementation discards the "option" part completely but it could be stored as a part of "properties" field.

```elisp
(when option
  (setf properties (plist-put properties :option option)))
```

nobiot:
> This implementation discards the "option" part completely but it could be stored as a part of "properties" field.

Sure. Done with [acdb153](https://github.com/org-roam/org-roam/pull/2496/commits/acdb153275a355f836e93455457ab6f891704696). 

Can you please test this on your end, too? My regression tests confirm that it works, but I may be missing something:

- I can remove and add search-options with no problem. 
- Combining with `:outline` works as well (both with `:outline nil` or with headlines).

PS. What is the use of `search-option` stored that you have in mind? I cannot think of a good use. Is it for "just in case"?

aragaer:
> PS. What is the use of `search-option` stored that you have in mind? I cannot think of a good use. Is it for "just in case"?

Mostly "just in case", but I can see having a custom backlinks section where "search-option" is used to display additional information of where exactly that backlink points at.

I've tested and it works for me.

aragaer:
Here's an example of using search-option in backlink display (my backlinks buffer is heavily customized):
backlinks

jethrokuan (approved):

dustinfarris:
@nobiot would you mind rebasing this on latest main?  part of this change has been merged already in #2432

nobiot:
> @nobiot would you mind rebasing this on latest main? part of this change has been merged already in #2432

How about this? Would this work for you?

dustinfarris:
I got you.  😃 

Thanks for this!  ❤️

PR #2498, opened 2025-01-11
Fix emacsql version requirement

manphiz:
* The Package-Version of emacsql around Feb 28, 2023 was actually 3.1.1.50-git.
* Using date as version value is too large and makes it unable to find a suitable candidate.


jethrokuan:
can you pull latest and check? I merged a recent PR to depend on the newer 4.0.0 release.

manphiz:
I think that should work.  Thanks.  Closing.

PR #2499, opened 2025-01-11
Set org-roam-directory to a non-existent path to ensure robust test

manphiz:
* In some rare case, the first expect of org-roam-file-p may return t when running in some environment (e.g. during autopkgtest in Debian).
* Setting org-roam-directory to a non-existent path ensure that the first check is always nil as expected.

### DIFF
--- a/tests/test-org-roam.el
+++ b/tests/test-org-roam.el
@@ -28,6 +28,7 @@
 
 (describe "org-roam-file-p"
   (it "checks if given file respects criteria"
+    (setq org-roam-directory "/non-existent")
     (expect (org-roam-file-p "tests/roam-files/family.org") :to-equal nil)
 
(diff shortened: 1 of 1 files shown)


dustinfarris:
Thanks for this.  We need to figure out why CI is failing (for all branches not just this one).

manphiz:
> Thanks for this. We need to figure out why CI is failing (for all branches not just this one).

Thanks for taking a look.  To be clear, this test failure was in the autopkgtest in a Debian buildd, which is separate from the Github CI.

dustinfarris:
Would you mind rebasing this and see if CI passes now

manphiz:
Rebased.  Also rechecked that the current HEAD still need this patch.  Also thanks for fixing Github CI :)

dustinfarris:
Great thank you! ❤️

PR #2509, opened 2025-03-13
(perf): Deprecate link :outline properties

meedstrom:
1. In `org-roam-db-insert-link`, stop adding `(list :outline OLP...)` 
   to every link object.

1. In `org-roam-node-insert-section`, reconstruct outline
   from one of its other arguments, `source-node`,
   permitting the `properties` argument to be nil.

   1. In callers of above function, don't bother to send 
      a `:properties` argument, since it won't be used.

1. Update schemata so link properties are not marked `:not-null`,
   making it legal for this field to be NULL.
   
   2. It may interest you to know that the `citations` table
      already was free of this restriction, so this change just affects
      the `links` table.

-------

As it happens, turning sexps into strings is one of the computationally
expensive steps in EmacSQL.

With an enormous number of links in the database, that's a lot of
`(:outline nil)` to stringify into `"(:outline nil)"`.

Its only use was in `org-roam-node-insert-section`, where the information
is very cheap to reconstruct.  It's already in one of the other
arguments!

This has passed unnoticed because `org-roam-db-sync` has other
performance tarpits, but it will probably be noticed eventually after
those get fixed.

---------

###### Full disclosure

Since many months ago, I've had a hack in my package
[org-node-fakeroam](https://github.com/meedstrom/org-node-fakeroam/blob/main/org-node-fakeroam.el),
to work around this performance issue, because it takes longer
to generate an org-roam.db that includes these link properties.

You can Ctrl+F `org-node-fakeroam--mk-link-props` if interested.

Now I'm working on a new package
[org-node-db.el](https://github.com/meedstrom/org-node/blob/dev/org-node-db.el) 
aiming to be 100% compatible with org-roam, using the same table schemata and all.  Which means I don't get to do the hack, and therefore this pull request.  Please consider it.

Benchmarks

- with the link props: ~1.00s to build org-roam.db
- without the link props: ~0.60s

Off-topic:  A related matter is the data-type of `atime` and `mtime` in the files table.  They are currently Lisp time values, and could easily be integers or floats instead.  Let me know if you would welcome a pullreq about that.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -108,7 +108,7 @@ ROAM_REFS."
   :type '(alist))
 
 ;;; Variables
-(defconst org-roam-db-version 18)
+(defconst org-roam-db-version 19)
 
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
   "Database connection to Org-roam database.")
@@ -214,7 +214,7 @@ The query is expected to be able to fail, in this situation, run HANDLER."
        (source :not-null)
        (dest :not-null)
        (type :not-null)
-       (properties :not-null)]
+       properties]
       (:foreign-key [source] :references nodes [id] :on-delete :cascade)))))
 
 (defconst org-roam-db--table-indices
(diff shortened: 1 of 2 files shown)


dustinfarris:
lgtm ❤️

meedstrom:
I'd like to extend an apology for not spotting that it had an use.

PR #2517, opened 2025-04-12
Cache closed timestamp property in node and db

hiecaq:
Add `closed` timestamp to node definition and database table.

There is no `org-get-closed-time` in org mode upstream at the time of writing, so the implementation of `org-roam-db-get-closed-time` ports the functionalities of `org-get-{scheduled,deadline}-time`.

Given that we've already been doing so for `scheduled` and `deadline`, I believe caching closed timestamp too might be helpful.

For example, we can filter nodes based on their closed time.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -108,7 +108,7 @@ ROAM_REFS."
   :type '(alist))
 
 ;;; Variables
-(defconst org-roam-db-version 20)
+(defconst org-roam-db-version 21)
 
 (defvar org-roam-db--connection (make-hash-table :test #'equal)
   "Database connection to Org-roam database.")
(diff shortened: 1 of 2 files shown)


dustinfarris:
Hi @hiecaq would you mind resolving the conflicts in this branch, then we'll get it in.

hiecaq:
Hi @dustinfarris I've rebased with the main branch. Thanks in advance!

hpfr (approved):
Addressed line length linting and rebased, LGTM

edit: slightly cumbersome to revert database version bumps or switch between testing branches, leaving this til after other 2.4.0 PRs

meedstrom:
LGTM as well.

> edit: slightly cumbersome to revert database version bumps or switch between testing branches, leaving this til after other 2.4.0 PRs

good catch

PR #2518, opened 2025-05-06
FIX org-roam-alias-add check for already exsting alias

Hixer05:
Adding an alias which was already take will result in the new node "eating" the reference of the previous one.
TODO: maybe add an interactive mode such that other names appear to avoid too much fail&retry when setting an alias.

Should i also update the changelog? I've seen the changes linked to already existing pull-request. This is my first pr on this project and github.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -1133,9 +1133,11 @@ and when nil is returned the node will be filtered out."
   "Add ALIAS to the node at point."
   (interactive "sAlias: ")
   (let ((node (org-roam-node-at-point 'assert)))
-    (save-excursion
-      (goto-char (org-roam-node-point node))
-      (org-roam-property-add "ROAM_ALIASES" alias))))
+    (if (member alias (org-roam--get-titles))
+        (message "Alias %s already exists in the Org-roam database." alias)
+      (save-excursion
+        (goto-char (org-roam-node-point node))
+        (org-roam-property-add "ROAM_ALIASES" alias)))))
 
 (defun org-roam-alias-remove (&optional alias)
   "Remove an ALIAS from the node at point."


dustinfarris:
Hi thanks for your PR.  Would you mind adding a test for this?  No need to update the changelog that will be auto-generated later.

dustinfarris:
Hi @Hixer05 are you able to add a test for this?

meedstrom:
It looks good, I don't know that this needs any test, code-wise. What possible regression to test for?

However... The premise isn't true for everyone.  Some people do want duplicate node titles, and presumably duplicate aliases too. 

I know because in org-node+[org-mem](https://github.com/meedstrom/org-mem), there is the setting `org-mem-do-warn-title-collisions`, which I see many users setting to nil. They work around titles "eating each other" by having the equivalent of e.g.

```elisp
(setq org-roam-node-display-template "${olp} ${title}")
```

so that bare titles/aliases no longer need to be unique.

AFAIK, we do not have any option similar to `org-mem-do-warn-title-collisions`, right? If we did, I'd suggest that this PR gate the error-check on such an option. As it stands, I'm not sure we can merge it.

meedstrom:
Hm, actually, this could be a simple Y/N prompt.  So when user adds an alias, org-roam asks "are you sure?" if there's a conflict.

meedstrom:
@Hixer05 if you want to be listed as a contributor on Github, feel free to make a new PR with this code

```elisp
(defun org-roam-alias-add (alias)
  "Add ALIAS to the node at point."
  (interactive "sAlias: ")
  (let ((node (org-roam-node-at-point 'assert)))
    (when (or (not (called-interactively-p 'any))
              (not (member alias (org-roam--get-titles)))
              (y-or-n-p (format "Alias \"%s\" already exists, really add here too? "
                                alias)))
      (save-excursion
        (goto-char (org-roam-node-point node))
        (org-roam-property-add "ROAM_ALIASES" alias)))))
```

I can add tests after that and merge. ;)

I'll make a PR from [my own branch](https://github.com/org-roam/org-roam/compare/main...meedstrom:org-roam:check-alias2) if I don't hear back from you in some weeks.

PR #2519, opened 2025-05-09
dailies: prevent matching .org_archive files

rgkirch:
`org-roam-dailies--list-files` would return `.org_archive` files along with `.org` files. The real issue is that `org-roam-dailies-goto-previous-note` can step into the archive file but not step out of the archive file. This could be fixed by not visiting the archive file or by fixing whatever issue prevents `org-roam-dailies-goto-previous-note` from stepping out of the archive file. this commit updates it to not consider archive files.

### DIFF
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -269,7 +269,7 @@ negative, find note N days in the future."
   "List all files in `org-roam-dailies-directory'.
 EXTRA-FILES can be used to append extra files to the list."
   (let ((dir (expand-file-name org-roam-dailies-directory org-roam-directory))
-        (regexp (rx-to-string `(and "." (or ,@org-roam-file-extensions)))))
+        (regexp (rx-to-string `(seq (literal ".") (or ,@org-roam-file-extensions) eos))))
     (append (--remove (let ((file (file-name-nondirectory it)))
                         (when (or (auto-save-file-name-p file)
                                   (backup-file-name-p file)


dustinfarris:
Hi thanks for surfacing this issue.  I _think_ the right solution would be to fix the underlying issue of moving through an archive file — otherwise we are effectively saying you can't archive dailies, correct?

rgkirch:
I would rather not step into the archive files when moving through dailies. I archived the heading because I don't want to see it in my dailies but I also don't want to delete it. If `org-roam-dailies-goto-previous-note` includes archive files then archiving headings means that I approach twice as many dailies files as I archive things. I'd rather have a separate way to toggle between the org and org_archive file for a given day.

> effectively saying you can't archive dailies

You can archive headings anywhere in org and they go to the org_archive file. Ditto for dailies. You can archive headings out of a dailies file. Are you talking like, archive the whole dailies file so that you no longer have the non-archived version?

rgkirch:
I was able to reproduce this in vanilla doom emacs and then I tried vanilla emacs and I could reproduce it there too.
`M-x doom/sandbox`
```
;; Welcome to the sandbox!
;;
;; This is a test bed for running Emacs Lisp in another instance of Emacs that
;; has varying amounts of Doom loaded:
;;
;; - vanilla Emacs (nothing loaded)         C-c C-c
;; - vanilla Doom (only Doom core)          M-x doom--run-vanilla-doom
;; - Doom + modules - your private config   M-x doom--run-vanilla-doom+
;; - Doom + modules + your private config   M-x doom--run-full-doom
;;
;; This is done without sacrificing access to installed packages. Use the sandbox
;; to reproduce bugs and determine if Doom is to blame.
```
then I run `M-x doom--run-vanilla-doom+` which is `Doom + modules - your private config`
then I set my `org-roam-directory` and `org-roam-dailies-directory`
```
(setq org-roam-directory (file-name-concat org-directory "pages/")
      org-roam-dailies-directory "journals/")
```
then I jump to today `org-roam-dailies-goto-today`, make a heading `org-meta-return`, archive it `org-archive-subtree-default`, navigate to it, and I can't navigate out of it via `org-roam-dailies-goto-next-note` or `org-roam-dailies-goto-previous-note`. I just get the message `user-error: Not in a daily-note`.
Then I opened vanilla emacs via `C-c C-c` and followed the same steps and got the same result except that I had to also `(require 'org-roam-dailies)`.

meedstrom (approved):

PR #2524, opened 2025-06-07
perf: suppress extra org features in the temp buffer

dustinfarris:


dustinfarris:
@akashpal-21 could you take a quick look at this before I merge it?

akashpal-21:
I do not have my computer with me right now so couldn't check sorry. Will get back 4 hours later.

dustinfarris:
> I do not have my computer with me right now so couldn't check sorry. Will get back 4 hours later.

No worries.  Whenever you’re able.

PR #2525, opened 2025-06-17
fix: run org-roam-db-clear-file on file deletion via vc-delete-file

okomestudio:
This PR proposes a fix to missing database autosync when files are deleted through `vc-delete-file`.

Currently, `org-roam-db-autosync-mode` sets up a before-advice with `delete-file` to maintain cache consistency. When an Org is under version control, however, file deletion may be done through `vc-delete-file`, which might not use `delete-file` to perform deletion. In such a case, cache will be in an inconsistent state (which can be fix by manually running `org-roam-db-sync`). 

Therefore, this PR implements an around-advice with `vc-delete-file` to ensure cache consistency when the function is used but not `delete-file`.


dustinfarris:
looks great, thank you!

PR #2527, opened 2025-06-29
doc: Update build system

hpfr:
Jethro wrote the original docs build based on Magit’s docs build. Jonas has updated it in the half-decade since. He deprecated ox-texinfo+ since Org 9.6 includes enough of its features.

Org-roam is simpler than Magit and has less available maintainers. Remove some of the unused features inherited from Magit:
- remove AUTHORS.md which was not kept updated past the initial implementation
- remove epub and pdf targets which were not deployed
- remove Elisp-related variables since Org-roam uses Eldev

The result is simple enough to consolidate default.mk into doc/Makefile.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,27 @@ jobs:
       - name: Test
         run: make test
 
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install deps
+        run: sudo apt install texinfo
+
+      - name: Build docs
+        run: make html
+
+      - name: Upload
+        uses: actions/upload-artifact@v4
+        with:
+          name: Docs
+          if-no-files-found: error
+          # TODO: optimize images
+          path: |
+            doc/
+            !doc/images/
+
 # Local Variables:
 # eval: (outline-minor-mode)
 # End:


hpfr:
The diff is pretty nasty, side by side view of the Makefiles is better since it’s a substantial refactor. Also since the variable/function/symbol documentation features of ox-texinfo+ are slightly different in upstream Org 9.6, there are a bunch of small tweaks to the org.

Also worth noting: to generate the texi for a tagged release, run `VERSION=N make -C doc release-texi`. This is necessary since we’re committing the texi so the tag can’t be queried at build time.

hpfr:
@jethrokuan is the only one who ever touched any of this and it’s been a while. Once I found the Magit build I was more confident about why things are the way they are but if you are around I’d appreciate your review!

manphiz:
Came from #2528.  The part of org update to drop ox-texinfo+ covers what I was trying to do in #2528 and looks good to me.

meedstrom:
Looking forward to this.

Question: how/where is https://www.orgroam.com/manual.html hosted?  I assume it corresponds to `doc/index.html`.

I think if that page survives, that's pretty much all we need to know that this PR is fine?  The tests obviously survive.

hpfr:
> Question: how/where is https://www.orgroam.com/manual.html hosted? I assume it corresponds to `doc/index.html`.
>
> I think if that page survives, that's pretty much all we need to know that this PR is fine? The tests obviously survive.

`workflows/docs.yml` runs `make html` and deploys the resulting `doc/` as static files to GitHub pages. So you can verify by confirming `make html` produces `doc/manual.html`. Once we get an approval I’m good to merge here as well!

dustinfarris:
I temporarily added this branch to trigger the workflow to test it ([`9de9c05`](https://github.com/org-roam/org-roam/pull/2527/commits/9de9c0588066958a091eec30e6f6d51af2009717)). Looks like it's having some trouble:

https://github.com/org-roam/org-roam/actions/runs/20624414126/job/59232388087

hpfr:
tl;dr: fixed now

> I temporarily added this branch to trigger the workflow to test it ([`9de9c05`](https://github.com/org-roam/org-roam/pull/2527/commits/9de9c0588066958a091eec30e6f6d51af2009717)). Looks like it's having some trouble:
> 
> https://github.com/org-roam/org-roam/actions/runs/20624414126/job/59232388087

Good idea. Seems like the token is invalid.
```
remote: Invalid username or token. Password authentication is not supported for Git operations.
```

I followed the upgrade guide in https://github.com/JamesIves/github-pages-deploy-action/discussions/592. I took another look at the README and it seems like it’s recommended not to use a manually defined token anymore, there is a default token (which the stale action already uses).

I removed the `token:` line entirely and [deploy works now](https://github.com/org-roam/org-roam/actions/runs/20645019265/job/59281688530). https://www.orgroam.com/manual.html now shows `2.3.1-devel`. I then dropped the temporary commit so this should be ready for review and merge again.

meedstrom:
A question since I swear I could use `make test` locally before, and now I can't.

I see that it used to download eldev with a deleted script `github-eldev` that invoked curl among other things:

```bash
curl -fsSL https://raw.githubusercontent.com/doublep/eldev/f111d19cda305e5e8fcb70a5675b87173041cb68/bin/eldev > $ELDEV_BIN_DIR/eldev
```

(Though I don't quite get how just typing `make test` invoked `github-eldev` in turn, maybe I've just forgotten doing something manually?)

It's a bit bothersome that eldev isn't on common repos, so I can't just `apt-get install eldev`.

Whoever does `make test` locally, how do you do it?  

I can adapt your tips to write some documentation. Ideally when `make test` fails, it should echo some instructions for how to set up the test suite.

PR #2528, opened 2025-06-30
Doc build improve

manphiz:

ox-texinfo+ has declared obsoletion.  I tried to migrate to upstream org texinfo as suggested by ox-texinfo+ README, and dropped ox-texinfo+ from Makefile.

Please review.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -66,7 +66,7 @@ clean:
 ## Release management ################################################
 
 ORG_ARGS  = --batch -Q $(ORG_LOAD_PATH)
-ORG_ARGS += -l ox-extra -l ox-texinfo+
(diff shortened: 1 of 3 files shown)


hpfr:
I was sure this was a spambot, since I did the exact same thing a couple days ago as part of https://github.com/org-roam/org-roam/pull/2527. But I looked at your profile and you seem to be a real person; it’s just a strange coincidence. I think you caught a couple things with the defun support I missed which I incorporated now. Would appreciate your review there!

manphiz:
What a coincidence!  I had been working on this for a few days and only pushed yesterday.  Thanks for letting me know.  I'll take a look at #2527.

Issue #2529, opened 2025-07-01
org-roam-link-replace-all breaks before-save hooks


fuzzbomb:
### Description
The recent change in https://github.com/org-roam/org-roam/commit/fc8638759b9d8e1725ff58a2b5f983679be0cf3d produces an error.

It's a malformed regex problem. It's now using a literal "[[" instead of `org-link-bracket-re`, but I think some escaping of the square brackets is needed.

Consequently, other before-save-hook functions don't run. (In my case, it stops `org-roam-timestamps-mode` and `ws-butler-before-save` from working.)

As a workaround, I've set `org-roam-link-auto-replace` to nil. I don't use `roam:` links generally.

#### Steps to Reproduce
TO COME

#### Backtrace
TO COME

#### Actual Results
`Before-save hook error: (invalid-: "Unmatched [ or [^")`

### Environment
TO COME

hpfr:
Sorry, I got sloppy with my review and the author evidently didn’t test the updated version. That change should have included switching to `search-forward`. Will fix

fuzzbomb:
Right-oh. Ditching the regex makes things simpler.

I've just tested this and it's fixed the problem. Both of my `ws-butler` and `org-roam-timestamps` hooks run as expected.

Thanks for the quick fix.

hpfr:
Thanks for the quick report! I have been sharply reminded of MELPA unstable users, I used Doom and now Guix so was never that close to the bleeding edge.

fuzzbomb:
It's all part of the fun :-)

PR #2530, opened 2025-07-01
Fix link replacement optimization

hpfr:
We are searching for a string literal for speed, so don’t treat it as a regexp.

Fixes: fc8638759b9d ("feat: Limit link replacement scope")
Fixes: #2529

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -810,10 +810,9 @@ Assumes that the cursor was put where the link is."
 (defun org-roam-link-replace-all ()
   "Replace all \"roam:\" links in buffer with \"id:\" links."
   (interactive)
-  (let ((org-roam-link-prefix (concat "[[" org-roam-link-type ":")))
-    (org-with-point-at 1
-      (while (re-search-forward org-roam-link-prefix nil t)
-        (org-roam-link-replace-at-point)))))
+  (org-with-point-at 1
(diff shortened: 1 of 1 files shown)


dustinfarris:
would it make sense to add a test for this?

hpfr:
> would it make sense to add a test for this?

Missed this. Yeah, tests certainly wouldn’t hurt. I will need to get familiar with Buttercup. I have also noted that I should actually run the branch locally before merging, I forgot a bunch of MELPA users live on main :sweat:.

dustinfarris:
it's part of the hazing process :P 

I don't know if you're an ai user, but Claude is pretty helpful for writing tests like this

hpfr:
I poke around with the web interfaces where I don’t have to log in :sweat_smile:. I have been meaning to try gptel. I guess you mean something more “agentic” like Aidermacs?

dustinfarris:
gptel can be used with Claude, among other ai providers, as can aidermacs.  but I don't mean to send you down a rabbit hole! the time spinning all of that up would probably be more than the time saved by it writing a test for you.  I can pitch in here if you don't already have something going.

hpfr:
Oh, no I don’t have something going yet! I’d be happy to review whatever you come up with; otherwise I’ll take a look this weekend.

dustinfarris:
> whatever you come up with

#2531 

I'll take credit if you're good with it, otherwise it was Claude.

PR #2544, opened 2025-09-29
make entry nodes "opt in"

dustinfarris:
See discussion in #2543

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -847,7 +847,9 @@ of the template are similar to ~org-capture~ templates.
    automatically chooses this template for you.
 2. The template is given a description of ~"default"~.
 3. ~plain~ text is inserted. Other options include Org headings via
-   ~entry~.
+   ~entry~. Note that ~entry~ type captures create regular Org headings
+   without IDs by default. To create an Org-roam node with an ID, use
+   the ~:entry-node t~ option.
 4. Notice that the ~target~ that's usually in Org-capture templates is missing
    here.
 5. ~"%?"~ is the template inserted on each call to ~org-roam-capture-~.
@@ -1220,6 +1222,10 @@ Here is a sane default configuration:
          :target (file+head "%<%Y-%m-%d>.org"
                             "#+title: %<%Y-%m-%d>\n"))))
 #+end_src
+
+Note: The ~entry~ type capture above creates a regular Org heading without an
+ID. If you want each daily entry to be an Org-roam node with its own ID, add
+~:entry-node t~ to the template.
 
 See [[*The Templating System][The Templating System]] for creating new templates.
 
--- a/doc/org-roam.texi
+++ b/doc/org-roam.texi
@@ -1219,7 +1219,9 @@ automatically chooses this template for you.
 The template is given a description of @code{"default"}.
 @item
 @code{plain} text is inserted. Other options include Org headings via
-@code{entry}.
+@code{entry}. Note that @code{entry} type captures create regular Org headings
+without IDs by default. To create an Org-roam node with an ID, use
+the @code{:entry-node t} option.
 @item
 Notice that the @code{target} that's usually in Org-capture templates is missing
 here.
@@ -1702,6 +1704,10 @@ Here is a sane default configuration:
          :target (file+head "%<%Y-%m-%d>.org"
                             "#+title: %<%Y-%m-%d>\n"))))
 @end lisp
+
+Note: The @code{entry} type capture above creates a regular Org heading without an
+ID@. If you want each daily entry to be an Org-roam node with its own ID, add
+@code{:entry-node t} to the template.
 
 See @ref{The Templating System} for creating new templates.
 
@@ -2421,5 +2427,5 @@ When GOTO is non-nil, go the note without creating an entry."
 
 @printindex vr
 
-Emacs 30.1 (Org mode 9.7.29)
+Emacs 30.2 (Org mode 9.7.34)
 @bye
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -63,9 +63,11 @@ description   A short string describing the template, which will be shown
               during selection.
 
 type       The type of entry. Valid types are:
-               entry       an Org node, with a headline.  Will be filed
-                           as the child of the target entry or as a
-                           top level entry.  Its default template is:
+               entry       an Org heading.  Will be filed as the child
+                           of the target entry or as a top level entry.
+                           Use :entry-node t to create an ID for this
+                           heading, making it an org-roam node.
+                           Its default template is:
                              \"* %?\n %a\"
                item        a plain list item, will be placed in the
                            first plain list at the target location.
@@ -134,6 +136,11 @@ The following options are supported for the :target property:
 
 The rest of the entry is a property list of additional options.  Recognized
 properties are:
+
+ :entry-node         When set to t for entry-type captures, creates an ID for
+                     the captured entry heading. When nil or not specified,
+                     no ID is created for the entry. Only applies to templates
+                     with type 'entry'.
 
  :prepend            Normally newly captured information will be appended at
                      the target location (last child, last table line,
@@ -390,7 +397,7 @@ This variable is populated dynamically, and is only non-nil
 during the Org-roam capture process.")
 
 (defconst org-roam-capture--template-keywords (list :target :id :link-description :call-location
-                                                    :region)
+                                                    :region :entry-node)
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")
 
 ;;; Main entry point
@@ -578,12 +585,13 @@ capture target."
                target-entry-p (and (derived-mode-p 'org-mode) (org-at-heading-p))))))
     ;; Setup `org-id' for the current capture target and return it back to the
     ;; caller.
-    ;; Unless it's an entry type, then we want to create an ID for the entry instead
+    ;; For entry type: only create ID if :entry-node is t
     (pcase (org-capture-get :type)
       ('entry
-       (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
-       (org-roam-capture--put :new-node-p t)
-       (setq id (org-roam-node-id org-roam-capture--node)))
+       (when (org-roam-capture--get :entry-node)
+         (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
+         (org-roam-capture--put :new-node-p t)
+         (setq id (org-roam-node-id org-roam-capture--node))))
       (_
        (save-excursion
          (goto-char p)
@@ -610,7 +618,7 @@ capture target."
   (advice-remove #'org-capture-place-template #'org-roam-capture-run-new-node-hook-a))
 
 (defun org-roam-capture--create-id-for-entry ()
-  "Create the ID for the new entry."
+  "Create the ID for the new entry heading."
   (org-entry-put (point) "ID" (org-roam-capture--get :id))
   (advice-remove #'org-capture-place-entry #'org-roam-capture--create-id-for-entry))
 
--- a/tests/test-org-roam-capture.el
+++ b/tests/test-org-roam-capture.el
@@ -58,38 +58,76 @@
      (org-roam-capture--fill-template (lambda () "foo"))
      :to-equal "foo")))
 
-(describe "org-roam-capture entry-type ID creation"
-  (it "creates ID for entry-type captures"
-    (let* ((temp-dir (make-temp-file "org-roam-test" t))
-           (test-file (expand-file-name "test.org" temp-dir))
-           (org-roam-directory temp-dir)
-           (org-roam-capture--node (org-roam-node-create :id (org-id-new)))
-           (org-roam-capture--info (make-hash-table :test 'equal))
-           capture-id)
-      (unwind-protect
-          (progn
-            ;; Create initial file content
-            (with-temp-file test-file
-              (insert "#+TITLE: Test File\n\n* Parent Heading\n:PROPERTIES:\n:ID: parent-id\n:END:\n"))
+(describe "org-roam-capture :entry-node option"
+  :var ((temp-dir) (org-roam-directory) (org-roam-db-location))
 
-            ;; Mock org-capture context and get-target
-            (cl-letf* (((symbol-function 'org-capture-get)
-                        (lambda (prop)
-                          (pcase prop
-                            (:type 'entry)
-                            (_ nil))))
-                       ((symbol-function 'org-roam-capture--get-target)
-                        (lambda () `(file ,test-file))))
+  (before-each
+    (setq temp-dir (make-temp-file "org-roam-test" t))
+    (setq org-roam-directory temp-dir)
+    (setq org-roam-db-location (expand-file-name "org-roam.db" temp-dir))
+    (org-roam-db-sync))
 
-              ;; Call the setup function
-              (with-current-buffer (find-file-noselect test-file)
-                (org-roam-capture--setup-target-location)
-                (setq capture-id (org-roam-capture--get :id)))
+  (after-each
+    (delete-directory temp-dir t))
 
-              ;; Verify ID was created and stored for entry type
-              (expect capture-id :not :to-be nil)
-              (expect (org-roam-capture--get :new-node-p) :to-be t)))
-        (delete-directory temp-dir t))))
+  (it "does not create ID for entry-type capture without :entry-node option"
+    (let* ((test-file (expand-file-name "test-parent.org" temp-dir))
+           (org-roam-capture-templates
+            '(("t" "test" entry "* ${title}"
+               :target (file "test-parent.org")
+               :unnarrowed t))))
+
+      ;; Create parent file with an existing heading
+      (with-temp-file test-file
+        (insert "#+TITLE: Parent File\n\n* Existing Heading\n"))
+
+      ;; Mock the node selection to return a new node
+      (cl-letf (((symbol-function 'org-roam-node-read)
+                 (lambda (&rest _)
+                   (org-roam-node-create :title "New Entry Without ID"))))
+
+        (org-roam-capture)
+
+        ;; Finalize the capture
+        (org-capture-finalize))
+
+      ;; Verify the captured entry exists but has no ID
+      (with-temp-buffer
+        (insert-file-contents test-file)
+        (goto-char (point-min))
+        (re-search-forward "^\\* New Entry Without ID$")
+        (let ((has-id (org-entry-get (point) "ID")))
+          (expect has-id :to-be nil)))))
+
+  (it "creates ID for entry-type capture with :entry-node t option"
+    (let* ((test-file (expand-file-name "test-parent-with-id.org" temp-dir))
+           (org-roam-capture-templates
+            '(("t" "test" entry "* ${title}"
+               :target (file "test-parent-with-id.org")
+               :entry-node t
+               :unnarrowed t))))
+
+      ;; Create parent file with an existing heading
+      (with-temp-file test-file
+        (insert "#+TITLE: Parent File\n\n* Existing Heading\n"))
+
+      ;; Mock the node selection to return a new node
+      (cl-letf (((symbol-function 'org-roam-node-read)
+                 (lambda (&rest _)
+                   (org-roam-node-create :title "New Entry With ID"))))
+
+        (org-roam-capture)
+
+        ;; Finalize the capture
+        (org-capture-finalize))
+
+      ;; Verify the captured entry has an ID
+      (with-temp-buffer
+        (insert-file-contents test-file)
+        (goto-char (point-min))
+        (re-search-forward "^\\* New Entry With ID$")
+        (let ((has-id (org-entry-get (point) "ID")))
+          (expect has-id :not :to-be nil)))))
 
   (it "creates ID at target for non-entry-type captures"
     (let* ((temp-dir (make-temp-file "org-roam-test" t))


akashpal-21:
```diff
--- org-roam-capture.el	2025-09-30 06:36:17.727822920 +0530
+++ org-roam-capture-suggestionv1.el	2025-09-30 07:28:01.361391615 +0530
@@ -65,8 +65,8 @@
 type       The type of entry. Valid types are:
                entry       an Org heading.  Will be filed as the child
                            of the target entry or as a top level entry.
-                           Use :entry-node t to create an ID for this
-                           heading, making it an org-roam node.
+                           Use :entry-parent t to toggle id creation
+                           from child to parent entry.
                            Its default template is:
                              \"* %?\n %a\"
                item        a plain list item, will be placed in the
@@ -137,10 +137,10 @@
 The rest of the entry is a property list of additional options.  Recognized
 properties are:

- :entry-node         When set to t for entry-type captures, creates an ID for
-                     the captured entry heading. When nil or not specified,
-                     no ID is created for the entry. Only applies to templates
-                     with type 'entry'.
+ :entry-parent       When set to t for entry-type captures, creates an ID for
+                     the captured entry's :target. When nil or not specified,
+                     ID is created for the entry specified in the template.
+                     Only applies to templates with type 'entry'.

  :prepend            Normally newly captured information will be appended at
                      the target location (last child, last table line,
@@ -340,6 +340,7 @@
                                          ((const :format "%v " :time-prompt) (const t))
                                          ((const :format "%v " :tree-type) (const week))
                                          ((const :format "%v " :unnarrowed) (const t))
+					 ((const :format "%v " :entry-parent) (const t))
                                          ((const :format "%v " :table-line-pos) (string))
                                          ((const :format "%v " :kill-buffer) (const t))))))))

@@ -397,7 +398,7 @@
 during the Org-roam capture process.")

 (defconst org-roam-capture--template-keywords (list :target :id :link-description :call-location
-                                                    :region :entry-node)
+                                                    :region :entry-parent)
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")

 ;;; Main entry point
@@ -585,22 +586,21 @@
                target-entry-p (and (derived-mode-p 'org-mode) (org-at-heading-p))))))
     ;; Setup `org-id' for the current capture target and return it back to the
     ;; caller.
-    ;; For entry type: only create ID if :entry-node is t
-    (pcase (org-capture-get :type)
-      ('entry
-       (when (org-roam-capture--get :entry-node)
-         (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
-         (org-roam-capture--put :new-node-p t)
-         (setq id (org-roam-node-id org-roam-capture--node))))
-      (_
-       (save-excursion
-         (goto-char p)
-         (unless (org-entry-get p "ID")
-           (org-roam-capture--put :new-node-p t))
-         (setq id (or (org-entry-get p "ID")
-                      (org-roam-node-id org-roam-capture--node)))
-         (setf (org-roam-node-id org-roam-capture--node) id)
-         (org-entry-put p "ID" id))))
+    ;; For entry type: target the entry itself (or its parent).
+    (if (and (eq (org-capture-get :type) 'entry)
+	     (null (plist-get (org-capture-get :org-roam) :entry-parent)))
+	(progn
+	  (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
+	  (org-roam-capture--put :new-node-p t)
+	  (setq id (org-roam-node-id org-roam-capture--node)))
+      (save-excursion
+        (goto-char p)
+        (unless (org-entry-get p "ID")
+          (org-roam-capture--put :new-node-p t))
+        (setq id (or (org-entry-get p "ID")
+                     (org-roam-node-id org-roam-capture--node)))
+        (setf (org-roam-node-id org-roam-capture--node) id)
+        (org-entry-put p "ID" id)))
     (org-roam-capture--put :id id)
     (org-roam-capture--put :target-entry-p target-entry-p)
     (advice-add #'org-capture-place-template :before #'org-roam-capture--set-target-entry-p-a)
```

Suggested iteration over this 

1. Keep the new behaviour default
Rationale: If user is invoking capture from org-roam-node-insert this is the behaviour that they would want.

2. Do not remove id creation for entry types - instead option on where to place the id - at the container level (target/parent) or at the entry level.

3. (plist-get (org-capture-get :org-roam) :entry-parent) is required to access org-roam-specific keywords in my end org-capture-get results in always nil.

4. suggestion to not use pcase for if then else branching (only two real bifurcation of paths)

wdyt? 


```
      '(("t" "Test entry-parent"
         entry
         "* h1\n %?"
         :target (file+datetree "test.org" month)
         :entry-parent t
         :immediate-finish t
         :jump-to-captured t)))


* 2025
** 2025-09 September
:PROPERTIES:
:ID:       20250930T072614.925841
:END:
*** h1


      '(("t" "Test entry"
         entry
         "* h1\n %?"
         :target (file+datetree "test.org" month)
         :immediate-finish t
         :jump-to-captured t)))


* 2025
** 2025-09 September
*** h1
:PROPERTIES:
:ID:       20250930T072727.063910
:END:
```

akashpal-21:
Also suggestion to remove conditional new-node flag setting depending on "new" id creation. It may make semantic sense to only update the flag if new id is generated - but it doesnt pass algorithmic logic - since the only new-node hook that runs is to add :ROAM_REF: if any - this doesnt require unique ID since one ID can store multiple refs. Unless this serves some other purpose to block the flag -- everytime capture is running - the new-node flag should be updated else hooks would fail.

```
 --- org-roam-capture-suggestionv1.el	2025-09-30 07:28:01.361391615 +0530
+++ org-roam-capture-suggestionv2.el	2025-09-30 09:30:28.749375552 +0530
@@ -595,8 +595,7 @@
 	  (setq id (org-roam-node-id org-roam-capture--node)))
       (save-excursion
         (goto-char p)
-        (unless (org-entry-get p "ID")
-          (org-roam-capture--put :new-node-p t))
+        (org-roam-capture--put :new-node-p t)
         (setq id (or (org-entry-get p "ID")
                      (org-roam-node-id org-roam-capture--node)))
         (setf (org-roam-node-id org-roam-capture--node) id)
```

anpandey:
> Do not remove id creation for entry types - instead option on where to place the id - at the container level (target/parent) or at the entry level.

Sometimes it's desirable to specify a `file+head+olp` target but not have IDs created for either the target or the entry itself, so if you go this route specifying where IDs are inserted should probably be decoupled from the target specification.

IMO modification of already existing headings beyond the captured entry itself might result in unexpected/surprising behavior, especially if `unnarrowed` is set to `nil` (the default). This becomes a little more ambiguous when headings outside the target are created by the capture template itself such as `datetree` or `olp` targets that don't yet exist.

Perhaps a three-way distinction might make sense here:

1. Don't create an ID anywhere
2. Create an ID just for the entry
3. Create an ID for the entry, the target, and all implicitly created headings.

akashpal-21:
> modification of already existing headings beyond the captured entry itself might result in unexpected/surprising behavior,

Correct - but this is unavoidable due to the juggling game that org-roam-capture and org-capture plays - (org-capture-mode) does not track modifying :PROPERTIES: drawer outside scope - so discard capture may not discard all changes - but these are usually unharmful.

> Perhaps a three-way distinction might make sense here:

It is trivial to implement this - but should org-roam enable creation of entries without putting ID anywhere - arent we back to org-capture territory? And if this is to be given - this should not be the default atleast IMO.

```
--- org-roam-capture.el	2025-10-01 10:12:52.787882637 +0530
+++ org-roam-capture-suggestionv1.el	2025-10-01 10:53:41.252175503 +0530
@@ -63,9 +63,11 @@
               during selection.

 type       The type of entry. Valid types are:
-               entry       an Org node, with a headline.  Will be filed
-                           as the child of the target entry or as a
-                           top level entry.  Its default template is:
+               entry       an Org heading.  Will be filed as the child
+                           of the target entry or as a top level entry.
+                           Use :entry-target to specify where to attach
+                           the ID at.
+                           Its default template is:
                              \"* %?\n %a\"
                item        a plain list item, will be placed in the
                            first plain list at the target location.
@@ -135,6 +137,14 @@
 The rest of the entry is a property list of additional options.  Recognized
 properties are:

+ :entry-target       For entries - either the ID can be attached to the:
+                     a. lowest entry level, (Default)
+                        :entry-target entry or nil,
+                     b. parent heading specified in :target,
+                        :entry-target parent; or
+                     c. no ID is assigned,
+                        :entry-target none or anything else.
+
  :prepend            Normally newly captured information will be appended at
                      the target location (last child, last table line,
                      last list item...).  Setting this property will
@@ -333,6 +343,11 @@
                                          ((const :format "%v " :time-prompt) (const t))
                                          ((const :format "%v " :tree-type) (const week))
                                          ((const :format "%v " :unnarrowed) (const t))
+					 ((const :format "%v " :entry-target)
+					  (choice :value entry
+						  (const :tag "Entry"  entry)
+						  (const :tag "Parent" parent)
+						  (const :tag "None"   none)))
                                          ((const :format "%v " :table-line-pos) (string))
                                          ((const :format "%v " :kill-buffer) (const t))))))))

@@ -390,7 +405,7 @@
 during the Org-roam capture process.")

 (defconst org-roam-capture--template-keywords (list :target :id :link-description :call-location
-                                                    :region)
+                                                    :region :entry-target)
   "Keywords used in `org-roam-capture-templates' specific to Org-roam.")

 ;;; Main entry point
@@ -578,21 +593,23 @@
                target-entry-p (and (derived-mode-p 'org-mode) (org-at-heading-p))))))
     ;; Setup `org-id' for the current capture target and return it back to the
     ;; caller.
-    ;; Unless it's an entry type, then we want to create an ID for the entry instead
-    (pcase (org-capture-get :type)
-      ('entry
-       (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
-       (org-roam-capture--put :new-node-p t)
-       (setq id (org-roam-node-id org-roam-capture--node)))
-      (_
-       (save-excursion
-         (goto-char p)
-         (unless (org-entry-get p "ID")
-           (org-roam-capture--put :new-node-p t))
-         (setq id (or (org-entry-get p "ID")
-                      (org-roam-node-id org-roam-capture--node)))
-         (setf (org-roam-node-id org-roam-capture--node) id)
-         (org-entry-put p "ID" id))))
+    ;; For entry type: target ID to the entry itself, or its parent, or none.
+    (let ((type (org-capture-get :type))
+	  (entry-target (plist-get (org-capture-get :org-roam) :entry-target)))
+      (if (and (eq type 'entry)
+               (not (eq entry-target 'parent)))
+	  (when (memq entry-target '(nil entry))
+	    (org-roam-capture--put :new-node-p t)
+	    (setq id (org-roam-node-id org-roam-capture--node))
+	    (advice-add #'org-capture-place-entry :after
+			#'org-roam-capture--create-id-for-entry))
+	(save-excursion
+          (goto-char p)
+          (org-roam-capture--put :new-node-p t)
+          (setq id (or (org-entry-get p "ID")
+                       (org-roam-node-id org-roam-capture--node)))
+          (setf (org-roam-node-id org-roam-capture--node) id)
+          (org-entry-put p "ID" id))))
     (org-roam-capture--put :id id)
     (org-roam-capture--put :target-entry-p target-entry-p)
     (advice-add #'org-capture-place-template :before #'org-roam-capture--set-target-entry-p-a)
@@ -610,7 +627,7 @@
   (advice-remove #'org-capture-place-template #'org-roam-capture-run-new-node-hook-a))

 (defun org-roam-capture--create-id-for-entry ()
-  "Create the ID for the new entry."
+  "Create the ID for the new entry heading."
   (org-entry-put (point) "ID" (org-roam-capture--get :id))
   (advice-remove #'org-capture-place-entry #'org-roam-capture--create-id-for-entry))
  ```

```
      '(("t" "Test entry-parent"
         entry
         "* h1\n %?"
         :target (file+datetree "test.org" month)
         :immediate-finish t
         :jump-to-captured t)))


* 2025
** 2025-10 October
*** h1
:PROPERTIES:
:ID:       20251001T104631.684338
:END:

      '(("t" "Test entry-parent"
         entry
         "* h1\n %?"
         :target (file+datetree "test.org" month)
         :immediate-finish t
	 :entry-target parent
         :jump-to-captured t)))


* 2025
** 2025-10 October
:PROPERTIES:
:ID:       20251001T104721.672668
:END:
*** h1

      '(("t" "Test entry-parent"
         entry
         "* h1\n %?"
         :target (file+datetree "test.org" month)
         :immediate-finish t
	 :entry-target none
         :jump-to-captured t)))


* 2025
** 2025-10 October
*** h1
```

wdyt? 

P.S. In the code - the only time new-node hooks arent run is when user blocks ID targetting 
`(:entry-target none)`

**Comment by anpandey** [comment]
> It is trivial to implement this - but should org-roam enable creation of entries without putting ID anywhere - arent we back to org-capture territory?

The argument is that `file+olp+head` can accomplish this with the ID being inserted at the file-level, and ID insertion template being controlled by the user. There isn't an analogue for this in org-capture.

**Comment by dustinfarris** [comment]
@akashpal-21 

> 1. Keep the new behaviour default
> Rationale: If user is invoking capture from org-roam-node-insert this is the behaviour that they would want.

Sounds reasonable.

> 2. Do not remove id creation for entry types - instead option on where to place the id - at the container level (target/parent) or at the entry level.

I think I understand the motivation here, but not aligned with the approach.  IMO your suggested change relies on broken behavior where IDs were being misplaced.  I think we should explicitly support datetree headings as nodes (not sure what kind of effort that would take).

> 3. (plist-get (org-capture-get :org-roam) :entry-parent) is required to access org-roam-specific keywords in my end org-capture-get results in always nil.

Possibly n/a given my response to number 2.

> 4. suggestion to not use pcase for if then else branching (only two real bifurcation of paths)

👍 

> remove conditional new-node flag setting depending on "new" id creation

I need to take time to think through any implications here — as you've said, probably harmless, but I don't want to rush it in and don't want to hold up this PR for it.

---

@anpandey 

> This becomes a little more ambiguous when headings outside the target are created by the capture template itself such as datetree or olp targets that don't yet exist.

Yes, this is my feeling as well.  We need a separate discussion I think just centered around this.

> Don't create an ID anywhere
> ...
> The argument is that file+olp+head can accomplish this with the ID being inserted at the file-level, and ID insertion template being controlled by the user.

Would the user use `plain` type in this case?

**Comment by akashpal-21** [comment]
@dustinfarris 

I am personally inclined towards anything you decide - since I can always force my way through in my own instance. But I sincerely do not understand the current implementation you proposed - 

please correct if I am mistaken - but you propose a change wherein it is user customisable whether ID will be created. Is this what users are complaining such as in #2551 ? or is user complaining to get the old "broken" behaviour?

Fundamentally - I do not think I am in a position to decide what is broken and for whom, etc. 

I can only advice - the final call is always yours since you are the admin. I only wish you think through things; many people rely on this library and over the years - we have come to define ways to get work done in specific ways - changes drastically alters people's workflows in ways that might not be obvious in the first instance, so its best to think from a third person perspective rather than "it works fine for me!" approach.

Best.

**Comment by dustinfarris** [comment]
@akashpal-21 

Thanks for the honest feedback.

> the final call is always yours since you are the admin.

Just to be clear for anyone else reading this, I'm not pretending to be a BDFL or anything of the sort.  This project needed an admin and I volunteered to fill that gap.  I'm happy to be a decision-maker if needed, but in no way intend to impose my own opinions unilaterally; and I truly appreciate when I'm called out on things that appear that way — I'm not immune to blind spots.

> you propose a change wherein it is user customisable whether ID will be created.

Not quite.  My POV is this: org-roam-capture should create a new node — full stop.  Whether that node is at the file level or the entry level I think should be customizable — and I'm not sure if this PR takes a positive step in that direction.  What special treatment (if any) should be given to `olp` and `datetree` targets is something I have not fully considered and would love to have a discussion about that.

My main goal with this PR is to give users an emergency hatch out of the breaking change in ed94524964a0b03538a84ae03c89ec61527ffe7d but that seems to have only been partially accomplished here.

> Is this what users are complaining such as in #2551 ? or is user complaining to get the old "broken" behaviour?

This change only partially address #2551 in that it removes the entry ID creation.  It does not account for the fact that this effectively leaves us nodeless — which doesn't seem right, so more to do here.

> its best to think from a third person perspective rather than "it works fine for me!" approach.

It may not be obvious but I am trying to think holistically about the design of org-roam and very much respect our users and their workflows.  I really do appreciate your feedback.  You seem to have a very mature mindset and I'm glad you're willing to take time to participate here.  If you ever want to connect 1:1 to talk things through I'd be more than happy to.

**Comment by akashpal-21** [comment]
> Not quite. My POV is this: org-roam-capture should create a new node — full stop. Whether that node is at the file level or the entry level I think should be customizable — and I'm not sure if this PR takes a positive step in that direction.

This is what I am getting at too,
For that you'd need to create a cond here for ex - 

```
(cond
      ((and (eq (org-capture-get :type) 'entry)
	    (plist-get (org-capture-get :org-roam) :entry-node))
       (advice-add #'org-capture-place-entry :after #'org-roam-capture--create-id-for-entry)
       (org-roam-capture--put :new-node-p t)
       (setq id (org-roam-node-id org-roam-capture--node)))
      
      ((eq (org-capture-get :type) 'entry)
       (setq p (point-min))
       (setq id (org-roam-capture--create-id p)))
      
      (t
       (setq id (org-roam-capture--create-id p))))
```

where

```
(defun org-roam-capture--create-id (p)
  "Create or assign an Org-roam ID at point P.
If no ID exists, mark node as new and assign it. Returns the ID."
  (save-excursion
    (goto-char p)
    (unless (org-entry-get p "ID")
      (org-roam-capture--put :new-node-p t))
    (let ((id (or (org-entry-get p "ID")
                  (org-roam-node-id org-roam-capture--node))))
      (setf (org-roam-node-id org-roam-capture--node) id)
      (org-entry-put p "ID" id)
      id)))
```


see - my point what that we are already using p to track the pos where the template expansion would occur - my idea was to re-use the value of p irrespective of whether it was entry or non-entry if user desired. But all we are doing here is ignoring (point) defined in date-tree if it was used - because it often chooses point-min for all others - effectively making it a file level id creation.

If you agree to re-use p if user wants to break out then we would have enough degrees of freedom - that is my only point!

**Comment by anpandey** [comment]
>Would the user use plain type in this case?

Not necessarily - A user might prefer to keep the ID at the file level, but still capture to certain, predefined headlines within that file (indeed this is something that I use within my own daily templates).

**Comment by akashpal-21** [comment]
Also another regression if p is not reused - :target node will be made superfluous for entry type, and I think this would also bork the ID stored in the database with respect to the entry - for example

If document structure is as follows

```
:PROPERTIES:
:ID: xyz
:END:
#+TITLE: helloworld

* node
:PROPERTIES:
:ID: xyz1
:END:
```

and user decides to :target node (xyz1) when doing capture with entry - the associated ID that will be stored in the database would be of the file level


```
:PROPERTIES:
:ID: xyz
:END:
#+TITLE: helloworld

* node
:PROPERTIES:
:ID: xyz1
:END:

** Capture
```

in database - capture would be related with :ID xyz instead of xyz1! Which is breaking the org structure that user expects!

A brief breakdown of how 'p' is determined for all target types is illustrated below
```
Target_Type,p_value
file,(goto-char (point-min))
file+olp,(point-min)
file+head,if new-file -> (point-max) else -> (goto-char (point-min))
file+head+olp,(point-min)
file+datetree,(point)
node,(org-roam-node-point node)
```

Note - for file+head, for new-file point-max and point-min are equivalent unless user forcefully creates olp structure in head

```
(when new-file-p
         (org-roam-capture--put :new-file path)
         (insert (org-roam-capture--fill-template head 'ensure-newline))
         (setq p (point-max)))
```

In which case - capture would create another ID when recapturing in the same file - at the file level (tries targeting head in new capture if hand is forced; else targets file)

**Comment by akashpal-21** [comment]
on further analysis - it seems this convulated logic of choosing min or max depending on new-file or not for file+head was inserted in #2bdeb7e (https://github.com/org-roam/org-roam/commit/2bdeb7ed2523ff7830cb00dc24e329da1dcd7ef9)

The older approach was more straight forward

```
      (`(file+head ,path ,head)
       (setq path (org-roam-capture--target-truepath path)
             new-file-p (org-roam-capture--new-file-p path))
       (set-buffer (org-capture-target-buffer path))
       (when new-file-p
         (org-roam-capture--put :new-file path)
         (insert (org-roam-capture--fill-template head 'ensure-newline)))
       (widen)
       (setq p (goto-char (point-min))))
```

This maybe off topic - but since I was discussing how p goes through - it seems pertinent to mention this.

The new approach doesn't look elegant to me - but appears to be a fast hack to get the work done. The new logic looks prone to cause bugs later.

FrederickGeek8:
My two cents:
I think we all need to be _really_ careful about "breaking workflows" (e.g. #2551) in this discussion and what comes of it. What I'm seeing here is discussions here about _changing defaults_ instead of unbreaking[^1] **breaking changes** and adding more **breaking changes**, in the sense of needing (in my opinion) a **major release** (in the semver parlance).

I believe adding new functionality is fine, but I don't think that should be the default. It can be opt-in for sure, but if new behavior _replaces_ that which other users and libraries rely on, then this should be part of a new _major_ release.

**I strongly believe the "correct" behavior for the time being is the old behavior**[^2]. Add the new behavior as opt-in, and then when org-roam is ready for a _major release_ then it can become the default.[^3]

I want to end with saying that I really appreciate this project and have an immense amount of respect for all that you do. I cannot speak for all users nor dependent library maintainers, but I think the current state of the repo and the discussion here is firmly in the territory of breaking changes & major releases. We should respect that and remember to treat it as such.

[^1]: "Unbreaking" here meaning either reverting the recent breaking changes or making the new defaults explicitly opt-in.
[^2]: As in <=[v2.3.1](https://github.com/org-roam/org-roam/releases/tag/v2.3.1). Most of these discussions (including #2543) revolve around the wake of changes introduced in #2056 which I have seen crop up in Issue tickets outside of just #2551.
[^3]: With changes to relied-upon behavior (i.e. breaking changes), I would also expect to see more than a footnote in a changelog as well as a migration guide. A major release (i.e. v3) would also raise flags for dependent users and libraries to not blindly upgrade, and would also prevent certain versioning systems from automagically upgrading.

akashpal-21:
@FrederickGeek8 I concur on reverting all changes made to org-roam-capture.el in the meantime, it needs proper discussion before being implemented in the library.

akashpal-21:
@FrederickGeek8 

https://github.com/org-roam/org-roam/issues/2552

See 2552 for a fork from 2.3.1 where the new behaviour is merely opt in. Although I think a decision should still be taken there whether to manually and arbitrarily turn off :target-entry-p nil in file+head.

hpfr:
@dustinfarris @meedstrom @akashpal-21 @dmgerman

There have been a number of [capture issues opened](https://github.com/org-roam/org-roam/issues?q=is%3Aissue%20created%3A%3E%3D2025-09-22%20label%3A%222.%20capture%22) since #2056 was merged. There are more reports on the Discourse.

Short term: Leaning towards reverting #2056 (plus the follow up commits) for the reasons @FrederickGeek8 mentioned.

Mid term: We should figure out how many [new capture issues](https://github.com/org-roam/org-roam/issues?q=is%3Aissue%20created%3A%3E%3D2025-09-22%20label%3A%222.%20capture%22) are due to that change, close any duplicates, and then maybe try again (#2056 was supposed to fix https://github.com/org-roam/org-roam/issues/2039), this time with test cases based on the issues.

Long term: Capture is really brittle. It’s probably going to take someone becoming an expert in the current implementation to be able to fix existing issues without introducing regressions. It looks like @dmgerman did a lot in https://github.com/org-roam/org-roam/discussions/2573. If it breaks the capture template API or changes behavior of existing templates in the wild, it’s probably not something we can consider short term, but long term, it might not be possible to fix all the existing capture issues without a major version bump?

Pipe dream: to be honest I think the whole string-based design of Org capture templates is not great, s-expressions are better (e.g. https://github.com/minad/tempel)

Aside: if you’re affected by this like @FrederickGeek8, consider pinning org-roam. And you can’t rely on Doom’s pins in cases like this, Henrik catches a lot but edge cases in runtime behavior like this aren’t really going to be visible to him when bumping pins and checking that packages run.

WDYT?

akashpal-21:
@hpfr sounds good to me, since we are playing catch and throw with org-capture, it is by design something brittle, introducing new features with backward compatibility seems the only pragmatic path without significant refactor. Best. Would take some chore work to clean up all the issues.

dustinfarris:
> if new behavior replaces that which other users and libraries rely on, then this should be part of a new major release.

fwiw we _can_ do a major release to melpa stable (which is still on v2.3.1 before these breaking changes were made).  not saying this is what we should do, but just want to point out that anyone on melpa stable should be unaffected so far.

I did call out these changes and issues that have stemmed from them to Henrik after he bumped doom's org-roam pin.  Have not heard back though.

---

> Leaning towards reverting https://github.com/org-roam/org-roam/pull/2056

Not opposed.  My tl;dr of the discussion in #2543 is that:

1. My goal when merging #2056 was to repurpose `entry` — which at the time appeared to me to be mostly duplicative of `plain` as far as org-roam-capture is concerned — so that a node can be captured either as a "file", or as an "entry"; however
2. org-capture `entry` carries additional behavior that some org-roam-capture workflows may have been depending on

dustinfarris:
Revert PR: #2586

dmgerman:
Please see #2587 for an alternate proposal.

PR #1982, opened 2021-11-23
Fix org-roam-node-slug for compatibility with future Emacs release #1981

pcrama:
Emacs commit 3f096eb3405b2fce7c35366eb2dcf025dda55783 introduced `string-glyph-compose` and `string-glyph-decompose`, autoloading these instead of the (still existing) `ucs-normalize-NFC-region` and `ucs-normalize-NFD-region`.

There are three cases:

- Emacs where these transitions have not happened yet (e.g. 27.1):  `ucs-normalize-NFC-region` and `ucs-normalize-NFD-region` are still  autoloaded, aliasing the new names to them will keep them usable
  and the code still works.
- Emacs where the new functions are defined (not yet released): the new names  are autoloaded, no aliases are installed and the code still works.
- A (hypothetical?) Emacs where `string-glyph-compose` and  `string-glyph-decompose` are renamed.  If `ucs-normalize-NFC-region` and `ucs-normalize-NFD-region` do not get their autoloaded status back, the  aliasing will happen but the functions not autoloaded and the code will  break in the same way as in #1981

###### Motivation for this change

Fix #1981

PR #2553, opened 2025-10-30
refactor: simplify org-roam-node-slug

meedstrom:
### Easier to read
The previous iteration was very hard to read.  But that's just my opinion.

### Remove autoload

There was no need to shim in an autoload for the Emacs 29 `string-glyph-compose` alias. (Background: https://github.com/org-roam/org-roam/pull/1982)  Packages shouldn't do that sort of thing, it's unhygienic. 

This PR changes that so we use the function name `ucs-normalize-NFC-string`, which is safe across all versions. Of course, you have to do `(require 'ucs-normalize)` too, but that's in line with convention, way more so than adding new definitions.

### Separate `org-roam-node-slugify`

It's often useful IME to be able to slugify any text string, so this separates out a function that does that, and lets the node method just wrap that.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -33,6 +33,7 @@
 ;;
 ;;; Code:
 (require 'crm)
+(require 'subr-x)
 (require 'org-roam)
 
 ;;; Options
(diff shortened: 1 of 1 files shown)


meedstrom:
For what it's worth, I just got `string-glyph-compose` into Compat, so on the next stable Compat release, we could switch back to using that over `ucs-normalize-NFC-string`.  But that'd better wait a couple of months, b/c package.el doesn't really look at dependency versions AFAIK (so may not update all dependencies when updating org-roam).

meedstrom:
It won't hurt to merge https://github.com/org-roam/org-roam/pull/2563 first.

EDIT: Ok, that's done! Checks re-run succeessfully.

PR #2554, opened 2025-11-02
build: add compat.el

meedstrom:

PR https://github.com/org-roam/org-roam/pull/2546 uses `ensure-list`, which came with Emacs 28.

A backport for that (and many other nice things) are in compat.el, so unless we bump the minimum Emacs version, let's just depend on compat and minimize toil.

This is a uncontroversial library, a stub of it is even part of Emacs from Emacs 30 onwards.

Bonus: we'll be able to get rid of `org-roam-dolist-with-progress` in favor of `dolist-with-progress-reporter`.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -6,7 +6,7 @@
 ;; URL: https://github.com/org-roam/org-roam
 ;; Keywords: org-mode, roam, convenience
 ;; Version: 2.3.1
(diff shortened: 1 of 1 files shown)


akashpal-21:
+1
It would be good to have another person such as @meedstrom having merging rights, given this repo is understaffed.

dustinfarris:
> It would be good to have another person such as @meedstrom having merging rights, given this repo is understaffed.

Good suggestion.  I've reached out to @meedstrom via email.

PR #2555, opened 2025-11-03
feat: new command org-roam-db-explore

meedstrom:

Very easy tool to include, very handy.


Delapouite:
thanks!

Issue #2547, opened 2025-10-09
org-roam-db-sync fails at ~96% with "too many open files" (5400+ files, 72MB)


wito-ch:
### Description

`org-roam-db-sync` consistently fails at ~96% with "too many open files" error when processing a Zettelkasten of 5,428 org files (72 MB). This occurs despite maximized system limits (ulimit -n: 524288) and various workaround attempts including chunked processing.

I've been using org-roam for years with a growing collection (currently +30-50 files/month). The sync worked fine with smaller collections but now consistently fails at approximately 5,200 of 5,428 files.

**Note:** Both macOS Tahoe 26 and Emacs 30.2 are very recent releases (2024/2025). This might be related to new OS restrictions or Emacs compatibility issues.

#### Steps to Reproduce

1. Create a collection of ~5,400 org files with org-roam nodes
2. Set `org-roam-directory` to the collection directory
3. Delete or backup existing `org-roam.db`
4. Run `M-x org-roam-db-sync`
5. Observe failure at ~96% progress

**Alternative reproduction with chunked processing:**

```bash
#!/bin/bash
CHUNK_SIZE=300
TOTAL=5428

for ((i=0; i<$TOTAL; i+=CHUNK_SIZE)); do
  emacs --batch -l ~/.emacs --eval "
    (progn
      (require 'org-roam)
      (setq org-element-use-cache nil)
      (setq org-roam-directory \"$HOME/myCloud/org/\")
      (org-roam-db-update-file fullpath))
  "
done

Even chunked processing with separate Emacs processes fails at ~96%.

What I’ve tried:

✅ Maximized system limits (kern.maxfilesperproc: 524288)
✅ Increased launchctl limits
✅ Disabled org-element-cache (setq org-element-use-cache nil)
✅ Tested Emacs 29.4 and 30.2
✅ Multiple chunking strategies (200-500 files per chunk)
✅ Fresh org-roam.db
✅ Minimal Emacs config
All approaches fail at ~96% with the same error.

Backtrace
Error (org-roam): Failed to process /path/to/file.org with error (file-error "Creating pipe" "Too many open files"), skipping...

The error repeats for remaining files after ~5,200 files have been processed.

Additional error context:

Error occurs consistently at the same point (~96%)
System file descriptor limits are not exhausted according to lsof
No other applications are consuming excessive file descriptors
Emacs process shows normal file descriptor usage until the error

Expected Results
[...]

PR #2558, opened 2025-11-03
Faster `org-roam-list-files` (based on directory-files-and-attributes)

meedstrom:
EDIT: See even newer alternative PR: #2608 

This is an alternative implementation of: https://github.com/org-roam/org-roam/pull/2546

It's probably better.

Some differences:

- Evalling `(org-roam-list-files)` now takes me 0.01-0.05s!  (As opposed to 4-5s)
  - That's when I set `org-roam-list-files-commands` to `nil`, of course. Probably there is no need to keep that user option any more (https://github.com/org-roam/org-roam/pull/2558#issuecomment-3510828729).

- Evalling `(org-roam-db-sync)` now takes me 0.10s (when there is nothing to update)
  - Thus it becomes realistic to just throw this in as a preamble to any function or command that may need to ensure a fresh DB.

- Does not use `org-roam-file-p` directly to filter the output of `org-roam-list-files`, because it is not necessary -- several facts, such as being a descendant of `org-roam-directory`, can be safely assumed!   The `org-roam-file-p` predicate is more useful for checking any wild file name in other programming contexts.
  - As a consequence, there is no use of `file-truename` nor `file-relative-name`, so `file-name-handler-alist` would not affect the runtime much (the earlier PR discussed overriding that setting for perf gains).

- Splits `org-roam-file-exclude-regexp` into 4 separate options, which are both more friendly to modification by users, and easier to optimize with.

- Controversial: the performance overhead in `org-roam-db-sync` from comparing SHA1 content hashes is eliminated, because it now just checks filesystem mtime.  I **do not know** if that is sane (feedback please), but I can't see why not.
  - EDIT: Chesterton's Fence satisfied! A legit reason to check the SHA1 is for cases like when the mtime of all files has changed (mass rename), but the content has not, incurring a unnecessary re-sync of them all. 
  
    We are safe from that because this PR does not modify the subroutine `org-roam-db-update-file`, which refuses to re-sync if the SHA1 is the same. It was never necessary to additionally compute SHA1 in `org-roam-db-sync`.

    There's a gotcha here though! Now when there exists such files, i.e. unchanged but with a changed mtime, the DB still holds their old mtimes and thus will attempt a `org-roam-db-update-file` of each of them, at every `org-roam-db-sync`.  This may seem to be a regression until you remember that `org-roam-db-update-file` will just check the SHA1 and then no-op, so the perf of `org-roam-db-sync` cannot become worse than it is currently, because it currently calculates the SHA1 of every file every time anyway.

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -175,10 +175,8 @@ responsibility to ensure that."
   :type '(repeat string)
   :group 'org-roam)
 
-(defcustom org-roam-list-files-commands
-  (if (member system-type '(windows-nt ms-dos cygwin))
-      nil
-    '(find fd fdfind rg))
+;; TODO: Deprecate, the elisp is faster now.
+(defcustom org-roam-list-files-commands nil
   "Commands that will be used to find Org-roam files.
 
 It should be a list of symbols or cons cells representing any of


akashpal-21:
Haven't checked the code, but regarding the hashing - I have had eliminated this much before from my system

```
;; (defun org-roam-db--file-hash (file-path)
;;   "Compute the hash of FILE-PATH."
;;   (with-temp-buffer
;;     (set-buffer-multibyte nil)
;;     (insert-file-contents-literally file-path)
;;     (secure-hash 'sha1 (current-buffer))))

(defun org-roam-db--file-hash (file-path)
  "Compute the hash of FILE-PATH concatenated with its modification time."
  (secure-hash 'sha1 (format "%s %s" file-path (nth 5 (file-attributes file-path)))))
```

This hasn't caused me any issues whatsoever.
Don't see any reason to check file content hash when there could be more cheap check for modification 

Unless something constantly corrupts these attributes for some users!

dmgerman:
It looks great. One suggestion:

instead of using dolist and cl-loop use mapcar. This turns the code from mutable to immutable and makes it easier to read (in my opinion) especially if you extract the mapcar larger lambdas (like in org-roam--list-recursive-subdirs) 

mapcan does mutate the original list to the result list. I think that is desirable in this case. Otherwise it should be replaced with mapcar. the advantage of mapcan is that it does not create a new list. See the documentation for each. The code below uses mapcan

For example:

     (defun org-roam--process-subdir-entry (name dir deny-re follow-symlinks)
       "Process a single directory entry NAME in DIR.
     Returns a list of the directory and its subdirectories if it matches the criteria.
     DENY-RE is the exclusion regex pattern.
     FOLLOW-SYMLINKS determines whether to follow symlinks."
       (when (and (directory-name-p name)
                  (not (member name '("./" "../"))))
         (let ((full-name (concat dir name)))
           (when (and (not (string-match-p deny-re full-name))
                      (or follow-symlinks (not (file-symlink-p full-name))))
             (cons full-name
                   (org-roam--list-recursive-subdirs full-name follow-symlinks))))))
     
     (defun org-roam--list-recursive-subdirs (dir &optional follow-symlinks)
       "Return a list of DIR, its subdirs, sub-subdirs and so on.
     FOLLOW-SYMLINKS means to include dirs that are symlinks and search them
     for subdirs too, but can result in an infinite loop.
     
     Affected by user options:
     - `org-roam-dir-exclude-literals'
     - `org-roam-dir-exclude-regexps'"
       (let ((deny-re (org-roam--dir-deny-re)))
         (mapcan
          (lambda (name)
            (org-roam--process-subdir-entry name dir deny-re follow-symlinks))
          (file-name-all-completions "" dir))))

meedstrom:
Thank you for your kind words. I don't think I agree on easier to read, and you'll have to explain what you mean by turning it from mutable to immutable.  AFAIK these are identical in effect:

```elisp
(cl-loop for x in '(1 2 3 4)
         collect (+ x 10))
```

```elisp
(let (result)
  (dolist (x '(1 2 3 4))
    (push (+ x 10) result))
  (nreverse result))
```

```elisp
(mapcar (lambda (x)
          (+ x 10))
        '(1 2 3 4))
```

I generally prefer cl-loop as my go-to tool, since it also has destructuring (the `(file . attr)` here):

```elisp
;; return directories in current dir
(cl-loop for (file . attr) in (directory-files-and-attributes ".")
         if (eq t (file-attribute-type attr))
         collect file)
```

and many other things, but destructuring in particular is really annoying to lack when using mapcar and friends. Anyway, just a tangent :)

meedstrom:
Btw, the design of `org-roam--list-recursive-subdirs` is essentially copy-pasted from the built-in `directory-files-recursively` and simplified.  That one uses dolist.

Not that the built-ins are well-known for ease of reading... but that's why it's how it is.

akashpal-21:
Unrelated but I too prefer cl-loop. I don't get why some library creators insist on excluding cl - for example the first thing that is written in `dash` library readme is that it excludes `cl` and so on.
Cl-loop is very simple and nonverbose.

I think a reader will find a cl-loop much easy to reason about than something built from first principles.

dmgerman:
> you'll have to explain what you mean by turning it from mutable to immutable. AFAIK these are identical in effect:

I am of the opinion that mutating variables increases the probability of bugs (as one has to keep track of how the variable is mutated).

That is why I think it is not very good idea to **define bindings equal to nil** and then assign them. That means that until they are assigned, they cannot be used, which  is what I mean as increasing the mental load on whoever is reading the code.

That is why I prefer do-loop and mapcar. But do-loop is a macro, so we might need to see if the macro creates garbage. mapcar is low level and very fast.

In this case, I suspect that mapcan will be slightly faster.

But I am happy with cl-loop. Much better than using do-list. cl-loop is basically for python people ;) as it mimics a comprehension.

dmgerman:
This might be relevant to speeding up org-roam

https://mahmoodsh.com/efficiently_parsing_org_files.html

meedstrom:
> This might be relevant to speeding up org-roam
> 
> https://mahmoodsh.com/efficiently_parsing_org_files.html

As yantar92 pointed out on the [Reddit discussion](https://old.reddit.com/r/emacs/comments/1oky4hy/efficiently_parsing_orgmode_files/nmfkuxk/), that article does one thing wrong: `(let ((major-mode 'org-mode)))`.  Alas, you should actually call `(org-mode)` to guarantee it's sane.

But yes, some of the other tricks are good.  I said last month (https://github.com/org-roam/org-roam/issues/2547#issuecomment-3387838208) that it'd make sense to rework `org-roam-db-sync` to not use `org-roam-with-file`, but rather `org-roam-with-temp-buffer`.

meedstrom:
Finally, the checks pass. The tests did good.

meedstrom:
Added debatable fifth commit that sets `org-roam-list-files-commands` to nil by default. 

Since it is my impression these alternative code paths will be unnecessary -- they only ever existed because the elisp version was thought to be slow, right?  

The bottleneck all along seems to have been `org-roam-file-p`, which the other code paths still apply -- and the elisp path no longer does, hence being faster.

meedstrom:
More tests needed, stay tuned.  In particular, must apply exclude-regexps only to the part of filename after `org-roam-directory` (https://github.com/org-roam/org-roam/pull/2178)

EDIT Done: https://github.com/org-roam/org-roam/pull/2575.  This PR will need some changes to pass those.

meedstrom:
Tests pass.

meedstrom:
Two things I can do after this is merged:
- I can make a quick PR for simplifying `org-roam-file-p`, similar to https://github.com/org-roam/org-roam/pull/2546/commits/cabb0ee09dd503c85c1a2d7f8c50c53a0d45ab81.
- I can start work on some optional async code-flow like https://github.com/meedstrom/org-roam-async.  

  - It could remain an extension, but either way, I'd want to figure out how that would best work, before I want to touch `org-roam-node-read--completions` and all its hairy callees. 

    - Especially because it seems like there should be some kind of after-sync-hook which just runs on `after-save-hook` by default -- along with DB sync -- but can be deferred for later if async is enabled. Such a hook would be useful for caching completions as soon as the DB is updated. 
      
      I don't know if caching will be necessary, but anyway, first this PR and then some thinking about async and then we'll get to `org-roam-node-read--completions`.

meedstrom:
Finally figured out the reason for SHA1 content hashes. Edited the OP:

>   EDIT: Chesterton's Fence satisfied! A legit reason to check the SHA1 is for cases like when the mtime of all files has changed (mass rename), but the content has not, incurring a unnecessary re-sync of them all. 
>   
>   We are safe from that because this PR does not modify the subroutine `org-roam-db-update-file`, which refuses to re-sync if the SHA1 is the same. It was never necessary to additionally compute SHA1 in `org-roam-db-sync`.
> 
>   There's a gotcha here though! Now when there exists such files, i.e. unchanged but with a changed mtime, the DB still holds their old mtimes and thus will attempt a `org-roam-db-update-file` of each of them, at every `org-roam-db-sync`.  This may seem to be a regression until you remember that `org-roam-db-update-file` will just check the SHA1 and then no-op, so the perf of `org-roam-db-sync` cannot become worse than it is currently, because it currently calculates the SHA1 of every file every time anyway.

PR #2560, opened 2025-11-05
feat: Allow functions and variables in capture targets

nullterminator-fyi:
Create a function to get the string value of the target path if it's a variable or function, just like in regular org capture templates

For better subdirectory management, I use Denote as the prompt for note creation. In regular org-mode capture templates, Denote recommends using a variable named `denote-last-path` for the target file as org-mode can pick up variable and functions in targets. Such functionality is absent in org-roam, making capture templates unusable with Denote. This commit adds this functionality into org-roam capture templates.

(The evaluation function is a modified version of the function used in org-mode's code, so let me know if there is any licensing concerns.)

### DIFF
--- a/org-roam-capture.el
+++ b/org-roam-capture.el
@@ -592,6 +592,17 @@ capture target."
     (advice-add #'org-capture-place-template :before #'org-roam-capture--set-target-entry-p-a)
     (advice-add #'org-capture-place-template :after #'org-roam-capture-run-new-node-hook-a)))
 
+(defun org-roam-capture--eval-path (path)
+  "Expand functions, symbols and file names for PATH.
+If PATH is a string, return it as is. If it's a function, call it.
+If it's a variable, return it's value. In any other case, raise an error."
+  (let ((location (cond ((stringp path) path)
+                        ((functionp path) (funcall path))
+                        ((and (symbolp path) (boundp path)) (symbol-value path))
+                        (t nil))))
+    (or location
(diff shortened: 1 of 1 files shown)


meedstrom:
You may want to do `C-x h M-x untabify RET`.

Then amend and force-push.

nullterminator-fyi:
Thanks, should be fixed now.

hpfr:
We call `org-roam-capture--fill-template` on capture targets as well. I think org-capture already supports this via `%(EXP)`: https://www.gnu.org/software/emacs/manual/html_node/org/Template-expansion.html

Could you give that a try? Since we already do full template expansion I’d like to avoid expanding the interface here, even if it happens to be similar to upstream org-capture.

nullterminator-fyi:
`%(EXP)` indeed fills the usecase, albeit with some boilerplate code as it only accepts functions and for some reason doesn't work with lambdas in my experience. I can close the PR now.

One extra question: Is this behavior documented? I was looking through the documentation and reading the comments in `org-roam-capture.el`, but I never noticed anything about this. Maybe it was just me and I missed something, but otherwise I think this behavior should be noted somewhere users can easily read. Thanks.

hpfr:
It’s in the docstring of `org-roam-capture--target-truepath` and template examples using %-escapes for the target imply some expansion is happening, but I added a more explicit note in the manual in https://github.com/org-roam/org-roam/pull/2592. Thanks!

PR #2567, opened 2025-11-07
style: use pgformatter on the huge sql expression

meedstrom:

I find it easier to read. Agree?

It's a purely whitespace change (and some lowercase->uppercase), courtesy of the `pgformatter` tool.

EDIT: D'oh, it falls afoul of the 110-char line length maximum.

## Sidenote

I was interested to learn that this expression dates all the way from the release of v2, with only minor changes since. Original from the v2 commit can be found here: https://github.com/org-roam/org-roam/blob/aee3467b3e89ddf49393584586252c2fb4509433/org-roam.el

I feel like there ought to be a way to make this a lot shorter. At least, the EmacSQL expression used in v1 was short, but v1 made different assumptions of course: 

https://github.com/org-roam/org-roam/blob/756f6215b672e267f986a3d6e494f5309825b91a/org-roam.el#L905-L909

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -378,77 +378,78 @@ nodes."
 
 (defun org-roam-node-list ()
   "Return all nodes stored in the database as a list of `org-roam-node's."
-  (let ((rows (org-roam-db-query
-               "
+  (let ((rows (org-roam-db-query "
 SELECT
-  title,
-  aliases,
-
-  id,
-  file,
-  filetitle,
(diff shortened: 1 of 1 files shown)


meedstrom:
I see no good way to work around the 110-char lint failure.

EDIT: Possible solution: https://github.com/gonewest818/elisp-lint/pull/38  But we'll see if that garners interest upstream.

Issue #2532, opened 2025-07-20
finalizer failed: (wrong-type-argument sqlitep nil)


chaicurioquest:
### Description

The basic functionality is working however there is error message "`finalizer failed: (wrong-type-argument sqlitep nil)`" shows up while selecting templates from node. 

#### Steps to Reproduce
1. org-roam--build-cache-async output:
```
 Debugger entered--Lisp error: (void-variable org-roam--build-cache-async)
  eval(org-roam--build-cache-async t)
  #f(compiled-function () #<bytecode -0xc3c33354b8e532e>)()
  #f(compiled-function () #<bytecode -0x5db3e1955cb81d1>)()
  handler-bind-1(#f(compiled-function () #<bytecode -0x5db3e1955cb81d1>) (error) eval-expression--debug)
  eval-expression(org-roam--build-cache-async nil nil 127)
  funcall-interactively(eval-expression org-roam--build-cache-async nil nil 127)
  call-interactively(eval-expression nil nil)
  command-execute(eval-expression)
```
2. org-roam-find-file ouput: 
```
 debug(error (void-variable org-roam--build-cache-async) :backtrace-base eval-expression--debug)
  eval-expression--debug((void-variable org-roam--build-cache-async))
  eval(org-roam--build-cache-async t)
  #f(compiled-function () #<bytecode -0xc3c33354b8e532e>)()
  #f(compiled-function () #<bytecode -0x5db3e1955cb81d1>)()
  handler-bind-1(#f(compiled-function () #<bytecode -0x5db3e1955cb81d1>) (error) eval-expression--debug)
  eval-expression(org-roam--build-cache-async nil nil 127)
  funcall-interactively(eval-expression org-roam--build-cache-async nil nil 127)
  call-interactively(eval-expression nil nil)
  command-execute(eval-expression)
```

#### Backtrace

finalizer failed: (wrong-type-argument sqlitep nil)
Clearing removed files...done
Processing modified files...done

Org roam Configuration settings.

```
(use-package org-roam
  :straight t
  :init
  (setq org-roam-directory default-directory)
  (let ((roam-subdir (expand-file-name "roam" org-roam-directory)))
    (unless (file-exists-p roam-subdir)
      (make-directory roam-subdir t)))
  (setq org-roam-dailies-directory "roam")
  :bind (("C-c r f" . org-roam-node-find)
         ("C-c r i" . org-roam-node-insert)
         ("C-c r n" . org-roam-capture))
  :custom
[...]

PR #2571, opened 2025-11-08
Gather dirty load effects into a single mode

meedstrom:
In general, just loading a `*.el` file should have no effect on the Emacs session, other than defining functions and variables.   If a package needs to cause side effects, it is best practice to do so in a mode, not on load.

This PR moves all side effects that I know of, into a new mode `org-roam-core-mode`.

To avoid breaking things for users, it is enabled on load (heh). 

I know it's a bit iffy to add a whole new mode. There's no rush with this PR.

But I have myself had issues with org-roam's lack of hygiene not once, but *twice*, while developing [org-node](https://github.com/meedstrom/org-node):

1. The fact that the calendar is always marked according to which days have an `org-roam-dailies` entry... unconditionally. Even in `org-read-date` popups not called by org-roam!
 
   It means that org-node can't mark days its own way if the user has loaded `org-roam-dailies`.  User basically has to restart Emacs and be careful to do nothing that accidentally loads `org-roam-dailies`.

   - Offending lines are https://github.com/meedstrom/org-roam/blob/ecb9747f0e0be89607ebe2d4ec1a3fe84f940b9f/extensions/org-roam-dailies.el#L321-L322, but I didn't want to move them into the new `org-roam-core-mode`, as that would make it necessary to always load `org-roam-dailies`.  Instead, I made the function into a potential no-op dependent on the new variable `org-roam-enabled`, which is tied to that mode.

2. The fact that `org-roam-id-find` preferentially looks in the `org-roam-db` means that if the user has disabled `org-roam-db-autosync-mode`, you can get a strange situation where org-id might actually know the current, correct location of an ID, but clicking on an ID-link results in trying to visit the location where the (stale) `org-roam-db` thinks it is.  Sending the user to a new blank file, under the file-name that *used to be* associated with that ID.

   In my package, I tried to keep the `org-id-locations` table updated, and then trust that *clicking on ID-links does look up that table*. Imagine my confusion.

Anyway, neither of these would've been much of a problem if there was some mode that the user could easily disable.

### DIFF
--- a/extensions/org-roam-dailies.el
+++ b/extensions/org-roam-dailies.el
@@ -310,7 +310,8 @@ Return (MONTH DAY YEAR) or nil if not an Org time-string."
 
 (defun org-roam-dailies-calendar-mark-entries ()
   "Mark days in the calendar for which a daily-note is present."
-  (when (file-exists-p (expand-file-name org-roam-dailies-directory org-roam-directory))
+  (when (and org-roam-enabled
+             (file-exists-p (expand-file-name org-roam-dailies-directory org-roam-directory)))
     (dolist (date (remove nil
                           (mapcar #'org-roam-dailies-calendar--file-to-date
                                   (org-roam-dailies--list-files))))
--- a/org-roam-log.el
+++ b/org-roam-log.el
@@ -41,7 +41,6 @@
   (run-hooks 'org-roam-log-setup-hook))
 
 (add-hook 'org-roam-log-setup-hook #'org-roam--register-completion-functions-h)
-(add-hook 'org-log-buffer-setup-hook #'org-roam-log--setup)
 
 (provide 'org-roam-log)
 ;;; org-roam-log.el ends here
--- a/org-roam.el
+++ b/org-roam.el
@@ -93,6 +93,7 @@
 (require 'ol)
 (require 'org-element)
 (require 'org-capture)
+(require 'org-protocol)
 
 (require 'ansi-color) ; to strip ANSI color codes in `org-roam--list-files'
 
@@ -345,6 +346,38 @@ E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
       (when (and (file-readable-p file)
                  (org-roam-file-p file))
         (push file result)))))
+
+(defvar org-roam-enabled nil)
+(define-minor-mode org-roam-core-mode
+  "Configure Emacs for org-roam."
+  :group 'org-roam
+  :global t
+  (let ((node-protocol '("org-roam-node" :protocol "roam-node" :function org-roam-protocol-open-node))
+        (ref-protocol  '("org-roam-ref"  :protocol "roam-ref"  :function org-roam-protocol-open-ref)))
+    (if org-roam-core-mode
+        (progn
+          (setq org-roam-enabled t)
+          (advice-add 'org-id-find :before-until #'org-roam-id-find)
+          (add-hook 'org-log-buffer-setup-hook #'org-roam-log--setup)
+          (push node-protocol org-protocol-protocol-alist)
+          (push ref-protocol org-protocol-protocol-alist))
+      (setq org-roam-enabled nil)
+      (advice-remove 'org-id-find #'org-roam-id-find)
+      (remove-hook 'org-log-buffer-setup-hook #'org-roam-log--setup)
+      (setq org-protocol-protocol-alist (->> org-protocol-protocol-alist
+                                             (delete node-protocol)
+                                             (delete ref-protocol))))))
+
+;; In general, just loading a .el file should have no effect on the Emacs session,
+;; other than defining functions and variables.
+;; Enabling a mode is what is allowed to mutate other things,
+;; and the effects should be reversible by disabling the mode.
+
+;; However, org-roam has never obeyed that principle, so users won't expect
+;; to have to enable a new mode to make things work again after an update.
+;; Therefore, we enable the mode on load, avoiding any surprises on update.
+;; This at least provides something that the user can disable.
+(org-roam-core-mode 1)
 
 ;;; Package bootstrap
 (provide 'org-roam)
--- a/extensions/org-roam-protocol.el
+++ b/extensions/org-roam-protocol.el
@@ -113,6 +113,7 @@ See `org-roam-capture-templates' for the template documentation."
                                          ((const :format "%v " :kill-buffer) (const t))))))))
 
 ;;; Handlers
+;;;###autoload
 (defun org-roam-protocol-open-ref (info)
   "Process an org-protocol://roam-ref?ref= style url with INFO.
 
@@ -150,6 +151,7 @@ It opens or creates a note with the given ref.
      :templates org-roam-capture-ref-templates))
   nil)
 
+;;;###autoload
 (defun org-roam-protocol-open-node (info)
   "This handler simply opens the file with emacsclient.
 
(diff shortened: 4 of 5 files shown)


dustinfarris:
So in a perfect world the user would initiate side effects something like:

```lisp
(use-package org-roam
  :defer t
  :config
  (org-roam-core-mode))
```

Do I understand correctly?

meedstrom:
I think so, yes. https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html

> Simply loading a package should not change Emacs’s editing behavior. Include a command or commands to enable and disable the feature, or to invoke it.

Debatable what is meant by "editing behavior" though.

meedstrom:
Could also put it all in with `org-roam-db-autosync-mode`, as it already does more things than just set up DB auto sync (https://github.com/org-roam/org-roam/pull/2571#issuecomment-3508771631).

However, there are plenty of users who don't turn it on, so that's a bit of a no-go at this stage.

BTW, the name "org-roam-core-mode" is just my first idea. Most natural might have been simply "org-roam-mode" -- but that is reserved by the org-roam buffer.

dustinfarris:
> BTW, the name "org-roam-core-mode" is just my first idea. Most natural might have been simply "org-roam-mode" -- but that is reserved by the org-roam buffer.

Maybe `global-org-roam-mode`?  à la [`global-auto-revert-mode`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Auto-Revert.html)

Also, in the spirit of backwards compatibility I wonder if an opt-in variable to inhibit side effects until load would work?  Something like:

```elisp
(defvar org-roam-inhibit-on-load nil
  "When `nil', run side effects immediately when org-roam is loaded.
When non-nil, wait until `global-org-roam-mode' is enabled.")

;; ...

(unless org-roam-inhibit-on-load
  (warn "org-roam is configuring emacs on load.  To prevent this, set `org-roam-inhibit-on-load' to t")
  (global-org-roam-mode 1))
```

```elisp
(use-package org-roam
  :defer t
  :init
  (setq org-roam-inhibit-on-load t)
  :config
  (global-org-roam-mode))
```

meedstrom:
> Maybe global-org-roam-mode? à la [global-auto-revert-mode](https://www.gnu.org/software/emacs/manual/html_node/emacs/Auto-Revert.html)

To me that implies there exists a buffer-local `org-roam-mode`, of which that is merely the globalized variant.  Like how `global-auto-revert-mode` is just a trivial mode that turns on/off the local `auto-revert-mode` in all relevant buffers, which does the real work. (Well, that's how most globalized minor modes work, I don't know `auto-revert-mode` specifically)

As it definitely wouldn't be the globalized variant but something else entirely, that's not a good idea.

> Also, in the spirit of backwards compatibility I wonder if an opt-in variable to inhibit side effects until load would work? 

Interesting idea. Although users that want to inhibit it could allow it to turn on normally at load-time and then turn it off:

```elisp
(use-package org-roam
  :defer t
  :config
  ;; turn it off
  (org-roam-core-mode 0))
```

meedstrom:
Clarification: When I said `org-roam-db-autosync-mode` does more than only set up DB auto-sync, I meant this line in it:

```elisp
(add-hook 'find-file-hook #'org-roam-db-autosync--setup-file-h)
```  

which leads to

```elisp
(defun org-roam-db-autosync--setup-file-h ()
  "Setup the current buffer if it visits an Org-roam file."
  (when (org-roam-file-p) (run-hooks 'org-roam-find-file-hook)))
```

and that hook, `org-roam-find-file-hook`, is used by things that have nothing to do with DB autosync, such as the org-roam buffer.  From file `org-roam-mode.el`, we see that it's how it enables the automatic redisplaying of the buffer:

```elisp
(add-hook 'org-roam-find-file-hook #'org-roam-buffer--setup-redisplay-h)
```

So when a user turns off `org-roam-db-autosync-mode`, they find that the org-roam buffer also stops refreshing itself. 

Clearly a case could be made to move that first line I mentioned, from `org-roam-db-autosync-mode`, into the new `org-roam-core-mode` (or whatever we end up calling it).

meedstrom:
~~Another feature we can comfortably add, with the mode in place: apply a patch for the intermittent message "`wrong type argument: sqlitep nil`".~~

~~It's an upstream bug. Thanks to https://github.com/org-roam/org-roam/issues/2532#issuecomment-3457190149 for figuring it out~~

EDIT: Nevermind, upstreamed the fix https://github.com/magit/emacsql/pull/132

extensions/org-roam-dailies.el
```diff
@@ -310,7 +310,8 @@ Return (MONTH DAY YEAR) or nil if not an Org time-string."
 
 (defun org-roam-dailies-calendar-mark-entries ()
   "Mark days in the calendar for which a daily-note is present."
-  (when (file-exists-p (expand-file-name org-roam-dailies-directory org-roam-directory))
+  (when (and org-roam-enabled
```
hpfr, line 313: I think we might as well handle the dailies calendar marking with a dailies global minor mode replacing the top-level hook additions as well. We’ll have to enable it by default for now like the core mode you’re introducing.

org-roam-log.el
```diff
@@ -41,7 +41,6 @@
   (run-hooks 'org-roam-log-setup-hook))
 
 (add-hook 'org-roam-log-setup-hook #'org-roam--register-completion-functions-h)
```
hpfr, line 43: While this is modifying an org-roam hook, I think it still makes sense to move the side effect to the minor mode.

org-roam.el
```diff
@@ -346,6 +347,38 @@ E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
...
+        (progn
+          (setq org-roam-enabled t)
+          (advice-add 'org-id-find :before-until #'org-roam-id-find)
+          (add-hook 'org-log-buffer-setup-hook #'org-roam-log--setup)
+          (push node-protocol org-protocol-protocol-alist)
+          (push ref-protocol org-protocol-protocol-alist))
+      (setq org-roam-enabled nil)
+      (advice-remove 'org-id-find #'org-roam-id-find)
+      (remove-hook 'org-log-buffer-setup-hook #'org-roam-log--setup)
+      (setq org-protocol-protocol-alist (->> org-protocol-protocol-alist
```
hpfr, line 367: no dash anymore :) I like `(cl-callf2 delete node-protocol org-protocol-protocol-alist)`

org-roam.el
```diff
@@ -346,6 +347,38 @@ E.g. (\".org\") => (\"*.org\" \"*.org.gpg\")"
                  (org-roam-file-p file))
         (push file result)))))
 
+(defvar org-roam-enabled nil)
```
hpfr, line 350: Don’t we get `org-roam-core-mode` as a variable for free with `define-minor-mode`?

hpfr (changes requested):
Nice! Just some nits.

> Clearly a case could be made to move that first line I mentioned, from org-roam-db-autosync-mode, into the new org-roam-core-mode (or whatever we end up calling it).

Case is compelling enough to me!

org-roam.el
meedstrom: We do! For some reason I felt that it was more instructive to have this explicit defvar. But I'm not attached to it, so it can go.

org-roam-log.el
meedstrom: It's not only an org-roam hook, it's an org-roam-log hook, i.e. defined in the same file :) I don't think it counts as a side-effect then.  But it could be moved into the default value, at the defcustom on line 29.

extensions/org-roam-dailies.el
meedstrom: Yes, it could be simpler to reason about.  I was probably considering the ergonomics of having *two* new modes to toggle, if you're for some reason toggling a lot.  But that's not really our problem.

hpfr: Yeah, and we’ll enable it by default (for now at least), so it should just make it easier to disable.

org-roam-log.el
hpfr: Ah, thanks for clarifying! I like moving it to the default.

PR #2340, opened 2023-04-15
(fix): improve org-roam buffer fontification speed #1853

kadircancetin:
Related issue: #1853

When there are a lot of nodes to write on the org-roam-buffer, Emacs starts to hang. This improvement decreases the fontification run time from `1.6 sec` to `0.07 sec` for a 10 node example (in my org-mode setup).

The basic idea is that (org-mode) has lots of initial needs, and users can have special org-mode hooks. Instead of calling org-mode for every node fontification, we could initialize it once and use it later.

You can test the speed improvements without needing to pull changes with:

```elisp
(with-eval-after-load 'org-roam-mode

  (setq org-roam-fontify-buffer (get-buffer-create "*org-roam-fontify-buffer*"))
  (with-current-buffer org-roam-fontify-buffer (org-mode))

  (defun org-roam-fontify-like-in-org-mode (s)
    (with-current-buffer org-roam-fontify-buffer
      (erase-buffer)
      (insert s)
      (let ((org-ref-buffer-hacked t))
        (setq-local org-fold-core-style 'overlays)
        (font-lock-ensure)
        (buffer-string))))

  ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
  ;; Advice for runtime test
  
  (defun calculate-runtime (orig-fun &rest args)
    (let ((start-time (float-time)))
      (apply orig-fun args)
      (message "Runtime of `org-roam-backlinks-section` is %.6f seconds."
               (- (float-time) start-time))))

  (advice-add 'org-roam-backlinks-section :around #'calculate-runtime))

```

Test file
```org
:PROPERTIES:
:ID:       c5e11a88-78c0-433a-abfb-43a95f28376e
:END:
#+title: org roam test


* A
:PROPERTIES:
:ID:       9b073a73-0d3b-4309-97bc-bbb672deb07b
:END:
[[id:9b073a73-0d3b-4309-97bc-bbb672deb07b][A]]
[[id:6a76ce97-7ddb-490a-a27a-33f550818055][B]]
[[id:17159c89-1938-4311-84ed-37f8aac15aee][C]]
[[id:0f6b16dc-e275-4c39-aac9-ffc36a5bc977][D]]
[[id:be1b11e6-8b6d-4a50-bc6f-5ec841c14989][E]]
[[id:8efcfa66-0f43-45b3-9737-9aabe0d304aa][F]]
[[id:61d131ff-cc13-40ff-970d-ddbf483b7980][G]]
[...]

PR #2572, opened 2025-11-10
refactor: More sane `org-roam-with-temp-buffer`, slight perf

meedstrom:
# EDITED (Feb 2026)

This PR does two things: 

- a small change for performance
- a sanity fix

## Perf change

This matters little now that we won't even use this macro for generating backlink previews, since merging the excellent #2340.

It may have some impact if [as I've suggested](https://github.com/org-roam/org-roam/pull/2558#issuecomment-3505416637), we rework `org-roam-db-update-file` to make use of this macro (that is, the `org-roam-with-temp-buffer` macro). Which we definitely should, by the way, to avoid "too many files" error #2547.  

Then evalling `(org-roam-db-sync :force)` would create a temp buffer for every Org-roam file. How much this PR will matter in relative terms depends on how fast that'll be otherwise, which we don't know yet. 

In absolute terms on my system, the savings add up to just ~2 seconds across 2600 files. Meh.  And `(org-roam-db-sync :force)` takes ~120 seconds, so that's one pill in a barrel o'drugs for now.  FWIW, I think that 120 figure should be able to go down a lot.

Anyway, here's benchmark code you can run. Probably you see a bigger difference the more files you got in `org-agenda-files`.  

Without checking out the PR branch, compare these three benchmarks in your current Emacs:

```elisp

(setq FILES (org-roam-list-files))

(seq-let (total _ gc)
    (benchmark-run 1
      (let ((files (copy-sequence FILES)))
        (dolist (file files)
          (org-roam-with-temp-buffer file
            ;; Traverse elements to test the org-element cache... probably
            (while (= 0 (org-next-visible-heading 1))
              (org-roam-end-of-meta-data t))))))
  (- total gc))

(seq-let (total _ gc)
    (benchmark-run 1
      (let ((files (copy-sequence FILES))
            (org-agenda-files nil))
        (dolist (file files)
          (org-roam-with-temp-buffer file
            (while (= 0 (org-next-visible-heading 1))
              (org-roam-end-of-meta-data t))))))
  (- total gc))

(seq-let (total _ gc)
    (benchmark-run 1
      (let ((files (copy-sequence FILES))
            (org-agenda-files nil)
            (org-element-cache-persistent nil))
        (dolist (file files)
          (org-roam-with-temp-buffer file
            (while (= 0 (org-next-visible-heading 1))
              (org-roam-end-of-meta-data t))))))
  (- total gc))
```

For me with about 30 agenda files, binding `(org-agenda-files nil)` shaves ~40% off the runtime.

As for binding `(org-element-cache-persistent nil)`, that makes barely a difference on my testing today, but I remember it being a big deal once.  Could just be my emacs session's different.  

I can edit out that binding, but it should be harmless. The org-element cache should not try to persist cache for the contents of a temp buffer, it would never be reused!

## Sanity fix

Note I moved `(delay-mode-hooks (org-mode))` to *after* the file content is inserted.  

This is good hygiene because it allows the Org initialization to read file variables such as `#+STARTUP: odd`, so that functions like `org-do-promote` and `org-reduced-level` do the right thing.

### DIFF
--- a/org-roam-utils.el
+++ b/org-roam-utils.el
@@ -136,11 +136,13 @@ If FILE, set `default-directory' to FILE's directory and insert its contents."
     `(let ((,current-org-roam-directory org-roam-directory))
        (with-temp-buffer
          (let ((org-roam-directory ,current-org-roam-directory)
+               (org-agenda-files nil) ; perf
+               (org-element-cache-persistent nil) ; perf
                (org-inhibit-startup t))
-           (delay-mode-hooks (org-mode))
            (when ,file
              (insert-file-contents ,file)
              (setq-local default-directory (file-name-directory ,file)))
+           (delay-mode-hooks (org-mode))
            ,@body)))))
 
 ;;; Formatting


akashpal-21:
Can you also figure out where these are being queried in? I couldn't do a rg to figure out the dependency call.

```
[~/.e/e/org-roam-20250324.2140] ~> rg org-roam-with-temp-buffer
org-roam-utils.el
157:(defmacro org-roam-with-temp-buffer (file &rest body)

org-roam-mode.el
462:    (org-roam-with-temp-buffer file
```
```
[~/.e/e/org-roam-20250324.2140] ~> tail -n +459 org-roam-mode.el | head -n 11

(defun org-roam-preview-get-contents (file pt)
  "Get preview content for FILE at PT."
  (save-excursion
    (org-roam-with-temp-buffer file
      (org-with-wide-buffer
       (goto-char pt)
       (let ((s (funcall org-roam-preview-function)))
         (dolist (fn org-roam-preview-postprocess-functions)
           (setq s (funcall fn s)))
         s)))))

```
```
[~/.e/e/org-roam-20250324.2140] ~> rg org-roam-preview-get-contents
org-roam-mode.el
406:               (org-roam-preview-get-contents (org-roam-node-file source-node) point))
459:(defun org-roam-preview-get-contents (file pt)
```

```
[~/.e/e/org-roam-20250324.2140] ~> tail -n +371 org-roam-mode.el | head -n 39

(cl-defun org-roam-node-insert-section (&key source-node point properties)
  "Insert section for a link from SOURCE-NODE to some other node.
The other node is normally `org-roam-buffer-current-node'.

SOURCE-NODE is an `org-roam-node' that links or references with
the other node.

POINT is a character position where the link is located in
SOURCE-NODE's file.

PROPERTIES (a plist) contains additional information about the
link.

Despite the name, this function actually inserts 2 sections at
the same time:

1. `org-roam-node-section' for a heading that describes
   SOURCE-NODE. Acts as a parent section of the following one.

2. `org-roam-preview-section' for a preview content that comes
   from SOURCE-NODE's file for the link (that references the
   other node) at POINT. Acts a child section of the previous
   one."
  (magit-insert-section section (org-roam-node-section)
    (let ((outline (if-let ((outline (plist-get properties :outline)))
                       (mapconcat #'org-link-display-format outline " > ")
                     "Top")))
      (insert (concat (propertize (org-roam-node-title source-node)
                                  'font-lock-face 'org-roam-title)
                      (format " (%s)"
                              (propertize outline 'font-lock-face 'org-roam-olp)))))
    (magit-insert-heading)
    (oset section node source-node)
    (magit-insert-section section (org-roam-preview-section)
      (insert (org-roam-fontify-like-in-org-mode
               (org-roam-preview-get-contents (org-roam-node-file source-node) point))
              "\n")
      (oset section file (org-roam-node-file source-node))
      (oset section point point)
   ```

meedstrom:
Not sure what you mean, but currently org-roam-with-temp-buffer is used once for each backlink rendered.

meedstrom:
On a related matter, I recently learned it's probably sanest to insert file content before enabling org: https://github.com/nobiot/org-transclusion/pull/282#issuecomment-3676553675

may be worth a PR if this macro will be used more places.

akashpal-21:
I mean why are these agenda files being queried even? It is not clear from the dependency call. Could there be a more generic way rather than targeting org-agenda-files specifically?

meedstrom:
Oh yeah, that's in `org-mode` itself. Upstream issue. And @yantar92 does want to refactor that IIRC.

yantar92:
@meedstrom I am not sure what you are referring to.

meedstrom:
That enabling org in any buffer visits all the agenda files (iirc), very bad/unnecessary when you loop over many buffers just to use org's fontification or whatever.

yantar92:
Martin Edström ***@***.***> writes:

> That enabling org in any buffer visits all the agenda files (iirc), very bad/unnecessary when you loop over many buffers just to use org's fontification or whatever.

Enabling Org in any buffer does not visit all the agenda files, no.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

meedstrom:
It does something, with a measurable impact. I'm not at my computer at present.

PR #2586, opened 2025-12-27
revert!(capture): do not generate headline id for entry

dustinfarris:
Unwinding the breaking changes to org-roam-capture so that we can rethink the best path forward here.


akashpal-21:
Looks good to be merged. It is inversing the following:

https://github.com/org-roam/org-roam/commit/ed94524964a0b03538a84ae03c89ec61527ffe7d
https://github.com/org-roam/org-roam/commit/2bdeb7ed2523ff7830cb00dc24e329da1dcd7ef9
https://github.com/org-roam/org-roam/commit/41f9a10be587b47454ca8ef26f6a8247605cbe34

dustinfarris:
Thanks for the review @akashpal-21.  @hpfr or @meedstrom I'd appreciate it if one of you could sign off on this as well before we merge it.

meedstrom:
I'll take you guys' word that this reverts those commits, and agree in that case. Let's get this show on the road.

Some notes from reading the diff, FWIW.

1. The change to file `extensions/org-roam-dailies.el` is purely cosmetic, so it's fine for sure.

2. I wonder why `org-roam-capture--adjust-point-for-capture-type` doesn't just use `org-roam-end-of-meta-data`, instead of reinventing it as a while-loop. The function could be as short as

   ```elisp
   (defun org-roam-capture--adjust-point-for-capture-type (&optional pos)
    (goto-char (or pos (point)))
    (pcase (org-capture-get :type)
      (`plain
       (if (org-capture-get :prepend)
           (org-roam-end-of-meta-data t)
         (goto-char (org-entry-end-position))))))
   ```

   (I recognize that from my impementation: https://github.com/meedstrom/org-node/blob/3956804cc0cc19fba0d98b2eb9b76056feda77ae/org-node.el#L1264-L1266)

akashpal-21:
@meedstrom Maybe you can refactor that after this revert. That way we can follow things more smoothly.

You can manually diff with v2.3.1 to see if everything is alright from your side. It should revert to status quo 2.3.1. that would be a good check.

dustinfarris:
> 1. The change to file `extensions/org-roam-dailies.el` is purely cosmetic, so it's fine for sure.

Thanks Martin.  I dropped this piece from the change.

I agree with Akash, let's address any further refactoring as a followup.

meedstrom:
Agreed.

Issue #2474, opened 2024-09-15
Master Issue for Making Org-Roam Faster


jmay:
I am digging into the =org-roam= code, looking for ways to improve retrieval speed when you have a database containing thousands of nodes.

This issue is for collecting relevant comments, ideas, PRs ... hopefully to eventually yield a solution that will Resolve All The Problems Forever Amen. (Though probably not.)

Some existing Issues and PRs that relate to this:

https://github.com/org-roam/org-roam/issues/2330
https://github.com/org-roam/org-roam/pull/2341
https://github.com/org-roam/org-roam/issues/2383

My primary observation so far is that fetching the node list from the database is *not* the main cause of slowness. Most of the trouble seems to reside in converting those node lists into lists of completion candidates. So I'm starting out by working on caching of those candidates instead of the nodes. So that filtering and sorting still work without touching the existing function interface.

Thoughts, suggestions, warnings all welcome.

okomestudio:
This is a great idea. I too have recognized that what slows down `org-roam` substantially is building the completion candidates, and for my own use I have implemented a caching layer of hash table(s) to speed up node find queries. I was wondering if such a layer already exists in the code base (which I haven't read as thoroughly yet), but apparently it doesn't.

My notes currently consists of a bit less than 8k nodes (including the ones generated from as title aliases). With the caching layer, `org-roam-node-find` runs comfortably under a sub-second range. Without the layer, the "freeze" on each query would make it unusable for me, I think.

I'm also customizing the `org-roam-node-display-template` to add more info useful for filtering, so with the additional processing, it's very crucial to have a layer to help build completion candidates as quickly as possible.

It would be great if some kind of uniform API existed to make things cleaner in this regard.

jmay:
@okomestudio If you don't mind, please post your code! I have a very basic hash-table approach of my own, but cache invalidation is not working yet. I'm hoping to get something working that will speed things up for a majority of users, and then to do some careful refactoring after that.

okomestudio:
I've only relatively recently refactored the portion of my `init.el` related to the said caching layer, so it's really rudimentary and doesn't look very nice, and certainly not intended for public use:

https://github.com/okomestudio/org-roam-plugin-ok/blob/78a8d986642f502227e194fd07d78cab995db589/orp-ok-node.el#L44-L113

In this file, I do some other thing that's personal customization and not particularly helpful in general. 

[...]

PR #2589, opened 2025-12-30
feat: Cache org-roam-node-list for quicker searches

vherrmann:

Using `org-roam-node-find` takes a few seconds on my laptop (when its in powersave-mode). This PR should reduce the time by about 50% if the database was not changed since the last time `org-roam-node-list` was called.

This PR is a stripped-down variant of #2341 giving 50% of the benefit while touching far less code (⇒ better for maintainability).

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -352,7 +352,7 @@ nodes."
             (org-roam-node-aliases node) alias-info)))
   node)
 
-(defun org-roam-node-list ()
+(defun org-roam-node--get-list ()
   "Return all nodes stored in the database as a list of `org-roam-node's."
   (let ((rows (org-roam-db-query
                "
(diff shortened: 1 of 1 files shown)


akashpal-21:
#2241 Related.

dmgerman:
I think the change is sound.

See my analysis here (specially the bottom question): https://chatgpt.com/share/695402b0-a0fc-8009-90a6-27d0f649fc37

hpfr (commented):
This is missing `org-roam-db--get-mtime`.

Regardless, I’m still hesitant about this for the reasons described in https://github.com/org-roam/org-roam/pull/2341#issuecomment-3017118652. Caches should be a last resort once we’ve exhausted other avenues, in particular https://github.com/org-roam/org-roam/issues/2241 and https://github.com/org-roam/org-roam/issues/2474#issuecomment-2369959995. [Filesystem mtimes also have a number of caveats](https://apenwarr.ca/log/20181113).

org-roam-node.el
```diff
@@ -439,6 +439,22 @@ GROUP BY id
...
+  "Previous value of the nodes.")
+(defvar org-roam-node--list-mtime nil
+  "Last time `org-roam-node--list' was changed.")
+(defun org-roam-node-list (&optional force-refresh)
+  "Return all nodes stored in the database as a list of `org-roam-node's.
+
+Memoizes `org-roam-node--get-list'."
+  (if (and (not force-refresh)
+           org-roam-node--list-mtime
+           (time-less-p (org-roam-db--get-mtime) org-roam-node--list-mtime))
```
meedstrom, line 452: Should probably be

    (time-equal-p (org-roam-db--get-mtime) org-roam-node--list-mtime))

to eliminate clock-skew issues.

org-roam-node.el
```diff
@@ -439,6 +439,22 @@ GROUP BY id
...
+      org-roam-node--list
+    (setq org-roam-node--list (org-roam-node--get-list))
+    (setq org-roam-node--list-mtime (current-time))
```
meedstrom, line 455: Should probably be 

    (setq org-roam-node--list-mtime (org-roam-db--get-mtime))

so the timestamps are directly comparable with `time-equal-p`.

meedstrom (changes requested):

meedstrom:
> [Filesystem mtimes also have a number of caveats](https://apenwarr.ca/log/20181113).

Excellent reference, thank you for the link.

hpfr:
Well, the main point of my comment was that I’m hesitant to merge this in any form (even with your suggested changes) before exhausting other approaches :sweat_smile:

PR #2595, opened 2026-01-04
node: capf: Insert id: links instead of roam: links

### LINKED ISSUE #2188: Insert Org ID links via CAPF

hpfr:
The org-roam capfs offer existing nodes, so an id: link can be inserted
directly instead of a roam: link that will be replaced on save anyway.

This also tightens the scope of ‘org-roam-bracket-completion-re’ so
‘org-roam-complete-link-at-point’ only activates for relevant links and
‘completion-at-point’ doesn’t get passed nonsense strings like
‘id:$id][$description’ for completing node titles, resulting in “No
match” instead of trying other CAPFs. The new regexp is aligned more
closely with ‘org-link-bracket-re’ to reduce edge cases.

Fixes: #2188
Closes: #2219

### DIFF
--- a/doc/org-roam.org
+++ b/doc/org-roam.org
@@ -779,16 +779,14 @@ Completions within link brackets are provided by
 ~org-roam-complete-link-at-point~.
 
 The completion candidates are the titles and aliases for all Org-roam nodes.
-Upon choosing a candidate, a ~roam:Title~ link will be inserted, linking to node
-of choice.
+Choosing a candidate inserts an ID link to the chosen node.
 
 ** Completing anywhere
 
 The same completions can be triggered anywhere for the symbol at point if not
 within a bracketed link. This is provided by ~org-roam-complete-everywhere~.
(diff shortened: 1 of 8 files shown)


dustinfarris (commented):
should we be adding/updating any tests here given the behavior has changed slightly?

hpfr:
> should we be adding/updating any tests here given the behavior has changed slightly?

Yes, added some CAPF tests now!

meedstrom (approved):

hpfr:
Approved and tests pass after rebase, merging

PR #2606, opened 2026-02-09
refactor: Simplify org-roam-file-p

meedstrom:

It's better, don't you think?

(Resurrected part of an old PR that I closed, #2546 https://github.com/org-roam/org-roam/commit/cabb0ee09dd503c85c1a2d7f8c50c53a0d45ab81)

### DIFF
--- a/org-roam.el
+++ b/org-roam.el
@@ -207,32 +207,16 @@ FILE is an Org-roam file if:
 - It's located somewhere under `org-roam-directory'
 - It has a matching file extension (`org-roam-file-extensions')
 - It doesn't match excluded regexp (`org-roam-file-exclude-regexp')"
-  (when (or file (buffer-file-name (buffer-base-buffer)))
-    (let* ((path (or file (buffer-file-name (buffer-base-buffer))))
-           (relative-path (file-relative-name path org-roam-directory))
-           (ext (org-roam--file-name-extension path))
-           (ext (if (or (string= ext "gpg")
-                        (string= ext "age"))
-                    (org-roam--file-name-extension (file-name-sans-extension path))
-                  ext))
-           (org-roam-dir-p (org-roam-descendant-of-p path org-roam-directory))
-           (valid-file-ext-p (member ext org-roam-file-extensions))
-           (match-exclude-regexp-p
-            (cond
-             ((not org-roam-file-exclude-regexp) nil)
-             ((stringp org-roam-file-exclude-regexp)
-              (string-match-p org-roam-file-exclude-regexp relative-path))
-             ((listp org-roam-file-exclude-regexp)
-              (let (is-match)
-                (dolist (exclude-re org-roam-file-exclude-regexp)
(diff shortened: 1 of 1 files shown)


hpfr (changes requested):
Nice! You can make the `let` a `when-let*` and make `file` the first binding, so you
don’t needlessly generate `suffixes` before confirming `file` is valid. Not that
that’s too expensive, but it’s nice to have tightly scoped `let` bindings.

Also probably a good opportunity to test that `org-roam-file-p` respects
`org-roam-file-extensions`.

I’m first looking at `org-roam-descendant-of-p` thanks to this. We should
deprecate that as an alias of `file-in-directory-p`, which has proper TRAMP
handling.

meedstrom:
> Nice! You can make the `let` a `when-let*` and make `file` the first binding, so you don’t needlessly generate `suffixes` before confirming `file` is valid. Not that that’s too expensive, but it’s nice to have tightly scoped `let` bindings.

Yeah, I think it'll not usually be too expensive anyway, esp. if we change `org-roam-list-files` to not even use this function. AFAIK, the bottleneck is in `org-roam-descendant-of-p` anyway.  But you can feel free to refactor if you want.


> Also probably a good opportunity to test that `org-roam-file-p` respects `org-roam-file-extensions`.

Oh, no such test already? Yeah, should be simple to do. 

Just one thing I don't like about our current tests, too often when writing a new test you have to rebase a bunch of other PRs because you have to update the numbers for the count of files, count of nodes etc. All those tests that rely on such counts could be rewritten to use e.g. `seq-set-equal-p` (or Buttercup's something like ":to-have-the-same-items-as") that should be more robust to future PRs. Maybe I'll create an issue shortly.

> I’m first looking at `org-roam-descendant-of-p` thanks to this. We should deprecate that as an alias of `file-in-directory-p`, which has proper TRAMP handling.

I did try that a few months back and discovered a good reason to keep `org-roam-descendant-of-p`. It's faster. :)

hpfr:
> Just one thing I don't like about our current tests, too often when writing a new test you have to rebase a bunch of other PRs because you have to update the numbers for the count of files, count of nodes etc. All those tests that rely on such counts could be rewritten to use e.g. `seq-set-equal-p` (or Buttercup's something like ":to-have-the-same-items-as") that should be more robust to future PRs. Maybe I'll create an issue shortly.

Yeah, completely agree.

> > I’m first looking at `org-roam-descendant-of-p` thanks to this. We should deprecate that as an alias of `file-in-directory-p`, which has proper TRAMP handling.
> 
> I did try that a few months back and discovered a good reason to keep `org-roam-descendant-of-p`. It's faster. :)

Faster by a margin that warrants its existence?

```
ELISP> (benchmark-run 10 (org-roam-descendant-of-p (buffer-file-name) org-roam-directory))
(0.000260049 0 0.0)

ELISP> (benchmark-run 10 (file-in-directory-p (buffer-file-name) org-roam-directory))
(0.000502965 0 0.0)
```

That seems fast enough to me!

PR #2608, opened 2026-02-18
Faster `org-roam-list-files` for the third time (now based on `truename-cache`)

meedstrom:
This is a rehash of https://github.com/org-roam/org-roam/pull/2558, but outsources the grunt code to a separate library I wrote: [truename-cache](https://github.com/meedstrom/truename-cache
)  

I have another use-case for the same library, which will also switch soon (https://github.com/meedstrom/org-mem/pull/38).

I don't know if org-roam particularly wants to rely on it, that's up to you guys.

TODO:
- [ ] Add test that verifies `org-roam-file-p` on every file returned by `org-roam-list-files`.
- [ ] Add symlinks among the `tests/roam-files/` files.  At least one symlink should point to somewhere outside that directory.  One should be inaccessible. One should have a name ending in ".org" but actually point to a directory.

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -566,7 +566,22 @@ in `org-roam-db-sync'."
         (db-hash (caar (org-roam-db-query [:select hash :from files
                                            :where (= file $s1)] file-path)))
         info)
-    (unless (string= content-hash db-hash)
+    (if (string= content-hash db-hash)
+        ;; Rare case: `org-roam-db-sync' found the mtime did change, but
+        ;; now we find the content has not.  Instead of re-parsing the file
+        ;; anyway, update only the mtime.  This avoids the possibility of
+        ;; a mass rename incurring a long sync.
+        ;; We can also just no-op here!  Updating the mtime just prevents
+        ;; re-calculating the hash of all affected files on every future
+        ;; invocation of `org-roam-db-sync', so it is a perf fix only.
+        (let ((attr (file-attributes file-path)))
+          (org-roam-db-query [:update files
+                              :set ((= atime $s1)
+                                    (= mtime $s2))
+                              :where (= file $s3)]
+                             (file-attribute-access-time attr)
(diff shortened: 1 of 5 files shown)


hpfr (changes requested):
I am kind of curious what emacs-devel thinks of `truename-cache`. I wonder if others have wanted this before, or if they have other solutions.

I’m a bit wary to introduce a dependency, especially a zerover one in beta with a breaking changes warning. Obviously you’d be the one making the changes and could update org-roam, but still.

Making it optional would be a good middle ground. Leave all the existing shell command support, and just prefer truename-cache if it’s available?

meedstrom:
Yeah, there was curiosity about it on the mailing list, at least for use by project.el, see [what /u/shipmints said on reddit](https://www.reddit.com/r/emacs/comments/1rclv68/comment/o6zd8sb/).

I sympathize with not adding a dependency! It'd be nice if something like it got into Emacs 31.  In the meantime, it feels simpler from org-roam's perspective to just in-house the necessary code as in #2558. Do you agree?

Notes if adding the dependency:

There is at least one flaw in truename-cache which I've left unfixed, I was just lazy at first  but then it became a game of waiting for a reviewer to spot it ;-) But I'll fix it if org-roam wants to pull in the dependency after all. And add org-roam devs as co-contributors so y'all can push hotfixes if needed.

Another problem, even though org-mem has had the same dependency for 2 months now, is that org-mem does not support TRAMP/remote filenames, so those aspects of truename-cache have not been battle-tested.  Would ideally ask a LLM to come up with tests for such cases, or we should definitely do as you suggest and keep the existing shell command support.

hpfr:
I don’t think we should remove the shell command support at this stage. I’d rather not vendor the code and just use `(require 'truename-cache nil 'noerror)` to check if it’s available. That way individual users can opt into the cached version by simply installing `truename-cache`, and everyone else continues to get the current behavior.

PR #2609, opened 2026-02-20
Handle nil schema version in org-roam-db

benthamite:
## Summary

When using `emacsql-sqlite-builtin` (Emacs 29+), the `PRAGMA user_version` query can transiently return nil instead of an integer. This causes `org-roam-db--upgrade-maybe` to signal `(wrong-type-argument number-or-marker-p nil)` when it compares the version with `<`.

The fix guards the version with `integerp` before passing it to `org-roam-db--upgrade-maybe`. When the version is not an integer, the connection is closed (so the next call retries cleanly via the existing `emacsql-live-p` check) and a descriptive error is signaled.

## Backtrace

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  org-roam-db--upgrade-maybe(#<emacsql-sqlite-builtin-connection ...> nil)
  org-roam-db()
  org-roam-db-sync()
  apply(org-roam-db-sync nil)
  timer-event-handler([t 0 1200 0 t org-roam-db-sync nil idle 0 nil])
```

## Environment

- Emacs 30.2 with built-in SQLite (`emacsql-sqlite-builtin`)
- org-roam v2 (commit 18dfe81)

### DIFF
--- a/org-roam-db.el
+++ b/org-roam-db.el
@@ -128,7 +128,11 @@ Performs a database upgrade when required."
         (when init-db
           (org-roam-db--init conn))
         (let* ((version (caar (emacsql conn "PRAGMA user_version")))
-               (version (org-roam-db--upgrade-maybe conn version)))
+               (version (if (integerp version)
+                            (org-roam-db--upgrade-maybe conn version)
+                          (emacsql-close conn)
+                          (error "Org-roam: could not read schema version from %s"
+                                 org-roam-db-location))))
           (cond
            ((> version org-roam-db-version)
             (emacsql-close conn)


meedstrom (approved):

meedstrom:
Approved. Any idea why it can return nil?

benthamite:
Unfortunately, I have no idea.

hpfr:
Even with a new, empty database, I still always get 0:
```elisp
(emacsql-with-connection (db (emacsql-sqlite-open (make-temp-file "user-version-test-")))
  (caar (emacsql db "PRAGMA user_version")))
```

The underlying `(sqlite-select (oref db handle) "PRAGMA user_version;")` also returns `((0))`.

I’m also generally fine with merging this, just confused.

@tarsius, would you have any insight on how `(emacsql db "PRAGMA user_version")` might transiently return nil as the author claims?

tarsius:
I don't think SQLite sets this to anything out-of-the-box. Forge, or rather `closql`, explicitly sets it to some number. If it doesn't do that yet, your package will have to do the same. Maybe SQLite has a compile time knob to set this to `0` for new databases, and that has been turned on on your machine but not those of your users?

hpfr:
`org-roam` also explicitly sets it on creation in `org-roam-db--init`, so that’s not the issue:
https://github.com/org-roam/org-roam/blob/903bd4ec56d29247990d005ed9052c201e18b812/org-roam-db.el#L219-L226

Thank you for the help!

Unfortunate that we can’t reproduce, but I’ll probably apply this soon as there’s little downside.

PR #2610, opened 2026-02-24
minor: Make docstring consistent with function behavior

okomestudio:

`org-roam-node-visit` delegates to `org-roam-node-open`, which does not return the visited buffer. The docstring provides an inconsistent description of the function behavior.

Alternatively, we might actually make `org-roam-node-visit` return the visited buffer. I'm not sure which is desired, but it may be better to keep the existing behavior.

### DIFF
--- a/org-roam-node.el
+++ b/org-roam-node.el
@@ -476,7 +476,7 @@ NODE, unless FORCE is non-nil."
   (org-fold-show-context))
 
 (defun org-roam-node-visit (node &optional other-window force)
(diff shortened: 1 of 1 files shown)


meedstrom:
Thanks!

meedstrom:
>  I'm not sure which is desired, but it may be better to keep the existing behavior.

Yeah, clearly no one has been able to depend on the documented behavior anyway.  But that kind of return value can be useful, so I'd support a PR to that effect if anyone writes it.

Issue #2584, opened 2025-12-07
PROPERTIES drawer with ID is inserted at end of generated node instead of at beginning


trharmon0:
### Description

I don't recall when it started but it has been a few months at least ... Recently when I create a new node the PROPERTIES drawer with ID is being inserted at the end of the node template.  I've looked at the docs and the only thing I can see which might have been causing this behavior is was using :if-new. From the comments in the elisp code, I have since switched to :target. Doing so did not resolve the behavior.

My org-roam-templates variable looks like so:

````elisp
(setopt org-roam-capture-templates '(("d" "default" plain
                                        "%?"
                                        :target (file+head "%<%Y%m%d%H%M%S>-${slug}.org" "#+title: ${title}\n#+category: ${title}\n\n* Timeline\n\n** %u ")
                                        :unnarrowed t
                                        :jump-to-captured nil)
````

which then produces templated output like so:

````
#+title: test node
#+category: test node

* Timeline

** [2025-12-06 Sat]  
:PROPERTIES:
:ID:       b681fec8-5e4d-47cb-b8d7-d28baa0947ae
:END:

````

#### Steps to Reproduce

1. trigger new node creation
2. C-c c to execute the node generation
3. review the generated node 

#### Backtrace

I don't believe this section to be relevant to this issue but I suppose I can find a way to generate a call trace if I'm wrong.

#### Expected Results

New node is generated from template with the PROPERTIES and the new node ID at the head of the generate node.

````
:PROPERTIES:
:ID:       b681fec8-5e4d-47cb-b8d7-d28baa0947ae
:END:

#+title: test node
#+category: test node

* Timeline

** [2025-12-06 Sat]  
````

#### Actual Results

See above.

### Environment

- Emacs: GNU Emacs 30.2 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.50, cairo version 1.18.4)
- Framework: N/A
- Org: Org mode version 9.7.11 (release_9.7.11 @ /usr/share/emacs/30.2/lisp/org/)
- Org-roam: v2.3.1-24-gf4ba41c
- sqlite-connector: emacsql-sqlite-builtin-connection

schrenker:
This does not happen on commit 89dfaef38b6caa3027f20f96a551dc8f194ac533 to which I reverted to avoid this problem.

Issue #2582, opened 2025-12-03
Updating org-roam to latest causes a bug in dailies capture.


larrasket:
#### Steps to Reproduce

Set your template to:
```elisp
(setq org-roam-dailies-capture-templates
        `(("d" "default" entry "* %<%H:%M> \n %?"
           :if-new
           (file+head
            "%<%Y-%m-%d>.org.gpg"
            ,(concat "#+title: %<%Y-%m-%d>\n#+DATE: <%<%Y-%m-%d>>"
                     "\n#+FILETAGS: journal\n"))
           :unnarrowed t))
        ;; org-roam-database-connector             'sqlite
        org-roam-dailies-directory              "stuff/")
```

And capture today.

#### Expected Results

Title has the org ID:
```org
:PROPERTIES:
:ID:       ea40u1m00sk0
:END:
#+title: 2025-12-03
#+DATE: <2025-12-03>
#+FILETAGS: journal
- tags :: [[roam:Journaling]]  
* 19:13 
 ```

#### Actual Results

Subheading has the orgid:
```org
#+title: 2025-12-03
#+DATE: <2025-12-03>
#+FILETAGS: journal
* 19:13 
:PROPERTIES:
:ID:       ea40u1m00sk0
:END:
 ```

dustinfarris:
If you change `entry` to `plain` you should get the behavior you expect

larrasket:
> If you change entry to plain you should get the behavior you expect
> 


Interesting, but why was the behavior change upon update?
—
Regards
Salih

Issue #2551, opened 2025-10-22
org-roam-dailies-capture-templates changed behavior


KyleBarton:
### Description
Hi there! After upgrading `org-roam` to the latest on elpa (`org-roam-20250927.2243`), it seems that my dailies capture templates are no longer function as before. I have two - the default template which matches the docs, and a "timestamp" template which adds a heading with the current time. Details in the reproduction steps below.

Before, the new Daily file would create a `PROPERTIES` drawer with the node id. Now, no such drawer is created at the file level, but a drawer with an ID is created within the new heading. Note that normal, non-daily org-roam capture templating continues to work fine.

#### Steps to Reproduce
1. Load emacs
2. Set the following daily capture templates:

```
(setq org-roam-dailies-capture-templates
          '(("d" "default" entry
             "* %?"
             :target (file+head "%<%Y-%m-%d>.org.gpg"
                                "#+title: %<%Y-%m-%d>\n"))
            ("t" "timestamped note" entry
             "* %<%H>:%<%M>\n%?"
             :target (file+head+olp "%<%Y-%m-%d>.org.gpg"
                                    "#+title: %<%Y-%m-%d>\n"
                                    ("Minutes")))))
```
3. Run `org-roam-dailies-capture-today`
4. For clarity, run `widen` in order to see the whole captured file

#### Expected Results

In the default case:

<img width="859" height="216" alt="Image" src="https://github.com/user-attachments/assets/b21d8586-cd5d-4b23-bad9-c93c6aeb8008" />

In the timestamp case:

<img width="859" height="275" alt="Image" src="https://github.com/user-attachments/assets/1c60270b-a14c-4541-a7f6-64f520cdb2ba" />

#### Actual Results

In the default case:

<img width="1344" height="278" alt="Image" src="https://github.com/user-attachments/assets/de932649-bdaa-4e1e-8409-e33a1d6c26bb" />

In the timestamp case:

<img width="1344" height="344" alt="Image" src="https://github.com/user-attachments/assets/c038bdad-9326-4afc-bf88-26893bd21c01" />

### Environment

```
- Emacs: GNU Emacs 30.2 (build 2, aarch64-apple-darwin24.6.0, NS appkit-2575.70 Version 15.6.1 (Build 24G90))
 of 2025-10-21
- Framework: N/A
[...]

Issue #2550, opened 2025-10-21
Error after recent changes to `org-roam-capture.el`


benthamite:
### Description

When I try to create a note for a bibtex entry using `orb-citar-edit-note` (from [org-roam-bibkex](https://github.com/org-roam/org-roam-bibtex/)), I get the error whose backtrace is below. If I checkout v2.3.1 (commit `7ce95a2`), the note is created successfully. Since a few commits since v2.3.1 affect `org-roam-capture.el`, it appears that some of these changes introduced a bug.

You can reproduce by evaluating something like

```emacs-lisp
(orb-citar-edit-note "Dafoe2018AIGovernanceResearch" '(("=key=" . "Dafoe2018AIGovernanceResearch") ("=type=" . "report") ("date" . "2018") ("author" . "Dafoe, Allan") ("institution" . "Future of Humanity Institute, University of Oxford") ("url" . "https://www.fhi.ox.ac.uk/wp-content/uploads/GovAI-Agenda.pdf") ("abstract" . "Artificial intelligence (AI) could be a rapidly developing potent general-purpose technology with both tremendous opportunities and substantial risks. Its future progress might accelerate in the coming decades, achieving superhuman capabilities in strategic and diverse domains. AI development could enable major advances in medicine, transportation, energy, education, science, economic growth, and environmental sustainability. However, significant governance challenges plausibly arise from its development, such as labor displacement, wealth inequality, privacy concerns, national security and safety risks, misuse, and unintended consequences – AI-generated abstract.") ("title" . "AI governance: a research agenda") ("langid" . "english") ("database" . "Tlön")))
```

but the bibtex entry "Dafoe2018AIGovernanceResearch" must exist for this to work.

#### Backtrace

```emacs-lisp
Debugger entered--Lisp error: (error "Capture template ‘d’: Wrong type argument: stringp, nil")
  error("Capture template `%s': %s" "d" "Wrong type argument: stringp, nil")
  org-capture(nil "d")
[...]

PR #2611, opened 2026-02-26
test: Add regression tests for capture ID placement

kovan:
Addresses #2591 - regression tests for capture bugs fixed by revert b2634a1.

7 focused regression tests covering #2550, #2551, #2582, and #2584:
- org-roam-capture--convert-template: entry/plain type conversion and property preservation, using pcase destructuring in assertions
- org-roam-capture integration: runs real capture templates through org-roam-capture-, finalizes the capture, and checks the resulting file contents
- ID placement coverage for plain file+head, entry file+head, entry file+head+olp, and existing file IDs

Verified:
- eldev -C --unstable -T test --file tests/test-org-roam-capture.el . - Ran 13 specs, 0 failed
- eldev -C --unstable -T test - Ran 58 specs, 0 failed

### DIFF
--- a/tests/test-org-roam-capture.el
+++ b/tests/test-org-roam-capture.el
@@ -58,6 +58,241 @@
      (org-roam-capture--fill-template (lambda () "foo"))
      :to-equal "foo")))
 
+;; Regression tests for capture regressions fixed by revert b2634a1.
+;; These cover issues #2584, #2582, #2551, and #2550.
+
+(describe "org-roam-capture--convert-template"
+  ;; Regression: #2550 — entry-type templates with third-party packages
+  ;; (e.g. org-roam-bibtex) produced "Wrong type argument: stringp, nil"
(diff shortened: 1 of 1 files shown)


hpfr (changes requested):
Can you tell the LLM to use `pcase` rather than a bunch of `nth` `nthcdr`?

Some examples:
https://github.com/org-roam/org-roam/blob/903bd4ec56d29247990d005ed9052c201e18b812/tests/test-org-roam-node.el#L323-L335

Also, a bunch of these tests mock internal implementation details of `org-roam-capture`. They dynamically override `org-roam-capture--get` and `org-capture-get` but that shouldn’t be necessary. I would think the tests could just set `org-roam-capture-templates` and run the top level `org-roam-capture` with the appropriate key, responding to prompts for input if the template requires, and then check that the capture buffer matches an expected string result.

PR #2613, opened 2026-02-27
feat(id): complete id links with org-roam nodes

### LINKED ISSUE #2137: Node completion in `org-insert-link` -> `id`
### Brief Abstract
(perhaps only when in a roam buffer), if `id` is selected as the link type in `org-insert-link`, offer completion over roam nodes (with `org-roam-node-read` and insert a link to the relevant id.

### Long Description
I use org both with and without org-roam and I want a really easy, simple workflow. Basically, I *do* want completion on node names when I'm inserting links in roam buffers, but I *don't* want to have to remember that I'm in a roam buffer and therefore use a special keybinding for `org-roam-node-insert`. I just want to insert link with org and have it Just Work™.

So, like altering the link following behaviour (as roam [already does](https://github.com/org-roam/org-roam/blob/3782e88d50f83c5b9fbb4b10df71df3f37e27156/org-roam-id.el#L91)), set org's built-in completion to play nice with roam nodes.

```elisp
(defun org-roam-id-complete (&optional initial-input filter-fn sort-fn require-match prompt)
  "Read an `org-roam-node', returning its id.

All args are passed to `org-roam-node-read'."
  (org-roam-node-id
   (org-roam-node-read
    initial-input filter-fn sort-fn require-match prompt)))

(org-link-set-parameters "id" :complete #'org-roam-id-complete)
```

Would go [here](https://github.com/org-roam/org-roam/blob/3782e88d50f83c5b9fbb4b10df71df3f37e27156/org-roam-id.el#L92). I included passing the optional args incase this function is useful in other lisp in the future.

This could be augmented with a check of `org-roam-buffer-p` if we wanted to avoid linking outside of roam buffers. I'm not sure if this is relevant, because org doesn't have a default completion mechanism for `id` links, so this implementation doesn't shadow anything.

#### Possible extensions
It would be nice to set the title of the node as the default/suggested description for the link, but this doesn't seem possible in a way which doesn't have to account for every link type at once (see [here](https://emacs.stackexchange.com/a/27860/34394) and the docstring of `org-link-make-description-function`).

Failing that, it might be a good idea to add the title to the history (or future history) of the relevant list, so that it can be easily pulled down as a description. This should be easy enough in the above function with `add-to-history`, but I don't know which list to add to. I could make the change work if I did though!

kovan:
Closes #2137.

## Summary
Adds `org-roam-id-complete`, then registers it as the `id` link completion function.

This makes `org-insert-link` with link type `id` complete over Org-roam nodes and insert an `id:<node-id>` link target.

## Changes
- `org-roam-id.el`
  - add `org-roam-id-complete`
  - register completion: `(org-link-set-parameters "id" :complete #'org-roam-id-complete)`
- `tests/test-org-roam-id.el`
  - unit tests for return value and argument pass-through
  - integration test asserting `id` link completion is registered

## Checks done
- `eldev -C --unstable -T test --file tests/test-org-roam-id.el .`
  - Result: `Ran 7 specs, 0 failed`
- completion parameter check:
  - `id` link completion parameter resolves to `org-roam-id-complete`
- completion behavior check:
  - completion function returns `"id:xyz-987"` when node-read returns that id

### DIFF
--- a/org-roam-id.el
+++ b/org-roam-id.el
@@ -26,6 +26,9 @@
 ;;; Code:
 (require 'org-id)
 
+(declare-function org-roam-node-read "org-roam-node")
+(declare-function org-roam-node-id "org-roam-node")
+
 (defun org-roam-id-at-point ()
   "Return the ID at point, if any.
 Recursively traverses up the headline tree to find the
@@ -60,6 +63,15 @@ With optional argument MARKERP, return the position as a new marker."
   "Obsolete alias - use `org-id-open' directly.")
 
 (advice-add 'org-id-find :before-until #'org-roam-id-find)
+
+(defun org-roam-id-complete (&optional initial-input filter-fn sort-fn require-match prompt)
+  "Read an `org-roam-node', returning an \"id:\" link.
+All args are passed through to `org-roam-node-read'."
+  (concat "id:"
+          (org-roam-node-id
+           (org-roam-node-read initial-input filter-fn sort-fn require-match prompt))))
+
+(org-link-set-parameters "id" :complete #'org-roam-id-complete)
 
 ;;;###autoload
 (defun org-roam-update-org-id-locations (&rest directories)
--- a/org-roam-id.el
+++ b/org-roam-id.el
@@ -28,6 +28,7 @@
 
 (declare-function org-roam-node-read "org-roam-node")
 (declare-function org-roam-node-id "org-roam-node")
+(defvar org-roam-link-type)
 
 (defun org-roam-id-at-point ()
   "Return the ID at point, if any.
@@ -71,7 +72,7 @@ All args are passed through to `org-roam-node-read'."
           (org-roam-node-id
            (org-roam-node-read initial-input filter-fn sort-fn require-match prompt))))
 
-(org-link-set-parameters "id" :complete #'org-roam-id-complete)
+(org-link-set-parameters org-roam-link-type :complete #'org-roam-id-complete)
 
 ;;;###autoload
 (defun org-roam-update-org-id-locations (&rest directories)
--- a/tests/test-org-roam-id.el
+++ b/tests/test-org-roam-id.el
@@ -74,6 +74,25 @@
       (expect (car location) :to-match ".*/tests/roam-files/family.org")
       (expect (cdr location) :to-equal 156))))
 
+(describe "org-roam-id-complete"
+  (it "returns an id link using org-roam node completion"
+    (let ((node (org-roam-node-create :id "abc-123")))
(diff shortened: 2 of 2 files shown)


hpfr (changes requested):
I’ve never been a fan of how `org-roam` presumes no other Org extensions use Org IDs, and this takes things a bit further in that direction. Let’s try to not take over all ID completion if we can, since users can use ID links outside of org-roam-directory.

Can we instead tie this to `roam`? That way, users choosing the `roam:` type when inserting links get Roam nodes, and `id:` still gives a clean slate. The completion function can still return the ID prefixed with `id:` to give the final form of the link.

(and individual users can still opt into `id:` completion in their config with another `org-link-set-parameters` call)

meedstrom:
I was gonna ask why someone would even want to complete id: manually, but I suppose it makes sense in the case that you have manually assigned short IDs like 1234.

I agree about keeping it opt-in.  Heck, it need not be limited to org-roam-directory, users could still filter from all their org-id-locations pretty quickly if their special IDs have a standard prefix like "n" (for "note") so they can complete to "id:n1234".

EDIT: Ah, I see from #2137 it's for completion inside of an `org-insert-link` minibuffer. Well, then perhaps it must be limited to org-roam-directory after all.

org-roam-id.el
```diff
@@ -61,6 +65,15 @@ With optional argument MARKERP, return the position as a new marker."
 
 (advice-add 'org-id-find :before-until #'org-roam-id-find)
 
+(defun org-roam-id-complete (&optional initial-input filter-fn sort-fn require-match prompt)
```
hpfr, line 68: None of these args will ever be used by `org-link--try-special-completion`. They can be replaced with simply `(&optional _arg)` to match the spec defined in `org-link-parameters`, and the tests checking the passthrough can be removed. For tests, see the capf tests, we should do something like that.

Let’s call the function `org-roam-complete-link` to match other custom link completion functions (`org-pdftools-complete-link`, `org-attach-complete-link`).

org-roam-id.el
```diff
@@ -61,6 +65,15 @@ With optional argument MARKERP, return the position as a new marker."
...
+  "Read an `org-roam-node', returning an \"id:\" link.
+All args are passed through to `org-roam-node-read'."
+  (concat "id:"
+          (org-roam-node-id
+           (org-roam-node-read initial-input filter-fn sort-fn require-match prompt))))
+
+(org-link-set-parameters org-roam-link-type :complete #'org-roam-id-complete)
```
hpfr, line 75: Note to self: block this PR on https://github.com/org-roam/org-roam/pull/2571 and move this into the mode

hpfr (changes requested):

Issue #2569, opened 2025-11-08
Proposal: use `require` in the conventional way


meedstrom:
## Background

We have at the end of org-roam.el, a list of modules to always be loaded together with the main file:

```elisp
;;; Package bootstrap
(provide 'org-roam)

(cl-eval-when (load eval)
  (require 'org-roam-compat)
  (require 'org-roam-utils)
  (require 'org-roam-db)
  (require 'org-roam-node)
  (require 'org-roam-id)
  (require 'org-roam-capture)
  (require 'org-roam-mode)
  (require 'org-roam-log)
  (require 'org-roam-migrate))

;;; org-roam.el ends here
```

Curious, isn't it?

Conventionally, these require-statements would go at the beginning of the file, not the end.

I surmise that they were put at the end because they rely on some definitions in the main file, which must exist first.

(Require-statements must be acyclic; either the main file loads a module, *or* a module loads the main file.  At least that's true if you're conventional and keep the requires at the beginning and don't use that `cl-eval-when` expression.)

What org-roam has done is essentially make all files the same file, with `org-roam.el` constituting the first few lines in it.

That's why `org-roam-node.el` can contain references to unknown functions like `org-roam-db-query` assuming it'll be defined later, just as if it had been a single file with the definition further down the file.

We don't hear the compiler complaints due to that `cl-eval-when` expression, but flycheck does draw a squiggly line under occurrences of `org-roam-db-query` in the file `org-roam-node.el`, if I have not loaded org-roam.

## Mapping it out

As this is, hypothetically, a [multi-file package](https://www.gnu.org/software/emacs/manual/html_node/elisp/Multi_002dfile-Packages.html), let's see how we would go about it conventionally to solve "who should require whom?"

You can categorize the `.el` files in a multi-file package into two mental buckets:

- *extensions*: those that load the main file
- *submodules*: those on which the main file depends
  - Submodules may also depend on other submodules, so there would be an implied necessary load-order.
  
    The programmer needs not know the exact order; it is resolved thanks to each submodule containing a comprehensive set of `require` statements.

Unfortunately, most files in this codebase have no require-statements.  So to reconstruct this information, I mapped out the implied dependencies:

(A handy trick for doing this yourself: in for example `org-roam-node.el`, type `M-x highlight-regexp RET org-roam-[^n][[:alnum:]-]* RET`)

```
- org-roam-db
  - it uses:
    - org-roam-node-at-point             from org-roam-node
[...]

PR #2621, opened 2026-05-31
Silence byte-compile warnings and add a byte-compile workflow

dannywillems:
This cleans up the byte-compilation warnings that show up when org-roam is
compiled (for example via `make lint` / `eldev compile`), without changing
any runtime behaviour, and adds a dedicated workflow so byte-compilation is
checked on its own.

### "the function ... is not known to be defined"

org-roam loads its sibling modules at runtime rather than at compile time:
`org-roam.el` requires the submodules inside a `(cl-eval-when (load eval) ...)`
form, and some pieces are wired up via `with-eval-after-load`. When each file
is byte-compiled in isolation, the cross-module functions it calls are not yet
defined, so the compiler warns. Added `declare-function` statements (matching
the pattern already used for `org-roam--directory-files-recursively`):

- `org-roam.el`: `org-roam-descendant-of-p`
- `org-roam-node.el`: `ucs-normalize-NFD-string`, `ucs-normalize-NFC-string`
  (required at runtime inside `org-roam-node-slugify`), `org-roam-capture-`,
  `org-roam-id-at-point`
- `org-roam-compat.el`: `org-roam-message`,
  `org-roam-update-org-id-locations`, `org-roam-capture--get`,
  `org-roam-capture--get-if-new-target-a`, `org-roam-db-autosync-mode`
- `org-roam-utils.el`: `org-roam-db--get-connection`
- `org-roam-db.el`: `org-roam-node-at-point`, `org-roam-id-at-point`

### "reference/assignment to free variable"

`org-roam-compat.el` touches `org-roam-capture--template-keywords` (a defconst
in `org-roam-capture`) inside a `with-eval-after-load 'org-roam-capture` block.
Added a bare `(defvar org-roam-capture--template-keywords)` forward
declaration. A defvar with no value only marks the symbol special for the
compiler and does not bind or alter the variable.

All of the above are purely informational for the byte-compiler and do not
change behaviour.

## CI

Added `.github/workflows/byte-compile.yml`, which byte-compiles and then loads
org-roam on Emacs 29.4, 30.1 and snapshot, plus a Dependabot config for the
GitHub Actions. The snapshot job is `continue-on-error` because it tracks the
daily Emacs development build. Warnings are not yet treated as errors because
of the residual warning below.

## Out of scope (left as residual)

- `org-roam-node.el`: `Warning: Unknown slot 'node'` in `org-roam-node-at-point`.
  The `node` slot belongs to the `org-roam-node-section` EIEIO class defined in
  `org-roam-mode.el`, which is compiled after `org-roam-node.el`. Resolving this
  cleanly would require reordering class definitions (a structural change), so
  it is left untouched here.


dannywillems:
This is an automated effort to help maintaining packages in the Emacs community. See https://x.com/dwillems42/status/2060720730338185699 and https://dannywillems.github.io/emacs-package-maintenance/

meedstrom:
> This is an automated effort to help maintaining packages in the Emacs community. See https://x.com/dwillems42/status/2060720730338185699 and https://dannywillems.github.io/emacs-package-maintenance/

Thank you! I love projects like yours, it tickles the civic engineer in me.

I'm fine with this PR, but 

- perhaps we can add a comment somewhere that these declares should become unnecessary once we get around to https://github.com/org-roam/org-roam/issues/2569
  - it's still a long time off given the pace of this project, so silencing warnings now is appreciated!
- question: is a byte-compile workflow actually necessary to silence warnings?

meedstrom:
> ## Out of scope (left as residual)
> * `org-roam-node.el`: `Warning: Unknown slot 'node'` in `org-roam-node-at-point`.
>   The `node` slot belongs to the `org-roam-node-section` EIEIO class defined in
>   `org-roam-mode.el`,

Aye, and I think that org-roam-mode.el will eventually move into `extensions/` subdirectory, so we're definitely not changing the load order. 

I see an invocation of magit-section-case here:  https://github.com/org-roam/org-roam/blob/903bd4ec56d29247990d005ed9052c201e18b812/org-roam-node.el#L235-L240

Ideally, org-roam-mode.el will be the only file that uses magit-section, so perhaps it can export a function called "org-roam-mode-section-at-point" or some such that we only call here if it is fboundp.

dannywillems:
Thanks for the warm reception!

- Added a comment noting these declarations should become unnecessary once #2569 lands (in org-roam-compat.el, referencing the other modules too).
- Good question on the workflow: no, a byte-compile workflow is not needed to silence the warnings; the `declare-function`/`defvar` forms do that on their own. Since org-roam already runs CI via `test.yml`, I have dropped the extra `byte-compile.yml` (and the dependabot file) from this PR, so it is now just the warning fixes plus that comment.
- On the `Unknown slot 'node'` residual: your suggestion of an `org-roam-mode-section-at-point` in org-roam-mode.el called via `fboundp` from org-roam-node.el is a clean way to avoid the cross-module slot reference without touching load order. I left it out of this PR to keep it scoped to warning silencing, but I am happy to open a follow-up for it if you would like.

gatuno9987 (approved):

